PR #9, opened 2015-04-06
Remove bundled node version from version command

jviotti:
We're not bundling node anymore.


abresas:
we are not? :S

jviotti:
@abresas Nope, we decided long ago to remove the bundling and instead make the installers take care of getting node in the user's system first.

abresas:
LGTM then

PR #10, opened 2015-04-08
Fix bug with login with token

jviotti:
The commit was merged from an older branch, in which settings was
still part of the SDK.


abresas:
LGTM

PR #20, opened 2015-04-17
Add lodash dependency to lib/elevate.coffee

jviotti:
Depedency was missing and thus elevating functions threw errors.


emirotin:
LGTM

PR #21, opened 2015-04-17
Bind progress update function to avoid context issues

jviotti:


lib/actions/os.coffee
```diff
@@ -167,7 +167,7 @@ exports.install =
 				return done() if not confirmed
 
 				bar = new visuals.widgets.Progress('Writing Device OS')
-				params.progress = bar.update
+				params.progress = _.bind(bar.update, bar)
```
emirotin: in Node you can also safely use native bind:

params.progress = bar.update.bind(bar)

emirotin:
LGTM

PR #23, opened 2015-04-20
Make device init and os download take app names instead of ids

jviotti:


jviotti:
@abresas Yeah, you're right. I'll implement `vcs.getApplicationName()`.

jviotti:
@abresas Updated with `vcs.getApplicationName()`. Please take another look.

abresas:
nice! good for merge

PR #26, opened 2015-04-20
Fix temporary os download output option in device init

jviotti:


abresas:
I suggest using https://github.com/raszi/node-tmp instead. We use it in other projects as well.

tmp.file() callback returns path, fd, and cleanup function that you can call in the end

```
tmp.file(function _tempFileCreated(err, path, fd, cleanupCallback) {
  if (err) throw err;

  console.log("File: ", path);
  console.log("Filedescriptor: ", fd);

  // If we don't need the file anymore we could manually call the cleanupCallback
  // But that is not necessary if we didn't pass the keep option because the library
  // will clean after itself.
  cleanupCallback();
});
```
It handles uncaught exceptions and other corner cases where you may have data leaks with

```
var tmp = require('tmp');
tmp.setGracefulCleanup();
```

Our images are huge files ~1GB so it's really sad if they are left on the user's device for no reason.

It uses `os.tmpDir` too internally.

abresas:
The code comment I pasted from the project suggests that it's not necessary to clean. It will clean by itself when the process exits, but better do it early IMO.

jviotti:
@abresas Implemented. Can you take another look please?

lib/actions/device.coffee
```diff
@@ -210,10 +211,26 @@ exports.init =
 			(confirmed, callback) ->
 				return done() if not confirmed
 				options.yes = confirmed
-				osAction.download.action(params, options, callback)
 
-			(outputFile, callback) ->
+				# Cleanup the temporary files even when an uncaught exception occurs
+				tmp.setGracefulCleanup()
```
emirotin: maybe this should be done at the very beginning of the file

package.json
```diff
@@ -66,6 +66,7 @@
     "resin-sdk": "^1.2.0",
     "resin-settings-client": "^1.0.0",
     "resin-vcs": "^1.2.0",
+    "tmp": "0.0.25",
```
emirotin: why using strict version here?

lib/actions/device.coffee
abresas: Yeah as @emirotin says, it's a module-level setting so just after require

package.json
abresas: I think npm install --save uses a strict version for some reason on tmp, I had it happen to me as well.
Maybe that's why it it happened here too. Better use ^

lib/actions/device.coffee
jviotti: FIxed.

package.json
jviotti: Fixed.

abresas:
LGTM

PR #29, opened 2015-04-27
Make devices command list all devices if no application option

jviotti:


jviotti:
Fixes https://github.com/resin-io/resin-cli/issues/17.

PR #40, opened 2015-05-05
Implement device await command

jviotti:


lib/actions/device.coffee
```diff
@@ -163,6 +163,46 @@ exports.supported =
...
 
+exports.await =
+	signature: 'device await <name>'
+	description: 'await for a device to become online'
+	help: '''
+		Use this command to await for a device to become online.
+
+		The process will exit when the device becomes online.
+
+		Notice that there is not time limit for this command, so it might run forever.
```
emirotin: "no time limit"

lib/actions/device.coffee
```diff
@@ -163,6 +163,46 @@ exports.supported =
...
+
+		You can configure the poll timeout with the --interval option (defaults to 3000ms).
```
emirotin: better to call it interval, not timeout

emirotin:
lgtm except of the small comments above

lib/actions/device.coffee
jviotti: Nice catch, fixed.

lib/actions/device.coffee
jviotti: Fixed.

PR #48, opened 2015-05-13
Make use of selfupdate

jviotti:


lib/actions/update.coffee
```diff
@@ -23,36 +20,7 @@ exports.update =
...
-
-					error.code is 'EPERM'
-					error.code is 'ACCES'
-				]
-					return president.execute(command, onUpdate)
-
-				return done(error)
-
-
+			console.info("Upgraded #{packageJSON.name} to version #{version}.")
```
emirotin: better to say "Updated" maybe?

jviotti: Yeah, much better. Fixed. Thanks for the suggestion.

PR #50, opened 2015-05-18
Point to a JavaScript file in package.json main property

jviotti:


emirotin:
lgtm

though I wanted to ask for long time: why do we publish all the built files? to make them independently `require`able? cause in case you export everything under a single namespace from the main file you can just have a single `src/main.js` that does `require('coffee-script/register'); module.exports = require('./lib/main');`

or are the files built for the sake of speed?

jviotti:
@emirotin Yeah, the main motivation was speed. Running the JavaScript files directly gave a small speed boost back then.

PR #54, opened 2015-05-27
Fix device await help not showing up

jviotti:
Since it was declared after device.info, doing:

  $ resin help device await

Showed the help page of the resin device command instead.


emirotin:
lgtm

PR #55, opened 2015-05-28
Add per device environment variable support

jviotti:


lib/actions/environment-variables.coffee
```diff
@@ -56,16 +71,25 @@ exports.remove =
 		Notice this command asks for confirmation interactively.
 		You can avoid this by passing the `--yes` boolean option.
 
+		If you want to eliminate a device environment variable, pass the `--device` boolean option.
+
 		Examples:
 
 			$ resin env rm 215
 			$ resin env rm 215 --yes
+			$ resin env rm 215 --device
```
Page-: How does this know which device to do it for?

lib/actions/environment-variables.coffee
```diff
@@ -31,19 +34,31 @@ exports.list =
 	]
 	permission: 'user'
 	action: (params, options, done) ->
-		resin.models.environmentVariables.getAllByApplication options.application, (error, environmentVariables) ->
-			return done(error) if error?
+		async.waterfall([
```
Page-: This would be nicer if you made use of the models promise interface, ie
```coffee
Promise.try ->
	if options.application?
		resin.models.environmentVariables.getAllByApplication(options.application, callback)
	else if options.device?
		resin.models.environmentVariables.device.getAll(options.device, callback)
	else
		throw new Error('You must specify an application or device')
.then (environmentVariables) ->
	if not options.verbose
		isSystemVariable = resin.models.environmentVariables.isSystemVariable
		environmentVariables = _.reject(environmentVariables, isSystemVariable)

	console.log visuals.widgets.table.horizontal environmentVariables, [
		'id'
		'name'
		'value'
	]
.nodeify(done)
```

jviotti: I'm making a note to make use of promises in all CLI commands that make use of SDK models functions, and will push another PR.

Page-: Ok, other than that it looks ok

PR #57, opened 2015-06-02
Ask for confirmation on app associate command

jviotti:


lib/actions/app.coffee
```diff
@@ -146,11 +146,15 @@ exports.associate =
 
 		This command adds a 'resin' git remote to the directory and runs git init if necessary.
 
+		Notice this command asks for confirmation interactively.
+		You can avoid this by passing the `--yes` boolean option.
```
Page-: What about `-y`?

jviotti: @Page- It's an alias to `--yes`, and will be displayed below automatically when doing `resin help app associate`.

PR #60, opened 2015-06-02
Open dashboardUrl instead of remoteUrl in auth login

jviotti:
When opening a browser, it currently used api.resin.io. Since the user usually login from dashboard.resin.io, and the hosts differed, the session was not shared between both of them.


Page-:
LGTM

emirotin:
LGTM. Also this change is the must as not only the hosts differ but since now you won't be able to login through API at all

PR #61, opened 2015-06-02
Fallback to nodejs binary if node was not found

jviotti:
Deals with Ubuntu's non standard node binary, which is called `nodejs`.


jviotti:
I confirm it works on a fresh Ubuntu 14.04 installation.

jviotti:
It works in OS X as well.

jviotti:
Doesn't work on Windows. I tried various workarounds but NPM will not make the correct symlink in Windows except the shebang is `#!/usr/bin/env node`.

PR #64, opened 2015-06-03
Implement config injection

jviotti:
Config injection support for the CLI.


lib/actions/device.coffee
```diff
@@ -266,27 +275,82 @@ exports.init =
...
-			(tmpPath, tmpFd, cleanupCallback, callback) ->
-				options.output = tmpPath
+			(results, callback) ->
+				console.info('Associating the device')
 
-				# TODO: Figure out how to make use of capitano.run()
-				# here given the complexity of converting network
-				# params object to string options
-				osAction.download.action params, options, (error, outputFile) ->
+				registerDevice.register pine, results.config, (error, device) ->
```
Page-: This can be:
```coffee
registerDevice.register(pine, results.config)
.then (device) ->
	results.config.deviceId = device.id
	params.uuid = results.config.uuid = device.uuid
.return result
.nodeify(callback)
```

jviotti: I prefer to hold on promises for now on the CLI. After porting the SDK to use promises (work in progress), I'll port the CLI as well.

Page-: The exposed interface wouldn't change, it'd just mean internal code can be a bit nicer by making use of interfaces exposed by libraries.

jviotti: Sure, I mean that I'll change the internal code to use promises all at once, to not mix both approaches yet.

Page-:
LGTM

PR #65, opened 2015-06-04
Fix app commands order of definition. Closes #62.

### LINKED ISSUE #62: >= 2 word command help doesn't work correctly
So this runs correctly:-

```bash
resin help app
Usage: app <name>

Use this command to show detailed information for a single application.

Examples:

	$ resin app MyApp
```

But if I run `resin help app rm` I get the same output:-

```bash
[~]$ resin help app rm                                                                                   
Usage: app <name>

Use this command to show detailed information for a single application.

Examples:

	$ resin app MyApp
```

Also in both cases I feel it would be useful to have flags like `-y` available and listed here :)

jviotti:
This caused `resin help app rm` erroneusly show the help page for `resin app`.

### DIFF
--- a/build/app.js
+++ b/build/app.js
@@ -78,8 +78,6 @@
 
   capitano.command(actions.app.list);
 
-  capitano.command(actions.app.info);
-
   capitano.command(actions.app.remove);
 
   capitano.command(actions.app.restart);
@@ -87,6 +85,8 @@
   capitano.command(actions.app.associate);
 
   capitano.command(actions.app.init);
+
+  capitano.command(actions.app.info);
 
   capitano.command(actions.device.list);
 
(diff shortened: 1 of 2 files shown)


Page-:
This feels like it should be fixed at the root in capitano, it's a rather unintuitive bug and took me a while to figure out why this change fixed it - presumably because it uses the first potential match, rather than the longest potential match.

Which makes me think capitano should either iterate the entire list for the longest match, or at the point of adding potential matches it can order them by the length of the matching string and keep the first matched behaviour (slightly slower at point of adding, slightly faster at the point of searching - better for using the commands in a programmatic way, since they only get parsed once but potentially called many times)

jviotti:
@Page- Can you move that comment to a Capitano's issue please? I'll merge this quick fix for now and will work in a definite solution later.

PR #67, opened 2015-06-09
Implement Quickstart command

jviotti:


build/actions/wizard.js
```diff
@@ -0,0 +1,89 @@
...
+            }, function(applicationName, callback) {
+              params.name = applicationName;
+              return callback();
+            }
+          ], callback);
+        }, function(callback) {
+          return capitano.run("device init --application " + params.name, callback);
+        }, function(deviceName, callback) {
+          params.device = deviceName;
+          console.log('Please install SD card in Raspberry Pi, Connect it to network and power it up.');
```
shaunmulligan: Is this hardCoded to the Raspberry Pi device type?

jviotti: @shaunmulligan Nice catch. I think I copy pasted this from your Google Doc and didn't pay attention to the hardcoded device type. I'll fix now.

jviotti: Fixed! It now uses the device name for the selected application.

Page-:
The power it up will also have to change, eg BBB - power up whilst holding the tiniest, hardest to hold button possible

jviotti:
@Page- I'll be printing the device instructions as the UI very soon (in this PR).

lib/actions/wizard.coffee
```diff
@@ -0,0 +1,118 @@
...
+
+						], callback
+
+					(application, callback) ->
+						return callback(null, application) if application?
+
+						form.ask
+							message: 'Choose a Name for your new application'
+							type: 'input'
+						, (applicationName) ->
```
jviotti: I believe you're missing a `.then ->` here.

lib/actions/wizard.coffee
```diff
@@ -0,0 +1,118 @@
...
+							Promise.promisifyAll(capitano).run "app create #{applicationName}".return(applicationName)
```
jviotti: You can promisify once at the top of the file and reuse the promisified functions everywhere. Also notice that promisified functions have an `Async` postfix:

```coffee
capitano = Promise.promisifyAll(require('capitano'))

...

capitano.runAsync("app create #{applicationName}")
```

jviotti: You should also put parenthesis around the command string passed to `.run()` since here you're trying to call a method `.return()` from a string:

```coffee
"app create #{applicationName}".return(applicationName)
```

lib/actions/wizard.coffee
```diff
@@ -0,0 +1,118 @@
...
+						params.name = applicationName
+						return callback()
+
+				], callback
+
+			(callback) ->
+				capitano.run("device init --application #{params.name}", callback)
+
+			(deviceName, callback) ->
+				params.device = deviceName
```
jviotti: Notice that `device init` returns a device name, however, device commands (for example `device await` a few lines below) now take uuids instead.

mkorakakis: So, i should use resin.models.device.getByName() in order to acquire the uuid, while having only the device name, and pass it to await()?

lib/actions/wizard.coffee
mkorakakis: undefined is not a function err msg when running this:
```
						form.ask
							message: 'Choose a Name for your new application'
							type: 'input'
						.then (applicationName) ->
							capitano.runAsync("app create #{applicationName}".return(applicationName))
						.nodeify(callback)
```

jviotti: You're making the same error here:

```coffee
capitano.runAsync("app create #{applicationName}".return(applicationName))
```

You're trying to call a `.return()` method of a string. Do this instead:

```coffee
capitano.runAsync("app create #{applicationName}").return(applicationName)
```

lib/actions/wizard.coffee
jviotti: @mkorakakis Yeah, that should work fine!

lib/actions/wizard.coffee
mkorakakis: Hmm, yeah i though that i should use the return parentheses in conjunction with those in the initial comment, anyway thanks!

lib/actions/wizard.coffee
mkorakakis: @jviotti just saw that multiple devices can have the same name in a app, so check the length of the returned array and do capitano.run("device await #{params.device}", callback) in async.forEach or something like that...

jviotti: @mkorakakis The correct approach will then be modifying `device init` to return a uuid instead (https://github.com/resin-io/resin-cli/blob/master/lib/actions/device.coffee#L421). Make sure all the commands that reuse device init (if any) are modified accordingly.

mkorakakis: @jviotti Yeah, but that will lead to displaying the deviceName while using the device.init and then using only the uuid in order to display device related messages in wizard.coffee. Are you OK with that?

jviotti: @mkorakakis You can do the translation from `uuid` to `name` when necessary in the quick start command, since it's much easier to go from `uuid` to `name` than from `name` to `uuid`.

PR #94, opened 2015-07-10
Upgrade Resin VCS dependencies

mkorakakis:


lib/actions/device.coffee
```diff
@@ -272,7 +272,7 @@ exports.init =
 
 			(callback) ->
 				return callback(null, options.application) if options.application?
-				vcs.getApplicationName(process.cwd(), callback)
+				vcs.getApplicationName(process.cwd(), callback).nodeify(callback)
```
jviotti: There is no need to pass the callback as an argument to the function, since it's being handled by `.nodeify()`.

lib/actions/app.coffee
```diff
@@ -175,13 +175,13 @@ exports.associate =
...
 				return done() if not confirmed
-				vcs.initialize(currentDirectory, callback)
+				vcs.initialize(currentDirectory, callback).nodeify(callback)
 
 			(callback) ->
 				resin.models.application.get(params.name, callback)
 
 			(application, callback) ->
-				vcs.addRemote(currentDirectory, application.git_repository, callback)
+				vcs.associate(currentDirectory, application.git_repository, callback).nodeify(callback)
```
jviotti: Same here, the callback should be removed from the arguments.

build/actions/device.js
```diff
@@ -192,7 +192,7 @@
           if (options.application != null) {
             return callback(null, options.application);
           }
-          return vcs.getApplicationName(process.cwd(), callback);
+          return vcs.getApplicationName(process.cwd(), callback).nodeify(callback);
```
jviotti: Callback issue here as well.

build/actions/app.js
```diff
@@ -129,11 +129,11 @@
           if (!confirmed) {
             return done();
           }
-          return vcs.initialize(currentDirectory, callback);
+          return vcs.initialize(currentDirectory, callback).nodeify(callback);
         }, function(callback) {
           return resin.models.application.get(params.name, callback);
         }, function(application, callback) {
-          return vcs.addRemote(currentDirectory, application.git_repository, callback);
+          return vcs.associate(currentDirectory, application.git_repository, callback).nodeify(callback);
```
jviotti: Callback issue here as well.

build/actions/app.js
```diff
@@ -129,11 +129,11 @@
           if (!confirmed) {
             return done();
           }
-          return vcs.initialize(currentDirectory, callback);
+          return vcs.initialize(currentDirectory, callback).nodeify(callback);
```
jviotti: Here too.

jviotti:
The commit message (and PR name) could be improved to more easily reflect the changes. Also, it's considered good practice to use an infinitive verb in the commit name ("upgrade" instead of "upgraded"), and capitalise the first work. For example:

> Upgrade Resin VCS to v2.0.0.

jviotti:
You're missing this `vcs` call in `examples.coffee`: https://github.com/resin-io/resin-cli/blob/master/lib/actions/examples.coffee#L94.

build/actions/examples.js
```diff
@@ -80,8 +80,7 @@
           return done(error);
         }
         console.info("Cloning " + example.display_name + " to " + destination);
-        vcs.clone(example.repository, destination, done);
-        return done();
```
jviotti: Thanks for deleting this one! :)

PR #95, opened 2015-07-14
Add registered_at UNIX epoch

jviotti:


lib/actions/device.coffee
```diff
@@ -324,6 +324,7 @@ exports.init =
 					# Associate a device
 					results.config.deviceId = device.id
 					results.config.uuid = device.uuid
+					results.config.registered_at = Math.floor(new Date() / 1000)
```
Page-: I'd rather use `Date.now()` to be clear we're working with an epoch value, rather than `new Date()` which, whilst it works for `/`, `-`, `*`, "fails" for `+`, eg `new Date() + 0` becomes `"Thu Jul 16 2015 01:48:53 GMT+0100 (GMT Daylight Time)0"`

jviotti: Fixed!

Page-:
LGTM

PR #110, opened 2015-07-22
Add promise support for Resin-SDK dependency

mkorakakis:


lib/actions/auth.coffee
```diff
@@ -134,11 +134,12 @@ exports.signup =
 				visuals.patterns.register(callback)
 
 			(credentials, callback) ->
-				resin.auth.register credentials, (error, token) ->
-					return callback(error, credentials)
+				resin.auth.register(credentials).then (token) ->
```
jviotti: You can use Bluebird's `.returns()` here: https://github.com/petkaantonov/bluebird/blob/master/API.md#returndynamic-value---promise.

lib/actions/device.coffee
```diff
@@ -208,16 +204,13 @@ exports.await =
...
-
+			resin.models.device.isOnline(params.name).then (isOnline) ->
 				if isOnline
 					console.info("Device became online: #{params.name}")
-					return done()
 				else
 					console.info("Polling device network status: #{params.name}")
 					setTimeout(poll, options.interval)
-
+			.nodeify(done)
```
jviotti: Putting `.nodeify()` here is not correct. The logic of this function is:

- Check if the device is online.
- If it is, exit.
- It it's not, wait for some seconds and go back to step 1.

But as it is now, the function calls `done()` after checking the device network status, without caring about it's results.

What I suggest is:

- Use `.catch(done)` in this line instead of `.nodeify(done)`, so we still react just to errors correctly, and manually call `return done()` if the device is online as it was before.

lib/actions/device.coffee
```diff
@@ -61,15 +61,12 @@ exports.info =
...
-
 		Examples:
-
 			$ resin device MyDevice
 	'''
 	permission: 'user'
 	action: (params, options, done) ->
-		resin.models.device.get params.name, (error, device) ->
-			return done(error) if error?
+		resin.models.device.get(params.name).then (device) ->
```
jviotti: Notice that now most of the device functions in the SDK take UUIDs instead of plain names. In those cases, can you rename the param to `uuid` here and in the capitano command definition, as well as modifying the command example above to use a fake uuid?

lib/actions/device.coffee
```diff
@@ -364,7 +357,7 @@ exports.init =
 				removeCallback(callback)
 
 			(callback) ->
-				resin.models.device.getByUUID(params.uuid, callback)
+				resin.models.device.getByUUID(params.uuid).nodeify(callback)
```
jviotti: This function doesn't exist in the SDK anymore. Since `resin.models.device.get()` handles a uuid now, you can use that function instead.

lib/actions/device.coffee
```diff
@@ -300,18 +293,18 @@ exports.init =
...
 
 					manifest: (callback) ->
 						console.info('Getting device manifest for the application')
-						resin.models.device.getManifestBySlug(application.device_type, callback)
+						resin.models.device.getManifestBySlug(application.device_type).nodeify(callback)
 
 					config: (callback) ->
 						console.info('Fetching application configuration')
-						resin.models.application.getConfiguration(options.application, networkOptions, callback)
+						resin.models.application.getConfiguration(options.application, networkOptions).nodeify(callback)
```
jviotti: This function doesn't exist in the SDK anymore. Can you install and reuse https://github.com/resin-io/resin-device-config instead here?

lib/actions/device.coffee
```diff
@@ -61,15 +61,12 @@ exports.info =
 	description: 'list a single device'
 	help: '''
 		Use this command to show information about a single device.
-
```
jviotti: Why are these lines being removed?

lib/actions/keys.coffee
```diff
@@ -31,17 +30,14 @@ exports.info =
 	description: 'list a single ssh key'
 	help: '''
 		Use this command to show information about a single SSH key.
-
```
jviotti: Why are these lines being removed?

lib/actions/logs.coffee
```diff
@@ -8,20 +8,13 @@ module.exports =
 	description: 'show device logs'
 	help: '''
 		Use this command to show logs for a specific device.
-
```
jviotti: Why are all these lines being removed in the help section?

lib/actions/logs.coffee
```diff
@@ -42,14 +35,13 @@ module.exports =
...
-		}, (error, message) ->
-			return done(error) if error?
+		}, (message) ->
 			if _.isArray(message)
 				_.each message, (line) ->
 					console.log(line.message)
 			else
 				console.log(message.message)
-			return done()
+	  ).nodeify(done)
```
jviotti: This should be indented one more level, to be aligned with `resin.logs.subscribe`.

lib/actions/logs.coffee
```diff
@@ -42,14 +35,13 @@ module.exports =
 	]
 	permission: 'user'
 	action: (params, options, done) ->
-		resin.logs.subscribe params.uuid, {
+		resin.logs.subscribe(params.uuid, {
```
jviotti: The parentheses are not needed here. We tend to omit them for multi line function calls.

jviotti: Notice that the logs interface completely changed, so the whole code below is no longer valid. See https://github.com/resin-io/resin-sdk/blob/master/DOCUMENTATION.md#resin.logs.subscribe. 

Also, since we have a different function to get the logs history now, the CLI behaviour should now be:

- Fetch history with `resin.logs.history()`.
- Print all messages in history.
- Subscribe to the pub nub channel in case the user included the `tail` option and print new messages in real time.

Also, we don't accept customising the history count anymore, so the `options.num` option should be removed from the capitano definition, and the `LOGS_HISTORY_COUNT` constant from the code.

lib/actions/device.coffee
```diff
@@ -208,16 +207,13 @@ exports.await =
 		options.interval ?= 3000
 
 		poll = ->
-			resin.models.device.isOnline params.name, (error, isOnline) ->
-				return done(error) if error?
-
+			resin.models.device.isOnline(params.uuid).then (isOnline) ->
 				if isOnline
-					console.info("Device became online: #{params.name}")
-					return done()
```
jviotti: Can you keep `return done()` here?

lib/actions/notes.coffee
```diff
@@ -30,4 +30,4 @@ exports.set =
 		if _.isEmpty(params.note)
 			return done(new Error('Missing note content'))
 
-		resin.models.device.note(options.device, params.note, done)
+		resin.models.device.note(options.device, params.note).nodeify(done)
```
jviotti: Can you edit the `options.device` description and the command example to use uuids as the other device commands?

lib/actions/logs.coffee
```diff
@@ -42,14 +30,18 @@ module.exports =
...
+		if options.tail?
+			resin.logs.history(params.uuid).then (lines) ->
+				for line in lines
+					console.log(line.message)
+			resin.logs.subscribe(params.uuid).then (logs) ->
+				logs.on 'line', (line) ->
+					console.log(line.message)
+				logs.on 'error', (error) ->
+					throw error
+	  .nodeify(done)
```
jviotti: There are indentation issues here.

lib/actions/logs.coffee
```diff
@@ -42,14 +30,18 @@ module.exports =
...
-			if _.isArray(message)
-				_.each message, (line) ->
+		if options.tail?
+			resin.logs.history(params.uuid).then (lines) ->
+				for line in lines
+					console.log(line.message)
+			resin.logs.subscribe(params.uuid).then (logs) ->
+				logs.on 'line', (line) ->
+					console.log(line.message)
+				logs.on 'error', (error) ->
```
jviotti: Can you pass `done` as a callback instead of throwing the error directly?

lib/actions/logs.coffee
```diff
@@ -42,14 +30,18 @@ module.exports =
...
+					throw error
+	  .nodeify(done)
+		else
+			resin.logs.history(params.uuid).then (lines) ->
```
jviotti: You're duplicating this logic from above. Since history is fetched independently on the state of the tail option, you can fetch the history outside the conditional block to remove the duplication.

lib/actions/device.coffee
mkorakakis: along with the console.info()?

lib/actions/logs.coffee
```diff
@@ -42,14 +30,18 @@ module.exports =
...
-		resin.logs.subscribe params.uuid, {
-			history: options.num or LOGS_HISTORY_COUNT
-			tail: options.tail
-		}, (error, message) ->
-			return done(error) if error?
-			if _.isArray(message)
-				_.each message, (line) ->
+		if options.tail?
+			resin.logs.history(params.uuid).then (lines) ->
+				for line in lines
```
jviotti: I'm thinking we could use Bluebird's `.each()` here to make it even more compact: https://github.com/petkaantonov/bluebird/blob/master/API.md#eachfunction-iterator---promise.

lib/actions/logs.coffee
mkorakakis: Sure, but then do i keep the throw error line or not?

jviotti: Remove `throw error` as well, since Capitano takes care of error handling if it's passed an Error instance to `done`.

lib/actions/device.coffee
```diff
@@ -208,16 +207,14 @@ exports.await =
 		options.interval ?= 3000
 
 		poll = ->
-			resin.models.device.isOnline params.name, (error, isOnline) ->
-				return done(error) if error?
-
+			resin.models.device.isOnline(params.uuid).then (isOnline) ->
```
Page-: This can be:
```coffee
		poll = ->
			resin.models.device.isOnline(params.uuid).then (isOnline) ->
				if isOnline
					console.info("Device became online: #{params.uuid}")
					return
				else
					console.info("Polling device network status: #{params.uuid}")
					return Promise.delay(options.interval).then(poll)
		poll().nodeify(done)
```

mkorakakis: done!

jviotti: Nice!

lib/actions/logs.coffee
```diff
@@ -42,14 +30,15 @@ module.exports =
...
+			return promise.nodeify(done)
+
+		promise.then ->
+			resin.logs.subscribe(params.uuid).then (logs) ->
+				logs.on 'line', (line) ->
 					console.log(line.message)
-			else
-				console.log(message.message)
-			return done()
+				logs.on 'error', (error,done) ->
```
jviotti: This should be:

```coffee
logs.on('error', done)
```

lib/actions/device.coffee
```diff
@@ -175,13 +174,12 @@ exports.supported =
 	'''
 	permission: 'user'
 	action: (params, options, done) ->
-		resin.models.device.getSupportedDeviceTypes (error, devices) ->
-			return done(error) if error?
+		resin.models.device.getSupportedDeviceTypes().then (devices) ->
 			_.each(devices, _.unary(console.log))
```
jviotti: Can we use Bluebird's `.each()` here like in the logs to avoid using `_.unary`?

mkorakakis: Sure, changing it now!

lib/actions/logs.coffee
```diff
@@ -42,14 +30,15 @@ module.exports =
...
+			return promise.nodeify(done)
+
+		promise.then ->
+			resin.logs.subscribe(params.uuid).then (logs) ->
+				logs.on 'line', (line) ->
 					console.log(line.message)
-			else
-				console.log(message.message)
-			return done()
+				logs.on('error', done) ->
```
jviotti: There shouldn't be a `->` here, just:

```coffee
logs.on('error', done)
```

The point is that when an error happens, `done` is called, so we pass the function directly as an argument.

This is a shortened version of:

```coffee
logs.on 'error', (error) ->
    done(error)
```

mkorakakis: fixed!

PR #118, opened 2015-07-27
Integrate resin-cli-form

mkorakakis:


lib/actions/app.coffee
```diff
@@ -44,7 +45,13 @@ exports.create =
 					return callback(new Error('You already have an application with that name!'))
 
 				return callback(null, options.type) if options.type?
-				visuals.patterns.selectDeviceType(callback)
+				resin.models.device.getSupportedDeviceTypes().then (supportedDeviceTypes) ->
+					form.ask
+						message: 'Device Type'
+						name: 'deviceType'
```
jviotti: `form.ask()` has a default value for a `name` (https://github.com/resin-io/resin-cli-form/blob/master/lib/form.coffee#L79), so it's not necessary here. Can you omit a explicit `name` property in all single `form.ask()` invocations for simplicity?

lib/actions/app.coffee
```diff
@@ -44,7 +45,13 @@ exports.create =
...
+						type: 'list'
+						choices: supportedDeviceTypes
+					.nodeify(callback)
```
jviotti: In this case, if `resin.models.device.getSupportedDeviceTypes()` happens to throw an error, you will miss it, because `.nodeify()` only applies to `form.ask()`. You can move `.nodeify()` to match `resin.models.device.getSupportedDeviceTypes()` instead.

lib/actions/app.coffee
```diff
@@ -134,9 +141,24 @@ exports.remove =
 	options: [ commandOptions.yes ]
 	permission: 'user'
 	action: (params, options, done) ->
-		visuals.patterns.remove 'application', options.yes, (callback) ->
-			resin.models.application.remove(params.name).nodeify(callback)
+		async.waterfall([
```
jviotti: There shouldn't be parenthesis surrounding this function call.

lib/actions/app.coffee
```diff
@@ -134,9 +141,24 @@ exports.remove =
...
+						message: 'Are you sure you want to delete the application?'
+						name: 'confirm'
+						type: 'confirm'
+						default: false
+					.nodeify(callback)
+
+			(confirmed, callback) ->
+				return callback() if not confirmed
+				resin.models.application.remove(params.name).nodeify(callback)
 		, done
```
jviotti: This `, done` here makes no sense. It should be the second argument to the `async.waterfall()` call instead.

lib/actions/app.coffee
```diff
@@ -169,7 +191,15 @@ exports.associate =
...
-				visuals.patterns.confirm(options.yes, message, callback)
+				if options.yes
+					return callback(null, true)
+				else
+					form.ask
+						message: message
+						name: 'confirm'
+						type: 'confirm'
+						default: false
+			    .nodeify(callback)
```
jviotti: Is this `.nodeify()` intended to match `form.ask()`? If so, there are indentation issues here.

lib/actions/app.coffee
```diff
@@ -169,7 +191,15 @@ exports.associate =
 					return callback(new Error("Invalid application: #{params.name}"))
 
 				message = "Are you sure you want to associate #{currentDirectory} with #{params.name}?"
-				visuals.patterns.confirm(options.yes, message, callback)
+				if options.yes
+					return callback(null, true)
+				else
+					form.ask
+						message: message
+						name: 'confirm'
```
jviotti: You can omit `name` here.

lib/actions/device.coffee
```diff
@@ -110,9 +111,24 @@ exports.remove =
...
-			resin.models.device.remove(params.uuid).nodeify(callback)
+		async.waterfall([
+
+			(callback) ->
+				if options.yes
+					return callback(null, true)
+				else
+					form.ask
+						message: 'Are you sure you want to delete the device?'
+						name: 'confirm'
```
jviotti: You can omit `name` here.

lib/actions/device.coffee
```diff
@@ -278,20 +294,62 @@ exports.init =
...
+					async.reject drives, drivelist.isSystem, (removableDrives) ->
+
+						if _.isEmpty(removableDrives)
+							return callback(new Error('No available drives'))
+
+						form.ask
+							message: 'Drive'
+							name: 'drive'
+							type: 'list'
+							default: _.map removableDrives, (item) ->
```
jviotti: This means that the default selection for a drive is a list of drives, which makes no sense. This property should be `choices` instead.

lib/actions/device.coffee
```diff
@@ -278,20 +294,62 @@ exports.init =
...
 			(device, callback) ->
 				params.device = device
 				message = "This will completely erase #{params.device}. Are you sure you want to continue?"
-				visuals.patterns.confirm(options.yes, message, callback)
+				if options.yes
+					return callback(null, true)
+				else
+					form.ask
+						message: message
+						name: 'confirm'
```
jviotti: `name` is not necessary here.

lib/actions/device.coffee
```diff
@@ -278,20 +294,62 @@ exports.init =
...
+					type: 'list'
+					choices: ['ethernet', 'wifi']
+				,
+					message: 'Wifi Ssid'
+					name: 'wifiSsid'
+					type: 'input'
+					when:
+						network: 'wifi'
+				,
+				  message: 'Wifi Key'
```
jviotti: There are indentation issues here.

lib/actions/device.coffee
```diff
@@ -278,20 +294,62 @@ exports.init =
...
 			(confirmed, callback) ->
 				return done() if not confirmed
 				return callback() if networkOptions.network?
-				visuals.patterns.selectNetworkParameters (error, parameters) ->
-					return callback(error) if error?
+				form.run [
+					message: 'Network Type'
+					name: 'network'
+					type: 'list'
+					choices: ['ethernet', 'wifi']
```
jviotti: There should be a space between a bracket and a quote.

lib/actions/environment-variables.coffee
```diff
@@ -85,12 +85,27 @@ exports.remove =
...
-				resin.models.environmentVariables.remove(params.id).nodeify(callback)
+		async.waterfall([
+
+			(callback) ->
+				if options.yes
+					return callback(null, true)
+				else
+					form.ask
+						message: 'Are you sure you want to delete the environment variable?'
+						name: 'confirm'
```
jviotti: `name` is not necessary here.

lib/actions/environment-variables.coffee
```diff
@@ -85,12 +85,27 @@ exports.remove =
 	]
 	permission: 'user'
 	action: (params, options, done) ->
-		visuals.patterns.remove 'environment variable', options.yes, (callback) ->
-			if options.device
-				resin.models.environmentVariables.device.remove(params.id).nodeify(callback)
-			else
-				resin.models.environmentVariables.remove(params.id).nodeify(callback)
+		async.waterfall([
```
jviotti: We don't use parenthesis for multiple lines function calls.

lib/actions/environment-variables.coffee
```diff
@@ -85,12 +85,27 @@ exports.remove =
...
+						default: false
+					.nodeify(callback)
+
+			(confirmed, callback) ->
+				return callback() if not confirmed
+				if options.device
+					resin.models.environmentVariables.device.remove(params.id).nodeify(callback)
+				else
+					resin.models.environmentVariables.remove(params.id).nodeify(callback)
 		, done
```
jviotti: This should be the second argument of `async.waterfall`.

lib/actions/keys.coffee
```diff
@@ -58,9 +59,24 @@ exports.remove =
...
-			resin.models.key.remove(params.id).nodeify(callback)
+		async.waterfall([
+
+			(callback) ->
+				if options.yes
+					return callback(null, true)
+				else
+					form.ask
+						message: 'Are you sure you want to delete the key?'
+						name: 'confirm'
```
jviotti: `name` is not necessary here.

lib/actions/keys.coffee
```diff
@@ -58,9 +59,24 @@ exports.remove =
...
+						type: 'confirm'
+						default: false
+					.nodeify(callback)
+
+			(confirmed, callback) ->
+				return callback() if not confirmed
+				resin.models.key.remove(params.id).nodeify(callback)
 		, done
```
jviotti: Same `done` issue as above.

lib/actions/keys.coffee
```diff
@@ -58,9 +59,24 @@ exports.remove =
 	options: [ commandOptions.yes ]
 	permission: 'user'
 	action: (params, options, done) ->
-		visuals.patterns.remove 'key', options.yes, (callback) ->
-			resin.models.key.remove(params.id).nodeify(callback)
+		async.waterfall([
```
jviotti: Parenthesis are not needed here.

lib/actions/plugin.coffee
```diff
@@ -86,9 +88,24 @@ exports.remove =
 	options: [ commandOptions.yes ]
 	permission: 'user'
 	action: (params, options, done) ->
-		visuals.patterns.remove 'plugin', options.yes, (callback) ->
-			plugins.remove(params.name, callback)
+		async.waterfall([
```
jviotti: Parenthesis are not needed here.

lib/actions/plugin.coffee
```diff
@@ -86,9 +88,24 @@ exports.remove =
...
+
+			(callback) ->
+				if options.yes
+					return callback(null, true)
+				else
+					form.ask
+						message: 'Are you sure you want to delete the plugin?'
+						name: 'confirm'
```
jviotti: `name` is not needed here.

lib/actions/app.coffee
mkorakakis: Is it possible to use as an alternative .catch(callback)?

lib/actions/app.coffee
mkorakakis: Yeah, regarding the indentation issues. Weird thing is that in Atom everything is seems fine.

lib/actions/app.coffee
mkorakakis: Yeah, missed it when i replaced the function call with the actual code. Though that it was attached to resin.models.device.remove().

lib/actions/app.coffee
mkorakakis: https://github.com/resin-io/resin-cli/blob/master/lib/actions/device.coffee#L266

lib/actions/app.coffee
jviotti: @mkorakakis They can *seem* the same from a visual standpoint. I suggest you to enable "Show Invisibles" in Atom settings: https://gist.github.com/timrwood/9242517#gistcomment-1181001, so you can spot the differences in the future.

lib/actions/app.coffee
jviotti: It'd be possible, but that means changing two places instead of one. By moving `.nodeify()` to the outer function, you only use it one and keep everything inside a promise chain.

lib/actions/app.coffee
jviotti: Hm, fair enough, not sure why I made that :/ Keep them if you like, we'll get rid of `async` anyway later on.

lib/actions/app.coffee
mkorakakis: Thanks! Really useful.

lib/actions/app.coffee
mkorakakis: Nope, its OK! I removed them completely.

PR #126, opened 2015-07-29
Fix #114

### LINKED ISSUE #114: Check if app exists before removing it "resin app rm <appName>"
Check if the input is valid i.e. the specific app actually exists/

mkorakakis:


jviotti:
As mentioned in https://github.com/resin-io/resin-cli/pull/127, let's move this checking to the SDK!

PR #127, opened 2015-07-29
Fix #123

### LINKED ISSUE #123: Check if device exists before removing it "resin device rm <uuid>"
The command doesn't check if the device actually exists before removing it.

mkorakakis:


lib/actions/device.coffee
```diff
@@ -126,7 +126,9 @@ exports.remove =
 
 			(confirmed, callback) ->
 				return callback() if not confirmed
-				resin.models.device.remove(params.uuid).nodeify(callback)
+				resin.models.device.get(params.uuid).then (device) ->
```
jviotti: Since we're not using the result of the `.get()`, function, can you omit the `(device)` argument?

jviotti:
I'm actually thinking that requiring clients to check that the resource exists before deleting is crazy. Maybe we can apply this fix directly into the SDK instead?

PR #129, opened 2015-07-30
Add email address to the returned information, when using whoami(). Fix #73.

### LINKED ISSUE #73: can whoami return email address aswell?
It maybe be nice if the `resin whoami` returns `username<my_mail@email.com>` if the email is available, this just helps to better identify the account, especially if the user has more than 1 account.

mkorakakis:
[Issue #73](https://github.com/resin-io/resin-cli/issues/73).

### DIFF
--- a/lib/actions/auth.coffee
+++ b/lib/actions/auth.coffee
@@ -177,5 +177,8 @@ exports.whoami =
 		resin.auth.whoami().then (username) ->
 			if not username?
 				throw new Error('Username not found')
-			console.log(username)
+			resin.auth.getEmail().then (email) ->
+				if not email?
+					throw new Error('Email not found')
+				console.log "Username: #{username}" + '\n' + "Email: #{email}"
 		.nodeify(done)
--- a/lib/actions/auth.coffee
+++ b/lib/actions/auth.coffee
@@ -5,6 +5,7 @@ async = require('async')
 resin = require('resin-sdk')
 settings = require('resin-settings-client')
 form = require('resin-cli-form')
+visuals = require('resin-cli-visuals')
 
 TOKEN_URL = url.resolve(settings.get('dashboardUrl'), '/preferences')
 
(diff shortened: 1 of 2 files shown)


lib/actions/auth.coffee
```diff
@@ -177,5 +177,11 @@ exports.whoami =
 		resin.auth.whoami().then (username) ->
 			if not username?
 				throw new Error('Username not found')
-			console.log(username)
+			resin.auth.getEmail().then (email) ->
+				if not email?
+					throw new Error('Email not found')
+				console.log """
+					Username: #{username}
```
jviotti: Can we use a vertical table here for consistency?

lib/actions/auth.coffee
```diff
@@ -177,5 +178,15 @@ exports.whoami =
 		resin.auth.whoami().then (username) ->
 			if not username?
 				throw new Error('Username not found')
-			console.log(username)
+			resin.auth.getEmail().then (email) ->
+				if not email?
```
jviotti: I don't think this check is necessary. I believe a user always has an email, and even if no, `visuals`'s table will add a blank space at the side of the EMAIL property, making it clear that there is no email.

Page-: I think certain social accounts (twitter?) don't always have emails associated

jviotti: @Page- True, but in that case the user would get:

```
== ACCOUNT INFORMATION
USERNAME: jviotti
EMAIL:                   
```

lib/actions/auth.coffee
```diff
@@ -177,5 +178,15 @@ exports.whoami =
...
+					throw new Error('Email not found')
+				console.log visuals.table.vertical
+					username: username
```
jviotti: You can use `{ username, email }` in this case.

lib/actions/auth.coffee
```diff
@@ -177,5 +178,10 @@ exports.whoami =
 		resin.auth.whoami().then (username) ->
 			if not username?
 				throw new Error('Username not found')
-			console.log(username)
+			resin.auth.getEmail().then (email) ->
+				console.log visuals.table.vertical {username, email}, [
```
jviotti: Please put spaces around the curly braces:

```coffee
{ username, email }
```

PR #153, opened 2015-08-14
Use `env_var_name` as the name property for devices env vars. Fix #120

### LINKED ISSUE #120: Command does not display valid output "resin envs --application/device"
Doesn't return anything when using the --application and in the case of --device doesn't display the variable's name.

```
C:\Users\korakas\Documents\GitHub\resin-cli>node ./bin/resin envs --application demo
Request error: Internal Server Error

C:\Users\korakas\Documents\GitHub\resin-cli>node ./bin/resin envs --device 8d5eb924251b5ac05913e4a1d
46396b832e62b8e76733df78ed607cc5a50b9
ID  NAME VALUE
652      5555
```

jviotti:
Application environment variables contain a property called `name`,
which is found as `env_var_name` on device environment variables
instead.

This fix maps the device environment variables to assign `name` to
`env_var_name` as a workaround, until the inconsistency is fixed in the
infrastructure side.

Fixes:

- https://github.com/resin-io/resin-cli/issues/120

### DIFF
--- a/build/actions/environment-variables.js
+++ b/build/actions/environment-variables.js
@@ -30,7 +30,10 @@
           if (options.application != null) {
             return resin.models.environmentVariables.getAllByApplication(options.application).nodeify(callback);
           } else if (options.device != null) {
-            return resin.models.environmentVariables.device.getAll(options.device).nodeify(callback);
+            return resin.models.environmentVariables.device.getAll(options.device).map(function(envVar) {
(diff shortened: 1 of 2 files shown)


jviotti:
I'm thinking this will be better fixed in the SDK instead, to keep the workaround close to the model. I'll close this pull request and implement there.

PR #167, opened 2015-08-18
Display msg when app/device does not have env variables. Fix #166.

### LINKED ISSUE #166: Error message instead of empty table for environment variables list command
If trying to get the environment variables for an application or device that doesn't have any custom ones, I  get an empty table:

```sh
$ ./bin/resin envs -a MuSync
ID NAME VALUE
```

Instead, we should print an error message explaining to the user that the selected resource doesn't have any environment variable.

For example:

```sh
$ ./bin/resin envs -a MuSync
No environment variables found
```

mkorakakis:
[Issue #166](https://github.com/resin-io/resin-cli/issues/166)

### DIFF
--- a/lib/actions/environment-variables.coffee
+++ b/lib/actions/environment-variables.coffee
@@ -44,6 +44,8 @@ exports.list =
 				throw new Error('You must specify an application or device')
 
 		.tap (environmentVariables) ->
+			if _.isEmpty(environmentVariables)
+				throw new Error('No environment variables found')
 			if not options.verbose
 				isSystemVariable = resin.models.environmentVariables.isSystemVariable
 				environmentVariables = _.reject(environmentVariables, isSystemVariable)
--- a/build/actions/environment-variables.js
+++ b/build/actions/environment-variables.js
@@ -37,6 +37,9 @@
         }
       }).tap(function(environmentVariables) {
         var isSystemVariable;
+        if (_.isEmpty(environmentVariables)) {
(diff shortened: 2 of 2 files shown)


lib/actions/environment-variables.coffee
```diff
@@ -44,15 +44,18 @@ exports.list =
...
-			if not options.verbose
-				isSystemVariable = resin.models.environmentVariables.isSystemVariable
-				environmentVariables = _.reject(environmentVariables, isSystemVariable)
-
-			console.log visuals.table.horizontal environmentVariables, [
-				'id'
-				'name'
-				'value'
-			]
+			if environmentVariables.length > 0
```
jviotti: A better way to formulate this is to check if the environment variables array is empty, and if so, throw an error right away. This also reduces indentation from the whole table printing block, and removes the need of an `else` block.

lib/actions/environment-variables.coffee
```diff
@@ -44,6 +44,8 @@ exports.list =
 				throw new Error('You must specify an application or device')
 
 		.tap (environmentVariables) ->
+			if environmentVariables.length == 0
```
jviotti: Using `_.isEmpty()` would convey the intention clearer.

jviotti: Just for future reference, we tend to use `is` instead of `==`.

jviotti:
LGTM

PR #175, opened 2015-08-20
Don't check for available updates when running as root

jviotti:
`update-notifier` persist its update check results in a file, which is
then read when running again the application.

If this file gets written when the application is being run as root, we
get ugly EPERM issues.


emirotin:
lgtm

PR #198, opened 2015-09-28
Use Promise.disposer() to make sure temp files are deleted

jviotti:


jviotti:
Can someone try provisioning a device with this branch?

PR #199, opened 2015-09-29
Prompt for select application if running device init with no arguments

jviotti:
Currently, if `device init` was ran without an application argument, we
attempted to get the application name from the current directory, given
it was a git repository.

This approach led to confusions from time to time, so now we prompt the
user to select one of it's own applications from a dropdown instead of
checking the current directory in this edge case.

Fixes: https://github.com/resin-io/resin-cli/issues/197


jviotti:
The CLI reference documentation needs to be regenerated first.

PR #231, opened 2015-10-15
Elevate permission in Windows

jviotti:
Fixes: https://github.com/resin-io/resin-cli/issues/222


jviotti:
This PR **cannot be merged** until https://github.com/SBoudrias/Inquirer.js/pull/297 is released to NPM, otherwise `os initialise` will break in Windows.

jviotti:
The inquirer commit was released to NPM, however this still needs some experimentation before merging (seems to still be breaking in some cases).

PR #262, opened 2015-11-12
Handle Windows permissions elevation automatically

jviotti:


jviotti:
Confirmed it works on Windows 10!

PR #297, opened 2016-01-18
Ensure env var values are strings

pcarranzav:
Otherwise, doing:
```
resin env add KEY 1 --application appname
```
fails with the message:
```
ResinRequestError: Request error: "value" is not a string: 1
```

### DIFF
--- a/lib/actions/environment-variables.coffee
+++ b/lib/actions/environment-variables.coffee
@@ -152,11 +152,11 @@ exports.add =
 					console.info("Warning: using #{params.key}=#{params.value} from host environment")
 
 			if options.application?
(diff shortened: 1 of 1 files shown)


jviotti:
@pcarranzav Nice catch! I think this should be implemented in the Resin SDK directly (in both environment variables `create` functions).

jviotti:
@pcarranzav I moved this fix to the SDK myself: https://github.com/resin-io/resin-sdk/pull/173, thanks for the catch!

pcarranzav:
Cheers! (It  was actually @shaunmulligan who found it :)

PR #299, opened 2016-01-21
Remove tests tasks and CI configuration

jviotti:
We don't have any tests in this repository


jviotti:
The tests fails because we didn't remove the project from Appveyor and Travis before deleting their configuration files.

PR #364, opened 2016-07-07
resin sync: filter out offline devices in interactive choosing dialog

### LINKED ISSUE #363: resin sync: filter offline devices in drop-down menu

lekkas:
Closes #363

### DIFF
--- a/lib/utils/patterns.coffee
+++ b/lib/utils/patterns.coffee
@@ -155,6 +155,8 @@ exports.inferOrSelectDevice = (applicationName) ->
 		if applicationName?
 			return resin.models.device.getAllByApplication(applicationName)
 		return resin.models.device.getAll()
+	.filter (device) ->
+		device.is_online
 	.then (devices) ->
 		if _.isEmpty(devices)
 			throw new Error('You don\'t have any devices')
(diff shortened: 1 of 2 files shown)


lib/utils/patterns.coffee
```diff
@@ -155,6 +155,8 @@ exports.inferOrSelectDevice = (applicationName) ->
 		if applicationName?
 			return resin.models.device.getAllByApplication(applicationName)
 		return resin.models.device.getAll()
+	.filter (device) ->
+		resin.models.device.isOnline(device.uuid)
```
jviotti: This would cause a new HTTP request to the API for every device on the list, which is not very efficient. Notice that you already have information about the device online state in the object array returned by `resin.models.device.getAll()` or `resin.models.device.getAllByApplication()`, which I believe is an integer `is_online` property.

lekkas: Very good point @jviotti - updated PR

jviotti:
Looks great, I'll merge it. Do you need this to be published ASAP? I saw you created a couple of CLI tickets, so happy to wait so we can batch them together into one release.

lekkas:
Thank you! No need to publish this ASAP, but merging would be great so that I can rebase work on top of it

lekkas:
Just saw it's merged, thanks again Juan!

PR #367, opened 2016-07-21
resin sync/ssh improvements

lekkas:
## New features
* Automatically **parse '.gitignore'** for file inclusions/exclusions from resin sync by default. Skip parsing with `--skip-gitignore`
* **Automatically save options** to `<sourceDirectory>/.resin-sync.yml` after every run
* Support **user-specified destination directories** with `--destination/-d` option
* Implement `--after` option to perform actions local (e.g. cleanup) after resin sync has finished
* Interactive dialog for destination directory, with `/usr/src/app` being the default choice

## Changes
* `resin sync` **`--source/-s`** option is mandatory if a `.resin-sync.yml` file is not found in the current directory
* `resin sync/ssh` now only accept `uuid` as an argument (`appName` has been deprecated)
* Always display interactive device selection dialog when uuid is not passed as an argument

## Fixes
* Disable ControlMaster ssh option (as reported in support)

### DIFF
--- a/lib/actions/sync.coffee
+++ b/lib/actions/sync.coffee
@@ -58,6 +58,7 @@ module.exports =
 			uuid: 7cf02a6
 			destination: '/usr/src/app'
 			before: 'echo Hello'
+			after: 'echo Done'
 			ignore:
 				- .git
 				- node_modules/
@@ -71,7 +72,7 @@ module.exports =
 		Examples:
 
 			$ resin sync 7cf02a6 --source '.' --destination '/usr/src/app'
-			$ resin sync 7cf02a6 -s '/home/user/myResinProject' -d '/usr/src/app' --before 'echo Hello'
+			$ resin sync 7cf02a6 -s '/home/user/myResinProject' -d '/usr/src/app' --before 'echo Hello' --after 'echo Done'
 			$ resin sync --ignore 'lib/'
 			$ resin sync --verbose false
 			$ resin sync
@@ -102,6 +103,11 @@ module.exports =
 			parameter: 'command'
 			description: 'execute a command before syncing'
 			alias: 'b'
+		,
+			signature: 'after'
+			parameter: 'command'
+			description: 'execute a command after syncing'
+			alias: 'a'
 		,
 			signature: 'port'
 			parameter: 'port'
--- a/lib/actions/sync.coffee
+++ b/lib/actions/sync.coffee
@@ -15,14 +15,12 @@ limitations under the License.
 ###
 
 module.exports =
-	signature: 'sync [destination]'
+	signature: 'sync [uuid]'
 	description: '(beta) sync your changes with a device'
 	help: '''
 		WARNING: If you're running Windows, this command only supports `cmd.exe`.
 
 		Use this command to sync your local changes to a certain device on the fly.
-
-		The `destination` argument can be either a device uuid or an application name.
 
 		You can save all the options mentioned below in a `resin-sync.yml` file,
 		by using the same option names as keys. For example:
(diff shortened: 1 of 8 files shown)


lib/actions/sync.coffee
```diff
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+
```
jviotti: I believe this was accidentally added.

lib/utils/helpers.coffee
```diff
@@ -52,3 +55,24 @@ exports.sudo = (command, message) ->
...
 
 	return president.executeAsync(command)
+
+# Loads '.resin-sync.yml' configuration from 'source' directory.
+# Returns the configuration object on success
+#
+# TODO: Use 'config.load()' method from `resin sync` when resin sync gets
+# integrated into resin CLI
+exports.loadConfig = (source) ->
+	configPath = path.join(source, '.resin-sync.yml')
```
jviotti: We shouldn't hardcode `.resin-sync.yml` here if we call the function `loadConfig()`, since that implies `.resin-sync.yml` is the configuration for the CLI as a whole. I propose either:

- Accepting the configuration file name as an argument.
- Changing the name to `loadSyncConfig()`.
- Moving this function to a private function in `sync.coffee`.

I personally prefer the third one.

lib/actions/sync.coffee
```diff
@@ -79,6 +91,8 @@ module.exports =
 		,
 	]
 	action: (params, options, done) ->
+		fs = require('fs')
```
jviotti: Are these imports required?

jviotti:
I didn't check in detail, but I believe some of your changes are missing the JS counter-part, but ignore me if I'm wrong. Also, you can run `npm run doc` to auto generate the documentation at `doc/cli.markdown`; otherwise it'll keep referring to the old command helps.

lib/actions/sync.coffee
lekkas: I generally run `resin-lint` to catch unused variables - we use all of the imported modules later on.

lib/utils/helpers.coffee
lekkas: Good point @jviotti , moved `loadConfig()` as a private function in `sync.coffee.

lib/actions/sync.coffee
lekkas: Yeap, fixed

lekkas:
Also added docs (thanks for reminding me, maybe we can add this in the `gulp build` process) and regenerated JS files. I've added the latter in a separate commit at the end, because having them along with the 'legit' coffeescript changes is makes life harder when reviewing the diffs.

lekkas:
@jviotti I forgot an important detail - please do not merge/publish this until we have the respective backend changes on production - I will ping you when it's up!

jviotti:
Yeah, transpiling is a PITA, I'd love to move away from CoffeeScript. Sure, let me know when everything is ready!

lekkas:
Thank you!

Tbh I like the syntactic sugar of CoffeeScript , but I agree that leaving it is probably the most sensible choice for the long term!

alexandrosm:
We might move from CS to TS -- so nobody gets what they want :P

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

jviotti:
@alexandrosm In any case, should we discuss a way to deal with transpilation in a way that doesn't annoy developers nor users? The current situation is awful. Executing CS or TS directly is an option for internal code, but not for open-source packages we publish to NPM. Adding a pre-publish hook to compile stuff led me to some issues in the past which made me chose to compile and commit the JS myself.

jviotti:
Just a comment: I'd love to make sure we get decent linting tools for whatever language we chose. ESLint for JS is magnitudes of times better than things like CoffeeLint. It makes a HUGE difference for me in other projects.

alexandrosm:
How does TSLint look on that spectrum?

https://palantir.github.io/tslint/

btw, I love their "develop" link. hint hint for Etcher :)

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lekkas:
@alexandrosm +1 for TS! 😆 

@jviotti What were the issues with adding pre-publishing hooks? 

I agree it'd be better to have transpilation automated. The current situation is a nuisance for the PR submitter and also doubles the review workload for the maintainer - because PR'ed transpiled code *must* be reviewed (or at least verified) as well

jviotti:
@lekkas Some I remember from the top of my head:

- I encountered some situations where the CoffeeScript would not compile correctly. The code would compile, but the intention was lost given some edge cases and weird CS usage, and I only noticed after publishing the module. I do a quick sanity review on the compiled JS since then.

- A combination of `.gitignore` and `.npmignore` subtle issues caused the CS to be compiled in the pre-publish hook, but not be included in the final package. It was hard to diagnose what was going on afterwards.

I believe there were other issues/annoyances, but I can't remember anymore.

lekkas:
I see, I'm thinking of adding a prepublish hook in `resin sync`, so I'll take note of these issues, thanks!

lekkas:
@jviotti this is ready for merging 👍

PR #384, opened 2016-08-09
Implement `device enableDeviceUrl/disableDeviceUrl/hasDeviceUrl/getDeviceUrl`

### LINKED ISSUE #383: There is access to "deviceURL" configurations from resin-cli

lifeeth:
Closes #383


lib/actions/device.coffee
```diff
@@ -186,6 +186,70 @@ exports.reboot =
 		resin = require('resin-sdk')
 		resin.models.device.reboot(params.uuid).nodeify(done)
 
+exports.enableDeviceUrl =
+	signature: 'device enableDeviceUrl <uuid>'
```
jviotti: Maybe we can use more human-like command names to be consistent with the other commands? Like `device enable url <uuid>`? Notice you can freely use multi-words commands.

Also, the device in `deviceUrl` is probably redundant with the initial `device` command namespace?

dfunckt: If we do change the command name, I think it's important to convey it's _public_. In the Dashboard and [docs](http://docs.resin.io/management/devices/#enable-public-device-url) we consistently call the feature "Public Device URL", which could probably map to `device public-url enable` and `device public-url disable` nicely (I'm unfamiliar with CLI's naming conventions but I think you get the picture).

lifeeth: That sounds good - Let me do that.

jviotti:
Looks great, I'll merge in a second. @lifeeth Do you need me to publish a new version ASAP, or can this wait to be chunked with other changes?

lifeeth:
@jviotti the sooner the better as we have a use for enabling the public url in testing.

jviotti:
@lifeeth I've published v4.3.0 to npm, which includes the device public url commands.

PR #404, opened 2016-10-26
Add a device shutdown command, and allow forcing reboot and shutdown

### LINKED ISSUE #401: Add a device shutdown command
The sdk does have a `resin.models.device.shutdown(uuid)`, but I couldn't find a `resin device shutdown` action here.

### LINKED ISSUE #402: Allow forcing a device reboot (and shutdown)
With resin-io/resin-supervisor#138, we'll make the supervisor stop the apps before rebooting or shutting down, which means that an update lock will prevent the action from happening and the endpoint will return a 423 status.
So the API endpoints for these actions now take a `force` option just like the `update` endpoint.
The device  reboot action should be able to take and pass this force option down to the SDK (which should also be modified to accept this option). If shutdown is implemented as per #401, then it should accept the option too.

pcarranzav:
Depends on resin-io/resin-sdk#229
(I will add a commit updating the SDK when that is merged)

Closes #401 
Closes #402

### DIFF
--- a/build/actions/command-options.js
+++ b/build/actions/command-options.js
@@ -73,4 +73,12 @@ limitations under the License.
     alias: 'k'
   };
 
+  exports.forceUpdateLock = {
+    signature: 'force',
+    parameter: 'force',
+    description: 'force action if the update lock is set',
+    boolean: true,
+    alias: 'f'
+  };
+
 }).call(this);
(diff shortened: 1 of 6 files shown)


jviotti:
Looks great! Just a minor comment: can you re-generate the documentation by running `npm run doc`?

pcarranzav:
@jviotti done, will merge after CI tests pass unless you object

jviotti:
This repository has no tests (only a linter, since its a module that just ties logic from other components), so feel free to merge right away.

pcarranzav:
Done, cheers!

jviotti:
@pcarranzav Do you want me to publish now?

pcarranzav:
Sounds good, the force option will have no effect until the newer supervisor is used, but there'll also be no need to use it.

jviotti:
I just published v5.2.0!

pcarranzav:
Thanks  @jviotti !

PR #405, opened 2016-10-28
Fix `Boolean options can't have parameters` error

jviotti:
This error was introduced as part of
`9cf42462c029e038e09efc961736946be8bfcb9b`, since the `forceUpdateLock`
option being used in the `reboot` command contains a `parameter`
property despite being declared a boolean.

Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

### DIFF
--- a/build/actions/command-options.js
+++ b/build/actions/command-options.js
@@ -75,7 +75,6 @@ limitations under the License.
 
   exports.forceUpdateLock = {
     signature: 'force',
-    parameter: 'force',
     description: 'force action if the update lock is set',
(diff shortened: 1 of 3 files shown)


pcarranzav:
Oops, sorry for that.
BTW, @jviotti, looks like you might have compiled with an older coffee-script? Otherwise I don't see why sync.js should change.

jviotti:
Oh, true. I'll create another PR re-building the CoffeeScript with the latest version.

PR #407, opened 2016-11-01
Actually add the shutdown command

pcarranzav:
When I did #404 I didn't realize I had to add it to the capitano app commands.


jviotti:
@pcarranzav Awesome, do you want me to publish a new version now?

PR #416, opened 2017-01-18
Support aufs resin sync

### LINKED ISSUE #415: Support resin sync on AUFS devices
Connects to https://github.com/resin-io/hq/pull/505

lekkas:
Must be merged after https://github.com/resin-io-modules/resin-sync/pull/100

Closes #415


lib/actions/ssh.coffee
```diff
@@ -92,7 +92,7 @@ module.exports =
 			Promise.props
 				username: resin.auth.whoami()
 				uuid: device.uuid		# get full uuid
-				containerId: resin.models.device.getApplicationInfo(device.uuid).get('containerId')
+				containerId: resin.models.device.getApplicationInfo(device.uuid).then(JSON.parse).get('containerId')
```
lekkas: Hey @pimterry @emirotin cc @wrboyce , I had to made this change because `resin.models.device.getApplicationInfo(device.uuid)` is, apparently, resolving with a `String` now (instead of an `Object`), so `containerId` was set to `undefined`. 

I'm a bit confused as to why, could the latest `resin-sdk` changes be related here?

pimterry: Ah, that's interesting. I don't think we've touched the definition of that method, but it's just returning the body of the result that `resin-request` returns, and that did have some big changes a little while back. None of the parts of the SDK that talk to the supervisor are tested, since they're tough to integration test, so it's hard to know when that broke.

I'll have a look into it now.

pimterry: @lekkas Very interesting indeed. It's not that resin-request doesn't parse JSON any more: hitting http://echo.jsontest.com/key/value with resin-request gives me a JS object back, but hitting `https://api.resin.io/supervisor/v1/apps/<my-app-id>` gives me a string.

Doing some more testing, the problem seems to be the content-type headers. If I hit `https://api.resin.io/supervisor/v1/apps/<my-app-id>`, it comes back with a json body, but `Content-Type: text/html`. This isn't the case for other endpoints: `https://api.resin.io/v1/application` comes back with `application/json` correctly for example, and resin-request returns a parsed object body from that.

This definitely seems broken, and we should fix in the API. The previous version of the SDK does still work, I'd guess because the previous version of resin-request ignored the content type header, and maybe it just parsed every response? We could do something like that ourselves too as a quick fix, but I'd much rather fix this in the API properly if that's easy.

pimterry: Yes, we used to use `request` inside `resin-request`, and pass `json: true` which forces parsing all response bodies as JSON. We now use the content-type, which is correct behaviour. Sounds like this isn't a new change in the API, it's just been exposed by changes in resin-request.

This needs fixing before the cli can go out, since it'll be broken for any supervisor interactions. The SDK is already out with this so these methods are broken there, but it's only out as a prerelease, so that's not a disaster. Who should fix this? Happy to take a look myself, but I haven't touched the API yet, so I'll need some pointers from somebody.

lekkas: Nice find @pimterry! @Page- and @emirotin are the go-to experts for anything API-related. The fix looks simple; iirc there's a `proxy()` middleware function in the API code that handles request proxying to the device supervisor and my guess is this is the culprit wrt incorrect content-type reponses. But let's take this back to the team for more discussion, because I'm not sure if changing the response type will break anything else (e.g. the `python-sdk` ?)

If you feel like diving into the API feel free to send a PR, but in any case, let's create an issue there to make sure this doesn't slip through.

emirotin: Interesting case. So with `fetch` we have to rely on the headers while `request` may have some heuristics.

I think the problem is here
```
		if responses.length is 1
			[{ statusCode }, body] = responses[0]
			return res.status(statusCode).send(body)
		res.status(207).json(multiResponse(responses))
```

The single response is forwarded but the header is not.

lekkas:
@emirotin @pimterry verified that this PR supports AUFS resin sync on both staging and production. Am I good to merge? The `resin ssh` bug mentioned in this thread was already existing, so we will be able to publish a new resin-cli version after the next `resin-api` deployment

emirotin:
I have no objections.

PR #424, opened 2017-03-03
resin ssh: handle undefined uuid parameter

### LINKED ISSUE #423: resin ssh: handle undefined uuid parameter
Passing `undefined` to `resin.model.device.has()` breaks in `resin-sdk@^6.0.0-beta2`

lekkas:
Closes #423


lib/actions/ssh.coffee
```diff
@@ -78,11 +78,15 @@ module.exports =
 
 		verbose = if options.verbose then '-vvv' else ''
 
-		resin.models.device.has(params.uuid).then (isValidUUID) ->
-			if isValidUUID
-				return params.uuid
+		Promise.try ->
```
emirotin: Looks like can be simplified?

```
Promise.try -> 
  return false if not params.uuid
  return resin.models.device.has(params.uuid)
.then (uuidExists) ->
  return params.uuid if uuidExists
  return patterns.inferOrSelectDevice()
.then -> ...
```

lekkas: Fixed

Issue #420, opened 2017-02-07
Automatically detect and report user errors in resin-cli


pimterry:
@stefan-mihaila is looking at creating a standard Resin-wide package for this, which we can add to resin-cli when it's ready.

pimterry:
This now exists, we should try it out: https://github.com/resin-io-modules/resin-raven

pimterry:
Now deployed. I'm leaving a week for manual initial triage and ignoring of super-common issues, and to check this is all working nicely, then moving Sentry to log in UI-ops in Front, and closing.

PR #450, opened 2017-03-17
Add sentry error tracking

pimterry:
Connects to #420.

This is a first pass for discussion. Interesting caveats:

* I'm not filtering errors, so we'll get things like connection errors, device not found, permission denied, etc. I'm expecting there won't be too many categories of these, so we can ignore them in sentry. That lets us keep these out of the way, but also keep the data if we really want it, to look for trends or similar.
* We've got the Sentry DSN inline. I'm pretty sure there's no security implications of this, and it's effectively unavoidable, since we're running client side. I don't think we have any existing built-in config we could use for this, but if we do and I've missed it, it'd be nice to move this there.
* This can delay cli exit while it waits for the error to get reported (this is required: if you quit immediately, not errors get reported). I've set this to time out after a second, but the exact timeout size is pretty debatable.
* This also catches errors by developers running locally during development. Annoyingly I can't think of an easy way to avoid this. We could detect non-global usage, but locally-installed usage isn't that unusual, and I can't think of a clean way to distinguish that from development. Ideas welcome, or we can just do this and not worry about it, I'm not super concerned.


lib/errors.coffee
```diff
@@ -26,4 +30,9 @@ exports.handle = (error) ->
 		message = error.stack
 
 	patterns.printErrorMessage(message)
-	process.exit(error.exitCode or 1)
+
+	captureException(error)
+	.timeout(1000)
+	.catch(->)
```
pimterry: This is an odd thing to complain about. Yes, houndbot, that's exactly what I intended. How do I do this without annoying the bot?

pimterry:
We should also talk about what we'll do with these errors once we've got them. I'm happy to triage them into github issues for now, I'm expecting there'll be a bit of chaos initially, and then we can come up with a more sensible process (probably via a front inbox) once that's settled down.

lib/errors.coffee
emirotin: Try this `.catch -> # ignore`?

emirotin:
Sentry is sensibly grouping them where possible and has nice GH integration.
Lucian is doing the similar triage for the UI

pimterry:
Hound issues are fixed now - if somebody could approve when you're happy that'd be great.

lib/app.coffee
```diff
@@ -14,6 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+Raven = require('raven')
+Raven.disableConsoleAlerts()
+Raven.config(
+	'https://56d2a46124614b01b0f4086897e96110:6e175465accc41b595a96947155f61fb@sentry.io/149239'
```
emirotin: I'd prefer this being in separate config

pimterry: Me too, but I don't think we have any. Do we?

emirotin: For now I would just create the normal `config.coffee` file on the top level and move it there. Later it can be transformed into more advanced config or replaced with some build-time logic

pimterry: Sure, that sounds sensible. Done.

PR #453, opened 2017-03-22
41 select os version

emirotin:
This is the actual implementation of the OS version support in the CLI.
It can work in one of the following modes:

* default (no version specified, or `--version default` specified): tries to use the most recent stable version (that is without pre-release tag) if available, or the most recent version if all available version are pre-release (like for the TS boards?)
* exact semver version number can be specified
* semver range can be provided as well (like ^1.0.0)
* `'latest'` will be the newest version including pre-releases
* `'recommended'` will be the newest stable version, or failure if there are no stable versions
* `'menu'` will show the interactive menu with all available versions


lib/actions/os.coffee
```diff
@@ -16,25 +16,76 @@ limitations under the License.
...
+	if isRecommended
+		result += ' (recommended)'
+	return result
+
+resolveVersion = (deviceType, version) ->
+	if version isnt 'menu'
+		return Promise.resolve(version)
+
+	form = require('resin-cli-form')
+	resin = require('resin-sdk-preconfigured')
```
pimterry: Why aren't these top-level requires?

lib/actions/os.coffee
```diff
@@ -43,9 +94,26 @@ exports.download =
...
+					or the newest version otherwise (if all existing
+					versions for the given device type are pre-release)')
+				return
+			return resolveVersion(params.type, version)
+		.then (resolvedVersion) ->
+			version = resolvedVersion
+			if version isnt 'default'
+				displayVersion = " #{version}"
+		.then ->
+			manager.get(params.type, version)
```
pimterry: Ok, you can simplify this, and fix my previous comment en route I think. If you drop the `version` intermediate variable entirely and just use `options.version` instead, you can drop the `version = 'default'` above and just `return 'default'`, drop the `.then` just here, and directly use `resolvedVersion` here instead of `version`.

lib/actions/os.coffee
```diff
@@ -43,9 +94,26 @@ exports.download =
...
+			if not version
+				version = 'default'
+				console.warn('OS version is not specified, using the default version:
+					the newest stable (non-pre-release) version if available,
+					or the newest version otherwise (if all existing
+					versions for the given device type are pre-release)')
+				return
+			return resolveVersion(params.type, version)
+		.then (resolvedVersion) ->
+			version = resolvedVersion
```
pimterry: There's some strange behaviour here. If `options.version` isn't set, we set `version` to 'default' above, but we then implicitly return undefined, so `resolvedVersion` is undefined, so we overwrite `version` with undefined here. That means version can never be default in the check below, and this is currently actually defaulting to `latest` behaviour (which is what `manager.get` will do with an undefined version).

lib/actions/os.coffee
emirotin: For performance reasons the CLI lazy-loads the dependencies. node caches them anyway, but they're only requested when first needed. Page and Juan has implemented this change to aid the specific performance issues.

lib/actions/os.coffee
emirotin: Oh, that's the mistake I made when refactoring the code. 'default' should be returned above

lib/actions/os.coffee
pimterry: Huh, I see, interesting. Fair enough then :+1:

Issue #452, opened 2017-03-21
Issues downloading and installing new OS versions for raspberry-pi


mscharley:
I'm having issues with provisioning new devices:

```
$ resin device init -va Foo
Getting device operating system for raspberry-pi
Downloading Device OS [========================] 100% eta 39s  
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: incorrect data check
    at Zlib._handle.onerror (zlib.js:370:17)
```

> resin-cli: 5.5.0
> node: 6.10.0
> npm: 3.10.10

mscharley:
The images that are downloaded are valid .img files - I can mount them locally completely fine.

lekkas:
Hey @mscharley , 

You can also track https://github.com/resin-io/resin-cli/issues/451 , which was opened as a response to your issue. Cheers!

emirotin:
@lekkas so I'm currently working on making the OS v1 downloadable via the CLI as requested in #451, but it seems that the rest of the CLI would still not work with the OS v2 downloads (as highlighted by this issue). Am I right?

emirotin:
Hey @mscharley I've just published the CLI v5.6.0, can you give it a try?

Here's the full story behind the issue and what was done:
- we've recently published two preview releases of the resinOS v2.0.0 for the RPi family
- the CLI was always downloading the most recent release which always used to be v1 until recently
- it looks (I didn't have time to investigate it yet) the CLI configuration doesn't work with OS v2 yet
- so you got the new OS image and configuration failed
- I've made 2 changes: first, you can now specify the version, check `resin help os download` for the variety of options, second, the default version is the most recent **non-pre-release** now, which would be the most recent v1 for RPi which is what it was previously

So now if you try to run the `device init` it will download the OS v1 and _should_ work.
Please let us know if it's the case.

But we still need to check if the OS configuration actually works with resinOS v2, and fix it if not.

mscharley:
`resin device init` is now working fine with 5.6.0 to install OS v1.24.1

A minor point of feedback though that's only slightly tangential - it'd be great if the CLI had an option to select which version as one of the options in the wizard if no option was given explicitly.

Currently our application isn't compatible with the 2.x OS either, leading to redirect loops if it ever got flashed.

emirotin:
Great to know!

Yeah, the OS version selection can and will be added to the wizard, so far I've implemented all the required primitives.

mscharley:
Actually, I spoke too soon.

[...]

PR #461, opened 2017-03-24
isolate the sudo-runnable command

### LINKED ISSUE #456: expecting an array or an iterable object but got [object Null]
https://github.com/resin-io/resin-cli/issues/452#issuecomment-288661711

My test account has passwordless sudo. When this code is ran
```
if os.platform() isnt 'win32'
  console.log('Type your computer password to continue')

return president.executeAsync(command)
```

it prints the message and then the control returns to the main program after a while.
During the `device init` the program then crashes with `expecting an array or an iterable object but got [object Null]` which I don't yet know where it's coming from.

Front logo [Front conversations](https://app.frontapp.com/open/top_1cbr)

emirotin:
This isolates the command that needs sudo (writing tot he actual physical disk) from the userland operations that require user authentication, like collecting the info from the API.

Also fixes the bluebird bug and updates a bunch of dependencies

Closes #456, #452


lib/actions/device.coffee
```diff
@@ -392,27 +392,23 @@ exports.init =
...
-
-							helpers.sudo([ 'os', 'initialize', temporalPath, '--type', application.device_type ], message)
-
+						configureCommand = "os configure #{tempPath} #{device.uuid}"
+						if options.advanced
+							configureCommand += ' --advanced'
+						capitano.runAsync(configureCommand)
+						.then ->
+							osInitCommand = "os initialize #{tempPath} --type #{application.device_type}"
+							capitano.runAsync(osInitCommand)
```
pimterry: Nit: the intermediate variable here doesn't add anything, it just makes me think this is used elsewhere, but it doesn't seem to be...

lib/actions/device.coffee
```diff
@@ -392,27 +392,23 @@ exports.init =
...
-							'''
-
-							helpers.sudo([ 'os', 'initialize', temporalPath, '--type', application.device_type ], message)
-
+						configureCommand = "os configure #{tempPath} #{device.uuid}"
+						if options.advanced
+							configureCommand += ' --advanced'
+						capitano.runAsync(configureCommand)
+						.then ->
+							osInitCommand = "os initialize #{tempPath} --type #{application.device_type}"
```
pimterry: Do we need quotes around `#{tempPath}`, here and above?

lib/actions/util.coffee
```diff
@@ -0,0 +1,37 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+###
+
+# These are internal commands we want to be runnable from the outside
+# One use-case for this is spawning the minimal operation with root priviledges
+
+exports.osInit =
+	signature: 'util osinit <image> <type> <config>'
```
pimterry: Could we use `internal` instead of `util`? Makes it clearer what's going on imo, both for cli users and in the code.

lib/utils/helpers.coffee
```diff
@@ -47,17 +47,14 @@ exports.stateToString = (state) ->
...
-exports.sudo = (command, message) ->
-	command = _.union(_.take(process.argv, 2), command)
-	console.log(message)
-
+exports.sudo = (command) ->
 	if os.platform() isnt 'win32'
-		console.log('Type your computer password to continue')
+		console.log('If asked please type your computer password to continue')
 
+	command = _.union(_.take(process.argv, 2), command)
```
pimterry: I'm not sure why, but it looks like this worked before without this change, with similar arguments to what we're now using. What's changed?

lib/utils/helpers.coffee
```diff
@@ -70,3 +67,21 @@ exports.getManifest = (image, deviceType) ->
...
+	step.on('stdout', _.bind(process.stdout.write, process.stdout))
+	step.on('stderr', _.bind(process.stderr.write, process.stderr))
+
+	step.on 'state', (state) ->
+		return if state.operation.command is 'burn'
+		console.log(exports.stateToString(state))
+
+	bar = new visuals.Progress('Writing Device OS')
+
+	step.on('burn', _.bind(bar.update, bar))
```
pimterry: Why don't we just use native `.bind`? (here and above)

lib/actions/device.coffee
emirotin: It's a specially built temp path, very unlikely to have spaces. But can put of course

lib/actions/device.coffee
emirotin: Nope, it's just to separate the things: first we build the command, then we use it. I prefer this style when the long strings manipulation is happening.

lib/utils/helpers.coffee
emirotin: I have no idea how this used to work, really. 
It was sudo-ing the command that tries to fetch some data using the SDK. 
The change that I've done only sudo-es the command that writes to the drive given all the required info.

lib/utils/helpers.coffee
emirotin: We can of course, I've just copied this verbatim. Had the same  thought, but didn't want to put more changes.

PR #466, opened 2017-03-29
Add ability to build and deploy image locally using resin-cli

CameronDiver:
Using `resin build` a user can now build an image on their own docker
daemon. The daemon can be accessed via a local socket, a remote host and
a remote host over a TLS socket. Project type resolution is supported.
Nocache and tagging of images is also supported.

Using `resin deploy` a user can specify either a build context or an image, and it will be uploaded to the resin registry, and deployed to the fleet.

Note: Adding all local actions to the same PR for continuity. If this ends up being too unwieldy I'll separate them out.

Depends-on:
    https://github.com/resin-io/resin-builder/pull/129/files
    https://github.com/resin-io/resin-api/pull/162

Connects-to: hq/pull/732
Change-type: minor


lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+			alias: 'a'
+		},
+		{
+			signature: 'tag'
+			parameter: 'tag'
+			description: 'The alias to the generated image'
+			alias: 't'
+		},
+		{
+			signature: 'nocache'
```
pimterry: It's a little odd that every other option is camel-case, except this one.

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+		throw new Error("Both a local docker socket and docker host have been provided. Don't know how to continue.")
+	else
+		# None provided, assume default docker local socket
+		connectOpts['socketPath'] = '/var/run/docker.sock'
+
+	# Now need to check if the user wants to connect over TLS
+	# to the host
+	connectOpts['ca'] = options.ca if options.ca?
+	connectOpts['cert'] = options.cert if options.cert?
+	connectOpts['key'] = options.key if options.key?
```
pimterry: I think if you don't provide all three of these, Dockerode will [silently not use TLS](https://github.com/apocas/docker-modem/blob/master/lib/modem.js#L68). We should probably throw an error to force you to provide all three or none.

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+	]
+	action: (params, options, done) ->
+		# The default build context is the current directory
+		params.context ?= '.'
+
+		# Tar up the directory, ready for the build stream
+		tarDirectory(params.context)
+		.then (tarStream) ->
+			hooks =
+					buildSuccess: (image, layers) ->
```
pimterry: Looks like this whole block has an extra indent.

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+						if options.tag?
+							console.log("Tagging image as #{options.tag}")
+						done()
+					buildFailure: (err) ->
+						done("Build error: #{err.message}")
```
pimterry: I think generally it's better to fail with the actual error we get here, so we can get the stack trace later (with `DEBUG` set, or in Sentry).

If you need to change the message you could:
* Console.error your own text, and then throw the exception as is.
* Make a new error with whatever you like, but copy `err.stack` across.
* Change the message on the error instance and rethrow it (feels a little nasty, but I don't have a good reason why that matters)
* Add specific cases to [resin-cli-errors](https://github.com/resin-io-modules/resin-cli-errors), which translates some errors into nice friendly messages already.

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+					buildStream: (stream) ->
+						info = getBundleInfo(options)
+						# info is either a promise, or is undefined
+						if info?
+							info
+							.catch (err) -> done("Error: #{err.message}")
```
pimterry: As above: we should avoid losing errors.

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+							.then ([arch, deviceType]) ->
+								# Perform type resolution on the project
+								bundle = new resolve.Bundle(tarStream, deviceType, arch)
+								resolve.resolveBundle(bundle, resolve.getDefaultResolvers())
+							.then (resolved) ->
+								console.log("Building #{resolved.projectType} project")
+								# Send the resolved tar stream to the docker daemon
+								resolved.tarStream.pipe(stream)
+							.catch(done)
```
pimterry: Why do we need two `catch` cases, with slightly different behaviour?

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+						else
+							console.log('Warning: No architecture/device type or application information provided')
+							console.log('  No project type resolution will be performed.')
```
pimterry: Will this make sense to end users? Might want to add something like `A vanilla Docker build will be run` (or maybe something better phrased, but you get the idea), or maybe a link to our build docs for context.

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+				opts = {}
+
+				if options.tag?
+					opts['t'] = options.tag
+				if options.nocache?
+					opts['nocache'] = true
+				# Pass done as the error handler so any propagated errors
+				# hit the main handler
+				builder.createBuildStream(opts, hooks, done)
+		.catch(done)
```
pimterry: The flow to `done` here is quite confusing! The hooks call it, we also provide it as an error callback, and then we have a catch-all `catch` here that calls it even after that. I don't know what happens if we call done more than once, but it definitely feels like we shouldn't.

Can we do this with proper promises instead, and just cleanly map it back to the done callback at the end?

For example, if you have `new Promise((resolve, reject) =>` inside the `.then (tarStream)` here, you can call `resolve` or `reject` from the hooks, pass `reject` as the error handler for build.createBuildStream, and all this hooks business will turn back into just another step in this promise chain.

You can then put a single `nodeify(done)` at the end to go back to capitano. That handles success or any errors here or in the previous promise code, all in one place. If you start heading down that route, I expect there'll be more places where consistently using promises simplifies things a little too, although I can't see them off the top of my head.

lib/actions/build.coffee
```diff
@@ -0,0 +1,224 @@
...
+resolve = require('resin-bundle-resolve')
+klaw = require('klaw')
+path = require('path')
+fs = require('mz/fs')
+_ = require('lodash')
+
+# When we do TLS enabled docker builds, node sometimes complains
+# about the certificates not being fully authorized, this stops
+# that.
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
```
CameronDiver: Definitely shouldn't be committed, not sure why it was still there (had to be there for my testing). It definitely should be something that the user turns on though, fully agree.

lib/actions/build.coffee
CameronDiver: It's because `nocache` is the docker standard way of passing it, but I've deviated away from how docker names other things already so I might as well keep my inconsistencies consistent :p

lib/actions/build.coffee
CameronDiver: This is for errors originating from the docker daemon though, so the stack trace tells us nothing but that the error came from the docker daemon itself. Errors originating from `resin-docker-build` are handled differently, so it wouldn't be one of those.

lib/actions/build.coffee
CameronDiver: Ah I was having some issues with error propagation yesterday, this is an artifact of that. Will remove.

lib/actions/build.coffee
CameronDiver: Good point, I kinda take it for granted at this point.

lib/actions/build.coffee
CameronDiver: Disregard this, I wasn't aware that the stack trace is hidden without `DEBUG` (which is the behaviour I was emulating by not printing it)

lib/utils/helpers.coffee
```diff
@@ -70,3 +69,17 @@ exports.getManifest = (image, deviceType) ->
 	.then(JSON.parse)
 	.catch ->
 		resin.models.device.getManifestBySlug(deviceType)
+
+exports.getAppInfo = (application) ->
+	Promise = require('bluebird')
+	_ = require('lodash')
+	resin = require('resin-sdk-preconfigured')
+	return resin.models.application.get(application)
```
Page-: It'd probably be nicer to do:
```coffee
Promise.join resin.models.application.get(application), resin.models.config.getDeviceTypes(), (app, config) ->
	...

File: lib/utils/helpers.coffee [hunk 11]
```diff
@@ -70,3 +69,17 @@ exports.getManifest = (image, deviceType) ->
 	.then(JSON.parse)
 	.catch ->
 		resin.models.device.getManifestBySlug(deviceType)
+
+exports.getAppInfo = (application) ->
+	Promise = require('bluebird')
+	_ = require('lodash')
+	resin = require('resin-sdk-preconfigured')
```
Page-: These three requires are already included at the module level

lib/utils/docker.coffee
```diff
@@ -0,0 +1,201 @@
...
+					if info?
+						info.then ([arch, deviceType]) ->
+							# Perform type resolution on the project
+							bundle = new resolver.Bundle(tarStream, deviceType, arch)
+							resolver.resolveBundle(bundle, resolver.getDefaultResolvers())
+						.then (resolved) ->
+							console.log("Building #{resolved.projectType} project")
+							# Send the resolved tar stream to the docker daemon
+							resolved.tarStream.pipe(stream)
+						.catch(reject)
```
Page-: This looks weird to me, I think it's correct but it is throwing up red flags - can you just double check it's sensible

lib/utils/docker.coffee
```diff
@@ -0,0 +1,201 @@
...
+		Promise.resolve(streamToPromise(klaw(dir)))
+			.filter((item) -> not item.stats.isDirectory())
+			.map((item) -> item.path)
+
+	pack = tar.pack()
+	getFiles(dir)
+	.map (file) ->
+		relPath = path.relative(path.resolve(dir), file)
+		Promise.all([relPath, fs.stat(file), fs.readFile(file)])
+	.map ([filename, stats, data]) ->
```
Page-: This would be better as 
```coffee
.map (file) ->
	relPath = path.relative(path.resolve(dir), file)
	Promise.join relPath, fs.stat(file), fs.readFile(file), (filename, stats, data) ->
		pack.entryAsync({ name: filename, size: stats.size }, data)
```
because that way all the `fs` accesses will be parallel with the packing, right now it will do all the fs access in parallel and *then* do all the packing in parallel, waiting for the first map to finish before starting the second

lib/utils/docker.coffee
```diff
@@ -0,0 +1,201 @@
...
+exports.tarDirectory = tarDirectory = (dir) ->
+	Promise = require('bluebird')
+	tar = require('tar-stream')
+	klaw = require('klaw')
+	path = require('path')
+	fs = require('mz/fs')
+	streamToPromise = require('stream-to-promise')
+
+	getFiles = ->
+		Promise.resolve(streamToPromise(klaw(dir)))
```
Page-: I think you could avoid needing to wrap this by using [any-promise registrating: `require('any-promise/register/bluebird')`](https://github.com/kevinbeaty/any-promise#usage-with-registration)

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+								else if obj.type == 'success'
+									resolve([appName, obj.image, token, url])
+
+					.finally ->
+						require('fs').unlink(tmpPath)
+			.catch(reject)
+		.then ([appName, imageName, token, url]) ->
+			console.log("Successfully pushed image to registry: #{imageName}")
+			console.log('Deploying to fleet...')
+			Promise.try(-> updateFleet(appName, imageName, token, url))
```
Page-: There's no need for the `Promise.try` here?

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+		.then ->
+			console.log('Done')
+			done()
```
Page-: `.asCallback(done)` can replace this and the catch below

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+		tmp = Promise.promisifyAll(require('tmp'))
+		resin = require('resin-sdk-preconfigured')
+
+		dockerUtils = require('../utils/docker')
+
+		# Ensure the tmp files gets deleted
+		tmp.setGracefulCleanup()
+
+		docker = dockerUtils.getDocker(options)
+		new Promise (resolve, reject) ->
```
Page-: There's no need to use `new Promise` here, you create a promise immediately inside via `Promise.try`

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+					params.context = context
+					if build
+						# Instead of an image, the user has specified to build an image
+						imagePromise = dockerUtils.runBuild(params, options, getBundleInfo)
+					else
+						imagePromise = Promise.resolve(imageName)
+
+					imagePromise
+					.then (imageName) ->
+						Promise.all([
```
Page-: Promise.join is preferable for cases like this 
```coffee
Promise.join
	dockerUtils.bufferImage(docker, imageName, tmpPath)
	...
	(stream, token, username, url, size) ->

File: lib/actions/build.coffee [hunk 19]
```diff
@@ -0,0 +1,81 @@
...
+		},
+		{
+			signature: 'noCache'
+			description: "Don't use docker layer caching when building"
+			boolean: true
+		},
+	]
+	action: (params, options, done) ->
+		dockerUtils.runBuild(params, options, getBundleInfo)
+		.then -> done()
```
Page-:
```coffee
.asCallback(done)

File: lib/actions/build.coffee [hunk 20]
```diff
@@ -0,0 +1,81 @@
...
+	if options.application?
+		# An application was provided
+		Promise.try(-> helpers.getAppInfo(options.application))
+		.then (app) ->
+			return [app.arch, app.device_type]
+	else if options.arch? and options.deviceType?
+		return Promise.resolve([options.arch, options.deviceType])
+	else
+		# No information, cannot do resolution
+		return
```
Page-: This function should always return a promise if it's going to sometimes return a promise, that way we can rely on `func().then()` always being valid, rather than failing horribly when this branch is hit, I'd recommend using `Promise.method`:
```coffee
exports.getBundleInfo = getBundleInfo = Promise.method (options) ->

File: lib/actions/build.coffee [hunk 21]
```diff
@@ -0,0 +1,81 @@
...
+# of this action
+dockerUtils = require('../utils/docker')
+
+exports.getBundleInfo = getBundleInfo = (options) ->
+	Promise = require('bluebird')
+	helpers = require('../utils/helpers')
+
+	if options.application?
+		# An application was provided
+		Promise.try(-> helpers.getAppInfo(options.application))
```
Page-: Do we need this `Promise.try`? Doesn't `helpers.getAppInfo` return a promise already?

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+		process.stdout.cursorTo(0)
+		process.stdout.write("Uploaded #{percent.toFixed(1)}%")
+		console.log() if percent == 100
+	, timeout
+
+# A regex for parsing;
+# * A hostname e.g. registry.resin.io
+# * An appname, which consists of 1 or more alphanumeric chars
+# * A commit hash (40 alphanumeric characters)
+resinImageRegex = /.*\/(\w+\/)(\w{40})/
```
dfunckt: AFAICS you only need this regex to extract the commit hash so it can be simplified to `.+\/(\w{40})`. Otherwise. if you need to do any kind of validation, I'd rewrite this as a function like so:

```
parseCommitHash = (image) ->
  parts = image?.split('/')
  return null if parts?.length isnt 3 // or throw error
  return parts.pop()
```

lib/actions/build.coffee
emirotin: Also let's have explicit return here as we do below

lib/actions/build.coffee
```diff
@@ -0,0 +1,81 @@
+# Imported here because it's needed for the setup
+# of this action
+dockerUtils = require('../utils/docker')
+
+exports.getBundleInfo = getBundleInfo = (options) ->
```
emirotin: I don't think this is imported / reused anywhere, so can be not exported?

lib/actions/build.coffee
emirotin: Right, or `return Promise.resolve()` can be used, but the `method` is neater, can save two `resolve's`. Also I would throw here as we cannot handle this case anyway, and rely on `Promise.catch` in the build runner to capture it, log the error and do the rest of the failure handling.
Right now the build runner is mixing promises with synch paradigm which is not great.

lib/actions/build.coffee
```diff
@@ -0,0 +1,81 @@
...
+
+		You must provide either an application or a device-type/architecture
+		pair to use the resin Dockerfile pre-processor
+		(e.g. Dockerfile.template -> Dockerfile).
+
+		Examples:
+
+			$ resin build
+			$ resin build ./context/
+			$ resin build -d raspberrypi3 -A armhf
```
emirotin: I'd use long readable params here, help is for less experienced users IMO

lib/actions/build.coffee
```diff
@@ -0,0 +1,81 @@
...
+			alias: 'a'
+		},
+		{
+			signature: 'tag'
+			parameter: 'tag'
+			description: 'The alias to the generated image'
+			alias: 't'
+		},
+		{
+			signature: 'noCache'
```
emirotin: is it case-sensitive? Not *nix idiomatic I think, usually all params are lowercase. I don't know if Capitano handles dash-separated names, but `--nocache` also looks fine

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
+getBuilderPushEndpoint = (baseUrl, owner, app) ->
+	"https://builder.#{baseUrl}/v1/push?owner=#{owner}&app=#{app}"
```
emirotin: I'd do `encodeURIComponent()` here for the owner and the app name, more future-proof

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+
+parseInput = (params, options) ->
```
emirotin: Can as well be `Promise.method ->` to start the promise chain from here

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+	if not params.appName?
+		throw new Error('Need an application to deploy to!')
+	appName = params.appName
+	image = undefined
+	if params.image?
+		if (options.build or options.context?)
```
emirotin: parens are not needed

lib/actions/deploy.coffee
emirotin: Right, makes sense to encapsulate the match/extract/throw into a method

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+			signature: 'context'
+			parameter: 'context'
+			description: 'The context to use when building the image'
+			alias: 'c'
+		}
+	]
+	action: (params, options, done) ->
+		Promise = require('bluebird')
+		request = require('request')
+		tmp = Promise.promisifyAll(require('tmp'))
```
emirotin: I tend not to use `promisifyAll` when we only need a method or two, you can check the usage of `tmp` in other modules here

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+		docker = dockerUtils.getDocker(options)
+		new Promise (resolve, reject) ->
+			# Check input parameters
+			Promise.try( -> parseInput(params, options))
+			.then ([appName, build, context, imageName]) ->
+				tmp.tmpNameAsync()
+				.then (tmpPath) ->
+
+					# Setup the build args for how the build routine expects them
+					options.appName = appName
```
emirotin: Not great to mutate the input params, not a real issue here but would be nicer to stay pure:
```
options = _.assign({}, options, { appName })
params = _.assign({}, params, { context })
```

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+							resin.auth.getToken()
+							resin.auth.whoami()
+							resin.settings.get('resinUrl')
+							dockerUtils.getImageSize(docker, imageName)
+						])
+						.then ([stream, token, username, url, size]) ->
+							url = url || process.env.RESINRC_RESIN_URL
+							post = request.post
+								url: getBuilderPushEndpoint(url, username, params.appName)
+								auth:
```
emirotin: Does it work like this? Nice :)

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+							.on 'data', (data) ->
+								handleMessage(data.toString())
+
+							# Set up upload reporting
+							pushProgress(size, post)
+
+							handleMessage = (data) ->
+								if process.env.DEBUG
+									console.log("Received data: #{data}")
+								obj = JSON.parse(data)
```
emirotin: Cannot `request` parse it or us?

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,168 @@
...
+								if obj.type == 'error'
+									reject(new Error("Remote error: #{obj.error}"))
+								else if obj.type == 'success'
+									resolve([appName, obj.image, token, url])
```
emirotin: `else`... throw something or at least log the unexpected response? Otherwise it'll hang forever

lib/actions/deploy.coffee
emirotin: And because of that simply `throw` from anywhere deep in the chain instead of fiddling with `reject`

lib/utils/docker.coffee
```diff
@@ -0,0 +1,201 @@
...
+		}
+	]
+
+exports.generateConnectOpts = generateConnectOpts = (opts) ->
+	connectOpts = {}
+	# Firsly need to decide between a local docker socket
+	# and a host available over a host:port combo
+	if opts.docker? and not opts.dockerHost?
+		# good, local docker socket
+		connectOpts['socketPath'] = opts.docker
```
emirotin: You can use the regular dot notation here and below

lib/utils/docker.coffee
CameronDiver: It seems to work, if the resolution fails, an error throws and the promise rejects, and it hits the global handler happily. Which part are you not sure on?

lib/utils/docker.coffee
emirotin: That way ^ you don't even have to pass the `filename` around, it's the statically resolved `relPath` above

lib/utils/docker.coffee
```diff
@@ -0,0 +1,201 @@
...
+
+	# The default build context is the current directory
+	params.context ?= '.'
+
+	# Tar up the directory, ready for the build stream
+	tarDirectory(params.context)
+	.then (tarStream) ->
+		new Promise (resolve, reject) ->
+			hooks =
+				buildSuccess: (image, layers) ->
```
emirotin: Unused param

lib/utils/docker.coffee
```diff
@@ -0,0 +1,201 @@
...
+					if options.tag?
+						console.log("Tagging image as #{options.tag}")
+					resolve(image)
+				buildFailure: reject
+				buildStream: (stream) ->
+					info = getBundleInfo(options)
```
emirotin: This is the place where we can just `.then` and return early if the bundle info is not defined

lib/utils/docker.coffee
```diff
@@ -0,0 +1,201 @@
...
+	Docker = require('dockerode')
+	Promise = require('bluebird')
+	connectOpts = generateConnectOpts(options)
+	# Use bluebird's promises
+	connectOpts['Promise'] = Promise
+	new Docker(connectOpts)
+
+exports.getImageSize = (docker, image) ->
+	docker.getImage(image).inspectAsync()
+	.then (data) ->
```
emirotin: NIT, can be `.then ({ Size }) -> Size`, or `.get('Size')`

lib/utils/helpers.coffee
emirotin: I would actually move them into the methods where they're required if possible

lib/actions/deploy.coffee
CameronDiver: I'm still getting my head around promises, but how would I then resolve when necessary as on this line: https://github.com/resin-io/resin-cli/pull/466/files/d12d94cd6504e4b8c9fe92cc7acf2e87ac4d9042#diff-c72c99d4ed405cbcbf5450d9fb46da8eR156 ?

emirotin: Ah, you need to wrap the callback (Promise.fromCallback) or promisify the part that currently requires the callback (wrap the stream-related stuff into a promise)

lib/actions/deploy.coffee
CameronDiver: `request` can if I set `json: true` in the request options, but then `request` tries to parse the image on the way to the builder, which is actually a tar archive so it fails. If you know how I can get `request` to only parse on the way back that would be much better.

lib/app.coffee
```diff
@@ -33,6 +33,9 @@ events = require('./events')
 plugins = require('./utils/plugins')
 update = require('./utils/update')
 
+# Assign bluebird as the global promise library
+require('any-promise/register/bluebird')
```
CameronDiver: It's for `streamToPromise`, otherwise it uses native promises and messes things up, a lot.

CameronDiver: I mean I guess I could always wrap `streamToPromise` in a bluebird promise but I could see that causing issues on codepaths that aren't followed often, and therefore not tested as well.

pimterry: Ah, ok. There is https://www.npmjs.com/package/stream-to-promise-agnostic, but it looks like a fork that's falling behind their master, and I don't think it's really worth the effort. Fine to ignore this :+1:

pimterry: Although actually a comment here blaming stream-to-promise might be nice

CameronDiver: Good idea, will add, thanks Tim.

lib/actions/deploy.coffee
```diff
@@ -0,0 +1,159 @@
...
+
+				# Setup the build args for how the build routine expects them
+				options = _.assign({}, options, { appName })
+				params = _.assign({}, params, { context })
+
+				if build
+					# Instead of an image, the user has specified to build an image
+					imagePromise = dockerUtils.runBuild(params, options, getBundleInfo)
+				else
+					imagePromise = Promise.resolve(imageName)
```
Page-: This can be done as
```coffee
Promise.try ->
	if build
		# Instead of an image, the user has specified to build an image
 		dockerUtils.runBuild(params, options, getBundleInfo)
	else
		imageName
.then (imageName) ->

CameronDiver: Much more subtle than my pattern! Thanks.

Issue #479, opened 2017-04-11
resin local docs


craig-mulligan:
We don't have `resin local` docs on the repo. Looks like they used to be generated with capitano, @lekkas did you give this a try? 

I don't mind looking into it just wanted to check if there are major hurdles?

craig-mulligan:
For reference https://github.com/resin-os/resinos/pull/267

lekkas:
Hey @craig-mulligan ,

Right, I missed to create/publish them when adding `resin local` commands. Normally `npm run doc` produces docs from comments, however I just tried to run it and it didn't produce docs for `resin local` commands because these are grouped in a folder (`actions/local`) and `extras/capitanodoc/index.coffee` doesn't seem to scan nested folders for actions.

So this seems like a quick fix in `extras/capitanodoc/index.coffee` if you 're up for it a PR 🤘 !

craig-mulligan:
Hey @lekkas yep it was a quick fix 👍 see #480

lekkas:
Rad! 🎉

PR #480, opened 2017-04-11
Add resin local docs

craig-mulligan:
Connects to #479

Change-Type: patch


craig-mulligan:
I'm not sure if there are any other implications of adding entries to `capitanodoc.json`. I assume it's just used for docs but not sure, anyone know? or are there some tests I can run?

PR #482, opened 2017-04-14
update to modules that have native node v4 support

emirotin:
This removes the necessity to use the babel hack, so I've removed it as well.

Tested the device init operation myself with node v4, it worked.
@lekkas if you can test the local flash that would be cool.


emirotin:
v5.7.2 published with this fix

PR #487, opened 2017-04-19
Device api keys

### LINKED ISSUE #473: resin device register only does the old long UUIDs
With resinOS 2.0, the device UUIDs became half the length (~31~ 32 chars down from 62 characters). Running `resin device register AppName` only able to do the original 62 character version. Not sure if it's a problem, just highlight it, because this way images freshly downloaded from the Dashboard and images configured with the CLI will have different behaviour for 2.0.

What would be the right default behaviour? Or this just doesn't have any real consequences?

Front logo [Front conversations](https://app.frontapp.com/open/top_20dt)

Page-:
Depends on https://github.com/resin-io/resin-sdk/pull/305

Closes #473


lib/actions/device.coffee
```diff
@@ -128,25 +128,40 @@ exports.register =
 		Examples:
 
 			$ resin device register MyApp
+			$ resin device register MyApp --uuid 83cb718aee3abe63d2e0a0016ea49399cdc5b3f9b879dbb01acf73f28f5085
+			$ resin device register MyApp --uuid 83cb718aee3abe63d2e0a0016ea49399cdc5b3f9b879dbb01acf73f28f5085 --device-api-key existingDeviceKey
```
Page-: Ideas on what I should do about this? The example kinda needs to be this long :|

emirotin: Err, ignore?:)

lib/actions/device.coffee
```diff
@@ -128,25 +128,40 @@ exports.register =
...
 
-		resin.models.application.get(params.application).then (application) ->
-
-			Promise.try ->
-				return options.uuid or resin.models.device.generateUniqueKey()
-			.then (uuid) ->
+		Promise.join(
+			resin.models.application.get(params.application)
+			options.uuid or resin.models.device.generateUniqueKey()
+			options.deviceApiKey ? resin.models.device.generateUniqueKey()
```
emirotin: We use `or` in the previous line and `?` here for the same purpose, so maybe let's make them consistent?

lib/actions/os.coffee
```diff
@@ -173,7 +174,9 @@ exports.configure =
 
 				return form.run(questions, { override })
 			.then (answers) ->
-				init.configure(params.image, params.uuid, answers).then(helpers.osProgressHandler)
+				Promise.resolve(params.deviceApiKey ? resin.models.device.generateDeviceKey(params.uuid))
+				.then (deviceApiKey) ->
+					init.configure(params.image, params.uuid, deviceApiKey, answers).then(helpers.osProgressHandler)
```
emirotin: The `then` in the end can be on the came level as the previous one?

Page-: Yeah, I didn't even notice the `.then` there as I wasn't changing it, moved

lib/actions/device.coffee
Page-: Yeah, it's just it also prints the error on every build as well, which sucks

pimterry: It's just using coffeelint, right? You should be able to add coffeelint ignore markers for individual lines. From http://www.coffeelint.org/:

```coffeescript
# coffeelint: disable=max_line_length
foo = "some/huge/line/string/with/embed/#{values}.that/surpasses/the/max/column/width"
# coffeelint: enable=max_line_length
```

emirotin:
Are we good here @Page-?

pimterry:
@Page- are we good to merge this?

Page-:
It's not good to merge because it requires resinos >= 2.0.3 for support for the deviceApiKey property

pimterry:
@Page- this is blocking some other changes, since we need to pull in new versions of `resin-device-config` elsewhere, and sadly this fix is required to do that, because of the breaking changes there.

What do we need to do to get this mergeable in a state where it'll work for both resinos 2.0.3+ devices and older devices? Can you take a look at it, or give me an outline of how it should work for older devices so I can finish it off myself?

Page-:
@pimterry the problem with this is it switches to provisioning keys for the device, which requires the device to do a key exchange in order to work.  Right now that will kind of work on resinos 2.0.3; the supervisor will work, but the vpn will not update until the device is rebooted, https://github.com/resin-os/meta-resin/pull/660 deals with that. And it won't work at all on any 1.x or any 2.x <2.0.3 as they do not do key exchange

pimterry:
@Page- I can't add you as a reviewer to this because you created the PR originally, but can you review it please?

I've pushed a bunch of updates which, combined with [the resin-device-init change](https://github.com/resin-io-modules/resin-device-init/pull/24) from yesterday should let us safely handle these three commands for both devices that do and don't support device api keys.

It's made safe by only using device api keys if you _explicitly_ offer one, with a docs warning for our support of that case, and falling back to application API keys otherwise.

I do want to add separate logic later for `os configure` at least, where we can potentially work out the exact image version you're using, and configure it completely appropriately. For now though, this stops all these features being 100% broken in the CLI for Resin OS 2.x, still lets you opt in to device api keys when you want them, and puts us on a good path to keep improving this and adding reconfix imminently.

Page-:
LGTM

PR #493, opened 2017-04-24
Upload logs when doing both build + deploy

CameronDiver:
Depends on https://github.com/resin-io/resin-builder/pull/146


lib/actions/deploy.coffee
```diff
@@ -123,34 +142,65 @@ module.exports =
...
+				Promise.join(
+					logs
+					token
+					url
+					buildId
+					username
+					params.appName
+					uploadLogs
+				)
+				.then ->
```
Page-: This can be
```coffee
.return('Successfully uploaded logs')

File: lib/utils/docker.coffee [hunk 2]
```diff
@@ -128,7 +130,7 @@ exports.runBuild = (params, options, getBundleInfo) ->
 				buildSuccess: (image) ->
 					if options.tag?
 						console.log("Tagging image as #{options.tag}")
-					resolve(image)
+					resolve([image, logs])
```
Page-: I think it'd be nicer to use `{ image, logs }` so that it's clear what each property means, rather than trying to map an index to a meaning (a similar idea to using `funcWithBoolOpts(true, false, false, true, false)` being much clearer with an object)

lib/actions/deploy.coffee
```diff
@@ -112,6 +126,11 @@ module.exports =
 			parameter: 'source'
 			description: 'The source directory to use when building the image'
 			alias: 's'
+		},
+		{
+			signature: 'nologs'
```
Page-: I'm not convinced by the name - this one makes me think that it would hide local logs - maybe `no-logs-upload`?

lib/actions/deploy.coffee
```diff
@@ -70,7 +84,7 @@ uploadToPromise = (request, size) ->
 			if obj.type?
 				switch obj.type
 					when 'error' then reject(new Error("Remote error: #{obj.error}"))
-					when 'success' then resolve(obj.image)
+					when 'success' then resolve([obj.image, obj.buildId])
```
Page-: I'm a big fan of objects for things like this (as mentioned in my other comment)

lib/actions/deploy.coffee
```diff
@@ -58,6 +63,15 @@ performUpload = (image, token, username, url, size, appName) ->
...
 	uploadToPromise(post, size)
 
+uploadLogs = (logs, token, url, buildId, username, appName) ->
+	request = require('request')
+	url = url || process.env.RESINRC_RESIN_URL
+	request.post
+		url: getBuilderLogPushEndpoint(url, buildId, username, appName)
+		auth:
+			bearer: token
+		body: Buffer.from(logs).toString('base64')
```
Page-: Is the manual base64-ing necessary?

lib/actions/deploy.coffee
```diff
@@ -6,6 +6,11 @@ getBuilderPushEndpoint = (baseUrl, owner, app) ->
 	escApp = encodeURIComponent(app)
 	"https://builder.#{baseUrl}/v1/push?owner=#{escOwner}&app=#{escApp}"
 
+getBuilderLogPushEndpoint = (baseUrl, buildId, owner, app) ->
+	escOwner = encodeURIComponent(owner)
+	escApp = encodeURIComponent(app)
+	"https://builder.#{baseUrl}/v1/pushLogs?owner=#{escOwner}&app=#{escApp}&buildId=#{buildId}"
```
Page-: It might be worth using `querystring.stringify({ owner, app, buildId })` - https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

package.json
```diff
@@ -73,6 +73,7 @@
     "resin-image-fs": "^2.1.2",
     "resin-image-manager": "^4.1.1",
     "resin-sdk-preconfigured": "^6.0.0",
+    "resin-settings-client": "^3.6.0",
```
Page-: This is unused?

lib/actions/deploy.coffee
CameronDiver: What would be the alternative? AFAICT `request` doesn't do that itself, nor does a `Buffer`.

lib/actions/deploy.coffee
CameronDiver: Good point, will change.

package.json
CameronDiver: Yup, accidentally added it from an upcoming PR.

lib/actions/deploy.coffee
```diff
@@ -58,6 +63,15 @@ performUpload = (image, token, username, url, size, appName) ->
 
 	uploadToPromise(post, size)
 
+uploadLogs = (logs, token, url, buildId, username, appName) ->
+	request = require('request')
+	url = url || process.env.RESINRC_RESIN_URL
```
emirotin: This is the wrong way to get this setting as it ignores the YAML configs. Look for resin-settings-client usages here and in resin-sdk-preconfigured.

lib/actions/deploy.coffee
```diff
@@ -2,9 +2,14 @@ Promise = require('bluebird')
 dockerUtils = require('../utils/docker')
 
 getBuilderPushEndpoint = (baseUrl, owner, app) ->
-	escOwner = encodeURIComponent(owner)
-	escApp = encodeURIComponent(app)
-	"https://builder.#{baseUrl}/v1/push?owner=#{escOwner}&app=#{escApp}"
+	querystring = require('querystring')
```
emirotin: does it auto-escape?

CameronDiver: It does:
```
> querystring.stringify({ foo: 'Will "this" get "escaped?"' })
'foo=Will%20%22this%22%20get%20%22escaped%3F%22'
```

lib/actions/deploy.coffee
CameronDiver: That's weird, `require('resin-settings-client').get('resinUrl')` is where the `url` value comes from, but I was experiencing weirdness where when setting the `RESINRC_RESIN_URL` env var to `resindev.io` that it wasn't being mirrored in the settings call. I've just re-tested and it is now, so will change.

lib/actions/deploy.coffee
```diff
@@ -130,35 +148,69 @@ module.exports =
...
+			docker = dockerUtils.getDocker(options)
+			# Check input parameters
+			parseInput(params, options)
+			.then ([appName, build, source, imageName]) ->
+				tmpNameAsync()
+				.then (tmpPath) ->
+
+					# Setup the build args for how the build routine expects them
+					options = _.assign({}, options, { appName })
+					params = _.assign({}, params, { source: source })
```
emirotin: Why the change from the shortcut syntax?

CameronDiver: rebase hell... will change.

lib/actions/deploy.coffee
```diff
@@ -61,23 +65,36 @@ performUpload = (image, token, username, url, size, appName, logStreams) ->
...
-			obj = JSON.parse(data)
-			if obj.type?
-				switch obj.type
-					when 'error' then reject(new Error("Remote error: #{obj.error}"))
-					when 'success' then resolve(obj.image)
-					when 'status' then logging.logInfo(logStreams, "Remote: #{obj.message}")
-					else reject(new Error("Received unexpected reply from remote: #{data}"))
-			else
-				reject(new Error("Received unexpected reply from remote: #{data}"))
+			Promise.try ->
```
Page-: I think this should be a normal try/catch, having this weird hybrid `Promise.try` inside a `new Promise` is really weird

lib/actions/deploy.coffee
Page-: From https://github.com/request/request
> body - entity body for PATCH, POST and PUT requests. Must be a Buffer, String or ReadStream. If json is true, then body must be a JSON-serializable object.

That makes me think you can just pass the `Buffer` object directly

emirotin: I think it will be sent as octet-stream, not as base64-encoded, so the change on the server side will be needed as well.

CameronDiver: Yeah server-side changes will definitely be needed. I just deployed the current code for now, and I'll spend more time integrating the logs.

PR #495, opened 2017-04-24
Respect source parameter in resin build

CameronDiver:
Upon changing the name of the source parameter from `context`, some
places weren't changed, this commit fixes that.

Change-type: patch

### DIFF
--- a/lib/actions/deploy.coffee
+++ b/lib/actions/deploy.coffee
@@ -19,11 +19,11 @@ parseInput = Promise.method (params, options) ->
 		options.build = false
 		image = params.image
 	else if options.build
-		context = options.source || '.'
+		source = options.source || '.'
 	else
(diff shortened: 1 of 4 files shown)


lib/actions/deploy.coffee
```diff
@@ -133,7 +133,7 @@ module.exports =
 
 				# Setup the build args for how the build routine expects them
 				options = _.assign({}, options, { appName })
-				params = _.assign({}, params, { context })
```
Page-: There's still a `context` variable available from
```coffee
 		parseInput(params, options)
 		.then ([appName, build, context, imageName]) ->
```
I'm guessing that should be dealt with as part of this PR?

CameronDiver:
I'll leave this unmerged for as long as possible without it becoming a blocker then, @pimterry or @emirotin if you do get time please do review and send me a ping.

PR #498, opened 2017-04-25
Add cloud-builder builder output parity to build and deploy

CameronDiver:
Change-type: minor


lib/actions/deploy.coffee
```diff
@@ -120,6 +123,11 @@ module.exports =
 		tmpNameAsync = Promise.promisify(tmp.tmpName)
 		resin = require('resin-sdk-preconfigured')
 
+		dockerUtils = require('../utils/docker')
```
emirotin: How did it work before if this requirement was missing?

CameronDiver: That's actually a merge error. The import is now at the top of the file so the options can be defined in the context of `dockerUtils.appendOptions` but when merging in some other code I did I've accidentally added it here. Will remove, good spot :+1:

PR #505, opened 2017-04-29
proxy support

emirotin:
Seems to work, at least all the common functionality.
Fixes https://github.com/resin-io/resin-cli/issues/478


capitanodoc.coffee
```diff
@@ -0,0 +1,107 @@
...
+
+		Then authenticate yourself:
+
+			$ resin login
+
+		Now you have access to all the commands referenced below.
+
+		## Proxy support
+
+		The CLI has the first-class support for HTTP(S) proxies.
```
pimterry: Extra 'the' here, and I'm not sure what 'first-class' means in terms of proxy support? I'd just drop it to "The CLI supports HTTP(S) proxies".

Issue #507, opened 2017-05-01
Add note for (Check available types with _$ resin devices supported)


craig-mulligan:
Moved from https://github.com/resin-io/docs/pull/403

pimterry:
Fixes ages ago, closing.

PR #508, opened 2017-05-01
Add note on how to get resin device types

craig-mulligan:
Connects to #507

Change-Type: patch

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -159,7 +159,7 @@ Examples:
 
 ### Options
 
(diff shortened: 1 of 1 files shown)


emirotin:
@craig-mulligan this will be lost next time we rebuild the doc with `npm run doc`

pimterry:
Damn, you're completely right: @craig-mulligan this needs to be in the corresponding Capitano per-action option descriptions for this! Sorry, I should've caught that.

PR #510, opened 2017-05-02
Highlight cache usage in resin build

CameronDiver:
This commit will highlight the usage of the cache when doing a docker
build via `resin build`, which not only helps the user understand what
the build is doing, but also achieves more parity with the cloud
builder.

Change-type: patch


lib/utils/docker.coffee
```diff
@@ -110,13 +110,34 @@ exports.tarDirectory = tarDirectory = (dir) ->
...
+		arrowTest = /^\s*-+>\s*(.+)/i
+		if (match = arrowTest.exec(message))
+			match[1]
+		else
+			false
+
+	es.map (data, cb) ->
+		msg = extractArrowMessage(data)
+		if msg and msg.toLowerCase() == 'using cache'
+			data = colors.bgGreen.black(msg)
```
pimterry: Playing around with `colors`, it looks like if `msg` already has a color this won't work. Are we sure it always doesn't?

CameronDiver: I'm fairly sure it won't, unless Docker changes it's output format, but in that case other changes will be needed anyway.

lib/utils/docker.coffee
```diff
@@ -110,13 +110,33 @@ exports.tarDirectory = tarDirectory = (dir) ->
...
+	{ EOL } = require('os')
+
+	extractArrowMessage = (message) ->
+		arrowTest = /^\s*-+>\s*(.+)/i
+		if (match = arrowTest.exec(message))
+			match[1]
+		else
+			undefined
+
+	es.map (data, cb) ->
```
Page-: `es.mapSync` would be nicer:
```coffee
es.mapSync (data, cb) ->
	msg = extractArrowMessage(data)
	if msg? and msg.toLowerCase() == 'using cache'
		data = colors.bgGreen.black(msg)
	return data + EOL

File: lib/utils/docker.coffee [hunk 3]
```diff
@@ -110,13 +110,33 @@ exports.tarDirectory = tarDirectory = (dir) ->
 		pack.finalize()
 		return pack
 
+cacheHighlightStream = ->
+	colors = require('colors')
```
Page-: Can you use `colors/safe` to avoid messing with the string prototype (you already use the "safe" api below)

PR #511, opened 2017-05-04
Add emulated build option to resin build

CameronDiver:
This commit adds the ability to run a Docker build for an architecture
which is not the host architecture, using qemu-linux-user. Currently
this is only supported for linux.

Added:
* Installation of qemu which supports propagated execve flags
* Copying of qemu binary into the build context
* Transposing the given Dockerfile to use the qemu binary
* Intercepting of the build stream, so the output looks *almost* exactly
  the same.

Change-type: minor


lib/utils/docker.coffee
```diff
@@ -244,3 +290,72 @@ exports.getDocker = (options) ->
...
+
+copyQemu = (context) ->
+	path = require('path')
+	fse = require('fs-extra-promise')
+	# Create a hidden directory in the build context, containing qemu
+	binDir = path.join(context, '.resin')
+	binPath = path.join(binDir, 'qemu-execve')
+
+	fse.accessAsync(binDir)
+	.catch (e) ->
```
Page-:
```coffee
.catch code: 'ENOENT', (e) ->
	fse.mkdir(binDir)

File: lib/utils/docker.coffee [hunk 2]
```diff
@@ -244,3 +290,72 @@ exports.getDocker = (options) ->
...
+installQemu = ->
+	request = require('request')
+	fs = require('fs')
+	zlib = require('zlib')
+
+	getQemuPath()
+	.then (qemuPath) ->
+		installStream = fs.createWriteStream(qemuPath)
+
+		qemuUrl = 'https://github.com/resin-io/qemu/releases/download/v2.5.50-resin-execve/qemu-execve.gz'
```
Page-: It might be nicer to separate the version out of this string to make it more obvious where we can change it?

lib/utils/docker.coffee
```diff
@@ -244,3 +290,72 @@ exports.getDocker = (options) ->
...
+	resin = require('resin-sdk-preconfigured')
+	path = require('path')
+	fs = require('mz/fs')
+
+	resin.settings.get('binDirectory')
+	.then (binDir) ->
+		# The directory might not be created already,
+		# if not, create it
+		fs.access(binDir)
+		.catch (e) ->
```
Page-:
```coffee
.catch code: 'ENOENT', (e) ->
	fse.mkdir(binDir)

File: lib/utils/docker.coffee [hunk 4]
```diff
@@ -244,3 +290,72 @@ exports.getDocker = (options) ->
...
 	docker.getImage(image).inspectAsync()
 	.get('Size')
+
+hasQemu = ->
+	fs = require('mz/fs')
+
+	getQemuPath()
+	.then (qemu) ->
+		fs.stat(qemu)
+	.then ->
```
Page-:
```coffee
.return(true)
.catchReturn(false)

File: lib/utils/docker.coffee [hunk 5]
```diff
@@ -244,3 +290,72 @@ exports.getDocker = (options) ->
...
+		fs.access(binDir)
+		.catch (e) ->
+			if e.code == 'ENOENT'
+				fs.mkdir(binDir)
+			else
+				throw e
+		.then ->
+			path.join(binDir, 'qemu-execve')
+
+platformNeedsQemu = ->
```
pimterry: This is really 'supports', not 'needs', right?

lib/utils/docker.coffee
CameronDiver: Good idea

lib/utils/docker.coffee
CameronDiver: It's technically both, osx doesn't require qemu for emulation, and windows isn't supported. I can change the name if you think it's misleading.

pimterry: Oh, ok, I didn't realise OSX doesn't need it at all. In that case this makes totally sense :smile:.

lib/utils/docker.coffee
```diff
@@ -244,3 +290,65 @@ exports.getDocker = (options) ->
...
+
+	qemuVersion = 'v2.5.50-resin-execve'
+
+	getQemuPath()
+	.then (qemuPath) ->
+		installStream = fs.createWriteStream(qemuPath)
+		qemuUrl = "https://github.com/resin-io/qemu/releases/download/#{qemuVersion}/qemu-execve.gz"
+		request(qemuUrl)
+		.pipe(zlib.createGunzip())
+		.pipe(installStream)
```
Page-: Should we be waiting on this stream to finish?

package.json
```diff
@@ -43,11 +43,13 @@
     "coffee-script": "~1.12.4",
     "columnify": "^1.5.2",
     "denymount": "^2.2.0",
+    "docker-qemu-transpose": "^0.2.1",
     "docker-toolbelt": "^1.3.3",
     "dockerode": "^2.4.2",
     "drivelist": "^5.0.16",
     "etcher-image-write": "^9.0.3",
     "global-tunnel-ng": "^2.0.0",
+    "fs-extra-promise": "^1.0.1",
```
Page-: `fs-extra-promise` and `mz`? Anywhere we can limit the number of these deps?

lib/utils/docker.coffee
```diff
@@ -137,15 +143,32 @@ exports.runBuild = (params, options, getBundleInfo, logStreams) ->
...
 	# The default build context is the current directory
 	params.source ?= '.'
 	logs = ''
+	# Only used in emulated builds
+	qemuPath = ''
 
-	# Tar up the directory, ready for the build stream
-	tarDirectory(params.source)
+	Promise.try ->
+		if options.emulated and platformNeedsQemu()
```
emirotin: can be `return if not (options.emulated and platformNeedsQemu())` and the rest of the method dedented

lib/utils/docker.coffee
CameronDiver: We need it to definitely be there when the build begins, as it gets copied into the build context. I feel like this is the safest method to ensure the binary will be present and fully downloaded.

package.json
CameronDiver: I've taken out fs-extra-promise.

lib/utils/docker.coffee
Page-: So is that a yes?

lib/utils/docker.coffee
```diff
@@ -266,3 +313,67 @@ exports.getDocker = (options) ->
...
+
+	fs.access(binDir)
+	.catch code: 'ENOENT', ->
+		fs.mkdir(binDir)
+	.then ->
+		getQemuPath()
+	.then (qemu) ->
+		read = fs.createReadStream(qemu)
+		write = fs.createWriteStream(qemu)
+		read.pipe(write)
```
Page-: Do we need to wait for this to finish?

lib/utils/docker.coffee
```diff
@@ -266,3 +313,67 @@ exports.getDocker = (options) ->
 exports.getImageSize = (docker, image) ->
 	docker.getImage(image).inspectAsync()
 	.get('Size')
+
+hasQemu = ->
+	fs = require('mz/fs')
+
+	getQemuPath()
+	.then (qemu) ->
+		fs.stat(qemu)
```
Page-: Any chance
```coffee
.then(fs.stat)
```
would work?

lib/utils/docker.coffee
CameronDiver: It is indeed.

CameronDiver: ...I'm an idiot.

lib/utils/docker.coffee
```diff
@@ -266,3 +313,73 @@ exports.getDocker = (options) ->
...
+	.catch code: 'ENOENT', ->
+		fs.mkdir(binDir)
+	.then ->
+		getQemuPath()
+	.then (qemu) ->
+		new Promise (resolve, reject) ->
+			read = fs.createReadStream(qemu)
+			write = fs.createWriteStream(qemu)
+			read
+			.pipe(write)
```
Page-: Doesn't this just read a file and write it over itself?

lib/utils/docker.coffee
```diff
@@ -266,3 +313,73 @@ exports.getDocker = (options) ->
...
+			.pipe(installStream)
+			.on('error', reject)
+			.on('finish', resolve)
+
+copyQemu = (context) ->
+	path = require('path')
+	fs = require('mz/fs')
+	# Create a hidden directory in the build context, containing qemu
+	binDir = path.join(context, '.resin')
+	binPath = path.join(binDir, 'qemu-execve')
```
Page-: What's the difference between this and the `getQemuPath`?

lib/utils/docker.coffee
```diff
@@ -266,3 +313,73 @@ exports.getDocker = (options) ->
...
+platformNeedsQemu = ->
+	os = require('os')
+	os.platform() == 'linux'
+
+installQemu = ->
+	request = require('request')
+	fs = require('fs')
+	zlib = require('zlib')
+
+	qemuVersion = 'v2.5.50-resin-execve'
```
Page-: This would be nice as a constant at the top of the module, so it's easy to see/change it

lib/utils/docker.coffee
```diff
@@ -156,15 +162,33 @@ exports.runBuild = (params, options, getBundleInfo, logStreams) ->
...
+		hasQemu()
+		.then (present) ->
+			if !present
+				logging.logInfo(logStreams, 'Installing qemu for ARM emulation...')
+				installQemu()
+		.then ->
+			# Copy the qemu binary into the build context
+			copyQemu(params.source)
+		.then (binPath) ->
+			qemuPath = binPath.split(path.sep)[1...].join(path.sep)
```
Page-: Why are we dropping the first element of the path?

CameronDiver: The path of the binary before this is:
```
[context location]/.resin/qemu
```
but inside of the build, this should be
```
.resin/qemu
```

lib/utils/docker.coffee
CameronDiver: `getQemuPath` is the global path of qemu on the system. `binPath` here is the path inside the local build context.

lib/utils/docker.coffee
CameronDiver: It does, I r idiot.

lib/utils/docker.coffee
CameronDiver: Agreed.

lib/utils/docker.coffee
```diff
@@ -182,32 +209,55 @@ exports.runBuild = (params, options, getBundleInfo, logStreams) ->
...
+					logThroughStream = es.through (data) ->
 						logs += data.toString()
 						this.emit('data', data)
 
-					stream
-					.pipe(throughStream)
+					if options.emulated and platformNeedsQemu()
+						buildThroughStream = transpose.getBuildThroughStream
+							hostQemuPath: qemuPath
+							containerQemuPath: './qemu-execve'
```
Page-: `qemu-execve` is hardcoded all over the place, should it be moved to a constant?

lib/utils/docker.coffee
```diff
@@ -156,15 +164,34 @@ exports.runBuild = (params, options, getBundleInfo, logStreams) ->
...
+		hasQemu()
+		.then (present) ->
+			if !present
+				logging.logInfo(logStreams, 'Installing qemu for ARM emulation...')
+				installQemu()
+		.then ->
+			# Copy the qemu binary into the build context
+			copyQemu(params.source)
+		.then (binPath) ->
+			console.log(binPath)
```
CameronDiver: Oh yeah I accidentally left it in from when testing to answer your previous review.

pimterry:
@CameronDiver What's blocking this now? Is it just @Page-'s approval to the latest updates?

Just found this again, because I've moved from my OSX to my shiny new Linux machine, and while `resin deploy` works great with the pi on OSX, it's currently  unusable on Linux without this patch.

CameronDiver:
Yup, my bad should have been chasing it up. To tell you the truth it's been open so long I completely forgot about it.

PR #514, opened 2017-05-10
Add progress bar for upload progress

dfunckt:
This replaces the "Uploading XX.X%" info string with a progress bar similar to the cloud builder, along these lines:

```
[Info]    [========>                                          ] 17.4%
```

Change-Type: patch


lib/actions/deploy.coffee
```diff
@@ -31,11 +31,21 @@ parseInput = Promise.method (params, options) ->
...
+	spaceCount = stepCount - barCount
+	bar = "[#{_.repeat('=', barCount)}>#{_.repeat(' ', spaceCount)}]"
+	return "#{bar} #{percentage.toFixed(1)}%"
+
 pushProgress = (imageSize, request, logStreams, timeout = 250) ->
 	logging = require('../utils/logging')
 	ansiEscapes = require('ansi-escapes')
 
-	logging.logInfo(logStreams, 'Initialising...')
+	logging.logInfo(logStreams, 'Initializing...')
```
dfunckt: Sorry about that :)

lib/actions/deploy.coffee
```diff
@@ -31,11 +31,21 @@ parseInput = Promise.method (params, options) ->
...
 
+# Builds and returns a Docker-like progress bar like this:
+# [==================================>               ] 64%
+renderProgress = (percentage, stepCount = 50) ->
+	_ = require('lodash')
+	percentage = Math.max(0, Math.min(percentage, 100))
+	barCount = stepCount * percentage // 100
+	spaceCount = stepCount - barCount
+	bar = "[#{_.repeat('=', barCount)}>#{_.repeat(' ', spaceCount)}]"
+	return "#{bar} #{percentage.toFixed(1)}%"
```
pimterry: This is now our third separate progress bar implementation. We've got one in the [docker-progress](https://github.com/resin-io/docker-progress/blob/master/index.coffee) that's almost but not _exactly_ the same as this, one in [resin-cli-visuals](https://github.com/resin-io-modules/resin-cli-visuals/blob/master/lib/widgets/progress.coffee), and this one.

It'd be nice to be consistent, and use one progress bar everywhere. Can we use the resin-cli-visuals bar we already use in the CLI [for image downloads](https://github.com/resin-io/resin-cli/blob/dede4bb3293a867f2dfdeabe0c9492ebff1acf7f/lib/actions/os.coffee#L111) for this too?

In a perfect world, I'd also quite like to change the UI to use that too, and have a standard resin style for our CLI interfaces (since Etcher also uses this one), but one step at a time :-).

lib/actions/deploy.coffee
```diff
@@ -31,11 +31,21 @@ parseInput = Promise.method (params, options) ->
 
 	return [appName, options.build, source, image]
 
+# Builds and returns a Docker-like progress bar like this:
+# [==================================>               ] 64%
+renderProgress = (percentage, stepCount = 50) ->
+	_ = require('lodash')
+	percentage = Math.max(0, Math.min(percentage, 100))
```
pimterry: Nit: Clearer imo as `_.clamp(percentage, 0, 100)`

lib/actions/deploy.coffee
pimterry: 😢

pimterry:
One more thing, sorry, you also need a changelog for this. The CLI doesn't have versionist yet, it's manual changelog updates for the moment.

lib/actions/deploy.coffee
dfunckt: Indeed, but unfortunately `_.clamp` is Lodash 4.

pimterry: Ah, very good point.

lib/actions/deploy.coffee
dfunckt: Agreed, however none of the two can be reused: it is a pity to pull in the whole docker-progress as a dependency just for this and cli-visuals requires an `eta` argument that's not convenient to calculate in this case. I think I'll leave it as is for now to get this out soon, but let's revisit sometime soon.

pimterry: Ok, that's fine then. I've filed an issue to fix that against the resin-cli-visuals: https://github.com/resin-io-modules/resin-cli-visuals/issues/42.

PR #515, opened 2017-05-10
Add ability to specify built-time variables for local build

dfunckt:
This allows users to specify build-time variables for local builds. Also available when building and deploying via `resin deploy`.

Change-Type: patch


lib/utils/docker.coffee
```diff
@@ -48,6 +48,12 @@ exports.appendOptions = (opts) ->
 			alias: 't'
 		},
 		{
+			signature: 'arg'
+			parameter: 'arg'
+			description: 'Set a build-time variable (eg. "-e \'ARG=value\'"). Can be specified multiple times.'
+			alias: 'e'
```
pimterry: Why `e`? `arg`/`a`?

CameronDiver: `-a` is taken on `resin build`, and e could stand for env-vars, but there's no mention of env-vars either so it could use some thought.

dfunckt: `a` is used as an alias to `application` and thought that given `v` (for "variable") is unavailable (didn't actually check, I just assume is used for an alias to `verbose/verbosity`), `e` (for "env") is the next best choice. Happy to change it if there are better ideas.

CameronDiver: I think `--env/-e` makes the most sense IMO

dfunckt: We do use `--env/-e` in `local push` and it's a very similar option, I just went by Docker's convention of calling them "Build Args": https://docs.docker.com/engine/reference/commandline/build/

lib/utils/docker.coffee
```diff
@@ -128,6 +134,19 @@ cacheHighlightStream = ->
 			data = colors.bgGreen.black(msg)
 		return data + EOL
 
+parseBuildArgs = (args, onError) ->
+	_ = require('lodash')
+	if not _.isArray(args)
+		args = [ args ]
+	obj = {}
```
CameronDiver: super NIT: I'd prefer this was something like `envVars = {}` rathan than `obj`.

lib/utils/docker.coffee
pimterry: I'd love `--build-arg` (to match docker exactly) and `-b`, but `b` is taken in `resin build`. `--build-arg`/`ba`?

CameronDiver: In the interest in consistency amongst our own utils, I think it should be env. `local push` is a Docker build command, just on device rather than locally. So if we changed it to `--build-arg` we'd have to change `local push` as well which is a breaking change.

dfunckt: How about `--arg/-A`? Not sure how Capitano handles option case though, will have to test.

pimterry: We've been talking about making big breaking changes to `local push` anyway though, so we could make this inconsistent in the short-term and then break `local push` to match shortly.

We don't necessarily need to worry about consistency there, if we think we are changing that anyway, as long as we think this is the right thing to do. If we don't of course, that's a different question :smiley:.

If we can't do build-arg, `A` sounds like an ok compromise to me though.

dfunckt: The thing with `--build-arg` is that its case is inconsistent with our other multi-word options (eg. `--dockerHost`) and the "build" part annoys me as superfluous (and it's the reason I omitted it in the first place). I'm fine overall with `--build-arg/-B` for the sake of consistency with Docker if others agree, however this highlights an issue with our use of names for options across the CLI.

pimterry: That is a good argument (if you'll pardon the pun). We definitely can't break the whole cli, so it'll at least have to be camel-case.

I do quite like `buildArg` over `arg` alone though - it feels clearer, and if you need it to be terse you can use the alias. And the `build` isn't always superfluous anyway, in `resin deploy --arg`, it's not clear whether the arg is for the build part or the deployed code. `resin deploy --buildArg` is very clear.

dfunckt: @CameronDiver I went with @pimterry's suggestion: `--buildArg/-B`.

dfunckt: Forgot to mention `-A` is used as an alias to `--arch`

PR #517, opened 2017-05-11
resin ssh proxy support

emirotin:


lib/app.coffee
```diff
@@ -38,6 +38,9 @@ catch
 # If that is not set as well the initialize will do nothing
 globalTunnel.initialize(proxy)
 
+# TODO: make this a feature of capitano
```
pimterry: Can you file an issue for this on capitano itself? Inline TODOs just get lost.

lib/actions/ssh.coffee
```diff
@@ -98,11 +78,20 @@ module.exports =
...
 				throw new Error('Did not find running application container') if not containerId?
 				Promise.try ->
+					sshProxyCommand = ''
+					proxyConfig = global.PROXY_CONFIG
+					if proxyConfig
+						{ proxyAuth } = proxyConfig
+						proxyHost = "-p #{proxyConfig.host}:#{proxyConfig.port}"
+						proxyAuth = if proxyAuth then "-P #{proxyAuth}" else ''
+						proxytunnelCommand = "proxytunnel #{proxyHost} #{proxyAuth} -d %h:%p"
+						sshProxyCommand = "-o ProxyCommand='#{proxytunnelCommand}'"
```
pimterry: With this, it's impossible to use an HTTP proxy with the CLI unless you also proxy SSH too. That means if you use a proxy at all, you can't SSH unless you install proxytunnel.

That's a bit annoying, as I expect most people using proxies use them for HTTP(S) only, not for 100% of the traffic from their machines. Anybody who isn't on a network with SSH blocked doesn't need this extra hassle. Could we warn and try anyway, instead of letting `ssh` itself fail here? We could detect if `proxytunnel` is in PATH, and do this if so, but otherwise just print a warning:

> [WARN] Proxy configuration set, but proxytunnel is not in your $PATH. Connecting with SSH directly...'

That would also give a clearer explanation for people who don't read the instructions and fail to install proxytunnel/set it up correctly when they do need it.

lib/actions/ssh.coffee
```diff
@@ -98,11 +78,20 @@ module.exports =
 			.then ({ username, uuid, containerId, proxyUrl }) ->
 				throw new Error('Did not find running application container') if not containerId?
 				Promise.try ->
+					sshProxyCommand = ''
+					proxyConfig = global.PROXY_CONFIG
+					if proxyConfig
+						{ proxyAuth } = proxyConfig
+						proxyHost = "-p #{proxyConfig.host}:#{proxyConfig.port}"
+						proxyAuth = if proxyAuth then "-P #{proxyAuth}" else ''
```
pimterry: Oh, also just spotted a couple more points here:

* We should escape and quote `proxyAuth`. I think the other params are ok, but (assuming this is a username:password pair?) this could easily contain a single quote or a space, both of which would break this. Might be other cases that could cause trouble too, I'm not sure.

* Is `-P` right? That's not a listed argument in the proxytunnel docs: http://proxytunnel.sourceforge.net/usage.php

emirotin: Good catch, I've used the -P from some article on the net, could it be renamed in a newer version of proxytunnel?

Speaking about escaping - good point as well, needs investigation

emirotin:
Some fixes implemented, remaining:
- escape shell args for user and password
- detect if proxytunnel is present, warn if not

emirotin:
The escaping lib I've used is pretty naive (it doesn't add outer quotes so it has to escape spaces and all the special chars), but I don't think it's a big deal

Here'a s sample output:
```
❯ https_proxy=https://us\'er:password@host:1234 ./bin/resin ssh 1234
ssh -t -o LogLevel=ERROR -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ControlMaster=no -o ProxyCommand=proxytunnel\ --proxy\=host\\\:1234\ --dest\=\\\%h\\\:\\\%p\ --user\=us\\\'er\ --pass\=password -p 123 aaa@ssh.resin.proxy enter 1234 xxx

```

PR #548, opened 2017-06-09
non-interactive device init

emirotin:


doc/automated-init.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Build the config file
+
+Interactive device provisioning process often includes collecting some extra device configuration, like the networking mode and wifi credentials.
+
+To skip this interactive step we need to buid this configuration once and save it to the JSON file for later reuse.
+
+Let's say we will place it into the `CONFIG_FILE` path, like _./resin-os/raspberrypi3-config.json_.
+
+We also need to put the OS image somewhere, let's call this path `OS_IMAGE_PATH`, it cam be something like _./resin-os/raspberrypi3-v2.0.6+rev1.prod.img_.
```
thgreasi: typo `can`

doc/automated-init.md
```diff
@@ -0,0 +1,113 @@
...
+
+1. First we need to download the OS image once. That's needed for building the config, and will speedup the consequent operations as the downloaded OS image is placed into the local cache.
+
+	Run:
+	```bash
+	resin os download DEVICE_TYPE --output OS_IMAGE_PATH --version OS_VERSION
+	```
+
+1. Now we're ready to build the config:
```
thgreasi: Use `2` for the second step :)

doc/automated-init.md
```diff
@@ -0,0 +1,113 @@
...
+
+	```bash
+	cat CONFIG_FILE
+	```
+
+## Done
+
+Now you're ready to run the command in the beginning of this guide.
+
+Please not again that all of these steps only need to be done once (unless you need to change something), and once all the parameters are collected the main init command can be run unchanged.
```
thgreasi: typo `note`

lib/actions/os.coffee
```diff
@@ -134,50 +144,125 @@ exports.download =
...
+			init.configure(params.image, params.uuid, answers).then(helpers.osProgressHandler)
 		.nodeify(done)
 
-initWarningMessage = '''
+exports.availableDrives =
+	# TODO: dedupe with https://github.com/resin-io-modules/resin-cli-visuals/blob/master/lib/widgets/drive/index.coffee
+	signature: 'os available-drives'
+	description: 'list available drives'
+	action: ->
+		Promise = require('bluebird')
```
thgreasi: I guess that we are going to be promises a lot. Could we move this to the top, like with lodash?

doc/automated-init.md
emirotin: No, I don't have to, that's the valid syntax for md lists if you don't want to keep updating the numbers when reordering or inserting / deleting items

lib/actions/os.coffee
emirotin: We use the local imports almost everywhere to speedup the runtime. Lodash is at the top because it's anyway imported in the top-level in some util file, and also because it's needed to build the option at start time.

emirotin: Ah, and node require's are cached, so multiple imports are not hitting the performance

doc/automated-init.md
thgreasi: Oh, I see, sorry for the noise, didn't know.
That's handy!

emirotin: No need to excuse :)

doc/automated-init.md
```diff
@@ -0,0 +1,113 @@
...
+
+You can configure the `resin` CLI command to be sudo-runnable without the password. Check [this post](https://askubuntu.com/questions/159007/how-do-i-run-specific-sudo-commands-without-a-password) for an example.
+
+### Extra initialization config
+
+As of June 2017 all the supported devices should not require any other interactive configuration.
+
+But by the design of our system it is _possible_ (though it doesn't look very likely it's going to happen any time soon) that some extra initialization options may be requested for the specific device types.
+
+If that is the case please raise the issue in the resin CLI repository and the maintainers will add the necessary options to build the similar JSON config for this step.
```
pimterry: I think we should remove this last section, and add some safeguards against this instead.

Generally, I think for any command we allow you to do non-interactively (e.g. `os configure` with `--config`, almost anything with `--yes`) we should throw if there's anything where we need to ask for input.

We could do this thoroughly and automatically with a CLI-wide non-interactive state flag, and making all interactive operations fail if that happens. Should make it harder for us to accidentally do this, and will mean we instantly get error reports if anybody ever hits it. Makes it easier for them to debug too: they don't end up with a broken process waiting for input, they clear an instant and clear error instead.

doc/automated-init.md
```diff
@@ -0,0 +1,113 @@
...
+
+#### Option 2: `NOPASSWD` directive
+
+You can configure the `resin` CLI command to be sudo-runnable without the password. Check [this post](https://askubuntu.com/questions/159007/how-do-i-run-specific-sudo-commands-without-a-password) for an example.
+
+### Extra initialization config
+
+As of June 2017 all the supported devices should not require any other interactive configuration.
+
+But by the design of our system it is _possible_ (though it doesn't look very likely it's going to happen any time soon) that some extra initialization options may be requested for the specific device types.
```
pimterry: I'm not sure this is possible, is it? If you pass `--config`, then `os configure` never calls `buildConfig`, which is the only place I can see in `device init` that might ask any questions. I'm not sure, but I suspect actually you'll get some obscure lower level error or weird behaviour when the config isn't sufficient. It'd actually be interesting to know for sure what happens in that case. Can you test it and see? Should just require removing a param from the generated config.

CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - `package-lock.json` for `npm5` users
+- Forced update to the newest resin-sdk (^6.4.1)
+- Allow OS version selection when doing `resin device init`
+- Actually tolerate the `--yes` param to `resin device init`
+- Allows passing `--drive` to `resin device init`
+- List detected drives with `resin os available-drives`
+- Add the `resin os build-config` method to pass the interactive config step once
+and reuse the built file for consequent `resin os configure` calls (added the new `--config` param to it),
```
pimterry: 'consequent' means something like 'resulting'. I think you want 'subsequent'

doc/automated-init.md
```diff
@@ -0,0 +1,113 @@
...
+
+Interactive device provisioning process often includes collecting some extra device configuration, like the networking mode and wifi credentials.
+
+To skip this interactive step we need to buid this configuration once and save it to the JSON file for later reuse.
+
+Let's say we will place it into the `CONFIG_FILE` path, like _./resin-os/raspberrypi3-config.json_.
+
+We also need to put the OS image somewhere, let's call this path `OS_IMAGE_PATH`, it can be something like _./resin-os/raspberrypi3-v2.0.6+rev1.prod.img_.
+
+1. First we need to download the OS image once. That's needed for building the config, and will speedup the consequent operations as the downloaded OS image is placed into the local cache.
```
pimterry: 'consequent' again

CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - `package-lock.json` for `npm5` users
+- Forced update to the newest resin-sdk (^6.4.1)
+- Allow OS version selection when doing `resin device init`
+- Actually tolerate the `--yes` param to `resin device init`
+- Allows passing `--drive` to `resin device init`
+- List detected drives with `resin os available-drives`
```
emirotin: I'm very positive about including it. It's a nice thing to have (because of the system drives filtering), and we already have this functionality. Yes it's not related to the OS per se, but it's the best NS we have as it's still related to `os initialize`

CHANGELOG.md
emirotin: Right

doc/automated-init.md
```diff
@@ -0,0 +1,113 @@
+# Provisioning Resin.io devices in automated (non-interactive mode)
+
+This document describes how to run the `device init` command in non-interactive mode.
+
+It requires collecting some preliminary information _once_.
+
+The final command to provision the device looks like this:
+
+```bash
+resin device init --app APP_ID --os-version OS_VERSION --drive DRIVE --config "$(cat CONFIG_FILE)" --yes
```
emirotin: We use the full JSON in other places already (like `internal osinit`) and I feel like it's just more flexible. Like if it takes the config as a string you can easily cat it from the file, but not vice versa.

doc/automated-init.md
emirotin: There are **two** config groups. One is config and includes things like network config.
It's covered by the `build-config` and `--config`.

The other is called _initialization_ and for almost every device it only includes the **drive** question (except Edison which doesn't have it at all). Potentially this section can have other questions too and they will be asked interactively. But as of now it's not the case.

doc/automated-init.md
emirotin: It's hard to handle because we often have lots of subcommands called from the original command.
There's no universal `--non-interactive` switch. `--config` covers the configuration step (see explanation above), and `--yes` only covers the _confirmations_ (Are you sure you want to explode your machine?). But if the info is missing right now the nested commands have no indication they should bail because they're operating in non-interactive mode.

lib/actions/device.coffee
```diff
@@ -374,6 +376,20 @@ exports.init =
...
 			alias: 'v'
 		}
+		_.assign({}, commandOptions.osVersion, { signature: 'os-version', parameter: 'os-version' })
+		{
+			signature: 'drive'
+			description: 'the drive to write the image to, like /dev/sdb.
+				Careful with this as you can erase your hard drive.
+				Check `resin os available-drives` for available options.'
+			parameter: 'drive'
+			alias: 'd'
```
emirotin: Probably yes

lib/actions/os.coffee
```diff
@@ -134,50 +144,125 @@ exports.download =
...
+		Examples:
+
+			$ resin os build-config ../path/rpi3.img raspberrypi3 --output rpi3-config.json
+			$ resin os configure ../path/rpi3.img 7cf02a6 --config "$(cat rpi3-config.json)"
+	'''
+	permission: 'user'
+	options: [
+		{
+			signature: 'advanced'
+			description: 'show advanced commands'
```
emirotin: Hm, did I write it?  ¯\_(ツ)_/¯ I obviously have meant "options"
yes I will improve both

lib/actions/os.coffee
```diff
@@ -134,50 +144,125 @@ exports.download =
...
+			boolean: true
+			alias: 'v'
```
emirotin: Probably yes

doc/cli.markdown
```diff
@@ -862,6 +909,14 @@ Examples:
 
 show advanced commands
 
+#### --config &#60;config&#62;
+
+stringified JSON with the device config, see `resin os build-config`
+
+## os available-drives
+
+undefined
```
emirotin: Hm, strange, why is it? Will investigate

CHANGELOG.md
pimterry: Ok, we can keep it, but I'd still rather have it outside `os`. If we can't come up with anything better then `resin available-drives` feels better than `resin os available-drives` to me (although I would rather namespace it if possible).

doc/automated-init.md
pimterry: It's definitely not the normal bash convention though. I can think of lots of tools that take config filenames on the command line, but not a single one that takes the full config raw.

I'm not convinced the flexibility is going to be frequently necessary for this, but I can see how it'd be nice. The normal convention for that is to pass `-` though, and interpret that as 'read this from stdin'.

Alternatively, even if you don't want to explicitly support that pattern, there's a whole bunch of standard bash conventions for creating temporary names that refer to pipes, which let you support exactly this case without any effort in our code at all: https://serverfault.com/questions/398514/pass-a-pipe-to-a-command-that-expects-a-filename

doc/automated-init.md
pimterry: Hm, ok. It's a bit dubious, but we could make `--yes` mean `--non-interactive`? I can't think of any case where you'd want the CLI to be interactive for input, but skip any confirmations, that seems very unusual. Then we could turn on the flag with `--yes`, allow you to pass it everywhere, and pass it into subcommands if it's set.

That's a bit extreme and needs some thought, so it's probably not worth it for now I guess, although I would quite like to find a nice solution that'd stop us shooting ourselves in the foot here in future. Can wait until after this PR though, since I don't think there's going to be a nice clean easy one, you're probably right.

lib/actions/os.coffee
emirotin: OK I'm innocent, I've just copied it from a different place below :P Anyway fixing now

emirotin:
Updated, please re-review

doc/automated-init.md
```diff
@@ -0,0 +1,112 @@
...
+	```bash
+	resin os versions DEVICE_TYPE
+	```
+	and pick the version that you need, like _v2.0.6+rev1.prod_.
+	_Note_ that even though we support _semver ranges_ it's recommended to use the exact version when doing the automated provisioning as it
+	guarantees full compatibility between the steps.
+
+1. `DRIVE`. Plug in your target medium (SD card or the USB stick, depending on your device type) and run
+	```bash
+	resin util available-drives
```
pimterry: Nice! Yep, I'm fine with this, good call.

Issue #549, opened 2017-06-09
resin deploy should gzip the stream before uploading


CameronDiver:
This is a simple fix, just adding a node gzip stream to the outgoing data, and also adding a gzip uncompress stream on the builder side.

pimterry:
@CameronDiver is there an issue for the uncompress side of this in the builder? Very easy to put into the CLI now, but tricky to test if it's working without the other side of it.

CameronDiver:
There is: https://github.com/resin-io/resin-builder/issues/187

PR #554, opened 2017-06-14
Gzip images when deploying

pimterry:
Connects to #549

This gzips the Docker image stream before uploading it. https://github.com/resin-io/resin-builder/issues/187 doesn't actually seem to be required - I think express automatically decodes the body if you send it with the right headers. Works fine with no changes in my devenv and in prod, and it's significantly quicker than without gzipping (in prod, for a simple node hello world, a full `resin deploy --build` takes me 1m47 vs 2m40 on master).

The gzipping here is very easy - the tricky bit is providing a useful progress bar whilst doing so. I've gone for tracking the progress of draining the raw stream (that's piped into the gzip stream that's piped into the request), since it's quite a bit quicker than gzipping the whole buffer up front, which is the only easy way to get the gzipped size I can think of. If it's different to the actual request progress, it's imperceptibly so.


lib/actions/deploy.coffee
```diff
@@ -186,21 +193,20 @@ module.exports =
 							{ image: imageName, log: '' }
 					.then ({ image: imageName, log: buildLogs }) ->
 						logs = buildLogs
-						Promise.join(
-							dockerUtils.bufferImage(docker, imageName, tmpPath)
+						Promise.all [
```
emirotin: Why switching here? `.join` is known t be more performant

lib/utils/docker.coffee
```diff
@@ -281,29 +281,20 @@ exports.runBuild = (params, options, getBundleInfo, logStreams) ->
...
-			.pipe(stream)
-	.then ->
-		new Promise (resolve, reject) ->
-			fs.createReadStream(tmpFile)
-			.on 'open', ->
-				resolve(this)
-			.on('error', reject)
+	imageMetadata = image.inspectAsync()
+
+	Promise.all([image.get(), imageMetadata.get('Size')])
```
emirotin: Same: `.join` can be used

lib/actions/deploy.coffee
CameronDiver: I personally do find this slightly clearer to read here, I wasn't very keen on the `Promise.join` when I wrote it, this at least makes it obvious that the `performUpload` is the function.

lib/utils/docker.coffee
CameronDiver: Conversely I find this is a little less clear - not by much mind you - but I would probably use a `.join` here.

lib/actions/deploy.coffee
pimterry: Ok, I'm going to leave this one

lib/utils/docker.coffee
pimterry: I've fixed this one

PR #556, opened 2017-06-15
Resin deploy ux improvements

pimterry:
A couple of other small changes I noticed along the way with the deploy gzipping stuff, pulled out as a separate PR. This PR:

* Shows the correct error if a build fails (instead of 'could not delete `bufferFile`', which happened before, because the `finally` step fails in that case)
* Shows an 'Initializing deploy...' message after the build, so you can tell what's going on during the couple of seconds gap while we get the docker image and buffer it to disk.
* Tweaks the docs slightly for clarity (mainly to be consistent about deploying _images_, not containers)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - `package-lock.json` for `npm5` users
 - Added ability to run an emulated build silently with `resin build`
 - Gzip images when uploading in `resin deploy`
+- Show a clear message immediately as the deploy starts, if we're deploying an image.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


CameronDiver:
This PR reminds me about an issue that @imrehg found, and I think it would kinda make sense to also handle it here. Basically he was using `resin deploy` and aiming it at a raspberry pi docker daemon. This is something that I never tested, but basically there is a significant delay with this method on the `image.get` call (exporting to a tar archive from the remote to the client/host), so we should add at least an output message there informing the user what is going on.

pimterry:
With this PR, you now get the `Initializing deploy...` message before that `image.get()` starts, so there's a little feedback here (although not in detail, and not a progress bar or anything). Is that what you were looking for, or something more in depth?

PR #557, opened 2017-06-15
Fix issue when using resin deploy with non-standard stdin

### LINKED ISSUE #539: clearLine is not a function error with resin deploy on windows
Running a `resin deploy` on Windows in a setup where the builder is a local-mode RPi3, with a command like:

```
resin deploy appname --build --dockerHost rpi3.local --dockerPort 2375  -s .
```
The build finishes successfully, unicorn is shown, and while waiting for the container to started to be uploaded, instead of getting a progress bar (after a wait), `[Initializing]` (or something similar) pops up for a brief moment, then I get this error:

```
TypeError: process.stdout.clearLine is not a function
    at Timeout._onTimeout (C:\Users\Gergely\AppData\Roaming\npm\node_modules\resin-cli\build\actions\deploy.js:84:20)
    at ontimeout (timers.js:386:14)
    at tryOnTimeout (timers.js:250:5)
    at Timer.listOnTimeout (timers.js:214:5)
```

Using resin-cli 5.10.1 on Windows 10, with node v7.10.0, in git bash. On Linux (same node and resin-cli versions) the same `deploy` command seems to work fine.

pimterry:
Connects to #539

With non-TTY inputs, and other weird stdin cases (e.g. git bash) functions like `process.stdin.clearLine()` are not defined. The resin-cli-visuals progress bar doesn't depend on these, so this PR moves to use that instead to fix #539, which currently breaks `resin deploy` for all git bash users on windows.

This also lets us neatly gives us an ETA for the image push too, for free (it's already being calculated by progress-stream).

Output looks like:

```bash
[Build]   Using cache
[Build]    ---> 5a029c04a533
[Build]   Successfully built 5a029c04a533

...unicorn...

[Info]    Initializing deploy...
Deploying [============            ] 50% eta 36s  
```

One downside is that this now doesn't include the coloured `[Info]` prefix. @CameronDiver how important is that? I could try to extend resin-cli-visuals to accept  arbitrary streams to write to, rather than using console.log, but it's a little fiddly, and most other commands don't use this log streaming anyway (and neither does the unicorn output). Thoughts?


lib/actions/deploy.coffee
```diff
@@ -31,30 +31,17 @@ parseInput = Promise.method (params, options) ->
...
-showPushProgress = (logStreams) ->
-	logging = require('../utils/logging')
-	logging.logInfo(logStreams, renderProgress(0))
-
-updatePushProgress = (percentage, logStreams) ->
-	logging = require('../utils/logging')
-	ansiEscapes = require('ansi-escapes')
+showPushProgress = ->
+	visuals = require('resin-cli-visuals')
+	progressBar = new visuals.Progress('Deploying')
```
CameronDiver: If we did want to add prefixes, this function could be used: https://github.com/resin-io-modules/stream-logger/blob/master/src/index.ts#L53, and maybe could switch to `[Success]` when completed.

lib/actions/deploy.coffee
```diff
@@ -31,30 +31,17 @@ parseInput = Promise.method (params, options) ->
...
+	progressBar.update({ percentage: 0 })
+	return progressBar
 
+updatePushProgress = (percentage, eta, progressBar) ->
```
dfunckt: I think this is unused now?

package.json
```diff
@@ -71,7 +71,7 @@
     "resin-cli-auth": "^1.1.3",
     "resin-cli-errors": "^1.2.0",
     "resin-cli-form": "^1.4.1",
-    "resin-cli-visuals": "^1.3.0",
+    "resin-cli-visuals": "^1.4.0",
```
emirotin: Needs rebasing, it's already on master

pimterry:
Sorry that took me a little while - this is now updated to add a prefix following @CameronDiver's suggestion. Output now looks like:

```
[Info]    Initializing deploy...
[Info]    Deploying [========================] 100% eta 0s            
[Info]    Remote: Accepting image for application picam, owner: gh_pimterry
```

Doesn't replace the progress bar with success, both because that's not what the builder does and because it's fiddly, but it's now at least a bit tidier and more consistent.

Rereview please - notably to make this happen I've refactored the logStreams code substantially, so that we can just pass around a logger object, rather than using the streams directly (because that made it impossible to get at the formatting logic).

Issue #562, opened 2017-06-20
resin build / deploy issue with key file


imrehg:
When using TLS CA/cert/key files, get an `Error: error:0906D06C:PEM routines:PEM_read_bio:no start line` error instead of the call properly authenticating with the remote Docker service

Here's the command running
```
DEBUG=1 resin deploy <project> \
	--build \
	--source . \
	--dockerHost <host> \
	--dockerPort <port> \
        --ca ca.pem \
	--cert cert.pem \
        --key key.pem
```

The output:
```
[Debug]   Connecting with the following options:
[Debug]   {
[Debug]     "host": "<host>",
[Debug]     "port": <port>,
[Debug]     "ca": "ca.pem",
[Debug]     "cert": "cert.pem",
[Debug]     "key": "key.pem"
[Debug]   }
Error: error:0906D06C:PEM routines:PEM_read_bio:no start line
    at Object.createSecureContext (_tls_common.js:89:17)
    at Object.exports.connect (_tls_wrap.js:1038:48)
    at Agent.createConnection (https.js:105:22)
    at Agent.createSocket (_http_agent.js:222:26)
    at Agent.addRequest (_http_agent.js:184:10)
    at new ClientRequest (_http_client.js:272:16)
    at Object.request (http.js:39:10)
    at Object.request (https.js:233:15)
    at Object.request (/usr/lib/node_modules/resin-cli/node_modules/agent-base/patch-core.js:52:20)
    at h.request (/usr/lib/node_modules/resin-cli/node_modules/docker-modem/lib/http.js:51:52)
    at Modem.buildRequest (/usr/lib/node_modules/resin-cli/node_modules/docker-modem/lib/modem.js:173:33)
    at Modem.dial (/usr/lib/node_modules/resin-cli/node_modules/docker-modem/lib/modem.js:168:8)
    at Docker.buildImage (/usr/lib/node_modules/resin-cli/node_modules/resin-docker-build/node_modules/dockerode/lib/docker.js:199:14)
    at Docker.tryCatcher (/usr/lib/node_modules/resin-cli/node_modules/bluebird/js/release/util.js:16:23)
    at Docker.ret [as buildImageAsync] (eval at makeNodePromisifiedEval (/usr/lib/node_modules/resin-cli/node_modules/bluebird/js/release/promisify.js:184:12), <anonymous>:13:39)
[...]

PR #563, opened 2017-06-20
Read ca files and convert to string before passing to the docker daemon

CameronDiver:
Before this commit, the docker daemon would recieve the filename of the
.pem files, which would be interpreted as the body and would fail. This
commit ensures that the actual body of the pem files are sent to the
daemon.

cc @imrehg 

Change-type: patch
Connects-to: #562

### DIFF
--- a/lib/utils/docker.coffee
+++ b/lib/utils/docker.coffee
@@ -70,36 +70,47 @@ exports.appendOptions = (opts) ->
 	]
 
 exports.generateConnectOpts = generateConnectOpts = (opts) ->
-	connectOpts = {}
-	# Firsly need to decide between a local docker socket
-	# and a host available over a host:port combo
-	if opts.docker? and not opts.dockerHost?
-		# good, local docker socket
-		connectOpts.socketPath = opts.docker
-	else if opts.dockerHost? and not opts.docker?
-		# Good a host is provided, and local socket isn't
-		connectOpts.host = opts.dockerHost
-		connectOpts.port = opts.dockerPort || 2376
-	else if opts.docker? and opts.dockerHost?
-		# Both provided, no obvious way to continue
-		throw new Error("Both a local docker socket and docker host have been provided. Don't know how to continue.")
-	else
-		# None provided, assume default docker local socket
-		connectOpts.socketPath = '/var/run/docker.sock'
+	Promise = require('bluebird')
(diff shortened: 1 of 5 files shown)


imrehg:
Tried this PR and does connect.

lib/utils/docker.coffee
```diff
@@ -70,36 +70,44 @@ exports.appendOptions = (opts) ->
...
+			# but not all
+			if not (opts.ca? and opts.cert? and opts.key?)
+				throw new Error('You must provide a CA, certificate and key in order to use TLS')
+
+			return Promise.all([fs.readFile(opts.ca), fs.readFile(opts.cert), fs.readFile(opts.key)])
+				.spread (ca, cert, key) ->
+					connectOpts.ca = ca.toString()
+					connectOpts.cert = cert.toString()
+					connectOpts.key = key.toString()
+					return connectOpts
```
pimterry: I think you can slightly simplify this to:

```coffeescript
Promise.props(ca: fs.readFile(opts.ca), cert: fs.readFile(opts.cert), key: fs.readFile(opts.key))
.then (result) ->
  _.mapValues result, value -> value.toString()
```

or even

```coffeescript
Promise.props(_.pick(opts, ['ca', 'cert', 'key']).map(fs.readFile))
.then (result) ->
  _.mapValues(result, value -> value.toString())
```

(although maybe that last step is more terse than it is clear, up to you)

CameronDiver: Huh TIL about `Promise.props`, thanks :)

lib/utils/docker.coffee
```diff
@@ -70,36 +70,44 @@ exports.appendOptions = (opts) ->
...
+		# Now need to check if the user wants to connect over TLS
+		# to the host
+
+		# If any are set...
+		if (opts.ca? or opts.cert? or opts.key?)
+			# but not all
+			if not (opts.ca? and opts.cert? and opts.key?)
+				throw new Error('You must provide a CA, certificate and key in order to use TLS')
+
+			return Promise.props(ca: fs.readFile(opts.ca), cert: fs.readFile(opts.cert), key: fs.readFile(opts.key))
```
emirotin: If you pass the 2nd encoding argument to `readFile` (set it to `utf8`) you shouldn't need `toString`

Issue #567, opened 2017-06-21
Print a warning if using an unsupport node version


pimterry:
The README currently says >= 4, but in Flowdock discussion we want to support only active LTS versions (so, as of April, only Node 6+). For users using older versions, we shouldn't outright block them (since most things are likely to still work with Node 4), but we should print a clear warning so they know to expect issues.

pimterry:
Fixed in 6.0.0, closing

PR #576, opened 2017-06-22
Remove Buffer polyfill, require Node v6+, and print warnings in older versions

pimterry:
Connects to #567

### DIFF
--- a/lib/app.coffee
+++ b/lib/app.coffee
@@ -29,9 +29,11 @@ Raven.setContext
 
 validNodeVersions = require('../package.json').engines.node
 if not require('semver').satisfies(process.version, validNodeVersions)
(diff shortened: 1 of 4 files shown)


lib/app.coffee
```diff
@@ -27,7 +27,12 @@ Raven.setContext
 		args: process.argv
 		node_version: process.version
 
-require('buffer-v6-polyfill')
+validNodeVersions = require('../package.json').engines.node
+if not require('semver').satisfies(process.version, validNodeVersions)
+	console.warn("Warning: this version of Node does not match the requirements of this package.
```
emirotin: NIT: can drop the parens here and use the multi-line string with triple quotes

Issue #575, opened 2017-06-22
Specify docker build additional options in resin build


jacintoArias:
Using the `resin build` command is key for my team in order to integrate our CI platform with resin, however our current building scripts call `docker build` with a set of particular options.

* We have tons of python packages, so we like to store docker-related files in a `/docker` folder, so we  need to trigger the build using the project root as context and specify the `Dockerfile` using `-f`

* We rely on the new experimental `--squash` feature to remove secrets and compress our images.

Right now I can't see a way to pass these parameters to the `resin build` command...

jacintoArias:
In case this is interesting for anyone else, adding `--squash` was easy, if you feel a pm would be a good idea just say the word.

The fork:
jacintoArias/resin-cli@9df617863c30003b8e967dc8630d523106611e14

`docker build -f` for referencing a `Dockerfile` inside a nested folder turns out to be a little more tricky, as I'm getting errors from the `resin-bunle-resolve` package when I pass a source dir without and explicit dockerfile.

For this matter we-ve just refactored our dir structure :p

pimterry:
@jacintoArias This is a great idea.

Yes, we'd happily accept a PR for that `--squash` change, that's looks great. We're also going to have a look into fixing `resin-bundle-resolve` to better support this, so we can add `-f` support too - watch this space.

pimterry:
I don't think we're looking at supporting arbitrary arguments right now, so now that `squash` is included I'm going to close this. There's a separate issue for the specific `-f` argument to specify the dockerfile explicitly over here: https://github.com/resin-io/resin-cli/issues/587

PR #577, opened 2017-06-24
Allow --squash option on resin build command

jacintoArias:
This partially solves #575


pimterry:
Rebased and merged. Thank @jacintoArias!

PR #584, opened 2017-06-27
Implement new device commands

horia-delicoti:
[Closes resin-io/resin-cli#583]


lib/actions/device.coffee
```diff
@@ -180,6 +180,21 @@ exports.remove =
...
+    help: '''
+           Use this command to ping a device.
+
+           Examples:
+
+                   $ resin device ping 23c73a1
+    '''
+    permission: 'user'
+    action: (params, options, done) ->
+        resin = require('resin-sdk-preconfigured')
```
pimterry: These errors are because all of your code is space-indented I think - this should all be tabbed.

lib/actions/device.coffee
```diff
@@ -293,6 +308,71 @@ exports.hasDeviceUrl =
 			console.log(hasDeviceUrl)
 		.nodeify(done)
 
+exports.getApplicationInfo =
+	signature: 'device application info <uuid>'
```
pimterry: Minor point, but I _think_ this should be `app` everywhere, not `application`, to be consistent with existing comments (`resin apps`, `resin app restart`).

lib/actions/device.coffee
```diff
@@ -293,6 +308,71 @@ exports.hasDeviceUrl =
...
+	'''
+	permission: 'user'
+	action: (params, options, done) ->
+		resin = require('resin-sdk-preconfigured')
+		resin.models.device.getApplicationInfo(params.uuid).then (appInfo) ->
+			console.log(appInfo)
+		.nodeify(done)
+
+exports.startApplication =
+    signature: 'device start application <uuid>'
```
pimterry: I'd rather this the sub-command here was a single word, rather than sort-of feeling like a namespace here (what else can I `start`?). `resin device start-app` sounds good to me.

Oooh, in fact, better actually: `resin device app start` (to fit with `resin device app info` above), and we can nest all device-specific application commands there.

lib/actions/device.coffee
```diff
@@ -293,6 +308,71 @@ exports.hasDeviceUrl =
...
+
+        Examples:
+
+            $ resin device start application 23c73a1
+    '''
+    permission: 'user'
+    action: (params, options, done) ->
+        resin = require('resin-sdk-preconfigured')
+        resin.models.device.startApplication(params.uuid).then (containerId) ->
+            console.log(containerId)
```
pimterry: Is there something a user wants the `containerId` for? If not, we should drop this `console.log`. If so, we should mention that this gets returned in the docs above.

lib/actions/device.coffee
```diff
@@ -293,6 +308,71 @@ exports.hasDeviceUrl =
...
+        .nodeify(done)
+
+exports.stopApplication =
+    signature: 'device stop application <uuid>'
```
pimterry: `device app stop` too

lib/actions/device.coffee
```diff
@@ -293,6 +308,71 @@ exports.hasDeviceUrl =
...
+
+        Examples:
+
+            $ resin device stop application 23c73a1
+    '''
+    permission: 'user'
+    action: (params, options, done) ->
+        resin = require('resin-sdk-preconfigured')
+        resin.models.device.stopApplication(params.uuid).then (containerId) ->
+            console.log(containerId)
```
pimterry: Almost certainly want to drop the console.log here too. Any reason you'd want the container id of the app you just stopped?

lib/actions/device.coffee
```diff
@@ -432,3 +512,51 @@ exports.init =
...
+
+		Examples:
+
+			$ resin device status 601317
+	'''
+	permission: 'user'
+	action: (params, options, done) ->
+		resin = require('resin-sdk-preconfigured')
+		resin.models.device.getStatus(params.device).then (status) ->
+			console.log(status)
```
pimterry: This returns a plain JSON object, right? That probably needs some formatting nicely - take a look at `resin device <uuid>` for a nice example.

lib/actions/device.coffee
```diff
@@ -432,3 +512,51 @@ exports.init =
...
+	permission: 'user'
+	action: (params, options, done) ->
+		resin = require('resin-sdk-preconfigured')
+		resin.models.device.purge(params.uuid).nodeify(done)
+
+exports.update =
+	signature: 'device update <uuid>'
+	description: 'trigger an update check on supervisor'
+	help: '''
+		Use this command to triggere an update check on supervisor
```
pimterry: Typo: trigger**e**
Typo: you need a 'the' before 'supervisor', or similar

lib/actions/device.coffee
```diff
@@ -293,6 +308,71 @@ exports.hasDeviceUrl =
...
+    '''
+    permission: 'user'
+    action: (params, options, done) ->
+        resin = require('resin-sdk-preconfigured')
+        resin.models.device.stopApplication(params.uuid).then (containerId) ->
+            console.log(containerId)
+        .nodeify(done)
+
+exports.restartApplication =
+    signature: 'device restart application <uuid>'
```
pimterry: `device app restart`

lib/actions/device.coffee
pimterry: This is wrong - this was me looking at the parameter (which _should_ be an object), not the response - that is indeed fine to log to the console if that's what you're looking for.

shaunmulligan:
@horia-delicoti can we close this PR?

horia-delicoti:
@shaunmulligan yes.

PR #601, opened 2017-07-26
Support the new OS with resin-sample.ignore connection file

### LINKED ISSUE #597: resin local configure should create a new connection file rather than editing `resin-sample`
In newer versions of resinOS 2.0, we will be removing or renaming the `resin-sample` file in `/boot/system-connections`. So before we do that, we need to make sure that the CLI can handle this, my suggestion is that the CLI should create a new system connection file in `/boot/system-connections` and not rely on `resin-sample` being there at all.

emirotin:
Shaun, I've decided to implement all the possible cases in this PR in case some useful NM options are later added by the OS:

- if the `resin-wifi` file exists (previously configured image or downloaded from the UI) it's used and reconfigured
- if the `resin-sample.ignore` exists it's copied to `resin-wifi`
- if the `resin-sample` exists it's reconfigured (legacy mode, will be removed later)
- otherwise, the new file is created

Tim, FYI I've decided to delete the npm5 lockfile for now but can regenerate it if you insist (I'm still using npm4 myself).

Fixes #597


lib/actions/local/configure.coffee
```diff
@@ -42,15 +45,133 @@ CONFIGURATION_SCHEMA =
...
+				image: target
+				partition: BOOT_PARTITION
+				path: "#{CONNECTIONS_FOLDER}/resin-sample.ignore"
+			,
+				image: target
+				partition: BOOT_PARTITION
+				path: "#{CONNECTIONS_FOLDER}/resin-wifi"
+			.thenReturn(null)
+
+		# Legacy mode, to be removed later
```
pimterry: Can you add an issue too, so we can track this?

pimterry: In fact, is this even true? Are we definitely going to be able to remove this any time soon? We can't do so until nobody is using any of the current ResinOS 2.X images, which sounds like it might be a long way away, so we'll need to keep supporting this `resin-sample` file indefinitely, no? I'm not sure what our OS support policy is.

lib/actions/local/configure.coffee
```diff
@@ -42,15 +45,133 @@ CONFIGURATION_SCHEMA =
...
+
+	[ipv4]
+	method=auto
+
+	[ipv6]
+	addr-gen-mode=stable-privacy
+	method=auto
+'''
+
+prepareConnectionFile = (target) ->
```
pimterry: This is a little complicated, and a quick summary comment above it (basically the bullet points from this PR's description, which are great) would be useful :+1:

lib/actions/local/configure.coffee
emirotin: Well, I think it's OK for CLI v4 to stop support resin OS < 2.1 or something like that. But indeed it won't happen too soon.

emirotin: Actually, it's fine to drop because the final fallback will create the identical fresh file

PR #610, opened 2017-08-04
Integrate resin-preload

zvin:
See it in action: https://asciinema.org/a/erp7Mq9uTSxYEeiABqhUklcEd

https://github.com/resin-io/resin-preload/pull/81 needs to be merged and released first, then I'll add `resin-preload` to `resin-cli`'s `package.json`.

Connects to #609 
Connects to https://github.com/resin-io/resin-preload/pull/81

Change-Type: minor


emirotin:
> then I'll add resin-preload to resin-cli's package.json.

you can add it now as `resin-io/resin-preload#71-integrate-with-resin-cli` so that this branch is fully functional

lib/actions/index.coffee
```diff
@@ -35,3 +35,4 @@ module.exports =
 	build: require('./build')
 	deploy: require('./deploy')
 	util: require('./util')
+	preload: require('./preload')
```
emirotin: I don't see `./preload` in this PR

zvin: I forgot to add it.
It is here now.
I also added `resin-io/resin-preload#71-integrate-with-resin-cli` to `package.json`.

zvin:
@CameronDiver I could't reuse docker options from https://github.com/resin-io/resin-cli/blob/master/lib/utils/docker.coffee#L12 as they include options like `tag`, `squash` or `nocache` that are related to build and not to a Docker connection.

emirotin:
@zvin you can of course extract the `appendConnectionOptions` part and reuse it where you need and in the original method. The order doesn't matter for capitano

lib/actions/preload.coffee
```diff
@@ -0,0 +1,208 @@
...
+preload = require('resin-preload')
+Promise = require('bluebird')
+resin = require('resin-sdk-preconfigured')
+
+dockerUtils = require('../utils/docker')
+
+LATEST = 'latest'
+
+selectApplication = (resin, form, deviceType) ->
+	resin.models.application.getAll({filter: {device_type: deviceType}})
```
emirotin: NIT: we use relaxed curly brackets (with spaces inside)

lib/actions/preload.coffee
```diff
@@ -0,0 +1,208 @@
...
+				value: app
+
+selectApplicationCommit = (resin, form, appId) ->
+	resin.models.build.getAllByApplication(appId)
+	.then (builds) ->
+		if builds.length == 0
+			expectedError('This application has no builds')
+		builds = builds.filter (build) ->
+			build.status == 'success'
+		builds = _.sortBy(builds, '-push_timestamp')
```
emirotin: Is it OK if you find no successful builds? Probably not too

Also why not filtering and ordering on the server side (using pine options)?

lib/actions/preload.coffee
```diff
@@ -0,0 +1,208 @@
...
+		choices = builds.map (build) ->
+			name: build.push_timestamp + ' - ' + build.commit_hash
+			value: build.commit_hash
+		choices.splice(0, 0, {'name': LATEST, 'value': LATEST})
```
emirotin: Or can be

```
DEFAULT_CHOICE = {'name': LATEST, 'value': LATEST}
choices = [ DEFAULT_CHOICE ].concat builds.map ...
```

lib/actions/preload.coffee
```diff
@@ -0,0 +1,208 @@
...
+			message: 'Select a build'
+			type: 'list'
+			default: LATEST
+			choices: choices
+
+disableAutomaticUpdates = (form, resin, application, commit) ->
+	if commit == LATEST or not application.should_track_latest_release
+		return Promise.resolve()
+	message = (
+		'\n' + # Multi-line messages look beter if they start after the "?"
```
emirotin: You should be able to use triple quotes instead of concatenation

lib/actions/preload.coffee
```diff
@@ -0,0 +1,208 @@
...
+						return selectApplicationCommit(resin, form, application.id)
+					.then (commit) ->
+						if commit != LATEST
+							options.commit = commit
+						return disableAutomaticUpdates(form, resin, application, commit)
+			.then(resin.auth.getToken)
+			.then (token) ->
+				options.apiToken = token
+				preload.run(docker, options)
+		.then (streams) ->
```
emirotin: `.spread (outStream, errStream) ->` (or whatever these two stream are)

zvin:
> @zvin you can of course extract the appendConnectionOptions part and reuse it where you need and in the original method. The order doesn't matter for capitano

@emirotin good idea, done

lib/actions/preload.coffee
zvin: I removed these completely.

lib/actions/preload.coffee
zvin: @emirotin good catch, done

lib/actions/preload.coffee
zvin: changed

lib/actions/preload.coffee
zvin: I didn't use because I use `streams.map` below, that would require to recreate an array like `[ stdout, stderr ]`

lib/actions/preload.coffee
zvin: done

lib/actions/preload.coffee
```diff
@@ -0,0 +1,175 @@
...
+
+selectApplicationCommit = (resin, form, appId) ->
+	options =
+		filter:
+			status: 'success'
+		orderby: 'push_timestamp desc'
+	resin.models.build.getAllByApplication(appId, options)
+	.then (builds) ->
+		if builds.length == 0
+			expectedError('This application has no builds')
```
emirotin: Now the message should be "no successful builds"

lib/actions/preload.coffee
emirotin: Can it then be `streams.forEach(stream -> stream.pipe(process.stdout))`?

zvin:
> Can it then be streams.forEach(stream -> stream.pipe(process.stdout))?

Sorry there was a mistake here, they should go to different places: `process.stdout` and `process.stderr`.

zvin:
@pimterry @CameronDiver @emirotin
Does anyone know how I could get only apps that have successful builds with pine ?

This appears to work:

```coffeescript
	resin.models.application.getAll
		filter:
			device_type: deviceType
			build:
				$any:
					$expr:
						build:
							status: 'success'
```

emirotin:
Yes, this looks valid

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+						$expr:
+							b:
+								status: 'success'
+			expand:
+				build:
+					$select: [ 'id', 'commit_hash', 'push_timestamp', 'status' ]
+					$orderby: 'push_timestamp desc'
+					# the filter is commented because it causes an api error
+					#$filter:
+					#	status: 'success'
```
pimterry: Does this comment still need to be here?

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+			return
+		resin.pine.patch
+			resource: 'application'
+			id: application.id
+			body:
+				should_track_latest_release: false
+
+expectedError = (message) ->
+	console.error(message)
+	process.exit(1)
```
pimterry: Nice! This is a good solution to some of the annoying noise in sentry (as well as avoiding extra console noise, which I assume is what you're going for).

It'd be good if it didn't live here though - this should probably be in utils somewhere really.

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+		applications.forEach (application) ->
+			application.build = application.build.filter (build) ->
+				build.status == 'success'
+		applications
+
+selectApplication = (resin, form, deviceType) ->
+	getApplicationsWithSuccessfulBuilds(resin, deviceType)
+	.then (applications) ->
+		if applications.length == 0
+			expectedError("You have no apps for a '#{deviceType}' device type.")
```
pimterry: This error isn't necessarily true: you'll get no applications returned if you do have apps for this device type, but they don't have any successful builds yet.

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+			alias: 'c'
+		}
+		{
+			signature: 'splash-image'
+			parameter: 'splashImage.png'
+			description: 'path to a png image to replace the splash screen'
+			alias: 's'
+		}
+		{
+			signature: 'dont-detect-flasher-type-images'
```
pimterry: This is very verbose. `--no-flasher-detection`?

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+		available version.
+		You need to disable this if you want to preload a specific build.
+		Do you want to disable automatic updates for this application?
+	'''
+	form.ask
+		message: message,
+		type: 'confirm'
+	.then (update) ->
+		if not update
+			return
```
pimterry: With this, if you reject the confirmation, we'll go on and give you a preloaded image that's not going to work. Is that definitely right? If we're sure that preloading is not going to work without this, we could just error out here, and avoid a little potential confusion from users.

If there is a plausible case for when you might want to do this, this is fine, but the message above makes it sound like this is something that's _always_ going to fail, and there's no point letting users shoot themselves in the foot like that.

pimterry: Even if we don't block this, it'd be nice to print something here to confirm they've agree they're going to do something potentially foolish:

> Warning: Leaving automatic updates enabled! Preloading may not work correctly.

(or something else more descriptive, I'm not sure exactly what the issue is here)

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+
+module.exports =
+	signature: 'preload <image>'
+	description: '(beta) preload an app on a disk image'
+	help: '''
+		Warning: "resin preload" requires Docker to be correctly installed in
+		your shell environment. For more information (including Windows support)
+		please check the README here: https://github.com/resin-io/resin-cli .
+
+		Use this command to preload an application to a local disk image.
```
pimterry: This should explain where the preloaded application comes from (i.e. Resin.io - they can't use local Docker images, until they've been deployed). Something like:

> Use this command to preload an application to a local disk image with a built commit from Resin.io. This can be used with cloud builds, or images deployed with `resin deploy`.

(Assuming that that is indeed true and tested).

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+limitations under the License.
+###
+
+_ = require('lodash')
+form = require('resin-cli-form')
+preload = require('resin-preload')
+Promise = require('bluebird')
+resin = require('resin-sdk-preconfigured')
+
+dockerUtils = require('../utils/docker')
```
pimterry: In the CLI we don't do top-level requires in actions, as a rule. To keep startup time down, wherever possible each action requires its dependencies inside the action itself, so we only end up loading the dependencies we definitely need. Otherwise every command loads every module that every other command might want to use, which really hurts UX.

`resin deploy` is a good example: https://github.com/resin-io/resin-cli/blob/master/lib/actions/deploy.coffee#L146-L152

zvin: Updated, except for `dockerUtils = require('../utils/docker')` which is used for `exports.options`.

lib/actions/preload.coffee
zvin: Yes until we find a fix.

lib/actions/preload.coffee
zvin: Applications with no successful builds are filtered out in getApplicationsWithSuccessfulBuilds: https://github.com/resin-io/resin-cli/blob/2aa72e8b16439f5ac378bb7ce5122d00cfd73b33/lib/actions/preload.coffee#L40-L45

lib/actions/preload.coffee
zvin: Preloading will work correctly.
The problem is that the device will start to download the latest build once it is online making the preloading useless.

I added this to the message.

lib/actions/preload.coffee
pimterry: Ah, yep, sorry. This is different behaviour to the uncommented `filter: build: ...` section above, got it.

Might be useful to explain this: "we manually filter out successful builds below - we should move that here once this API error is resolved"

lib/actions/preload.coffee
pimterry: Yep, which is why this error can be wrong.

If I have a raspberry-pi app with no successful builds, and I try to preload a raspberry-pi image, I'll get an error that says "You have no apps for a raspberry-pi device type". I do have a raspberry-pi app though, it's just it doesn't have any successful builds yet. That will be confusing. Am I missing something?

lib/actions/preload.coffee
pimterry: Based on this, it sounds like preloading + automatic updates isn't useless, and makes sense in lots of environments where you might use a device first time without reliable internet, or where you want to start using the device immediately while letting the image download slowly start up in the background (and later upgrade with a handover or something). Is that right?

If so, sounds to me like we should tone this message down further, since that seems like a pretty reasonable use case. I would have thought lots (most?) preloaded images really are being used to stop devices _requiring_ internet on first boot, not to avoid automatic deploys entirely. Could we tweak the message a bit further to something like:

```
This application is set to automatically update all devices to the latest available version.
This might be unexpected behaviour: with this enabled, the preloaded device will still
download and install the latest build once it is online.

Do you want to disable automatic updates for this application?
```

And then drop the warning, since it's not the scary result I thought it was.

Are my assumptions right, and does that make sense do you think?

pimterry: If that is right, we should also rename this method - e.g. `offerToDisableAutomaticUpdates` - since we're expecting it might not be disabled for lots of people.

lib/actions/preload.coffee
zvin: Good idea, moved to utils.patterns.

lib/actions/preload.coffee
zvin: This is something that shouldn't be used unless you know what you're doing.
I'd like it to keep a long obscure name.
Discussed here : https://github.com/resin-io/resin-preload/pull/81#discussion_r131387302

lib/actions/preload.coffee
zvin: You're right, I updated the error message to "You have no apps with successful builds for a '#{deviceType}' device type."

lib/actions/preload.coffee
zvin: That makes sense, I updated the message and removed the warning.

zvin: And renamed the function.

lib/actions/preload.coffee
zvin: help updated

lib/actions/preload.coffee
zvin: FIXME comment added

lib/actions/preload.coffee
pimterry: Fair enough, in that case I'm not too worried about it.

lib/actions/preload.coffee
```diff
@@ -0,0 +1,206 @@
...
+
+		options.image = params.image
+		options.appId = options.app
+		delete options.app
+
+		options.dontDetectFlasherTypeImages = options['dont-detect-flasher-type-images']
+		delete options['dont-detect-flasher-type-images']
+
+		dockerUtils.getDocker(options)
+		.then (docker) ->
```
CameronDiver: Could you add maybe a little vertical spacing to break up the logic a bit? I imagine I'm not the only one that struggles to parse walls of code.

lib/utils/docker.coffee
```diff
@@ -9,7 +9,7 @@ QEMU_BIN_NAME = 'qemu-execve'
 #
 # NOTE: Care MUST be taken when using the function, so as to
 # not redefine/override options already provided.
-exports.appendOptions = (opts) ->
+exports.appendConnectionOptions = appendConnectionOptions = (opts) ->
```
CameronDiver: Nice :+1:

Issue #594, opened 2017-07-17
"Unauthorized" uploading image to shared project


memolipd:
It seems to work on Resin Applications I have made myself but throws the error when pushing to any applications a different developer has made and shared with me.

This seems like a bug, since using other means I can push images to applications created by other developers and shared with me.

pimterry:
Hi @memolipd, this is due to a bug in our builder, which checked only if you were the owner of an application, for local pushes, not anybody with access to it. We've got the basics of a fix working now, just finishing it up at the moment - I'll ping back here once it's resolved and deployed. Thanks for the report!

pimterry:
Hi @memolipd, I'm afraid we forgot to update you here - we added support for this in August. If you want to deploy to a collaborators app, you can now use `resin deploy $APP_OWNER/$APP_NAME` and that should work as if it were your own. I've just opened a PR to also mention this in the documentation for the command, so it hopefully shouldn't catch people out in the future!

Thanks again for the report, let us know if you have any more trouble.

PR #617, opened 2017-08-11
Allow the looking up of applications with the owner

CameronDiver:
change-type: minor
Connects-to: #594

### DIFF
--- a/lib/utils/helpers.coffee
+++ b/lib/utils/helpers.coffee
@@ -96,7 +96,7 @@ exports.getAppInfo = (application) ->
 	resin = require('resin-sdk-preconfigured')
 	_ = require('lodash')
 	Promise.join(
-		resin.models.application.get(application),
+		getApplication(application),
 		resin.models.config.getDeviceTypes(),
 		(app, config) ->
 			config = _.find(config, 'slug': app.device_type)
@@ -105,6 +105,16 @@ exports.getAppInfo = (application) ->
 			app.arch = config.arch
 			return app
 	)
+
+getApplication = (application) ->
+	resin = require('resin-sdk-preconfigured')
+
+	# Check for an app of the form `user/application`, and send
+	# this off to a special handler (before importing any modules)
+	if (match = /(\w+)\/(\w+)/.exec(application))
+		return resin.models.application.getAppWithOwner(match[2], match[1])
+
+	return resin.models.application.get(application)
 
 # A function to reliably execute a command
 # in all supported operating systems, including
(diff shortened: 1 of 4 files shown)


lib/utils/helpers.coffee
```diff
@@ -93,10 +93,25 @@ exports.osProgressHandler = (step) ->
...
+	return getAppInfoFromPromise(resin.models.application.getAppWithOwner(appName, username))
+
+# Helper function to setup application info from a promise that resolves
+# to an application
+getAppInfoFromPromise = (appPromise) ->
 	resin = require('resin-sdk-preconfigured')
 	_ = require('lodash')
 	Promise.join(
-		resin.models.application.get(application),
+		appPromise,
```
pimterry: If you flip this refactoring on its head, it'll be much nicer.

Here you've pulled out the bit that varies as a wrapper, around the function's main logic. Instead, you can push down the bit that varies, and keep the core important logic as the top level. That looks something like:

```coffeescript
exports.getAppInfo = (application) ->
	_ = require('lodash')
	Promise.join(
		getApplication(application),
		resin.models.config.getDeviceTypes(),
		(app, config) ->
			config = _.find(config, 'slug': app.device_type)
			if !config?
				throw new Error('Could not read application information!')
			app.arch = config.arch
			return app
	)

getApplication = (application) ->

	# Check for an app of the form `user/application`, and send
	# this off to a special handler
	if (match = /(\w+)\/(\w+)/.exec(application))
		return resin.models.application.getAppWithOwner(match[1], match[2])

	return resin.models.application.get(application)
```

Easier to follow the structure that way, and you end up with more composable standalone parts, rather than tying everything together. Also strictly necessary if you have more than one refactoring like this, or you get an explosion of interlinked entry points. Eventually, `getApplication` might be something we even move into the SDK, which is super easy this way around.

It looks like you're also exporting `getAppInfoWithUser`, because you want to explicitly use that elsewhere? If so, you could extend the above to include an optional user argument, pass it to `getApplication`, and handle that there to get the same result. If not, that's fine too :smiley:

CameronDiver: Much, much better :ok_hand: definitely refactoring this.

Re: the exporting of `getAppInfoWithUser`, I was exporting it simply because it may be useful. I think if we do want it to be exported, your method is also nicer, but if we don't want it to be exported because it won't be used anywhere at first, that's also fine too. I think given the subtlety of your method, we might as well export it like that.

pimterry:
The SDK change has now been released.

Issue #618, opened 2017-08-11
Any command using dockerode with the default file socket will fail if RESINRC_PROXY is set.


zvin:
Try it:
```sh
RESINRC_PROXY="http://0.0.0.0:8118" ./bin/resin build /some/project/folder
```

results in 

```
EPIPE: write EPIPE

If you need help, don't hesitate in contacting us at:

  GitHub: https://github.com/resin-io/resin-cli/issues/new
  Forums: https://forums.resin.io
```

This happens because `global-tunnel-ng` proxies __all__ connections including those to local socket file like `/var/run/docker.sock`.

PR #619, opened 2017-08-11
Use forked global-scoket-ng that doesn't proxy connection to socket files.

zvin:
This is temporary, until https://github.com/np-maintain/global-tunnel/pull/9 is merged.

Connects to #618

Change-Type: patch


zvin:
@pimterry changelog entry added

Issue #620, opened 2017-08-15
description on package.json could be updated


flesler:
```json
"description": "Git Push to your devices",
```

Maybe something like `The official Resin CLI tool` like we have on the project homepage would be a better option.

pimterry:
Agreed! Good point, done: #621

PR #621, opened 2017-08-16
Update npm package description

pimterry:
Connects to #620

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "resin-cli",
   "version": "6.4.0",
-  "description": "Git Push to your devices",
+  "description": "The official resin.io CLI tool",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "resin-cli",
   "version": "6.4.0",
-  "description": "Git Push to your devices",
+  "description": "The official Resin CLI tool",
```
emirotin: Should it be `resin.io`?

pimterry: Indeed it should - updated.

We should also update the repo description to match. I've asked Apo to add me as an admin so I can do so, but feel free to do it yourself in the meantime, if you've got admin access.

emirotin: No I haven't

Issue #622, opened 2017-08-16
Shrink install size to improve install speed


pimterry:
Right now, a fresh install of resin-cli is noticeably ridiculously slow on my machine, because it installs 680MB of dependencies (!). Doesn't look like this is caused by any one single enormous dependency - there's 36 deps that more than 10MB each, which covers at least half of this straight away.

It looks like a major offender is lodash: a lot of packages use it, split between v3 and v4, and only v3 is deduped on my system, so v4 is duplicated in every dependency (at 14MB a pop).

This relates to #607 (though we'll need to fix it independently), and might be the cause of #561.

pimterry:
Looks like a big part of this is minimum file sizes actually: I was installing in my home dir, which is ecryptfs and seems to end up with a minimum file size of 12kb - installing on ext4 shrinks this down to 310MB (4kb minimum file size).

pimterry:
We've drastically shrunk this with the linked fixes, especially lodash. It'd be nice to do more of this in future, but for now I think it'll do.

PR #623, opened 2017-08-16
Upgrade to lodash v4

pimterry:
This is the main big step for #622 - with this done, we dedupe all v4 lodash installs, and only duplicate v3 installs which are a) far fewer b) far smaller. Once this is in, we should start going through our submodules and upgrading them to lodash v4 where possible, which saves about 5MB in every install.

In a quick test with `npm install --production` in my home directory (ecryptfs) this reduces the install size from 680MB to 280MB, or 310MB to 163MB (ext4).

I went through all the breaking changes in v4 individually to check our usage, and only found `findWhere` as a case that officially _should_ break. I've also done some quick basic testing of:
* Login
* resin apps/devices
* local scan/ssh/push
* build & deploy

Suggestions for anything else that definitely need including welcome, but I'm reasonably confident this is pretty low risk given all that.


pimterry:
Installation speed seems to be fairly proportional to package size - this more or less halves the CLI install time for me.

PR #624, opened 2017-08-16
Remove unused babel dependency

pimterry:
Another part of #622, just because it's an easy way to shave 10MB or so off the CLI install.

Seems that @CameronDiver put these in as part of the local build/deploy work (#466), but they don't look used and neither he nor I know any good reason why they should be here. Some very brief testing suggests everything works without them very happily - unless you know something we don't @emirotin we should drop these.


emirotin:
These deps used to be used to `require` some module from juanchi that used to use ES6 syntax (specifically default function arguments). The dependency was later republished in ES5-compatible form.

Issue #626, opened 2017-08-17
Don't preload a build twice.


zvin:
`resin preload` allows to preload the same build twice: it extends the partition for nothing.

PR #627, opened 2017-08-17
Don't try preloading the same build twice in an image.

zvin:
Requires https://github.com/resin-io/resin-preload/pull/87 to be merged and released first.

Connects to #626 

Change-Type: patch

### DIFF
--- a/lib/actions/preload.coffee
+++ b/lib/actions/preload.coffee
@@ -174,23 +174,22 @@ module.exports =
 				options.apiHost = settings.apiUrl
 
 				# Use the preloader docker image to extract the deviceType of the image
-				preload.getDeviceTypeSlug(docker, options)
+				preload.getDeviceTypeSlugAndPreloadedBuilds(docker, options)
 				.catch(preload.errors.ResinError, expectedError)
-			.then (deviceType) ->
-
+			.then ({ slug, builds }) ->
 				# Use the appId given as --app or show an interactive app selection menu
 				Promise.try ->
 					if options.appId
 						return preload.getApplication(resin, options.appId)
 						.catch(errors.ResinApplicationNotFound, expectedError)
-					selectApplication(expectedError, resin, form, deviceType)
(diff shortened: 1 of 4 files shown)


lib/actions/preload.coffee
```diff
@@ -203,16 +202,23 @@ module.exports =
...
+				.then ->
+
+					builds = builds.map (build) ->
+						build.slice(-preload.BUILD_HASH_LENGTH)
+					if options.commit in builds
+						console.log('This build is already preloaded in this image.')
+						process.exit(0)
+					# All options are ready: preload the image.
+					preload.run(resin, docker, options)
+					.catch(preload.errors.ResinError, expectedError)
```
pimterry: Doesn't matter immediately for this PR, but I'd really like a `ResinError` subclass/flag for errors that are issues in usage or environment, not functional unknown errors, so we can filter for those and behave differently (this would basically replace `expectedError`, and instead just filter errors in our standard error handling to handle that case differently).

We can ignore this for now - just a note for the future.

pimterry:
Sorry, bonus point I forgot: you include this in the changelog

zvin:
@pimterry the CHANGELOG already has " * Preload support".
As there was no release yet, I don't think we need to add anything.

lib/actions/preload.coffee
zvin: Also it would be nice if the sdk exported resin-errors.

pimterry:
Ah ok, very good point, this is fine then.

There is going to be a release _any second now_, but not until #631 gets merged, so you've just squeezed this in :smile:

zvin:
@pimterry if you can approve https://github.com/resin-io/resin-preload/pull/87 yes , otherwise no as we need resin-preload 3.0.0 to be released.

lib/actions/preload.coffee
pimterry: Excellent idea - filed: https://github.com/resin-io/resin-sdk/issues/365

Issue #629, opened 2017-08-17
preload fails with "builds is not defined"


BrianAdams:
The script crashes right after successfully building the resin-preload docker image.  Problem resolving the builds? 

>resin preload $(pwd)/resin.img --dockerHost docker --dockerPort 2375 --api-key $RESIN_PROD_APIKEY --app $appid --commit ${lastcommit}
```
Step 1/7 : FROM docker:17.06.1-ce-dind
 ---> 9e3d126eb84f
Step 2/7 : RUN apk update && apk add --no-cache python3 parted btrfs-progs docker util-linux inotify-tools sfdisk
 ---> Running in 6e3f11a93e47
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/x86_64/APKINDEX.tar.gz
v3.6.2-80-gd6beefe139 [http://dl-cdn.alpinelinux.org/alpine/v3.6/main]
v3.6.2-78-gc30bfa742f [http://dl-cdn.alpinelinux.org/alpine/v3.6/community]
OK: 8439 distinct packages available
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/x86_64/APKINDEX.tar.gz
(1/21) Installing device-mapper-libs (2.02.168-r3)
(2/21) Installing libseccomp (2.3.2-r0)
(3/21) Installing docker (17.05.0-r0)
Executing docker-17.05.0-r0.pre-install
(4/21) Installing inotify-tools (3.14-r2)
(5/21) Installing ncurses-terminfo-base (6.0-r8)
(6/21) Installing ncurses-terminfo (6.0-r8)
(7/21) Installing ncurses-libs (6.0-r8)
(8/21) Installing readline (6.3.008-r5)
(9/21) Installing parted (3.2-r5)
(10/21) Installing libbz2 (1.0.6-r5)
(11/21) Installing expat (2.2.0-r1)
(12/21) Installing libffi (3.2.1-r3)
(13/21) Installing gdbm (1.12-r0)
(14/21) Installing sqlite-libs (3.18.0-r0)
(15/21) Installing python3 (3.6.1-r2)
(16/21) Installing libfdisk (2.28.2-r2)
[...]

PR #630, opened 2017-08-18
Fix resin preload when a commit is provided.

zvin:
Connects-To #629

Change-Type: patch

### DIFF
--- a/lib/actions/preload.coffee
+++ b/lib/actions/preload.coffee
@@ -196,7 +196,7 @@ module.exports =
 					# Use the commit given as --commit or show an interactive commit selection menu
 					Promise.try ->
 						if options.commit
-							if _.map(builds, 'commit_hash').indexOf(options.commit) == -1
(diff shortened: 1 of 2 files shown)


lib/actions/preload.coffee
```diff
@@ -196,7 +196,7 @@ module.exports =
 					# Use the commit given as --commit or show an interactive commit selection menu
 					Promise.try ->
 						if options.commit
-							if _.map(builds, 'commit_hash').indexOf(options.commit) == -1
+							if _.map(application.build, 'commit_hash').indexOf(options.commit) == -1
```
dfunckt: I think this can be

```
if _.find(application.build, commit_hash: options.commit)
  ...
```

emirotin: Right, but with negation: `if !`

dfunckt: Good catch, yes.

zvin: :+1:

Issue #636, opened 2017-08-18
Help command: TypeError: _.object is not a function


BrianAdams:
When bringing up the command line help using today's master repo version (2ad789457f0a4c187912147519f0c9f04fa73522) of resin-cli, there is an error when it attempts to enumerate the possible options:

> resin help preload
```
Usage: preload <image>

Warning: "resin preload" requires Docker to be correctly installed in
your shell environment. For more information (including Windows support)
please check the README here: https://github.com/resin-io/resin-cli .

Use this command to preload an application to a local disk image with a
built commit from Resin.io.
This can be used with cloud builds, or images deployed with resin deploy.

Examples:
  $ resin preload resin.img --app 1234 --commit e1f2592fc6ee949e68756d4f4a48e49bff8d72a0 --splash-image some-image.png
  $ resin preload resin.img

Options:

TypeError: _.object is not a function
    at parse (/usr/local/lib/node_modules/resin-cli/build/actions/help.js:31:12)
    at /usr/local/lib/node_modules/resin-cli/build/actions/help.js:107:13
    at /usr/local/lib/node_modules/resin-cli/node_modules/capitano/build/state.js:31:14
    at /usr/local/lib/node_modules/resin-cli/node_modules/capitano/build/signature.js:106:16
    at /usr/local/lib/node_modules/resin-cli/node_modules/capitano/build/signature.js:148:20
    at /usr/local/lib/node_modules/resin-cli/node_modules/async/lib/async.js:181:20
    at Immediate.iterate [as _onImmediate] (/usr/local/lib/node_modules/resin-cli/node_modules/async/lib/async.js:262:13)
    at tryOnImmediate (timers.js:543:15)
    at processImmediate [as _immediateCallback] (timers.js:523:5)
```

zvin:
lodash v4.0.0 removed `_.object`

see  https://github.com/lodash/lodash/wiki/Changelog
> Removed _.object in favor of _.fromPairs or _.zipObject

PR #638, opened 2017-08-18
Use _.fromPairs instead of _.object which was removed in lodash 4.0.0

zvin:
Connects-To: #636

Change-Type: patch

### DIFF
--- a/build/actions/help.js
+++ b/build/actions/help.js
@@ -28,7 +28,7 @@ columnify = require('columnify');
 messages = require('../utils/messages');
 
 parse = function(object) {
-  return _.object(_.map(object, function(item) {
+  return _.fromPairs(_.map(object, function(item) {
     var signature;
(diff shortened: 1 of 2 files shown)


zvin:
@emirotin @pimterry any idea what might be wrong with the travis build ?

emirotin:
Super-strange and reproducible. No idea wtf

pimterry:
@zvin the travis problem is that it can't find `.travis.yml`, so it's trying to build it as a default ruby project, which obviously isn't going to work. If you rebase on top of master you'll pick up the `.travis.yml`, and that'll fix it.

pimterry:
It'd be good to merge this and get a 6.5.1 fixed release out as soon as we can - in the mean time I've reset `latest` on npm to 6.4.0, so new installs won't pick up the release with this bug in it by default.

Issue #628, opened 2017-08-17
Add some progress bars to `resin preload`


zvin:
progress bars:
 * docker build
 * images size fetching
spinners:
 * application download
 * image device type and builds extraction

PR #639, opened 2017-08-22
Add progress bars and spinners for resin preload.

zvin:
Depends-On: https://github.com/resin-io/resin-preload/pull/95

Connects-To: #628 

Change-Type: minor

### DIFF
--- a/lib/actions/preload.coffee
+++ b/lib/actions/preload.coffee
@@ -18,7 +18,9 @@ dockerUtils = require('../utils/docker')
 
 LATEST = 'latest'
 
-getApplicationsWithSuccessfulBuilds = (resin, deviceType) ->
+getApplicationsWithSuccessfulBuilds = (deviceType) ->
+	resin = require('resin-sdk-preconfigured')
+
 	resin.pine.get
 		resource: 'my_application'
 		options:
@@ -50,9 +52,17 @@ getApplicationsWithSuccessfulBuilds = (resin, deviceType) ->
 				build.status == 'success'
 		applications
 
-selectApplication = (expectedError, resin, form, deviceType) ->
-	getApplicationsWithSuccessfulBuilds(resin, deviceType)
+selectApplication = (deviceType) ->
+	visuals = require('resin-cli-visuals')
+	form = require('resin-cli-form')
+	{ expectedError } = require('../utils/patterns')
+
+	applicationInfoSpinner = new visuals.Spinner('Downloading list of applications and builds.')
+	applicationInfoSpinner.start()
+
+	getApplicationsWithSuccessfulBuilds(deviceType)
 	.then (applications) ->
+		applicationInfoSpinner.stop()
 		if applications.length == 0
 			expectedError("You have no apps with successful builds for a '#{deviceType}' device type.")
 		form.ask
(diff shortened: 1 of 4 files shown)


lib/actions/preload.coffee
```diff
@@ -50,9 +50,11 @@ getApplicationsWithSuccessfulBuilds = (resin, deviceType) ->
 				build.status == 'success'
 		applications
 
-selectApplication = (expectedError, resin, form, deviceType) ->
+selectApplication = (expectedError, resin, form, deviceType, applicationInfoSpinner) ->
```
pimterry: Why pass this in, rather than creating it inside this function?

lib/actions/preload.coffee
```diff
@@ -163,8 +169,7 @@ module.exports =
 		.then (docker) ->
 
 			# Build the preloader image
-			buildOutputStream = preload.build(docker)
-			buildOutputStream.pipe(process.stdout)
```
pimterry: Does this output ever contain anything useful? Might be worth keeping this pipe, but only in debug mode (i.e. `if process.env.DEBUG`)

zvin: That's usual docker build stuff.
Like `Step 1/7 .....`

I'd remove it completely, @jhermsmeier mentioned in an earlier PR that it might be useful for the caller.
OK for showing it only if DEBUG is enabled.
@pimterry what values are true for `process.env.DEBUG` ? 1, true, True and TRUE or anything as long as it is defined ?

lib/actions/preload.coffee
zvin: @pimterry I'd need to `require('resin-cli-visuals')` in that function if I didn't pass the spinner.
Not sure what's the best option.

zvin: Or pass resin-cli-visuals as a parameter.

lib/actions/preload.coffee
pimterry: >  @pimterry what values are true for process.env.DEBUG ? 1, true, True and TRUE or anything as long as it is defined ?

Yep, it's very simple & dumb: debug output should show if it's set to anything at all.

zvin: added

lib/actions/preload.coffee
pimterry: I'd either:
* Just require resin-cli-visuals here - it's cheap, clear, doesn't cause any problems
* Pull these functions out of this file into `utils`, in a separate module where you can do requires up front like a normal node module (which you can then import here from inside the action function). https://github.com/resin-io/resin-cli/blob/master/lib/utils/patterns.coffee is an example - it requires at the top-level, but that's not a problem because everywhere that requires it does so from inside an action.

Up to you which one, depends on whether you think moving these functions will be useful for other parts of the code/future changes.

zvin: I'm not sure moving these to `utils` right now would make sense.
If someday someone wants to reuse this function, he/she can move it to utils.
For example utils already has a `selectApplication` function but I couldn't reuse it because the app_name instead of the application itself.

I'll go with the simple solution: requiring resin-cli-visuals in selectApplication.

zvin: I'll do the same for `expectedError`, `resin` and `form` then.

zvin: done

PR #644, opened 2017-08-23
Check for uncommitted output

pimterry:
Connects to #573

With this change, if you commit any source files without also committing the build output, Travis will build, spot it, and fail.

This also conveniently means we're now actually running the build itself in CI, so Travis catch at least obvious compile-breaking errors in future.
---- Autogenerated Waffleboard Connection: Connects to #573


pimterry:
Ha, the build then fails, because we've generated a package-lock.json in travis that we haven't committed. Travis using npm@5 seems risky to me, given that we're not happy its stable enough to use locally, so I've bumped it down to 4.

automation/catch-uncommitted-output
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
lekkas: Can we rename the file to `catch-uncommitted-output.sh` to give a hint on its type?

automation/catch-uncommitted-output
```diff
@@ -0,0 +1,16 @@
...
+set -e
+
+npm run build
+
+if ! git diff HEAD --exit-code ; then
+	echo '\n** Uncommitted changes found after build - FAIL **'
+	exit 1
+elif test -n "$(git ls-files --exclude-standard --others)" ; then
```
lekkas: It'd be nice if we could avoid duplication of `git ls-files --exclude-standard --others` here

automation/catch-uncommitted-output
pimterry: Good idea, done

automation/catch-uncommitted-output
pimterry: Agree, done

Issue #646, opened 2017-08-25
Add an option to `resin preload` to ignore the image device type.


zvin:
Like https://github.com/resin-io/resin-preload/pull/97

PR #647, opened 2017-08-25
Add a --dont-check-device-type option for `resin preload`

zvin:
Connects-To: #646

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #646


zvin:
This is similar to https://github.com/resin-io/resin-preload/pull/97

PR #659, opened 2017-09-14
Move to using the catch-uncommitted npm package

pimterry:
See https://github.com/resin-io-modules/catch-uncommitted


pimterry:
Hound doesn't seem to be running, which is blocking this, and I have no idea why.

I'm actually fine with that though, because we lint in the build anyway (so Hound is just a kind of loud way to get inline feedback, now that we build PRs), and we've had a few annoying problems with it using the wrong config settings and things in the past. I'm going to stop drop it as a required status.

Issue #443, opened 2017-03-10
With a bad token, resin cli can become entirely broken


pimterry:
As part of reproducing #442, I changed the password for a test account to be purely numeric, and started reproducing and solving the issue. I had an old leftover token for this test account though, and with that in place, it was impossible (before or after that fix) to login or logout or see the help, or do anything with the cli, until the token is manually deleted from the disk. Not clear whether it's because the token has expired or because the account details were changed. Removing the token solves the issue.

```bash
> resin help
ResinRequestError: Request error: Unauthorized

If you need help, don't hesitate in contacting us at:

  GitHub: https://github.com/resin-io/resin-cli/issues/new
  Gitter: https://gitter.im/resin-io/chat
```

Haven't investigated in depth, but I think this is because we check `resin.auth.whoami()` when [logging mixpanel events](https://github.com/resin-io/resin-cli/blob/master/lib/events.coffee#L14), before we try to run any commands, and that explodes.

I've got a backup of the broken token I'm happy to share on demand, if that's useful.

pimterry:
Looks like this is related: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/9aaxkeZiIhX9p07X7d81Gy7FTDc

In that case, Joe's hitting a similar looking issue after recreating the devenv, but with the old token leftover.

pimterry:
Fixed in 6.6.8

PR #675, opened 2017-10-05
Ensure whoami failures (i.e. present but broken tokens) at startup don't break commands

pimterry:
Right now, if your token is invalid - i.e. expired, or for the wrong service (staging vs prod) or somehow corrupted - then every command fails, including `help` and `login`, because they try to get your username up front for analytics and that explodes. That's pretty bad. This stops everything exploding.

There's some other parts to complete this full, but once this is released in combination with https://github.com/resin-io/resin-sdk/pull/414 and https://github.com/resin-io-modules/resin-request/pull/105, this will solve #443 :champagne:


pimterry:
> How about also adding a cathReturn after events.trackCommand(cli) in app.coffee? Makes sense that Tracking errors should never prevent you to do your actual task.

Another excellent idea :+1:.

Done, in the definition of trackCommand itself rather than the calling code, since nobody should ever be waiting on that.

I've also added a timeout, so if it works but it's just a bit slow, we don't wait around. Thought about not waiting for the promise at all, but it'd be nice to increase the odds we'll get events for commands that run pretty quickly (`resin help`).

thgreasi:
Would it make sense to do something like this?
```coffee
Promise.all([
	events.trackCommand(cli)
	cli.global?.help ?
		capitanoExecuteAsync(command: "help #{cli.command ? ''}")
		capitanoExecuteAsync(cli)
])
```
So that the actual command can start the processing right away?

pimterry:
Nice, definitely, done.

lib/events.coffee
```diff
@@ -30,3 +30,5 @@ exports.trackCommand = (capitanoCommand) ->
 				resinUrl: resinUrl
 				platform: process.platform
 				command: capitanoCommand
+	.timeout(100)
```
thgreasi: Lets say that there is a command runs in 1ms and the tracking code for some reason needs 150ms (eg slow DNS or overcrowded wifi).
Is there a chance that the cli closes and everything gets disposed before the mixpanel tracking completes and we loose that stats?
Would it make any difference to increase the timeout to something a bit larger?
(I'm not sure if that makes sense, but I was thinking of it last night, for cli commands that can complete without any network requests.)

pimterry: > Would it make any difference to increase the timeout to something a bit larger?

Maybe. I've gone for a 100ms to avoid this issue for cases even when it's super fast, but in slow conditions that can still happen. It's not too bad though: I think as long as the full request has been sent, it doesn't matter if the response hasn't come back, so events should be tracked even if we time out while waiting for the mixpanel server & response. If we time out in TCP connection setup on the other hand, yes, we're stuck.

This problem is always going to be present up to a point though - eventually we have to decide whether we want to make users wait, or lose events, and I think I'm fairly happy dropping events occasionally in those cases. We could boost the timeout... I was trying to pick a time that's going to generally be sufficient, but isn't going to ever be noticeable if users do have to wait for it.

For now I've bumped it to a 500ms max. That could still time out, but it feels like it should successfully fire the event first, and if we block this beyond that point then we're well into noticeable experience, and even if we're going to lose events, I don't want to make users wait.

thgreasi: :+1:

PR #685, opened 2017-10-09
Document how to `resin deploy` to an app as a collaborator

pimterry:
Connects to #594 

Also wrapped the example deploy line in backticks, as the `<appName>` was being interpreted as HTML otherwise, and disappearing.


pimterry:
Ping @MoranF @CameronDiver

lib/actions/deploy.coffee
```diff
@@ -114,7 +114,10 @@ module.exports =
 	help: '''
 		Use this command to deploy an image to an application, optionally building it first.
 
-		Usage: deploy <appName> ([image] | --build [--source build-dir])
+		Usage: `deploy <appName> ([image] | --build [--source build-dir])`
+
+		To deploy to somebody else's app, on which you're a collaborator, use
```
CameronDiver: I think that this would be better as 
```
To deploy to an app on which you're a collaborator, use:
```

pimterry: Yep, that's better - updated

PR #687, opened 2017-10-16
Update resin-preload to 4.0.2 to support preloading Edison images

zvin:
Change-Type: patch


lib/actions/preload.coffee
```diff
@@ -100,16 +100,18 @@ offerToDisableAutomaticUpdates = (application, commit) ->
 			body:
 				should_track_latest_release: false
 
+progressBars = {}
```
pimterry: Is there a reason this is defined outside the action, but `spinners` is indeed? Personally, I'd put both inside, just to make it clearer what this relates to.

lib/actions/preload.coffee
```diff
@@ -157,9 +159,26 @@ module.exports =
...
+		spinners = {}
+
+		spinnerHandler = (event) ->
+			spinner = spinners[event.name]
+			if not spinner
+				spinner = spinners[event.name] = new visuals.Spinner(event.name)
+			if event.action == 'start'
+				spinner.start()
+			else
+				console.log()
```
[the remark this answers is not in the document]
zvin: We have several spinners. Some of them are quite fast.
When a spinner ends and another starts, the first one is erased.
If I output a newline here, the old spinner stays (but stops spinning) so we can see the history of what happened.

lib/actions/preload.coffee
zvin: There's no reason to keep it here, I just forgot to move it, it is fixed now.

Issue #700, opened 2017-10-25
Allow specifying `--commit latest` for `resin preload`


zvin:
Asked by a customer here https://app.frontapp.com/open/cnv_9wa3w9

- **resin-cli version:** all
- **Operating system and architecture:** all

Issue #600, opened 2017-07-26
`resin local configure` is broken if you answer "no" to setting advanced settings


emirotin:
Fails with `Missing variable hostname` which is probably coming from reconfix not getting this value (because we're not setting it, right) and not being able to skip configuring this file.

Reproducible with the fresh image that doesn't have the `hostname` prop in the config.json

Will fix after #597

emirotin:
May be related to the slightly outdated version of OS on my card, will check later with the freshest OS

PR #701, opened 2017-10-25
Allow specifying `--commit=latest` for `resin preload`

zvin:
Connects-To: #700

Depends-On: https://github.com/resin-io/resin-preload/pull/137

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #600
---- Autogenerated Waffleboard Connection: Connects to #700


zvin:
The build fails because resin-preload 4.0.9 is not released yet

Issue #705, opened 2017-10-27
Update resin-preload to support jetson-tx2


zvin:
Also remove the --dont-detect-flasher-type-images option which is no longer required as the flasher image detection has improved.

pimterry:
I think this is done now with that linked PR, right? I'm going to close it, shout if that's wrong.

Generally @zvin you don't need to create issues like this for the CLI. Issues here (and in the SDK etc) are good for planning work or bug fixes that need doing one day in the future, but you don't need one before opening a PR that's totally self-explanatory. We're not using waffle at all, mainly because there's no need to track done vs released when they're the same thing.

For things like this, you can just open a PR and get it merged, that's all that's required.

zvin:
Thanks @pimterry I forgot to close it.
I'll make PRs directly in the future.

PR #706, opened 2017-10-27
Update resin-preload to 5.0.0 to handle jetson-tx2

zvin:
Also remove the --dont-detect-flasher-type-images option

Connects-To: #705

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #705


zvin:
@pimterry I removed the --dont-detect-flasher-type-images option, that's why I've set the Change-Type to minor. Is that correct ?

pimterry:
Are there any cases where this might break existing behaviour for users using that flag? If not, this is fine, if so, this should be a major bump.

Either way, this should be in the top line of the commit message, so it appears in the changelog. Something like "Improve detection logic, and remove `dont-detect-flasher-type-images` option".

Generally these commit messages shouldn't focus on the "update X to C" kind of details - that's fine if you want to add extra detail, but the only important thing is that it fixes preload for the TX2.

zvin:
@pimterry I updated the commit message.
You can still specify any option you want but --dont-detect-flasher-type-images won't have any effect, so users won't be impacted by this.

Issue #691, opened 2017-10-19
Make UUID configuration optional in `os configure` 


brownjohnf:
As reported in the linked Front ticket, a user is attempting to configure an image, setting the wifi credentials and processor count, and then also preload an image. The user is able to configure via the Python SDK, but is unable to preload with the Python SDK. 

Conversely, the user is able to preload with the CLI, but is unable to configure (because the CLI configure command requires a UUID)

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_38e1)

pimterry:
@shaunmulligan I'd like to do this, since it should be fairly easy and it sounds like it's a blocker for the linked customer. The only sensible way of handling it I can see is a breaking CLI change though, is that a concern? I'd like to go from:

```
resin os configure <image> <uuid>
resin os configure <image> <uuid> [deviceApiKey]
```

to

```
resin os configure <image> --app <appName>
resin os configure <image> --device <uuid>
resin os configure <image> --device <uuid> --deviceApiKey [deviceApiKey]
```

I think that fits in with our long-term design ideas for the CLI, and it matches the signature for `config generate`. Happy to go for it?

shaunmulligan:
@pimterry , I think that makes sense. Are we marking breaking changes explicitly in the Change Log and also warn when something is deprecated?

pimterry:
@shaunmulligan We don't have any deprecations that I know of right now. Yes though, all breaking changes should be marked as such in the changelog.

I was expecting to just break the CLI API and jump here to v7, rather than trying to support both formats but deprecate the older one. Is that what you're suggesting? We could try to support both in the short-term, but it's fairly awkward in the code (capitano particularly doesn't support this particularly well), so I'd rather not. Since it's a breaking release nobody will automatically upgrade, so we could just break it and people can stay on v6 until they have a minute to make the small change and move to the new format. Do you think we need a more complicated deprecation process here?

pimterry:
Are you happy with the above? Should be quick and easy to do if so, I just need you to confirm.

shaunmulligan:
@pimterry that sounds good. Will the multicontainer stuff also require any breaking changes or can we also just keep that in the v7.x.x line? I just don't want to have regular breaks :)

pimterry:
[...]

PR #718, opened 2017-11-16
Allow configuring images for both device and just applications

pimterry:
Connects to #691 

This deprecated the old format, and prints a warning, but doesn't break anything. It adds two new ways to call this command:

```
$ resin os configure ../path/rpi.img --device 7cf02a6 --deviceApiKey <existingDeviceKey>

$ resin os configure ../path/rpi.img --app MyApp
```
---- Autogenerated Waffleboard Connection: Connects to #691


MoranF:
@pimterry what is ```deviceApiKey``` ?

pimterry:
_Excellent_ question, and this is not well documented right now, so let me explain. This is a bit long, sorry.

## Context

Firstly, these are different to _user_ api keys, which is what you've been working on (those are aiming to make it easy for users to create, manage and use their own api keys). This is all about the keys that live on devices, which they use to get their data from the api and update their status etc etc.

Once upon a time, every device got an _application_ api key. This was allowed to do more or less anything in the entire application the device lived in. That means it can register a new device (itself), get hold of the env vars etc, and pull the correct code and so on. Unfortunately, it also means it can see all the info about other devices in the same application, and I think potentially change their data too, which is bad. Especially because devices are out there in the world - somebody might grab one and steal the api key from the sd card, which would give them access to all the other devices in that application. Instead, we want the keys on the devices to have the minimum possible privileges, to avoid this.

Nowadays we want to fix this, by instead using provisioning keys & device api keys, both of which have very limited scope.

## Current Provisioning Workflow

A provisioning key is only allowed to register devices, for one application, and get the device api key for that new device. It's not allowed to do anything else.

A device api key is allowed to do almost anything, but only for one specific device.

The idea is that you can configure your OS images with a provisioning key, and then put that onto lots of different devices, and when they each start up they register themselves, get a device api key, and start using that (and delete the provisioning key immediately). That way you can use the same image for many devices, but they all end up with a unique device api key, and stealing the key from the device still only lets you affect the data for that one device.

Some of the work to get this working everywhere is still in progress, so we don't use provisioning keys all the time (old images don't support them). There's ongoing work to extend that though and implement this more and more completely.

## Pre-provisioning workflow

What's happening here is an extension of the normal provisioning workflow. Sometimes users want to _preregister_ devices.

Imagine you're provisioning 1000 devices, and you want to associate data with them before they start up. Maybe you want to add them all to your app db, so they can start up and immediately know what they're doing. You need to register them all in the API in advance, so you have UUIDs you can use to refer to these devices. To do that, you can do the provisioning steps above yourself, in advance. They you can start the devices up with a UUID & device api key already included.

To do that, you would run `resin device register` 1000 times, which gives you 1000 UUIDs and device api keys, and puts 1000 offline devices in your dashboard. You then generate the config for each of your 1000 devices, with the UUID and device api key pair for each device: `resin config generate --device <uuid> --deviceApiKey <apiKey> --output <uuid>.json`. When you actually set up the devices you run `resin config inject <uuid>.json --device <my-sd-card>` for each device to add the correct config, and they'll start up already configured as the expected device, and be able to skip the provisioning step entirely.

Phew. Make sense? There's quite a lot of context in here, but it's probably all useful in understanding what's going in the CLI generally, and this is knowledge we really need to share around the team more. Feel free to ping me with more questions!

PR #720, opened 2017-11-23
Start adding TypeScript to the CLI and stop committing build output

pimterry:
@dfunckt mentioned that he's going to be adding some CLI features soon, based on TS code elsewhere: https://www.flowdock.com/app/rulemotion/resin-tech/threads/w9ehtABDLGFpdKQ8bJLI2hVXuiA. That means it'd be useful to start the TS migration sooner rather than later, if possible.

This PR does that, largely based on the patterns in the SDK. It:

* Adds a minimal TS setup
* Converts a single tiny file, to check it works
* Stops committing the JS output of the build. It gets regenerated at publish time anyway, and it's just a recipe for conflicts and hassle.
* Moves from a `.npmignore` blacklist of files to not publish to a `files` whitelist in package.json.


package.json
```diff
@@ -37,7 +43,8 @@
     "gulp-coffee": "^2.2.0",
     "gulp-coffeelint": "^0.6.0",
     "gulp-shell": "^0.5.2",
-    "require-npm4-to-publish": "^1.0.0"
+    "require-npm4-to-publish": "^1.0.0",
+    "typescript": "^2.5.3"
```
thgreasi: Is there a reason we didn't got for ^2.6 ?

pimterry: This is what was in my npm5 cache, and I installed it whilst flying over Sardinia on Thursday :smiley:. Good reminder - I've updated it to the latest.

thgreasi: Wow, so npm5's offline cache works! That's awesome!

dfunckt:
Thank you @pimterry @thgreasi 💯

PR #721, opened 2017-11-23
Inline the entire resin-cli-auth module 

pimterry:
This is part of a general push to demodularize any code that isn't realistically reusable outside resin-cli, to make the codebase easier to manage and understand. Any code we have that nobody is going to want to use elsewhere should be moved into the core of the CLI itself.

Once this is merged, we can deprecate resin-cli-auth.

I've also removed all usages for underscore.string, since I noticed it as part of pulling in the dependencies from resin-cli-auth. It was a bit weird, and we don't really need it, so I've dropped it entirely. 

This PR will conflict a little with https://github.com/resin-io/resin-cli/pull/720, and the build probably won't pass until that's merged (because I haven't committed the build/ output, to limit the conflicts with that PR, which removes all build output). That's fine - as long as the tests pass and it only fails on uncommitted changes then here we're good, and everything should pass once the other PR is merged and this is rebased.


pimterry:
(Was failing with test failures, but this is now failing 'right' instead)

pimterry:
Now rebased on top of the now-merged #720, so this should be all passing and good to go if you're happy @MoranF.

Issue #607, opened 2017-08-03
package up the cli using pkg


shaunmulligan:
It would be great if the CLI could be a self contained, installable binary. We should be able to do this in the same way etcher.io does here: https://github.com/resin-io/etcher/blob/master/Makefile#L231

pimterry:
Duplicate: https://github.com/resin-io/resin-cli/issues/484

PR #728, opened 2017-12-14
*BREAKING*: Remove support for plugins entirely

pimterry:
There are very few plugins in real-world use afaict, we're not actively working on this at all, and the current approach breaks once we move to standalone node-less binary installation anyway.

This is a breaking change! I'm opening it for review now, to be merged into a v7 branch where I'm going to start collecting breaking changes for the next CLI release. I don't think that's coming super super soon, but we will want to ship this before/at the same time as standalone binaries become the preferred installation method, and there's definitely some other breaking changes coming soon too anyway.

Change-Type: major


shaunmulligan:
@pimterry I think the only thing we ever used the plugins for was `resin sync`, but I think that has since been included directly in the CLI, correct?

lurch:
> There are very few plugins in real-world use afaict

Do we know what plugins exist, and who / if they're being used by? I guess if there are any plugins that _are_ being actively used, their functionality could be incorporated into the CLI itself?
BTW I'm just asking out of curiosity, and as I know almost nothing about the resin-cli, feel free to ignore my comments :)

shaunmulligan:
@lurch I think Juanchi once made a demo plugin as sort of documentation, but we never pushed this plugin interface or documented it, so my guess is there 0 external ones.

pimterry:
Appveyor fails here because this branch isn't yet configured for appveyor at all - we can ignore that I think, and it'll magically start working once this is rebased over my upcoming other changes.

pimterry:
In terms of existing plugins, to work they'd have to be prefixed with `resin-plugin-`, and there's nothing with that name I can find that isn't created by us, and very very rarely used: https://www.npmjs.com/search?q=resin-plugin-

pimterry:
Merged into the v7 branch, but this will only actually be released once that branch is merged into master.

alexandrosm:
@pimterry -- has this come through an arch call? seems like a big change
that would warrant it.

Two things about plugins -- first, they've allowed us to increment
functionality into the cli (preload and sync) and second, they are part of
mine and petros' thinking about the future of resin. It would have been an
unpleasant surprise if I hadn't seen this.

I get the point about the bundling in the future, but I want to understand
the precise way in which plugins are getting retired.

Please put this on hold until the change is discussed.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

pimterry:
@alexandrosm I see, sorry about that. I asked @shaunmulligan about it, and he said it was fine: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/JuPnSagp_k7hgcZr5F_cEiOTZcZ. 

I'm not sure preloading did ever use this (certainly it was never been published on npm, which is effectively a requirement), and it doesn't look like anybody has published any plugins for at least 2 years now. If it's important for the future though we can definitely discuss it, and there's certainly ways we can make this work with packaging (note that packaging is actually here already, though still a little experimental: #607).

I'll open an arch call topic now, and make sure we resolve that before this gets shipped.

alexandrosm:
awesome, thanks!

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lekkas:
I've also raised a process call item to make sure that this doesn't happen again in the future - we need to have references to internal discussion threads (ideally product call discussion notes/decisions) in PRs like this that introduce breaking changes. After all,  we already do this for backend components, where links to arch call notes are generally included in PRs @pimterry @alexandrosm

Issue #484, opened 2017-04-17
Release standalone version (that ships with Node.js)


jviotti:
Various Etcher components that take care of the writing are going native. Since Etcher is also going to publish a CLI version, we sorted out a way to embed Node.js on the package,  code-sign it, and thus avoid any compilation need.

This will be part of Capitano in the future, but for now, take a look at the following Etcher scripts:

- Makefile target: https://github.com/resin-io/etcher/blob/master/Makefile#L253
- Bash script: https://github.com/resin-io/etcher/blob/master/scripts/build/node-package-cli.sh

The way it works is that we download a patched version of Node.js that locks its own entry point to an `index.js` file in the same file of the binary. 

In order to make use of this feature, we need to make sure the Resin CLI code can be concatenated into a single file.

emirotin:
The updated plan is to use pkg (as ethcer does atm, too).

cron410:
+1 I'm a n00b with NPM. I can't get it to work at all in the Windows 10 Bash environment. Yes I really should be using Linux for that.

pimterry:
A standalone cross-platform version is definitely coming eventually, hopefully before the end of the year. Watch this space!

pdcastro:
I think this issue has been implemented, no? What is missing?
https://www.balena.io/docs/reference/cli/#standalone-install

PR #729, opened 2017-12-14
Standalone binary for Resin-CLI

pimterry:
Connects to #484 

With this change, all releases automatically build & publish standalone zipped binaries of resin-cli to github. You can download the result and try it out at https://github.com/resin-io/resin-cli/releases. 

I've given this some _brief_ initial testing on Linux, Windows and OSX, and it certainly seems to run happily so far. I've just successfully logged in & run a `local push` on a windows machine without node or docker or anything installed, for example, which was my main goal.

I've marked this as experimental throughout the docs anyway: we'll want to test more thoroughly in future, there'll be some commands that make assumptions that don't quite work out any more, there's definitely a few further improvements we could make for usability (e.g. an actual installers, bundling `ssh`, maybe `rsync`...?), and the build process is a bit proof-of-concept in places. Still, I think this is in a state where we could start getting users testing it, and it's definitely a much easier installation process for some users (on Windows especially) than npm.

I've kept this as a separate PR to https://github.com/resin-io/resin-cli/pull/728 - this leaves a bunch of code in the CLI for plugins which is totally useless in the standalone format, but I think that's fine for now, whilst it's just an experimental alternative option.


automation/build-bin.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as fs from 'fs-extra';
+import * as filehound from 'filehound';
+import { exec as execPkg } from 'pkg';
+
+const ROOT = path.join(__dirname, '..');
+
+console.log('Building package...\n');
+
+execPkg([
+	'--target', 'host',
```
jviotti: Not sure if this is something you want to consider, but if the host system is GNU/Linux and has a pretty recent glibc version, then the CLI will only work on very modern GNU/Linux distributions (usually breaking Debian and some other modern + conservative ones).

You can use a Docker container with Ubuntu 12.04 or Debian Wheezy to ensure compatibility, as we do for the Etcher CLI.

lurch:
Will you be building 32-bit for Windows and Linux (like we do on Etcher) or will this be 64-bit only?

.gitignore
```diff
@@ -35,3 +35,5 @@ resinrc.yml
 
 /tmp
 build/
+build-bin/
+resin-cli-*.zip
```
thgreasi: NIT: would it make sense emit the zip files ain a `build-zip` folder to keep things tidy.

README.md
```diff
@@ -34,14 +37,37 @@ If you still want to use `cmd.exe` you will have to use a package manager like M
...
+If you don't have node or a working pre-gyp environment, you can still install the CLI as a standalone
+binary. **This is in experimental and may not work perfectly yet in all environments**, but it seems to work
+well in initial cross-platform testing, so it may be useful, and we'd love your feedback if you hit any issues.
+
+To install the CLI as a standalone binary:
+
+* Download the latest zip for your OS from https://github.com/resin-io/resin-cli/releases.
+* Extract the contents, putting the `resin-cli` folder somewhere appropriate for your system (e.g. `C:/resin-cli`, `/usr/local/lib/resin-cli`, etc).
+* Add the `resin-cli` folder to your `PATH`. (
+[Windows instructions](https://www.computerhope.com/issues/ch000549.htm),
```
thgreasi: From now and on, whenever I need this, I will visit the CLI docs instead of googling :)

lurch:
I dunno if this patch from Etcher is at all relevant to resin-cli? https://github.com/resin-io/etcher/blob/master/patches/cli/lzma-native-index-static-addon-require.patch

Also, as you may or may not be aware, there's a big stack of open issues with pkg https://github.com/zeit/pkg/issues

AFAIK @jviotti is our `pkg` expert ;-)

README.md
lurch: Just for the sake of comparison, here's the Etcher CLi instructions https://etcher.io/cli/
(although the Etcher CLI is officially still in beta, and I'm not sure how well tested it is)

jviotti:
I don't think the Resin CLI requires `lzma-native`, otherwise yes, that
patch will be needed, otherwise `node-pre-gyp` does some dynamic stuff
that confuses `pkg`.

On Fri, Dec 15, 2017 at 11:43:38AM +0000, Andrew Scheller wrote:
> I dunno if this patch from Etcher is at all relevant to resin-cli? https://github.com/resin-io/etcher/blob/master/patches/cli/lzma-native-index-static-addon-require.patch
> 
> Also, as you may or may not be aware, there's a big stack of open issues with pkg https://github.com/zeit/pkg/issues
> 
> AFAIK @jviotti is our `pkg` expert ;-)
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/resin-io/resin-cli/pull/729#issuecomment-351986031

-- 
Juan Cruz Viotti
Software Engineer

lurch:
Sorry if I confused people, I meant more "perhaps the ideas from this patch might be relevant to resin-cli", rather than suggesting resin-cli should be using a LZMA module ;-)

automation/build-bin.ts
pimterry: I'm not going to worry about this for this initial experimental release, but I've opened an issue to test and investigate this further later: https://github.com/resin-io/resin-cli/issues/738

.gitignore
pimterry: Good idea, done

pimterry:
@thgreasi Yes, all those logged warnings are expected. It seems to output the warnings regardless (annoyingly) but we then fix the underlying issues by copying in the various native modules etc afterwards.

@lurch I've started with just 32 bit since it's simpler, and I'd be surprised if we had a lot of users on non-64-bit machines. That's complete conjecture, and I'm definitely up for widening this in future. Do you have any stats from Etcher installs on platform usage that could help inform this?

@jviotti @lurch That pre-gyp patch is interesting. I don't think it _immediately_ affects this at least whilst it's marked experimental, since the only pre-gyp module I'm aware of is `ext2fs` (and the CLI works even if that isn't available), but I've opened an issue to check more thoroughly here: https://github.com/resin-io/resin-cli/issues/739.

I suspect we'll find all sorts of fun issues as we start trying to really use this, but as long as it _might_ be useful and doesn't break anything, I'd generally rather ship it with a big experimental warning, and iterate with real user testing & feedback coming in.

package.json
```diff
@@ -18,12 +18,26 @@
...
+      "node_modules/resin-sync/build/sync/*.js"
+    ],
+    "assets": [
+      "build/auth/pages/*.ejs",
+      "node_modules/resin-discoverable-services/services/**/*"
+    ]
+  },
   "scripts": {
-    "build": "gulp build && tsc && npm run doc",
+    "prebuild": "rimraf build/ build-bin/",
```
lurch: Should this also delete the `build-zip` folder?

pimterry: Oooh yes, good catch, done.

pimterry:
^^ Seems to have been a random failure, sadly, due to a token expiring during the tests? Haven't seen that before, very odd.

lurch:
> I've started with just 32 bit since it's simpler

I assume that was a typo? ;)

PR #734, opened 2017-12-15
Add analytics lib

izavits:
Use a single analytics library instead of having separate ones for Mixpanel and Sentry (Raven), etc.

Connects to: https://github.com/resin-io/resin-cli/issues/731

Connects to: https://github.com/resin-io/resin-cli/issues/703

Connects to: https://github.com/resin-io/resin-cli/issues/727


izavits:
@pimterry is there a way I can test Sentry? i.e. when you first included the raven lib, how did you validate that it was working?

In mixpanel we have the resin-dev project. Is there anything similar for sentry to see what is being logged there? 
thanks

pimterry:
@izavits just add some code that throws an error, after the point that Sentry initializes, and see if it turns up in https://sentry.io/resinio/cli/ within a few minutes.

There's no separate dev project at the moment. That's something I'd like to do as part of the bundling & packaging work, but right now errors during local dev end up in the same place as errors in production (though it's very easy to tell the difference if you look at the details).

lurch:
@pimterry If you come up with some way of testing Sentry without it going to the 'live' Sentry server, I think that'd be very useful for the Etcher project too. Ping @jhermsmeier @jviotti 
I previously wondered if we'd be able to make use of https://github.com/resin-io/resin-sentry for testing, but I know almost nothing about sentry so I don't even know if that's a sensible suggestion or not?

package.json
```diff
@@ -114,7 +112,8 @@
     "tmp": "0.0.31",
     "umount": "^1.1.6",
     "unzip2": "^0.2.5",
-    "update-notifier": "^2.2.0"
+    "update-notifier": "^2.2.0",
+    "analytics": "github:resin-io-playground/node.analytics"
```
lurch: Hmmm, does it "make sense" for a project in the `resin-io` org to have a dependency on a project in the `resin-io-playground` org? I dunno if there's any kind of policy around this ( @lekkas ?) but if `resin-cli` is depending on `node.analytics`, then perhaps it makes sense to move `node.analytics` from `resin-io-playground` to `resin-io-modules` ?

izavits: we will move node.analytics to another repo down the line. I want to make sure everything works as expected. I then support I'll move the analytics lib to another repo, publish to npm and then merge this PR.

lib/app.coffee
```diff
@@ -14,18 +14,20 @@ See the License for the specific language governing permissions and
...
-Raven.setContext
-	extra:
-		args: process.argv
-		node_version: process.version
+	disableConsoleAlerts: true
+Analytics.addIntegration(Raven)
+Analytics.initialize 'Sentry': ravenOptions
+Analytics.setContext Analytics.anonymize(extra:
+	args: process.argv
+	node_version: process.version)
```
pimterry: I think our general style guide for coffeescript is that you use brackets for one-line function calls, and no brackets for multiple line calls, so this shouldn't have brackets, and the line before should.

pimterry: Personally, I'd also drop the extra down a line, like:

```coffeescript
Analytics.setContext Analytics.anonymize
  extra:
    args: process.argv
    node_versionn: process.version
```

lib/app.coffee
```diff
@@ -14,18 +14,20 @@ See the License for the specific language governing permissions and
...
-.install (logged, error) ->
-	console.error(error)
-	process.exit(1)
-Raven.setContext
-	extra:
-		args: process.argv
-		node_version: process.version
+	disableConsoleAlerts: true
+Analytics.addIntegration(Raven)
+Analytics.initialize 'Sentry': ravenOptions
```
pimterry: What happens with errors in setup? The previous code caught them - was that unnecessary, or is this now not catching errors if Sentry fails to start up, or is something else dealing with this elsewhere?

lib/errors.coffee
```diff
@@ -17,10 +17,21 @@ limitations under the License.
...
+ravenOptions =
+	config: dsn
+	release: release
+	captureUnhandledRejections: true
+	disableConsoleAlerts: true
+analytics.addIntegration(Raven)
+analytics.initialize 'Sentry': ravenOptions
+analytics.setContext analytics.anonymize(extra:
+	args: process.argv
+	node_version: process.version)
```
pimterry: Why do we need to duplicate this between `app.coffee` and here? Does this only configure the `analytics` instance, not globally? That's slightly surprising, but I guess it makes sense.

It'd be good to put these in a separate file that's imported into both though, rather than duplicating it identically.

lib/errors.coffee
```diff
@@ -31,7 +42,10 @@ exports.handle = (error) ->
 
 	patterns.printErrorMessage(message)
 
-	captureException(error)
+	Promise.fromCallback((cb) ->
+		console.log('inside promise')
+		analytics.captureException(error)
+	)
```
pimterry: This promise never resolves: you never call `cb`, so this _always_ waits the second-long timeout below, which isn't good.

lib/errors.coffee
```diff
@@ -31,7 +42,10 @@ exports.handle = (error) ->
 
 	patterns.printErrorMessage(message)
 
-	captureException(error)
+	Promise.fromCallback((cb) ->
+		console.log('inside promise')
```
pimterry: Leftover console.log

package.json
pimterry: Yep, this is fine for review, but I'd definitely like to fix this and release node.analytics properly before we merge this PR.

lib/events.coffee
```diff
@@ -1,27 +1,32 @@
 _ = require('lodash')
-Mixpanel = require('mixpanel')
-Raven = require('raven')
+Analytics = require('analytics.node').core
+mixpanelIntegration = require('analytics.node').mixpanelIntegration
+ravenIntegration = require('analytics.node').sentryIntegration
 Promise = require('bluebird')
 resin = require('resin-sdk-preconfigured')
 packageJSON = require('../package.json')
+Analytics.addIntegration(ravenIntegration)
```
pimterry: Why is the raven configuration that's elsewhere not required here?

lib/events.coffee
```diff
@@ -1,27 +1,32 @@
...
 
 exports.getLoggerInstance = _.memoize ->
```
pimterry: Can we rename this to `getAnalyticsInstance`? That seems clearer now, given the code here.

pimterry:
You're going to get a few conflicts when you get to updating this, because https://github.com/resin-io/resin-cli/pull/743 is now merging. That converts a couple of these files to typescript. The exact same logic should still work fine, but you'll need to translate.

izavits:
@pimterry Thanks for the advice and comments. I'll start dealing with them tomorrow. OK, I'll try to translate to typescript. It'd be the first time :-)

pimterry:
Sounds good! Just ping me in flowdock if you need any help at all :+1:

lib/app.coffee
lurch: Yup: https://github.com/resin-io/process/blob/master/process/coffeescript-coding-standards.md

lib/errors.coffee
izavits: I have changed that. Although the .finally at the end behaved weirdly.

lib/app.coffee
izavits: I don't think we need all this setup in app.coffee after all. (unless I miss something)

lib/errors.coffee
izavits: I've removed those from app.coffee

izavits:
@pimterry can you take another look at this please?

LucianBuzzo:
Do we notify the users that we anonymously collect usage data? Is there any way for users to disable this? If not, I think we should probably consider it.

izavits:
Collecting usage data is something that we already do. I don't think there is a way in the CLI to disable this. Maybe this is something to consider for all the components because I think some time ago a user asked about the supervisor and what data it collects.

lib/events.ts
```diff
@@ -1,43 +1,51 @@
...
 		resinUrl: resin.settings.get('resinUrl'),
 		username: resin.auth.whoami().catchReturn(undefined),
-		mixpanel: getMixpanel(),
+		analytics: getAnalyticsInstance(),
 	})
-		.then(({ username, resinUrl, mixpanel }) => {
+		.then(({ username, resinUrl, analytics }) => {
 			return getMatchCommandAsync(capitanoCli.command).then(command => {
-				Raven.mergeContext({
+				Analytics.mergeContext({
```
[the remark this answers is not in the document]
pimterry: Is there any documentation anywhere on what `initialize` does, and what you're supposed to call on the global vs calling on the returned instance? I'm a little unclear about the structure of that, maybe if I understood that better this would be clearer.

pimterry:
@LucianBuzzo the 'should we always collect this' and 'how can users disable this' question is a good one, and is going to become pretty important pretty soon I think, since GDPR comes into force at the end of May. There's some ongoing discussion there in FD already: https://www.flowdock.com/app/rulemotion/r-operations/threads/Ig644kK_E-s_rOPaAZ1P6OqODap, https://www.flowdock.com/app/rulemotion/r-security/threads/6hBT25V6mqKRtLDYmHz3ZQs4REM.

I don't think we need to worry about it right now in this PR, but you're totally right that we'll probably need to do a bunch of work on this shortly (at least across the frontend, and probably quite a few other places too).

lib/app.coffee
```diff
@@ -14,19 +14,6 @@ See the License for the specific language governing permissions and
...
-Raven.config require('./config').sentryDsn,
-	captureUnhandledRejections: true
-	release: require('../package.json').version
-.install (logged, error) ->
-	console.error(error)
-	process.exit(1)
-Raven.setContext
-	extra:
-		args: process.argv
-		node_version: process.version
```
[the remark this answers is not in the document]
izavits: Alright, I'm changing it to the way you describe. Thanks!

lib/errors.ts
```diff
@@ -16,12 +16,31 @@ limitations under the License.
 
 import errors = require('resin-cli-errors');
 import patterns = require('./utils/patterns');
-import Raven = require('raven');
 import Promise = require('bluebird');
+import { core as analytics } from 'analytics.node';
+import { sentryIntegration as Raven } from 'analytics.node';
```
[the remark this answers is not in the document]
izavits: right, thanks :-)

lib/errors.ts
```diff
@@ -16,12 +16,31 @@ limitations under the License.
...
+import { sentryDsn } from './config';
+import { version } from '../package.json';
+const ravenOptions = {
```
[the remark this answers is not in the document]
izavits: yeah, it'd be better that way

lib/errors.ts
```diff
@@ -37,9 +56,12 @@ exports.handle = function(error: any) {
...
 	return captureException(error)
-		.timeout(1000)
 		.catch(function() {
 			// Ignore any errors (from error logging, or timeouts)
 		})
-		.finally(() => process.exit(error.exitCode || 1));
+		.finally(() => {
+			setTimeout(() => {
+				process.exit(error.exitCode || 1);
+			}, 1000);
```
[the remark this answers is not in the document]
izavits: alright, changing it!

lib/events.ts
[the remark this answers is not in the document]
izavits: yeah, I see your point. I'll change the names for now.
Regarding the other comment, initialize() calls the init/install method of the desired integration.

lib/events.ts
```diff
@@ -1,43 +1,51 @@
...
 					user: {
 						id: username,
 						username,
 					},
 				});
 
-				return mixpanel.track(`[CLI] ${command.signature.toString()}`, {
+				return analytics.track(`[CLI] ${command.signature.toString()}`, {
 					distinct_id: username,
```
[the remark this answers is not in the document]
izavits: I'm not sure I get this. We are still using the node version of mixpanel here.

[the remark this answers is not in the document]
pimterry: We're using the node version of mixpanel internally, but this code shouldn't care about that. We're building this analytics library so we can abstract away the details of the underlying analytics tool, and have a single interface that works identically for everything.

It looks like analytics.js expects us to call `analytics.identify` to configure it with a userId, and for all the built in analytics services it then automatically uses this id in the appropriate way for each service. I think that's what we should be doing here for this integratoin as well: calling `identify` once to set the user, and then having it automatically used in all future `track` calls.

pdcastro:
Closing this PR as stale (2 years old, a lot changed in the codebase).

Issue #736, opened 2017-12-18
Implement host OS access through CLI


afitzek:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_5q2h)

thgreasi:
@afitzek Should this be closed now that #737 is merged?

afitzek:
Yes

PR #737, opened 2017-12-18
Add ssh option for direct host OS access

afitzek:
Use the `--host` (short `-s`) option in the ssh command to access
the host OS of the device.
Direct host OS is enabled for devices with Resin OS >= 2.7.5.

Change-Type: minor
Connects-To: #736
Signed-off-by: Andreas Fitzek <andreas@resin.io>
---- Autogenerated Waffleboard Connection: Connects to #736


doc/cli.markdown
```diff
@@ -819,6 +820,10 @@ ssh gateway port
 
 increase verbosity
 
+#### --host, -H
+
+access host OS
```
thgreasi: I guess that we should document that this is
> enabled for devices with Resin OS >= 2.7.5.

lib/actions/ssh.coffee
```diff
@@ -44,6 +45,11 @@ module.exports =
 			boolean: true
 			description: 'increase verbosity'
 			alias: 'v'
+	,
+			signature: 'host'
+			boolean: true
+			description: 'access host OS'
```
thgreasi: I guess that we should document that this is supported
> for devices with Resin OS >= 2.7.5.

lib/actions/ssh.coffee
```diff
@@ -44,6 +45,11 @@ module.exports =
 			boolean: true
 			description: 'increase verbosity'
 			alias: 'v'
+	,
+			signature: 'host'
+			boolean: true
+			description: 'access host OS (for devices with Resin OS >= 2.7.5)'
+			alias: 'H'
```
pimterry: We try to share command line arguments in the CLI e.g. in https://github.com/resin-io/resin-cli/blob/master/lib/actions/command-options.coffee when they're used in multiple places - I think it'd be tidier to do that here for this and `local ssh`.

Notably, doing so highlights a difference: `local ssh` uses `-s` as the host alias, not `-H`. It'd be good to keep this consistent, and we can't change the `local ssh` format without a breaking change, so I think this should be `-s` too (though I wouldn't be surprised if we end up changing them both of together later, once we get to the big CLI redesign).

doc/cli.markdown
```diff
@@ -808,6 +808,7 @@ Examples:
 	$ resin ssh 7cf02a6
 	$ resin ssh 7cf02a6 --port 8080
 	$ resin ssh 7cf02a6 -v
+	$ resin ssh 7cf02a6 -H
```
zwalchuk: Example still shows -H, should be -s I believe?

thgreasi:
Woohoo!!
Would this also allow support agents connect to devices that have granted access?

afitzek:
@thgreasi Yes.

thgreasi:
Woohoo x2 !! :+1: :+1: :raised_hands: :heart: 
(We should add that to the support docs then.)

PR #748, opened 2018-01-07
convert windows paths to posix when passing to tar

ashirley:
Due to https://github.com/mafintosh/tar-stream/issues/3, the tar module
needs posix style paths but system-specific paths are being supplied

### DIFF
--- a/lib/utils/docker.coffee
+++ b/lib/utils/docker.coffee
@@ -150,7 +150,7 @@ exports.tarDirectory = tarDirectory = (dir) ->
 		relPath = path.relative(path.resolve(dir), file)
 		Promise.join relPath, fs.stat(file), fs.readFile(file),
 			(filename, stats, data) ->
-				pack.entryAsync({ name: filename, size: stats.size, mode: stats.mode }, data)
+				pack.entryAsync({ name: toPosixPath(filename), size: stats.size, mode: stats.mode }, data)
 	.then ->
 		pack.finalize()
 		return pack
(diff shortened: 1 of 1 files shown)


lib/utils/docker.coffee
```diff
@@ -150,7 +150,8 @@ exports.tarDirectory = tarDirectory = (dir) ->
 		relPath = path.relative(path.resolve(dir), file)
 		Promise.join relPath, fs.stat(file), fs.readFile(file),
 			(filename, stats, data) ->
-				pack.entryAsync({ name: filename, size: stats.size, mode: stats.mode }, data)
+				posixFilename = filename.replace(new RegExp('\\' + path.sep, 'g'), '/')
```
pimterry: Could you pull this conversion out into a separate function, so these become something like `toPosixPath(filename)`? It'd be nice to make it clearer why this is required.

pimterry:
Oh, and thanks for the contribution! This is a great find and fix, nice work :+1:

pimterry:
^ that error was because it turns out that actually our autoversioning doesn't work nicely with PRs from external forks at all, which is awkward. We've got proper fixes for this planned long-term, but not soon. I'm fudging around this by opening a separate PR with your same commit, but from a branch on this repo instead of a fork, and that seems to work (and marks _both_ PRs as merged, neatly).

This is just an fyi though - everything seems to be working for now, no need for you to do anything. This merged code is on master now, and should get autodeployed to npm imminently.

Issue #749, opened 2018-01-07
emulated builds on windows


ashirley:
- **resin-cli version:** 6.12.1 (and HEAD)
- **Operating system and architecture:** Windows x64

I have been trying to run an emulated build from windows. I have raised issue 748 on a related but orthogonal matter. There are 2 remaining issues. The first is that qemu is copied into the container but is not executable. The easiest solution to this is to always add it to the tar ball with 777 permissions. I am happy to either include this in 748 or raise a separate PR for it.

The other thing is the platformNeedsQemu function which means we only try to actually do any emulation on linux. I'm not sure what the intention of this function is but if it returns false, -e doesn't do anything except a console.log. 

I can either remove it entirely or change it to be based on `process.arch` so that we don't try to emulate if we are already running on the target architecture (although arguably, it is the user's responsibility to provide -e only when they mean it?), which ever you'd prefer. Again I can either roll this into 748 or put it in a separate PR.

Let me know what you'd prefer and I'll prepare a PR.

pimterry:
`platformNeedsQemu` returns false for windows mainly because emulation hasn't been tested on Windows, and we simply have no idea if it'll work (most of our developers, customers & users are on Linux or Mac). I'd love to add Windows support though, and I'm very happy if you're in a place to do that investigation and open some PRs to get it going!

I'd rather we keep PRs separate, and leave #748 as-is, but I'd love PRs making the improvements you're talking about, if you've tested them and you've got full emulation working for you locally.

Some context by the way: with #673 we're eventually aiming towards making this argument unnecessary, and turning emulation on or off automatically (maybe removing the arg entirely) based on the local platform and target architecture. I'm not sure what the Windows situation will be, but it does need to take the local platform into consideration, as the qemu setup isn't required on OSX.

ashirley:
Raised  #750 for adjusting the file permissions. Changes to `platformNeedsQemu` still remaining

pdcastro:
I believe this issue was fixed in CLI v11.14.2 (currently 11.18.3), PR #1464. I have successfully tested emulated builds on Windows 10 (using Docker for Windows 18.06.1). If you find that the issue hasn't been fixed, just add a comment and I'll be happy to reopen it and investigate further.

PR #750, opened 2018-01-09
adjust file permissions on windows

ashirley:
This is done by the standard docker client so that files which need to be
executable are even though that isn't suported on windows.

https://github.com/moby/moby/blob/94b8a116fbf1cd90e68d8f5361b520d326a66f9b/pkg/archive/archive_windows.go#L43

Change-Type: patch


lib/utils/docker.coffee
```diff
@@ -418,3 +428,13 @@ copyQemu = (context) ->
...
+
+# if required, change the permissions (mode) of files added to the tar stream
+# This is needed for windows as the execute bit isn't supported there
+chmodTarEntry = (mode) ->
+	os = require('os')
+	if os.platform() == 'win32'
+		mode |= 0o111
+		mode &= 0o755
+	else
+		mode
```
ashirley: I'm not too familiar with coffee script and this was the only way I could find to return `mode` unchanged on non-windows platforms. If there is a more idiomatic way of doing it, I'll change it.

lib/utils/docker.coffee
```diff
@@ -418,3 +428,13 @@ copyQemu = (context) ->
 	.then ->
 		fs.chmod(binPath, '755')
 	.return(binPath)
+
+# if required, change the permissions (mode) of files added to the tar stream
+# This is needed for windows as the execute bit isn't supported there
+chmodTarEntry = (mode) ->
```
pimterry: Can you add a comment linking to the corresponding moby code for this that you've included in the PR description?

lib/utils/docker.coffee
```diff
@@ -196,13 +196,23 @@ exports.runBuild = (params, options, getBundleInfo, logger) ->
...
 	qemuPath = ''
 
+	if os.platform() == 'win32'
+		logger.logWarn '''
+			SECURITY WARNING: You are building a Docker image from
+			Windows against a non-Windows Docker host.
+			All files and directories added to build context will have
+			'-rwxr-xr-x' permissions. It is recommended to double check
+			and reset permissions for sensitive files and directories.
+		'''
```
pimterry: Can you add a comment here explaining that this is imitating the Docker client behaviour too, and link to the corresponding code for that?

lib/utils/docker.coffee
pimterry: There's a variety of other ways to do this, but this is fine imo, don't worry about it (we're in the process of moving this to TypeScript anyway, so the exact details of this kind of thing don't matter too much)

pdcastro:
I understand that this PR addresses issue #749 (emulated builds on Windows). Without realising that this PR existed, I fixed the issue in another PR, #1464. Given that this PR appears stale (almost 2 years old), I am taking the liberty of closing it.

PR #751, opened 2018-01-09
Convert windows paths to posix when passing to tar

pimterry:
#748 is an external PR from a fork, so doesn't work with versionbot. This PR is the same code, which versionbot has conveniently already version bumped (somewhat surprisingly - I think it's because the branch has the same name), but living on a local branch instead.


lurch:
Adding a quick ping to @hedss just to make sure VersionBot doesn't fall over, or do something daft like double-version-bumping ;-)

PR #755, opened 2018-01-11
Convert lots of actions to TypeScript

pimterry:
Mainly focusing on converting the actions code (this is 1/2 of them, but mainly the smaller ones). Notes:

* I've split most separate actions into separate commits, which should be fairly easily reviewable individually
* This also upgrades to TypeScript 2.6.2, and SDK 8 (lots of coffeescript is still using `resin-sdk-preconfigured`, which is SDK 6, but that's slowwwwly dying)
* There's a few bug fixes this found - e.g. it looks like ssh proxying never worked because we took a non-promise function and never passed a callback - but in general, this shouldn't be a functional change
* I've ended up with some really cool stuff in the resin-cli-form types, which is fun. Usage (and maybe slightly the types themselves) might be of interest to @LucianBuzzo, as we start looking at form generation more generally.


pimterry:
After a bug got introduced in the last PR, and since we have near-zero test coverage I thought I'd do some manual testing of the main changed commands:

- [x] help [command...]
- [x] login
    - [x] web
    - [x] credentials
    - [x] auth token
    - [x] register (against staging)
- [x] apps
- [x] app <name>
- [x] logs <uuid>
- [x] sync [uuid]
- [x] ssh [uuid]
- [x] logout
- [x] signup (against staging)
- [x] whoami
- [x] app create <name>
- [x] app rm <name>
- [x] app restart <name>
- [x] note <|note>
- [x] util available-drives

That's just the basic successful case, so there might be some interesting corner cases, but in general it's looking good!

lib/actions/ssh.ts
```diff
@@ -0,0 +1,182 @@
...
+		const childProcess = await import('child_process');
+		const Bluebird = await import('bluebird');
+		const resin = (await import('resin-sdk')).fromSharedOptions();
+		const _ = await import('lodash');
+		const bash = await import('bash');
+		const hasbin = await import('hasbin');
+
+		// hasbin ignores the error-first convention, so we can't promisify automatically
+		const hasbinAsync = (bin: string) =>
+			new Promise<boolean>(resolve => hasbin(bin, resolve));
```
LucianBuzzo: Is there a way that we can avoid using native Promise *and* bluebird here? one of the other would be great

lib/auth/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @param {String} sessionToken - token
+ * @fulfil {Boolean} - whether is valid or not
+ * @returns {Promise}
+ *
+ * utils.isTokenValid('...').then (isValid) ->
+ *   if isValid
+ *     console.log('Token is valid!')
+ */
+export function isTokenValid(sessionToken: string) {
```
LucianBuzzo: This is pre-existing, but this would be a good time to make the implementation match the TSdoc here. The docs say its return type is `Promise<boolean>` but its actually more like `Promise<boolean | void | ???>`

typings/bash.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'bash' {
```
LucianBuzzo: What a strange and futuristic world we live in!

lib/actions/ssh.ts
```diff
@@ -0,0 +1,182 @@
...
+			signature: 'noproxy',
+			boolean: true,
+			description: `don't use the proxy configuration for this connection. \
+Only makes sense if you've configured proxy globally.`,
+		},
+	],
+	async action(params, options, done) {
+		const childProcess = await import('child_process');
+		const Bluebird = await import('bluebird');
+		const resin = (await import('resin-sdk')).fromSharedOptions();
```
thgreasi: Should we do `const { resin } = await import('../sdk');` like in `lib/actions/settings.ts`?

lib/actions/ssh.ts
```diff
@@ -0,0 +1,182 @@
...
+		const hasbin = await import('hasbin');
+
+		// hasbin ignores the error-first convention, so we can't promisify automatically
+		const hasbinAsync = (bin: string) =>
+			new Promise<boolean>(resolve => hasbin(bin, resolve));
+
+		const { getSubShellCommand } = await import('../utils/helpers');
+		const patterns = await import('../utils/patterns');
+
+		if (options.port == null) {
```
thgreasi: How about `if (!options.port)`?

lib/actions/ssh.ts
```diff
@@ -0,0 +1,182 @@
...
+		};
+
+		const uuidExists =
+			params.uuid && (await resin.models.device.has(params.uuid));
+		const selectedUuid = uuidExists
+			? params.uuid!
+			: await patterns.inferOrSelectDevice();
+
+		console.info(`Connecting to: ${selectedUuid}`);
+		const device = await resin.models.device.get(selectedUuid, {});
```
thgreasi: I'm not sure whether the empty options param is bc the sdk wasn't marking the param as optional, but if that's the case then this is a reminder to remove when the sdk is updated.

lib/actions/ssh.ts
```diff
@@ -0,0 +1,182 @@
...
+			containerId: options.host
+				? ''
+				: resin.models.device
+						.getApplicationInfo(device.uuid)
+						.get('containerId'),
+			proxyUrl: resin.settings.get('proxyUrl'),
+			hasTunnelBin: useProxy ? await hasbinAsync('proxytunnel') : false,
+		});
+
+		if (containerId == null) {
```
thgreasi: Could this be `if (!containerId)`?

lib/utils/patterns.ts
```diff
@@ -72,7 +73,7 @@ export function authenticate(options: {}): Promise<void> {
 }
 
 export function askLoginType() {
-	return form.ask({
+	return form.ask<'web' | 'register' | 'token' | 'credentials', 'list'>({
```
thgreasi: How about also having this (`'web' | 'register' | 'token' | 'credentials'`) as a type?

lib/utils/helpers.ts
```diff
@@ -102,7 +100,7 @@ export function osProgressHandler(step: InitializeEmitter) {
 		if (state.operation.command === 'burn') {
 			return;
 		}
-		console.log(exports.stateToString(state));
+		console.log(stateToString(state));
```
thgreasi: I guess that this wasn't working before.

lib/auth/server.ts
```diff
@@ -0,0 +1,147 @@
...
+import bodyParser = require('body-parser');
+import Promise = require('bluebird');
+import { resin } from '../sdk';
+import utils = require('./utils');
+
+function createServer(param?: { port?: number; isDev?: boolean }) {
+	if (param == null) {
+		param = {};
+	}
+	const { port, isDev } = param;
```
thgreasi: Is there really a chance of getting `null` param?
Could this be
```
function createServer({ port, isDev }: { port?: number; isDev?: boolean } = {}) { }
```
?

pimterry:
This is very outdated now - it got overrun by lots of changes for multicontainer, and never really caught up...

I'm going to close it, but if you're looking to do some more TS conversion in future @thgreasi there'll almost certainly be some good snippets in here that can help speed that up.

PR #763, opened 2018-01-25
Add docker-compose-aware builds and deployments

dfunckt:
This is part of the multi-container project, so it must not be merged before the backend has been deployed.

The PR looks massive, but in reality it's not: only the latest commit is relevant and for the most part is new code.

Legacy behaviour is mostly retained. The most notable change in behaviour is that invoking `resin deploy` without options is now allowed (see help string how it behaves).

In the latest commit there are also the following notable changes:

- Deploy/Build are promoted to primary commands
- Extracts QEMU-related code to a new file
- Adds a utility file to retrieve the CLI version and its parts
- Adds a helper that can be used to manipulate display on capable clients
- Declares several new dependencies. Most are already indirectly installed via some dependency

Other changes in separate commits include (see commit messages for full details):

- Update bundle-resolve and docker-build to latest
- Update docker-toolbelt
- Update YAML parser
- Fix invoking undefined method

Change-Type: minor


lib/actions/build.coffee
```diff
@@ -2,32 +2,65 @@
...
+	projectPath: the project root directory; must be absolute
+	buildEmulated
+	buildOpts: arguments to forward to docker build command
+###
+buildProject = (docker, logger, composeOpts, opts) ->
+	compose.loadProject(
+		logger
+		composeOpts.projectPath
+		composeOpts.projectName
+		opts.appName ? 'main')
```
CameronDiver: Can we get the closing paren on the next line please?

lib/actions/build.coffee
```diff
@@ -2,32 +2,65 @@
...
+			docker
+			logger
+			project.path
+			project.name
+			project.composition
+			opts.arch
+			opts.deviceType
+			opts.buildEmulated
+			opts.buildOpts
+			composeOpts.inlineLogs)
```
CameronDiver: Same here.

lib/utils/docker.coffee
```diff
@@ -131,210 +123,56 @@ exports.generateConnectOpts = generateConnectOpts = (opts) ->
...
+exports.generateBuildOpts = (options) ->
+	opts = {}
+	if options.tag?
+		opts['t'] = options.tag
+	if options.nocache?
+		opts['nocache'] = true
+	if options.squash?
+		opts['squash'] = true
+	if options.buildArg?
+		opts['buildargs'] = parseBuildArgs(options.buildArg)
```
CameronDiver: Why are you using the `['name']` form rather than `obj.name = ` ?

lib/utils/docker.coffee
```diff
@@ -131,210 +123,56 @@ exports.generateConnectOpts = generateConnectOpts = (opts) ->
...
+	# 	**New code must not use the *Async methods.**
+	#
+	Docker = require('docker-toolbelt')
+	Promise.promisifyAll Docker.prototype, {
+		filter: (name) -> name == 'run'
+		multiArgs: true
+	}
+	Promise.promisifyAll(Docker.prototype)
+	Promise.promisifyAll(new Docker({}).getImage().constructor.prototype)
+	Promise.promisifyAll(new Docker({}).getContainer().constructor.prototype)
```
CameronDiver: I'm not sure about docker-progress, but dockerode has promise support built-in now. Just don't pass a callback to the methods, and in the intialisation you can pass ` { Promise: require('bluebird') }` to get bluebird promises.

lib/actions/deploy.coffee
```diff
@@ -1,229 +1,183 @@
...
+						"Bearer #{auth}"
+						apiEndpoint
+						!opts.shouldUploadLogs)
+			)
+	.then ->
+		logger.logSuccess('Deploy succeeded!')
+		console.log()
+		console.log(doodles.getDoodle()) # Show charlie
+		console.log()
+	.catch (e) ->
```
thgreasi: NIT: Would it make sense to use `.tapCatch`?

lib/actions/deploy.coffee
```diff
@@ -1,229 +1,183 @@
...
+			if image? and options.build
+				throw new Error('Build option is not applicable when specifying an image')
+
+			Promise.join(
+				helpers.getApplication(appName)
+				helpers.getArchAndDeviceType(appName)
+				(app, { arch, device_type }) ->
+					app.arch = arch
+					app.device_type = device_type
+					return app
```
thgreasi: NIT: these three lines could be a single `_.assign(app, { arch, device_type })`.

lib/actions/deploy.coffee
```diff
@@ -1,229 +1,183 @@
...
+			)
+			.then (app) ->
+				[ app, image, !!options.build, !options.nologupload ]
+
+		.then ([ app, image, shouldPerformBuild, shouldUploadLogs ]) ->
```
thgreasi: > Would it make more sense to return an object & destructure it inside the `.then` instead of arrays with semantics based on the item position?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+			# Copy the qemu binary into the build context
+			qemu.copyQemu(path.join(projectPath, task.context))
+			.then (binPath) ->
+				transpose.transposeTarStream task.buildStream,
+					hostQemuPath: binPath
+					containerQemuPath: "/tmp/#{qemu.QEMU_BIN_NAME}"
+				.then (stream) ->
+					task.buildStream = stream
+				.return([ task, binPath ])
+	.map ([ task, qemuPath ]) ->
```
thgreasi: > Would it make more sense to return an object & destructure it

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+	releaseMod.create
+		client: client
+		user: userId
+		application: appId
+		composition: composition
+		source: "cli/#{versions.cli.major}"
+		commit: crypto.pseudoRandomBytes(16).toString('hex').toLowerCase()
+	.then ({ release, serviceImages }) ->
+		[ 'created_at', 'belongs_to__application', 'is_created_by__user', '__metadata' ].forEach (attr) ->
+			delete release[attr]
```
thgreasi: Would it make sense to `pick` only the properties that we are interested in instead of removing those that we are not?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+		_.keys serviceImages, (serviceName) ->
+			img = serviceImages[serviceName]
+			[ 'created_at', 'is_a_build_of__service', '__metadata' ].forEach (attr) ->
+				delete img[attr]
```
thgreasi: > Would it make sense to `pick` only the properties that we are interested in instead of removing those that we are not?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+					serviceImage.image_size = size
+					serviceImage.content_hash = digest
+					serviceImage.dockerfile = props.dockerfile
+					serviceImage.project_type = props.projectType
+					serviceImage.start_timestamp = props.startTime if props.startTime
+					serviceImage.end_timestamp = props.endTime if props.endTime
+					serviceImage.push_timestamp = new Date()
+					serviceImage.status = 'success'
+			)
+			.catch (e) ->
```
thgreasi: NIT: how about `.tapCatch`?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+				Promise.map images, ({ localImage }) ->
+					localImage.remove()
+		.tap (images) ->
+			return if skipLogUpload
+			logger.logInfo('Uploading build logs...')
+
+			Promise.map images, ({ serviceName, serviceImage, logs }) ->
+				releaseMod.updateImage(client, serviceImage.id, build_log: logs)
+				.catch (e) ->
+					logger.logError("Failed to upload build logs for '#{serviceName}' due to '#{e}'")
```
thgreasi: Just a heads-up that this doesn't rethrow and whether this was intentional.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+		.then ->
+			release.status = 'success'
+		.catch (e) ->
```
thgreasi: NIT: how about `.tapCatch`?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+				numSteps = match[2]
+			if step?
+				str = "Step #{step}/#{numSteps}: #{str}"
+				progress = parseInt(step, 10) * 100 // parseInt(numSteps, 10)
+
+		cb(null, { status: str, progress })
+
+pullProgressAdapter = (outStream) ->
+	return (e) ->
+		{ status, id, percentage, error, errorDetail } = e
```
thgreasi: NIT: how about moving the descructuring inside the function header, since `e` isn't used anywhere else?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+			return cb(null, { status: str })
+
+		if /^Successfully tagged /.test(str)
+			progress = undefined
+		else
+			if (match = stepRegex.exec(str))
+				step = match[1]
+				numSteps = match[2]
+			if step?
+				str = "Step #{step}/#{numSteps}: #{str}"
```
pimterry: This duplicates the `Step X/Y:` part of `str`, if present.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+		outStream.write({
+			status,
+			progress: percentage,
+			error: errorDetail?.message ? error
+		})
+
+createSpinner = ->
+	chars = '|/-\\'
+	index = 0
+	-> chars[(index++) % chars.length]
```
pimterry: Why aren't we using the existing spinner component?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+
+class BuildProgressUI
+	constructor: (tty, descriptors) ->
+		_ = require('lodash')
+		chalk = require('chalk')
+		through = require('through2')
+
+		services = _.map(descriptors, 'serviceName')
+		prefix = chalk.blue('[Build]') + '   '
```
pimterry: Why do we need to do manual prefixing etc etc here, instead of using/extending the existing logger that already handles that?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+		@_maxLineWidth = @_tty.currentWindowSize().width
+
+	_getServiceSummary: ->
+		_ = require('lodash')
+
+		services = @_services
+		serviceToDataMap = @_serviceToDataMap
+
+		_(services).map (service) ->
+			{ status, progress, error } = serviceToDataMap[service] ? {}
```
pimterry: With this, we're only showing the latest line from the docker build. For large `RUN` steps, that's ends up being `Step 3/6:  ---> Running in adebb1ebfc2a`, not the actual command being run. That's not useful, and it's going to happen all the time, for all the most expensive steps in people's dockerfiles. You do get useful output for the initial `FROM` and `COPY` etc, but I still think we need to improve this.

Could we easily show all the lines from the current step, not just literally the last line? It'd require us to put the status on a separate line below, but that's actually a bit clearer imo. Might be some tricky UX with the output changing in height though... Alternatively, we could special case `RUN` steps to provide better output? Need to think about how to handle long multiline RUN commands though...

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+# high-level function resolving a project and creating a composition out
+# of it in one go. if image is given, it'll create a default project for
+# that without looking for a project. falls back to creating a default
+# project if none is found at the given projectPath.
+exports.loadProject = (logger, projectPath, projectName, appName, image) ->
+	logger.logDebug('Loading project...')
+
+	Promise.try ->
+		if image?
+			logger.logInfo("Creating default resin-compose project with image: #{image}")
```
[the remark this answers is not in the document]
dfunckt: Yeah, I reused docker-compose terminology here because we lacked our own. Project is merely the folder which contains, well, a docker-compose.yml and related folders with Dockerfiles, etc. Happy to change to whatever we like (at least the most prominent use which is the option named `--project`). cc @shaunmulligan

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+			# Get the service-specific log stream
+			# Caveat: `multibuild.BuildTask` defines no `logStream` property
+			# but it's convenient to store it there; it's JS ultimately.
+			task.logStream = renderer.streams[task.serviceName]
+			task.logBuffer = []
+
+			# Setup emulation if needed
+			return [ task, null ] if task.external or not needsQemu
+			# Copy the qemu binary into the build context
+			qemu.copyQemu(path.join(projectPath, task.context))
```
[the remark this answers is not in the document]
dfunckt: I just need to write and export a new function that returns QEMU's path, as that's the only thing that's needed here. The first `copyQemu` call was hastily added to fix a bug and forgot to update this occurrence.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+				image.props.endTime = new Date(endTime)
+			docker.getImage(image.name).inspect().get('Size').then (size) ->
+				image.props.size = size
+			.return(image)
+		.tap (images) ->
+			summary = _(images).map ({ serviceName, props }) ->
+				[ serviceName, "Image size: #{humanize.filesize(props.size)}" ]
+			.fromPairs()
+			.value()
+			renderer.end(summary)
```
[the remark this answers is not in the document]
dfunckt: This predates the builder. I don't think having parity with the builder is feasible, given they only share a few core bits. I guess a table would be nice to have, but it should come in a future PR.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+			.finally ->
+				logger.logDebug('Untagging images...')
+				Promise.map images, ({ localImage }) ->
+					localImage.remove()
+		.tap (images) ->
+			return if skipLogUpload
+			logger.logInfo('Uploading build logs...')
+
+			Promise.map images, ({ serviceName, serviceImage, logs }) ->
+				releaseMod.updateImage(client, serviceImage.id, build_log: logs)
```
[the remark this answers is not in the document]
dfunckt: No reason, good point, I loosely followed what the previous command did, i.e uploading logs separately in the end.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+
+# utilities
+
+renderProgressBar = (percentage, stepCount) ->
+	_ = require('lodash')
+	percentage = _.clamp(percentage, 0, 100)
+	barCount = stepCount * percentage // 100
+	spaceCount = stepCount - barCount
+	bar = "[#{_.repeat('=', barCount)}>#{_.repeat(' ', spaceCount)}]"
+	return "#{bar} #{_.padStart(percentage, 3)}%"
```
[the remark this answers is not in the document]
dfunckt: It's very opinionated and doesn't fit here and can't really mend it to work. It assumes a prefix, for example, and the way it erases the line is incompatible with the way these commands do output.

lib/utils/compose.coffee
dfunckt: For the same reason I didn't use the progress bar.

lib/utils/compose.coffee
dfunckt: Good point, I don't really remember the reason now, will explore the code and add a comment.

lib/utils/compose.coffee
dfunckt: I think it's good enough for now and there's also the option to inline everything. I'd save improvements for subsequent PRs.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,732 @@
...
+
+		@_tty.clearLine()
+		@_tty.write(@_prefix)
+		if end and @_cancelled
+			@_tty.writeLine('Build cancelled')
+		else if end
+			serviceCount = @_services.length
+			serviceStr = if serviceCount is 1 then '1 service' else "#{serviceCount} services"
+			durationStr = humanize.relativeTime(@_startTime // 1000).replace(' ago', '')
+			@_tty.writeLine("Built #{serviceStr} in #{durationStr}")
```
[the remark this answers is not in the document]
dfunckt: Good point. In fact I struggled to provide "smart" durations but turns out they're not so smart. Will change to print a proper duration formatted as MM:SS.

dfunckt:
@pimterry regarding the app name error -- the reason it works with the old command is probably because it targets version 3 (i.e. not multicontainer) of the API. I'm not aware of any constraint to app names, will check and amend the PR appropriately. Good find.

Regarding the output, I've no idea why there's so many blank lines. In reality, both commands try hard to adjust to window changes or different sizes, because all this replacing of lines in the output would not work otherwise. I probably missed something, I'll investigate.

lib/utils/docker.coffee
dfunckt: I know, and I did leave a comment just above this function explaining that it's so we keep backwards compatibility with actions that use the promisified functions. New code I added only uses dockerode's methods, I just can't change and test every single use of docker outside the build/deploy commands this PR is for. The migration should be handled in separate PR and tested properly.

lib/utils/docker.coffee
dfunckt: It was like this before and didn't think changing it as it felt natural to me given the object is treated as a dictionary here. Have no strong opinion though, either way is fine by me, will change.

lib/actions/build.coffee
```diff
@@ -56,9 +89,45 @@ module.exports =
...
+				throw new Error('You must specify either an application or an arch/deviceType pair to build for')
+
+			if arch? and deviceType?
+				[ undefined, arch, deviceType ]
+			else
+				helpers.getArchAndDeviceType(application)
+				.then (app) ->
+					[ application, app.arch, app.device_type ]
+
+		.then ([ appName, arch, deviceType ]) ->
```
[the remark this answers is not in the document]
dfunckt: To me it wouldn't. I'd agree with you if this was the return value of the function, I don't really see a reason to bother naming each argument when the packing is done just one line above.

lib/actions/deploy.coffee
```diff
@@ -1,229 +1,183 @@
...
+		logger
+		composeOpts.projectPath
+		composeOpts.projectName
+		opts.app.app_name
+		opts.image)
+	.then (project) ->
+		# find which services use images that already exist locally
+		Promise.map project.descriptors, (d) ->
+			# unconditionally build (or pull) if explicitly requested
+			return d if opts.shouldPerformBuild
```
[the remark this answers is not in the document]
dfunckt: No, why should it? It's an object as defined in https://github.com/resin-io-modules/resin-compose-parse/blob/master/src/types.ts#L196-L199

lib/actions/deploy.coffee
dfunckt: It could, but I'd have to import lodash and I think there's no need for that (not that it would save anything, because it's already imported elsewhere, but "less imports > more imports" IMO :)

lib/utils/compose.coffee
dfunckt: The issue here is that these properties (being relations to resources, not "primitive" properties) are problematic when sent to the API (it fails in an awful way that leaves you scratching your head) so I'd rather make this clear.

lib/utils/compose.coffee
dfunckt: Yep :)

lib/actions/deploy.coffee
```diff
@@ -1,229 +1,183 @@
...
+
+		Promise.try ->
+			{ appName, image } = params
+
+			# look into "resin build" options if appName isn't given
+			appName = options.application if not appName?
+			delete options.application
+
+			if not appName?
+				throw new Error('Need an application to deploy to!')
```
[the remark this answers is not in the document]
dfunckt: Cool, it was just carried over from the old code.

lib/actions/build.coffee
```diff
@@ -37,7 +70,7 @@ module.exports =
 			$ resin build --docker '/var/run/docker.sock'
 			$ resin build --dockerHost my.docker.host --dockerPort 2376 --ca ca.pem --key key.pem --cert cert.pem
 	'''
-	options: dockerUtils.appendOptions [
+	options: removeSourceOption(dockerUtils.appendOptions compose.appendOptions [
```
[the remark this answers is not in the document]
dfunckt: Conceptually they belong together (I mean if other parts of code use compose.coffee) and it'd make sense for them to appear next to each other in the help string. Changed however, as per your suggestion.

lib/utils/compose.coffee
dfunckt: Right, so the reason is `logger.coffee` magically prefixes the log line with `[Build]` etc., but it doesn't work well with a spinner we're also showing because the whole line is replaced on each update.

dfunckt:
@pimterry I can't reproduce the uppercase app name issue -- both commands below work fine:

- `resin deploy AmpeLi --emulated --build`
- `resin deploy AmpeLi --emulated --build --projectName AmpeLi `

The error looks like it's from Docker trying either to do something locally or while interacting with the registry and receiving that error.

Can you think of any details that might help me reproduce it? Can you reproduce it yourself again?

lib/utils/compose.coffee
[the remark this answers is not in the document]
CameronDiver: The code can be stolen pretty easily actually, and that would be super nice (for example when resin-cli starts builds on the cloud builder), but I'm also happy with it waiting for a future PR.

lib/utils/compose.coffee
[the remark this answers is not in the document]
CameronDiver: That was mainly due to the implementation having two seperate endpoints, but I think with the setup we're doing now, it makes sense to push it as a single request. Also it's worth noting you'll have to make sure that the payload doesn't exceed 512Kb, otherwise the API will reject it.

dfunckt:
I've opened a new PR targeting the multicontainer branch here: https://github.com/resin-io/resin-cli/pull/782

There are no code differences between this PR and the other, for now. All comments left in this PR have been addressed (either with changes or replies) except the app name and narrow window issues @pimterry reports. Let's follow up on that PR.

PR #769, opened 2018-01-30
Require multicontainer SDK

zvin:
* require('resin-sdk') => multicontainer SDK
 * require('resin-sdk-preconfigured') => 6.15.0 SDK
 * all 'resin-sdk' requires replaced with 'resin-sdk-preconfigured'
 * resin-sdk-preconfigured TS typings are copy pasted from the current resin-sdk master

The idea is to progressively replace all 'resin-sdk-preconfigured'
requires with 'resin-sdk' (multicontainer sdk) and eventually remove
resin-sdk-preconfigured from package.json.

Change-Type: patch


lib/utils/config.ts
```diff
@@ -14,11 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import Promise = require('bluebird');
-import ResinSdk = require('resin-sdk');
+import ResinSdk = require('resin-sdk-preconfigured');
 import _ = require('lodash');
 import deviceConfig = require('resin-device-config');
 
-const resin = ResinSdk.fromSharedOptions();
+const resin = ResinSdk;
```
dfunckt: This is redundant -- why not remove this line as you did on the file above? Same for the other files.

package.json
```diff
@@ -131,8 +131,8 @@
     "resin-doodles": "0.0.1",
     "resin-image-fs": "^2.3.0",
     "resin-image-manager": "^5.0.0",
-    "resin-preload": "^5.0.0",
-    "resin-sdk": "^7.0.0",
+    "resin-preload": "beta",
+    "resin-sdk": "beta",
```
pimterry: We might reuse the `beta` tag in future for different SDK betas, so we shouldn't use that as our only specifier (otherwise, somebody installing this version of the CLI in the future could get a beta version of SDK 10, or anything).

`^9.0.0-beta` should work for the SDK (that'll match any SDK v9 releases, including all prereleases), and you can probably come up with something similar for resin-preload.

typings/resin-sdk-preconfigured.d.ts
```diff
@@ -1,5 +1,830 @@
...
+			id: number;
+			message: string | null;
+			project_type: string;
+			push_timestamp: string | null;
+			start_timestamp: string;
+			status: BuildStatus;
+			update_timestamp: string | null;
+
+			belongs_to__user: NavigationResource<User>;
+			belongs_to__application: NavigationResource<Application>;
```
pimterry: These types aren't actually correct... Looks like they're from SDK v7, but `preconfigured` is SDK v6.

Otoh, I don't care that much, since we should be removing the preconfigured SDK from everything that uses TS anyway. As a quick improvement though, can you just remove all the relationship properties (everything with `__`) completely? Those definitely won't work in SDK v6, and are the largest change, so it should be an easy way to catch anything that's _definitely_ going to break.

lib/utils/patterns.ts
```diff
@@ -18,12 +18,12 @@ import _ = require('lodash');
 import Promise = require('bluebird');
 import form = require('resin-cli-form');
 import visuals = require('resin-cli-visuals');
-import ResinSdk = require('resin-sdk');
+import ResinSdk = require('resin-sdk-preconfigured');
```
pimterry: Here too - we only ever look at device & app names & device types, which will still work totally fine with MC.

lib/utils/config.ts
zvin: I reverted this file back to using resin-sdk

lib/actions/settings.ts
```diff
@@ -27,7 +27,7 @@ Examples:
 	$ resin settings\
 `,
 	async action(_params, _options, done) {
-		const resin = (await import('resin-sdk')).fromSharedOptions();
+		const resin = await import('resin-sdk-preconfigured');
```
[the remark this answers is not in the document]
zvin: reverted

lib/events.ts
```diff
@@ -4,10 +4,9 @@ import _ = require('lodash');
 import Mixpanel = require('mixpanel');
 import Raven = require('raven');
 import Promise = require('bluebird');
-import ResinSdk = require('resin-sdk');
+import resin = require('resin-sdk-preconfigured');
```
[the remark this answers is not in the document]
zvin: reverted

lib/utils/config.ts
```diff
@@ -14,11 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import Promise = require('bluebird');
-import ResinSdk = require('resin-sdk');
+import ResinSdk = require('resin-sdk-preconfigured');
```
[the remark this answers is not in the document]
zvin: reverted

lib/utils/patterns.ts
zvin: I'm not sure of that, device.name was renamed to device.device_name for example.

package.json
zvin: updated to use sdk 9.0.0-beta and preload 6.0.0-beta

typings/resin-sdk-preconfigured.d.ts
zvin: relationship properties removed.
As you said this file is only here the time we remove resin-sdk-preconfigured.

lib/utils/patterns.ts
pimterry: Ah, good spot! In that case I expect you can probably just follow the TS errors and rename the field in a couple of places below to sort it, but otherwise I'm not too bothered by moving to preconfigured here if that's tricky for some reason.

We should probably quickly manually test a couple of the commands that use this either way, since it's currently v7, so MC or preconfigured would both require a major version bump in one direction or another.

pimterry:
I've added the no-checks label by the way - don't forget to use that and merge manually for all these PRs, or this'll get shipped into production when you merge :boom:.

lib/utils/patterns.ts
```diff
@@ -18,12 +18,12 @@ import _ = require('lodash');
...
 import chalk from 'chalk';
 import validation = require('./validation');
 import messages = require('./messages');
 
-const resin = ResinSdk.fromSharedOptions();
+const resin = ResinSdk;
```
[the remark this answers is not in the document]
zvin: This one sticks to resin-sdk-preconfigured for now as it has functions using device.name.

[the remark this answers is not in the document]
pimterry: Yes, I'm fine with that, but even so, this line now doesn't do anything, it just renames a variable. Neater to `import resin = require('resin-sdk-preconfigured')`, and remove this line entirely.

PR #770, opened 2018-01-30
Update the keys action to use the multicontainer SDK

zvin:
Change-Type: patch


zvin:
@pimterry I just wanted to get rid of resin-cli-preconfigured asap.

PR #772, opened 2018-01-30
Multicontainer preload: Update resin-preload to 6.0.0-beta4

zvin:
Change-Type: minor


doc/cli.markdown
```diff
@@ -1137,7 +1136,7 @@ id of the application to preload
 
 #### --commit, -c &#60;hash&#62;
 
-a specific application commit to preload, use "latest" to specify the latest commit
+a specific application release to preload, use "latest" to specify the latest release
```
pimterry: This now reads a little confusingly - they are still specifying a commit, it's just that internally we turn that into a specific release.

I think this change is fine if you prefix the whole line with "the commit hash for ..."

zvin: updated

PR #773, opened 2018-01-31
Reconfix integration.

abrodersen:
Initial work to use reconfix for the `local configure IMAGE` command. Currently uses a single schema to read and write configuration, but this will be expanded after some tests can be created.


abrodersen:
@pimterry When you have a chance, could you take a look at this PR?

extras/schemas/resin-os.json
```diff
@@ -0,0 +1,185 @@
...
+                                            },
+                                            "ipv6": {
+                                                "addr-gen-mode": "stable-privacy",
+                                                "method": "auto"
+                                            }
+                                        }
+                                    },
+                                    {
+                                        "value": false,
+                                        "template": {}
```
jviotti: It'd be nice to test this on an actual resin device. I want to make sure NetworkManager will not get confused by an empty file.

gulpfile.coffee
```diff
@@ -25,6 +27,10 @@ gulp.task 'coffee', ->
 		.pipe(coffee(bare: true, header: true))
 		.pipe(gulp.dest(OPTIONS.directories.build))
 
+gulp.task 'schemas', ->
+	gulp.src(OPTIONS.files.schemas, { cwd: 'extras' })
```
jviotti: Does this mean we're just copying the schemas to another directory? If so, could we put them on the final directory from the start?

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
 	}
 ]
 
+joinPath = (path) -> '/' + path.join('/')
```
jviotti: Keep in mind that Windows paths don't start with a slash.

jviotti: Ah, is this Reconfix specific? If so, can we make that clearer, or make the leading slash optional?

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
...
+	imagefs = require('resin-image-fs')
+
+	reconfix = new Reconfix({
+		read: (partition, path) ->
+			imagefs.readFile
+				image: target
+				partition: partition
+				path: joinPath(path)
+			.catch (err) ->
+				if err?.code == 'NOENT'
```
jviotti: Shouldn't this be `ENOENT`?

jviotti: Why is it necessary to check if `err` exists? It should always be present on a catch block, right?

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
...
+			.catch (err) -> {}
+		write: (partition, path, data) ->
+			imagefs.writeFile
+				image: target,
+				partition: partition,
+				path: joinPath(path)
+			, data.toString('utf-8')
+	})
+	readFileAsync(path.join(__dirname, '../../schemas/resin-os.json'))
+	.then (content) -> JSON.parse(content)
```
jviotti: I think you can just do `.then(JSON.parse)`

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
...
+	.then (json) ->
```
jviotti: Not critical, but I guess you could write this as:

```
.then(reconfix.loadSchema)
.return(reconfix)
```

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
...
+writeConfiguration = (config, target, type) ->
+	load(target, type)
+	.then (reconfix) -> reconfix.writeValues(config)
+
+
 getConfiguration = (data) ->
-	_ = require('lodash')
 	inquirer = require('inquirer')
 
+	console.log('read configuration: ' + JSON.stringify(data, '  '))
```
jviotti: I guess this can be removed?

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
...
+
 	# `persistentLogging` can be `undefined`, so we want
 	# to make sure that case defaults to `false`
 	data = _.assign data,
 		persistentLogging: data.persistentLogging or false
 
 	inquirer.prompt(inquirerOptions(data))
+	.then (answers) ->
+		keys = _.filter(_.keys(answers), (key) -> key != 'advancedSettings')
```
jviotti: I think this could be simplified to `_.keys(_.omit(answers, [ 'advancedSettings' ]))`

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
...
+		return _.pick(answers, keys)
```
jviotti: Hm, wait, so you're basically returning the original object, just without the `advancedSettings` property, right? If so `_.omit(answers, [ 'advancedSettings' ])` should do the trick

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
 	}
 ]
 
+joinPath = (path) -> '/' + path.join('/')
+
+load = (target, type) ->
+	{ Reconfix } = require('reconfix-preview')
```
jviotti: Lets push this package as `reconfix`.

jviotti: Is there any reason why the `reconfix-preview` package exports a `Reconfix` property with the actual class? Can we export the class directly?

lib/actions/local/configure.coffee
```diff
@@ -206,22 +143,20 @@ module.exports =
...
-				return if not isMounted
-				umountAsync(params.target)
-		.then (configurationSchema) ->
+		isMountedAsync(params.target).then (isMounted) ->
+			return if not isMounted
+			umountAsync(params.target)
+		.then ->
 			denymount params.target, (cb) ->
-				reconfix.readConfiguration(configurationSchema, params.target)
+				readConfiguration(params.target, '')
```
jviotti: What does an empty type mean here?

lib/actions/local/configure.coffee
```diff
@@ -206,22 +143,20 @@ module.exports =
...
 				.then(getConfiguration)
 				.then (answers) ->
-					if not answers.hostname
-						removeHostname(configurationSchema)
-					reconfix.writeConfiguration(configurationSchema, answers, params.target)
+					console.log('answers: ' + JSON.stringify(answers, '  '))
```
jviotti: I guess this can be removed

lib/actions/local/configure.coffee
```diff
@@ -206,22 +143,20 @@ module.exports =
...
+					return answers
+				.then (answers) ->
+					writeConfiguration(answers, params.target, '')
```
jviotti: Same here. Why do we use an empty path?

lib/utils/helpers.coffee
```diff
@@ -0,0 +1,134 @@
+###
```
jviotti: Looks like this file is not being used anywhere.

pimterry:
Sorry @abrodersen I know this is waiting on me. It is in my list, just busy with lots of multicontainer stuff at the mo! I'll definitely get to it this week, hopefully tomorrow.

extras/schemas/resin-os.json
```diff
@@ -0,0 +1,185 @@
...
+                    },
+                    "deviceType": {
+                        "type": ["string"],
+                        "mapping": ["/deviceType"]
+                    },
+                    "userId": {
+                        "type": ["string"],
+                        "mapping": ["/userId"]
+                    },
+                    "appUpdagePollInterval": {
```
lekkas: It would be nice to use the "description" attribute to add more context and additional information (e.g. here we could add more info about the value that's expected here (is it seconds, milliseconds?)

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
 	}
 ]
 
+joinPath = (path) -> '/' + path.join('/')
+
+load = (target, type) ->
```
lekkas: Can we rename this function to something a bit more descriptive (e.g. `loadReconfix`, or something else)

lib/actions/local/configure.coffee
lekkas: Related, where would be the best place to look for Reconfix documentation?

gulpfile.coffee
abrodersen: The files are copied to the `build` directory, which is not in source control.

lib/actions/local/configure.coffee
abrodersen: These paths are used by `resin-image-fs`, which means they are always UNIX style.

lib/actions/local/configure.coffee
abrodersen: `NOENT` is what the error type that `resin-image-fs` emits. It would be best to standardize (or document) the error interface of `resin-image-fs`.

lib/actions/local/configure.coffee
```diff
@@ -95,100 +67,65 @@ inquirerOptions = (data) -> [
...
+			.then (content) -> Buffer.from(content, 'utf-8')
+			.catch (err) -> {}
+		write: (partition, path, data) ->
+			imagefs.writeFile
+				image: target,
+				partition: partition,
+				path: joinPath(path)
+			, data.toString('utf-8')
+	})
+	readFileAsync(path.join(__dirname, '../../schemas/resin-os.json'))
```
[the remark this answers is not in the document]
abrodersen: I didn't know this! Still, having `load` return an asynchronous `Promise` makes calling it cleaner.

lib/actions/local/configure.coffee
abrodersen: I think that looks better.

lib/actions/local/configure.coffee
pimterry: In that case, it'd be good to add a comment clarifying this here

lib/actions/local/configure.coffee
abrodersen: Yes, this is what I was looking for.

lib/actions/local/configure.coffee
abrodersen: Yes, we can. When I first wrote the export, I thought the module had to be a map.

lib/utils/helpers.coffee
```diff
@@ -0,0 +1,134 @@
...
+	resin = require('resin-sdk-preconfigured')
+
+	# Attempt to read manifest from the first
+	# partition, but fallback to the API if
+	# we encounter any errors along the way.
+	imagefs.read
+		image: image
+		partition: 1
+		path: '/device-type.json'
+	.then(rindle.extractAsync)
```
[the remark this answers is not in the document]
abrodersen: @pimterry There is a `readFile` method that simply returns a `Promise` with a string value containing the contents. However, when I try to call this method, I get TypeScript errors saying it does not exist. Where are the TypeScript definitions for `resin-image-fs`?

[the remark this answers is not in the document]
pimterry: That's because resin-image-fs itself doesn't have any types, it's just got the adhoc ones that are in this codebase, which only covers the stuff we're already using, and doesn't know anything about different versions of resin-image-fs. The file you're looking for is here: https://github.com/resin-io/resin-cli/blob/master/typings/resin-image-fs.d.ts

Don't know how familiar with TS you are, let me know you need a hand.

extras/schemas/resin-os.json
abrodersen: In the current implementation, there is no description field. The schema format is going to change significantly in the next version.

pimterry: There's a typo in the schema here: `appUpdagePollInterval`

.travis.yml
```diff
@@ -6,6 +6,8 @@ node_js:
   - "6"
 before_install:
 - npm -g install npm@4
+- curl https://sh.rustup.rs -sSf | sh -s -- -y
+- source $HOME/.cargo/env
```
pimterry: This is going to be a problem. End users `npm install` the CLI on their own machines, often Windows machines, and they're not going to have Rust & Cargo installed. This (and the travis output) suggests that's not going to work, is that right?

I think we did discuss this briefly before, sorry I didn't call it out properly on the previous review. There's two options I think: hold this PR until we have CLI bundling completed (so users don't ever run `npm install` themselves), or set up node pre-gyp to precompile and publish reconfix as a binaries that users can use immediately.

The 2nd is probably more practical, if you've got time to set it up, and we'll want it if we want to use reconfix in any other node projects anyway, so we don't have to add a rust build env to every other project. We already do this for other native CLI packages like node-ext2fs. https://github.com/resin-io/node-ext2fs/pull/11 + https://github.com/resin-io/node-ext2fs/pull/13 are a full setup example I think, which should take you most of the way.

To get this working, you need:
* Versionbot set up (we should do this regardless)
* A CI build for each target platform/node version (see node-ext2fs)
* Enough Node configuration to build the project (we should have this already)
* A bit of npm boilerplate for node-pre-gyp & node-pre-gyp-release (see node-ext2fs)

The end goal is that on every tagged commit to master (from versionbot), a binary for every target platform & node version is built, and published to github releases. On every install of the package (including from CLI users), we look in github releases for a prebuilt binary, and use that immediately if there's a compatible one available (or try to build from scratch if not).

This is a non-trivial chunk of work, might be worth more discussion, feel free to ping me. @jviotti might have opinions on this too.

jviotti: Doing prebuilds is definitely an option (we usually favor `prebuild` rather than `node-pre-gyp` given the latter does some weird dynamic stuff that doesn't play nice with PKG) but proper PKG integration is something that would need to do anyways, right? How far are we with that? I remember seeing that you got this working at this point.

@abrodersen I can help you out setting this up for the Node.js bindings, but we should also do the PKG thing asap here.

lib/actions/local/configure.coffee
```diff
@@ -95,8 +66,48 @@ inquirerOptions = (data) -> [
...
+		read: (partition, path) ->
+			imagefs.readFile
+				image: target
+				partition: partition
+				path: joinPath(path)
+			.catch (err) ->
+				if err.code == 'NOENT'
+					return Buffer.alloc(0)
+				else
+					throw err
```
jviotti: Why do we throw an error here if we're going to swallow it two lines later?

lib/actions/local/configure.coffee
```diff
@@ -95,8 +66,48 @@ inquirerOptions = (data) -> [
...
+			.then (content) -> Buffer.from(content, 'utf-8')
+			.catch (err) -> {}
```
jviotti: I don't think we should be swallowing all errors here.

lib/actions/local/configure.coffee
```diff
@@ -95,8 +66,48 @@ inquirerOptions = (data) -> [
...
+				image: target,
+				partition: partition,
+				path: joinPath(path)
+			, data.toString('utf-8')
+	})
+	schema = require('../../../extras/schemas/resin-os.json')
+	reconfix.loadSchema(schema)
+	Promise.resolve(reconfix)
+
+readConfiguration = (target) ->
```
jviotti: These 2 functions are only used in one place as far as I can see, so maybe worth inlining them?

lib/actions/local/configure.coffee
```diff
@@ -95,8 +66,48 @@ inquirerOptions = (data) -> [
 	}
 ]
 
+# Since these paths are used by `resin-image-fs`, we can safely
+# assume they are always UNIX style.
+joinPath = (path) -> '/' + path.join('/')
+
+loadReconfix = (target) ->
+	{ Reconfix } = require('reconfix-preview')
```
jviotti: Did we update the npm module to just `reconfix`?

package.json
```diff
@@ -129,10 +131,10 @@
     "resin-device-init": "^4.0.0",
     "resin-docker-build": "^0.4.0",
     "resin-doodles": "0.0.1",
-    "resin-image-fs": "^2.3.0",
+    "resin-image-fs": "^4.1.1",
     "resin-image-manager": "^5.0.0",
     "resin-preload": "^5.0.0",
-    "resin-sdk": "^7.0.0",
+    "resin-sdk": "^7.2.4",
```
jviotti: Was this update necessary for this PR? Can it be done on a separate PR?

lib/actions/local/configure.coffee
```diff
@@ -95,8 +66,48 @@ inquirerOptions = (data) -> [
 	}
 ]
 
+# Since these paths are used by `resin-image-fs`, we can safely
+# assume they are always UNIX style.
+joinPath = (path) -> '/' + path.join('/')
```
jviotti: Manually joining strings look fragile. Node.js provides `path.join()` in the `path` built-in module (`require('path')`). If we always want to use UNIX-style, then you can use `path.posix.join()`.

jviotti: Maybe something like this?

```js
path = require('path')
joinPath = (fragments) -> path.posix.join(path.posix.sep, ...fragments)
```

lib/actions/local/configure.coffee
jviotti: Can we export `Reconfix` directly, rather than exporting an object containing a `Reconfix` property? I should ideally be able to do `Reconfix = require('reconfix')`

package.json
```diff
@@ -13,7 +13,8 @@
     "bin/",
     "build/",
     "doc/",
-    "lib/"
+    "lib/",
+    "extras/"
```
jviotti: Where does this directory come from?

.travis.yml
pimterry: We have bundled packages published with all releases, but they don't work perfectly I think, and they're not well tested. It's in the readme, but marked as experimental. Needs finishing, but I've only really got time for quick & essential bug fixes on the CLI right now.

shaunmulligan:
@jviotti and @zrzka I assume this PR can be closed as we will integrate the new reconfix later?

zrzka:
@shaunmulligan yes, it's obsolote.

PR #782, opened 2018-02-06
Multicontainer: Add docker-compose-aware builds and deployments

dfunckt:
(Initial review is at https://github.com/resin-io/resin-cli/pull/763 -- this PR targets the multicontainer branch)

This is part of the multi-container project, so it must not be merged before the backend has been deployed.

The PR looks massive, but in reality it's not: only the latest commit is relevant and for the most part is new code.

Legacy behaviour is mostly retained. The most notable change in behaviour is that invoking `resin deploy` without options is now allowed (see help string how it behaves).

In the latest commit there are also the following notable changes:

- Deploy/Build are promoted to primary commands
- Extracts QEMU-related code to a new file
- Adds a utility file to retrieve the CLI version and its parts
- Adds a helper that can be used to manipulate display on capable clients
- Declares several new dependencies. Most are already indirectly installed via some dependency

Other changes in separate commits include (see commit messages for full details):

- Update bundle-resolve and docker-build to latest
- Update docker-toolbelt
- Update YAML parser
- Fix invoking undefined method

Change-Type: minor


lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+			# multibuild parses the composition internally so any tags we've
+			# set before are lost; re-assign them here
+			task.tag ?= [ projectName, task.serviceName ].join('_')
+			if d.image.context?
+				d.image.tag = task.tag
+
+			# configure build opts appropriately
+			task.dockerOpts ?= {}
+			_.merge(task.dockerOpts, buildOpts)
+			_.merge(task.dockerOpts, { t: task.tag })
```
zvin: NIT: this could be
```coffeescript
_.merge(task.dockerOpts, buildOpts, t: task.tag)
```

lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+
+	releaseMod.create
+		client: client
+		user: userId
+		application: appId
+		composition: composition
+		source: "cli/#{versions.cli.major}"
+		commit: crypto.pseudoRandomBytes(16).toString('hex').toLowerCase()
+	.then ({ release, serviceImages }) ->
+		[ 'created_at', 'belongs_to__application', 'is_created_by__user', '__metadata' ].forEach (attr) ->
```
zvin: NIT: you could use `_.omit` here

lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+authorizePush = (tokenAuthEndpoint, auth, registry, images) ->
+	_ = require('lodash')
+	querystring = require('querystring')
+	request = require('request')
+	getAsync = Promise.promisify(request.get)
+
+	if not _.isArray(images)
+		images = [ images ]
+
+	query = querystring.stringify
```
zvin: NIT: you can do this using resin sdk's request like here https://github.com/resin-io/resin-preload/blob/154-multicontainer-preload/lib/preload.js#L453-L464

and avoid requiring request and querystring

lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+			release.status = 'failed'
+		.finally ->
+			runloop = runSpinner(tty, spinner, "#{prefix}Saving release...")
+			release.end_timestamp = new Date()
+			releaseMod.updateRelease(client, release.id, release)
+			.finally(runloop.end)
+
+# utilities
+
+renderProgressBar = (percentage, stepCount) ->
```
zvin: Don't we already have something like this in resin-cli-visuals?

lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+pullProgressAdapter = (outStream) ->
+	return ({ status, id, percentage, error, errorDetail }) ->
+		if status?
+			status = status.replace(/^Status: /, '')
+		if id?
+			status = "#{id}: #{status}"
+		if percentage is 100
+			percentage = undefined
+		outStream.write({
+			status,
```
zvin: NIT: unnecessary comas, parenthesis and braces

lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+			progress: percentage,
+			error: errorDetail?.message ? error
+		})
+
+createSpinner = ->
```
zvin: We already have spinners in resin-cli-visuals. What can't they be used?

lib/utils/qemu.coffee
```diff
@@ -0,0 +1,88 @@
...
+	path.relative(context, binPath)
+
+exports.copyQemu = (context) ->
+	path = require('path')
+	fs = require('mz/fs')
+	# Create a hidden directory in the build context, containing qemu
+	binDir = path.join(context, '.resin')
+	binPath = path.join(binDir, QEMU_BIN_NAME)
+
+	fs.access(binDir)
```
zvin: Edge case: is binDir is a file the following will fail

lib/utils/qemu.coffee
```diff
@@ -0,0 +1,88 @@
...
+hasQemu = ->
+	fs = require('mz/fs')
+
+	getQemuPath()
+	.then(fs.stat)
+	.return(true)
+	.catchReturn(false)
+
+getQemuPath = ->
+	resin = require('resin-sdk-preconfigured')
```
zvin: We should update this to use `resin-sdk` instead.

lib/utils/qemu.coffee
```diff
@@ -0,0 +1,88 @@
...
+
+	getQemuPath()
+	.then (qemuPath) ->
+		new Promise (resolve, reject) ->
+			installStream = fs.createWriteStream(qemuPath)
+			qemuUrl = "https://github.com/resin-io/qemu/releases/download/#{QEMU_VERSION}/#{QEMU_BIN_NAME}.gz"
+			request(qemuUrl)
+			.pipe(zlib.createGunzip())
+			.pipe(installStream)
+			.on('error', reject)
```
zvin: NIT: this won't catch request or gunzip errors.

pimterry:
To continue a couple of points from my previous review:

* I can no no longer reproduce the uppercase app name bug, oddly. No idea what's going on there, but I guess we can ignore it until we hit it again.

* Progress bar/spinner: fair enough if they don't do what you want right now, but medium-term we're aiming to make the CLI much more componentized and tie into Rendition, so we're going to want to remove these in the fairly short-term future. Can you file bugs against https://github.com/resin-io-modules/resin-cli-visuals for the specific changes you'd need to make them usable here? I can go through and patch those up and dedupe that later from there, once the rest of MC is out the way.

* Output formatting: fine if we don't match the builder for now, although the final table of services output particularly would be really nice if we can.

I'll do a proper rereview of the latest changes now, but with a quick skim it's looking fairly good already.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+	versions = require('./versions')
+
+	client = releaseMod.createClient({ apiEndpoint, auth })
+
+	releaseMod.create
+		client: client
+		user: userId
+		application: appId
+		composition: composition
+		source: "cli/#{versions.cli.major}"
```
pimterry: This is useful data, but it doesn't match the current format (which uses `local` for all CLI builds), which means right now on MC we don't show local builds as local in the UI.

What's the reason we want the CLI major version in here? We could put in a patch there to fix this fairly easily, but on closer thought I am a little surprised by this.

pimterry:
Btw the build is failing here because of an odd `ent` typings issue (will be fixed by #786), but after that it'll also still fail because you haven't updated the docs here. You need to run `npm run build` to do a full build, and commit the changes.

lib/utils/qemu.coffee
dfunckt: Yeah, this file was merely code extracted from docker.coffee that I felt deserved its own place. In any case, I've updated it to try to create the folder unconditionally.

lib/utils/compose.coffee
dfunckt: Yeah, Tim also mentioned this here: https://github.com/resin-io/resin-cli/pull/763#discussion_r165714077

lib/utils/compose.coffee
dfunckt: Not sure TBH, I felt like it'd be nice to have a way to identify the origin with more granularity but I can't justify it anymore. I've changed it to "local" and removed the "version.coffee" file as well.

CameronDiver: I really like the idea of being able to specify the version of component which created the images/releases. Maybe this should be another field in the DB? (Of course this can be added after the initial release, as long as we make it optional.)

lib/utils/compose.coffee
```diff
@@ -0,0 +1,746 @@
+Promise = require('bluebird')
+path = require('path')
+
+exports.appendProjectOptions = appendProjectOptions = (opts) ->
+	opts.concat [
+		{
+			signature: 'projectName'
```
CameronDiver: This has always been `source` or `context` in the past, context would be my preference, but project feels weird.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,746 @@
...
+						rawStream = stream.pipe(buildThroughStream)
+					else
+						rawStream = stream
+					# `stream` sends out raw strings in contrast to `task.progressHook`
+					# where we're given objects. capture these strings as they come
+					# before we parse them.
+					rawStream
+						.pipe(captureStream)
+						.pipe(buildProgressAdapter(inlineLogs))
+						.pipe(task.logStream)
```
CameronDiver: You should be consistent on whether you indent method chaining or not, for example on https://github.com/resin-io/resin-cli/pull/782/files#diff-d863ebb012fba8e05d698376062052b1R115 you don't

lib/utils/compose.coffee
```diff
@@ -0,0 +1,746 @@
...
+		if id?
+			status = "#{id}: #{status}"
+		if percentage is 100
+			percentage = undefined
+		outStream.write
+			status: status
+			progress: percentage
+			error: errorDetail?.message ? error
+
+createSpinner = ->
```
CameronDiver: I know it's been said but I want to +1 using the CLI visuals we've already implemented.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,746 @@
...
+
+		@_services.forEach (service, index) =>
+			str = _.padEnd(@_prefix + chalk.bold(service), @_prefixWidth)
+			str += serviceToStrMap[service]
+			if @_maxLineWidth?
+				str = truncate(str, @_maxLineWidth)
+			@_lineWidths[index] = strlen(str)
+
+			@_tty.clearLine()
+			@_tty.writeLine(str)
```
CameronDiver: It would be much nicer to get the table here that the builder uses, I'll even seperate it out into a seperate module for you if you want.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,746 @@
...
+		streams = _(services).map (service) ->
+			stream = through.obj (event, _enc, cb) ->
+				eventHandler(service, event)
+				cb()
+			stream.pipe(outStream)
+			[ service, stream ]
+		.fromPairs()
+		.value()
+
+		offset = 10 # account for escape sequences inserted for colouring
```
CameronDiver: I'd much prefer this kind of stuff was added into the logging utils, rather than being done by hand at individual call sites.

lib/utils/compose.coffee
dfunckt: I don't understand? The flag is named `--projectName` and I don't see how it relates to source, context or project.

lib/utils/compose.coffee
dfunckt: This has nothing to do with the logger. This is to correctly right-pad service names so their logs left-align.

lib/utils/compose.coffee
dfunckt: Yes, it's planned for the near future, after resin-cli-visuals is patched be less opinionated and usable in a context such as here where there's multiline output being constantly updated.

lib/utils/compose.coffee
dfunckt: That would be nice, but it'll have to come in a subsequent PR. The command already presents a summary of useful info and it's good enough for now.

lib/utils/compose.coffee
CameronDiver: I know, but that feels like it should be part of the logger to me, rather than repeated several times where you're doing custom output

lib/utils/compose.coffee
CameronDiver: My mistake, I interpreted this incorrectly - LGTM.

lib/actions/deploy.coffee
```diff
@@ -1,229 +1,194 @@
...
+						opts.app.id
+						userId
+						"Bearer #{auth}"
+						apiEndpoint
+						!opts.shouldUploadLogs
+					)
+			)
+	.then (release) ->
+		logger.logSuccess('Deploy succeeded!')
+		logger.logSuccess("Release ID: #{release.commit}")
```
pimterry: This prints `Release ID: undefined` for me.

pimterry: We shouldn't call this the release id, because that exists, and it's a different thing (which we'd like to avoid users using, but they'll end up seeing in places anyway, especially if they're using the API).

At the same time, we obviously don't want 'commit' in here since that'd make no sense. `Release hash:`?

dfunckt: Yeah, that's a tricky one. I think "hash" suggests you'll get repeatedly the same value for the same input and I'd rather avoid this. I'm inclined towards "release identifier" because that's it is really, a way to refer to the release. What do you think?

pimterry: >  I think "hash" suggests you'll get repeatedly the same value for the same input 

True, that is awkward. I still think it's going to be confusing that release identifier != `release.id` though. `Release key`? `Release name`? Naming is hard.

Could just copy the UI, which puts this value in a column called 'Release', and print `Release: ` all by itself.

dfunckt: Neither "key" and "name" feel ideal to me. Let's follow the UI then :)

dfunckt: Ah, good catch, I forgot to update a `tap` to `then`.

lib/utils/compose.coffee
```diff
@@ -0,0 +1,747 @@
...
+	reporters = progress.aggregateProgress(images.length, renderer)
+
+	Promise.using tty.cursorHidden(), ->
+		Promise.map images, ({ serviceImage, localImage, props, logs }, index) ->
+			Promise.join(
+				localImage.inspect().get('Size')
+				progress.push(localImage.name, reporters[index], opts).finally(renderer.end)
+				(size, digest) ->
+					serviceImage.image_size = size
+					serviceImage.content_hash = digest
```
pimterry: If I repeatedly rebuild the same thing, with the entire build coming from the cache, the build is super fast but I still end up doing the entire image upload from scratch. I've checked the API though, and I do end up with the same digest for the image, so in principle we should be able to detect that.

This doesn't need to be in this PR, but as a future improvement is there an easy way we could avoid the image upload entirely in those cases?

dfunckt: Yes, that's true unfortunately. I did try keeping images around between consecutive deploys, but that didn't help. The issue is either that docker doesn't keep around the metadata needed to determine that layers have been pushed already (which means simply pulling the previous images will fix the issue) or there's something with the way the API authenticates access to the registry. Didn't get round to test either theory, but I suspect it should be straight-forward.

PR #787, opened 2018-02-08
Remove support for CLI plugins

pimterry:
This has actually been PRd and approved before (#728), see that PR for details. That went onto a `v7` branch as a staging branch for the next breaking release. This PR puts it onto the multicontainer branch instead - mostly just as an FYI that I'd like to include this in the MC release.


lib/app.coffee
```diff
@@ -74,7 +74,6 @@ resin = require('resin-sdk-preconfigured')
 actions = require('./actions')
 errors = require('./errors')
 events = require('./events')
-plugins = require('./utils/plugins')
```
dfunckt: Shouldn't "utils/plugins.coffee" be removed as well?

pimterry: Oooh, well done - that was in the previous PR, but we've since autoconverted it to TS, so when I rebased that change got lost completely. Good spot, fixed :+1:

PR #792, opened 2018-02-15
Multicontainer

zvin:


pimterry:
Just so we're clear, **this depends on the MC API & builder reaching production, and must not be merged before that happens**. When we merge this PR, it will ship the multicontainer CLI to production.

zvin:
Please wait before merging this, I'd like to release resin-preload as non beta first.

pimterry:
Yep, just approving so it's ready to go - we're not going to merge until the API is in production and we're happy everything is stable.

Issue #765, opened 2018-01-26
CLI should build on npm 5


lekkas:
We require npm4 for the build process and it is stated nowhere in the README. It's implied with the `require-npm4-to-publish` script/dependency but it's easy to miss because the prepublish script is not invoked when building the cli locally. 

see: https://www.flowdock.com/app/rulemotion/public-s-community/threads/PtlnyJqSUPmiRZdNC8oUKUNFOFU

- **resin-cli version: all**
- **Operating system and architecture:**

lekkas:
Marked as bug because this is slowing down contributors who want to perform trivial changes like cli doc updates.

pimterry:
@lekkas can you confirm that the current master with a fresh npm install now builds for you locally? If so, I think we can close this.

lekkas:
On it

lekkas:
@pimterry `npm run build` is now working flawlessly on my machine

PR #793, opened 2018-02-16
Tweak TS & add missing deps that may cause build failures in some envs

pimterry:
This connects to #765. I'm not really sure what's causing the underlying issue, so this is some best efforts guesswork Really the CLI should build & install totally fine with npm5 already, and it seems to do locally on my machine.

There are a few small things related to the errors that appeared in https://www.flowdock.com/app/rulemotion/public-s-community/threads/PtlnyJqSUPmiRZdNC8oUKUNFOFU that weren't quite right though, so I've tightened things up a little to make sure everything's working properly there. A few of those errors look like lodash usage getting `any` typed - the lodash types dependency is already present in principle, but only due to it being required elsewhere in the tree, so I've added it here explicitly, in case that was causing trouble.

@lekkas if you could check out this branch locally and confirm it builds for you that'd be very helpful.

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #765


lekkas:
@pimterry sorry I saw this a bit too late

PR #812, opened 2018-03-15
Fix getting window size when there’s no TTY attached

dfunckt:
Change-Type: patch


lib/utils/tty.coffee
```diff
@@ -13,7 +13,8 @@ module.exports = (stream = process.stdout) ->
 	updateWindowSize()
 
 	currentWindowSize = ->
-		# always return a copy
+		# always return a copy.
+		# width/height can be null if no TTY.
```
pimterry: `undefined`, not `null`

lib/utils/compose.coffee
```diff
@@ -619,7 +619,7 @@ class BuildProgressUI
 
 	_clear: ->
 		@_tty.deleteToEnd()
-		@_maxLineWidth = @_tty.currentWindowSize().width
+		@_maxLineWidth = @_tty.currentWindowSize()?.width
```
pimterry: I don't think you need this `currentWindowSize` is always defined still, it's just the width that might not be, and that's fine.

dfunckt: Oh right, that's was the whole point, to call `currentWindowSize()` and be sure there's always width/height props, even if null/undefined.

PR #813, opened 2018-03-15
Properly generate consistent working anchors for both our md output & resin docs

### LINKED ISSUE #634: Docs links for commands with special characters don't work
Lots of the contents links in https://github.com/resin-io/resin-cli/blob/master/doc/cli.markdown don't work.

For example, `resin ssh [deviceIp]` links to `#local-ssh-deviceip-` but the real anchor is `#local-ssh-deviceip` (no following `-`).

Special characters cause problems too `deploy <appName> [image]` links to `#deploy-60-appname-62-image-`, instead of `#deploy-appname-image`

`local scan` works fine (`#local-scan`)

### LINKED ISSUE #809: Links in docs table of contents no longer match header anchor links
The anchor links that are generated for the headers in the CLI docs no longer container 60 and 62 for the brackets, both in GitHub and on our docs page. It seems like this is a behavior change in the markdown renderers, but it means the link generation for the table of contents needs to be updated.

pimterry:
I still have no idea why this changed, and the logic here looks like it's always been very odd. Right now, we generate markdown, which gets transformed into HTML in one way for github's UI, and in a slightly different way when we build the real docs for the docs site, and we just hope that that'll work out.

That's been a bit annoying for a while, since it means even before the current issues, the contents links only worked on the docs site, not the docs on github (#634).

Rather than try to work out what's changed here, I've made the logic more sensible:
* We now generate much simpler anchor slugs for commands (`#device-rm-uuid`)
* These match what github would generate for the given command's titles anyway, which is nice, and they're way more readable (vs `device-rm-60-uuid-62-` before).
* We then explicitly add an actual anchor tag to every command with the exactly correct slug, rather than depending on how anchor tags are generated when the HTML is generated elsewhere.

Seems to work for me in the github UI for this branch (https://github.com/resin-io/resin-cli/blob/fix-help-links/doc/cli.markdown), and I'm pretty sure it'll work in the docs site too, but I can't confirm 100% until we build it there (and I don't know how that works)

Fixes #809 
Fixes #634 
Change-Type: patch


pimterry:
@zwalchuk This is merged now - can you rebuild the CLI docs page, and check whether the links now work correctly?

zwalchuk:
@pimterry unfortunately, not working still. It takes the entire html tag and tries to put it in the link (for example, https://docs.resin.io/reference/cli/#-a-name-envs-a-envs). I'll think just updating the TOC links and not having the anchors should work in GitHub and on the docs site. I'll update with another PR. It's rendering differently if I run locally too, so I'm not sure what exactly is handling HTML generation that has changed.

zwalchuk:
@pimterry actually, is it easy for you to roll back this change, and then reapply just the changes in the lines ~40 of markdown.ts, and not the ones ~ line 9 (and include the regenerated cli.markdown)? I think that will do it.

Issue #694, opened 2017-10-20
Getting help on nested commands


Page-:
Running `resin help device` gives
```
Usage: device <uuid>

Use this command to show information about a single device.

Examples:

        $ resin device 7cf02a6
```
and I was expecting it to also list the nested commands like `resin device rename`, `resin device init`, etc.  I eventually realised I could get those via `resin help --verbose` but I was expecting it to show up under `resin help device` as they're nested in that namespace

pdcastro:
CC: @srlowe because of recent discussions on `balena help` conversion to oclif

srlowe:
This is mostly solved. 
e.g. `balena help os` now outputs:
```
os commands include:
  os build-config <image> <device-type>  build an OS config and save it to a JSON file
  os configure <image>                   configure a previously downloaded balenaOS image
  os download <type>                     download an unconfigured OS image
  os initialize <image>                  initialize an os image for a device
  os versions <type>                     show available balenaOS versions for the given device type

Run balena help -v for a list of all available commands,
 or balena help <command> for detailed help on a specific command.
```

However, the edge case is when the topic is also a command (e.g. `balena device` is a topic and a command, whereas `balena os` is only a topic.  Will try to fix the edge cases.

PR #816, opened 2018-03-18
Adding list of nested commands on 'help' command

rssilva:
Solves #694 

screen shot 2018-03-18 at 8 08 34 pm

- Exposing capitano instance on global through _capitano variable
- Adding method to list all nested commands if available on the help
command

### DIFF
--- a/lib/actions/help.coffee
+++ b/lib/actions/help.coffee
@@ -91,7 +91,23 @@ command = (params, options, done) ->
 			console.log('\nOptions:\n')
 			print(parse(command.options))
 
+		if params.command
+			listNested(params.command, command.signature)
+
 		return done()
+
+listNested = (commandName) ->
+	list = []
+
+	for command in global._capitano.state.commands
+		parameters = command?.signature?.parameters
+		parameterName = parameters?[0].parameter
+
+		if parameterName is commandName and not (/[^a-z]{1,}/g).test(parameters?[1])
+			list.push("#{parameters?[1]?.parameter}")
+
+	if list.length
+		console.log("also available for '#{commandName}':", list.join(', '))
 
 exports.help =
 	signature: 'help [command...]'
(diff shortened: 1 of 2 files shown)


lib/actions/help.coffee
```diff
@@ -91,8 +91,24 @@ command = (params, options, done) ->
...
 		return done()
 
+listNested = (commandName) ->
+	list = []
+
+	for command in global._capitano.state.commands
+		parameters = command?.signature?.parameters
+		parameterName = parameters?[0].parameter
+
+		if parameterName is commandName and not (/[^a-z]{1,}/g).test(parameters?[1])
```
pimterry: What cases is the `not (/[^a-z]{1,}/g).test(parameters?[1])` test catching?

lib/actions/help.coffee
```diff
@@ -91,8 +91,24 @@ command = (params, options, done) ->
...
+			list.push("#{parameters?[1]?.parameter}")
+
+	if list.length
+		console.log("also available for '#{commandName}':", list.join(', '))
```
pimterry: Tiny point, but could `also` have a capital `A` here?

lib/actions/help.coffee
```diff
@@ -91,8 +91,24 @@ command = (params, options, done) ->
...
 
+listNested = (commandName) ->
+	list = []
+
+	for command in global._capitano.state.commands
+		parameters = command?.signature?.parameters
+		parameterName = parameters?[0].parameter
+
+		if parameterName is commandName and not (/[^a-z]{1,}/g).test(parameters?[1])
+			list.push("#{parameters?[1]?.parameter}")
```
pimterry: It looks like this could be undefined, though I'm not 100% sure when... I think in that case we should probably avoid pushing anything at all.

PR #818, opened 2018-03-19
Restore legacy deployment method

dfunckt:
This mostly reverts the removal of the legacy deploy code that pushed image tars via the builder. It’s needed for users to avoid having to switch between CLI versions in order to push to legacy apps as well.

It also adds a warning if deploying a multicontainer project to an incompatible app.

Note: this also pins resin-sdk to 9.0.0-beta14 as I couldn’t get it to install otherwise — npm would always install 9.0.0-beta9 instead (npm is probably naive enough to just do a lexicographical sort, beta9 sorts after beta14).

Change-Type: minor


lib/utils/deploy-legacy.coffee
```diff
@@ -0,0 +1,145 @@
...
+showPushProgress = (message) ->
+	visuals = require('resin-cli-visuals')
+	progressBar = new visuals.Progress(message)
+	progressBar.update({ percentage: 0 })
+	return progressBar
+
+forwardStatusMessageFromRemote = (logger, msg) ->
+	msg.split(/\r\n|\n/).forEach (line) ->
+		if /^Warning: This endpoint is deprecated/.test(line)
+			return
```
pimterry: I don't love this... If we're effectively undeprecating the legacy endpoint, shouldn't we just undeprecate it in the builder itself?

It's going to be ridiculous & annoying when really really want to deprecate it in future because it's actually going to be 100% removed, and we have to make sure our message doesn't include the above or some CLIs won't show it.

lib/utils/deploy-legacy.coffee
```diff
@@ -0,0 +1,145 @@
...
+				return
+
+			if obj.type?
+				switch obj.type
+					when 'error' then reject(new Error("Remote error: #{obj.error}"))
+					when 'success' then resolve(obj)
+					when 'status' then forwardStatusMessageFromRemote(logger, obj.message)
+					else reject(new Error("Received unexpected reply from remote: #{data}"))
+			else
+				reject(new Error("Received unexpected reply from remote: #{data}"))
```
pimterry: Nit: Does this need to be a special case? I would've thought `obj.type = undefined` would've hit the else case above anyway.

dfunckt: Ah, you're right, I merely reverted the removal of the old code making only the needed changes in order to work. Will change this.

dfunckt: In fact I don't think it should even raise -- what's wrong with receiving messages it doesn't know? -- the exception prevents extending the builder push endpoint. But on the other CLI v6 does that as well, so I'm not sure if it makes sense stopping raising.

pimterry: Yeah, good point. I think sticking with reject is fine for now, since it's consistent with before and it doesn't limit us any more than we already are, but yeah if we were building this again in future we might well want to do differently.

lib/actions/build.coffee
```diff
@@ -21,6 +20,12 @@ buildProject = (docker, logger, composeOpts, opts) ->
 		composeOpts.projectName
 	)
 	.then (project) ->
+		if project.descriptors.length > 1 and not opts.app.application_type?[0]?.supports_multicontainer
+			logger.logWarn(
+				'Target application does not support multiple containers.\n' +
+				'Continuing with build, but you will not be able to deploy.'
```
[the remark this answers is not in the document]
dfunckt: I don't see the reason to stop. I expect users building code for deployment are doing `resin deploy --build`, in which case the command stops before long. I think letting users play with `resin build` to be fair -- building an image may be for other purposes apart from deployment, eg. preparing their code for MC.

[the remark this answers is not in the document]
pimterry: Yeah, ok. I don't feel strongly, let's stick with this.

PR #822, opened 2018-03-20
Don't include anchor tags in docs, as it breaks the docs site

pimterry:
More changes after https://github.com/resin-io/resin-cli/pull/813. I've stripped out the actual anchor tags, and updated this to match what the docs seem to want more closely, can you try this as-is @zwalchuk, and see if it works?

It now doesn't work in the docs on github (they don't end up with trailing `-`s for headings like `app <name>`, whereas the docs generation does seem to), but that's not a big problem as long as it works for the docs site. 

Also @zwalchuk can you let me what you do to regenerate the CLI docs pages? This will do as a short-term fix, if it works, but I'd like to dig into it and really work out what's going on here, so we can fix this properly.

### DIFF
--- a/automation/capitanodoc/markdown.ts
+++ b/automation/capitanodoc/markdown.ts
@@ -4,9 +4,7 @@ import * as utils from './utils';
 import { Document, Category, Command } from './doc-types';
 
 export function renderCommand(command: Command) {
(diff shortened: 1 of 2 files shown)


pimterry:
Superseded by #823

PR #823, opened 2018-03-20
Remove explicit anchor links in CLI docs

zwalchuk:
Our docs markdown renderer doesn't process explicit anchor tags, as it generates its own. The script that generates the markdown has been updated to not include these tags and to properly build the TOC links.

Change-type: patch


zwalchuk:
@pimterry The reason for the fork and not adding you as a reviewer is because I don't have write access to the CLI repo, which has unfortunately caused this same issue with other repos. I'll request that access and then do this properly.

zwalchuk:
Handled in #823

Issue #851, opened 2018-04-09
Error: Cannot find module 'tar-stream'


pimterry:
https://sentry.io/resinio/cli/issues/518175738/

```
Error: Cannot find module 'tar-stream'
  File "module.js", line 469, in Function.Module._resolveFilename
  File "module.js", line 417, in Function.Module._load
  File "module.js", line 497, in Module.require
  File "internal/module.js", line 20, in require
  File "/opt/local/lib/node_modules/resin-cli/build/utils/compose.js", line 115, in tarDirectory
    tar = require('tar-stream');
...
(12 additional frame(s) were not displayed)

Error: Cannot find module 'tar-stream'
```

Looks like we're depending on this transitively, but not explicitly, which might be messing something up somewhere.

pimterry:
Fixed in v7.2.4

mohit-aicg:
Still getting this error in angular 12

Issue #845, opened 2018-04-04
CLI does not install on arm64


joshughes:
- **resin-cli version:**
Latest
- **Operating system and architecture:**
Debian Stretch node 8

```
FROM arm64v8/node:8-stretch

RUN apt-get update && apt-get install -y build-essential libconfig-dev
RUN npm install resin-cli -g --production --unsafe-perm

CMD ['/usr/local/bin/resin']
```

I have an arm64 build environment to create arm64 build containers. I would like to have the resin-cli tool available there but it fails to install. 

```

> ext2fs@0.1.1 install /usr/local/lib/node_modules/resin-cli/node_modules/ext2fs
> node-pre-gyp install --fallback-to-build

[91mnode-pre-gyp [0m[91mERR![0m[91m [0m[91mTried to download(404): https://github.com/resin-io/node-ext2fs/releases/download/v0.1.1/bindings-v0.1.1-node-v57-linux-arm64.tar.gz[0m[91m 
[0m[91mnode-pre-gyp [0m[91mERR! [0m[91mPre-built binaries not found for ext2fs@0.1.1 and node@8.11.1 (node-v57 ABI, glibc) (falling back to source compile with node-gyp) 
[0mmake: Entering directory '/usr/local/lib/node_modules/resin-cli/node_modules/ext2fs/build'
  CC(target) Release/obj.target/ext2fs/deps/e2fsprogs/lib/ext2fs/alloc.o
[91m../deps/e2fsprogs/lib/ext2fs/alloc.c:12:20: fatal error: config.h: No such file or directory
 #include "config.h"
                    ^
compilation terminated.
[0mext2fs.target.mk:177: recipe for target 'Release/obj.target/ext2fs/deps/e2fsprogs/lib/ext2fs/alloc.o' failed
make: Leaving directory '/usr/local/lib/node_modules/resin-cli/node_modules/ext2fs/build'
```

pimterry:
That's true, but I'm afraid I don't have an nice fix for you.

Right now, we don't have an easy way to publish prebuilt binaries for ARM64 (which is what you'd need for ext2fs, and likely a few other submodules), so you'll need to build them yourself. Our [travis config](https://github.com/resin-io/node-ext2fs/blob/master/.travis.yml) might be useful to see the environment we're building in for x64. If you can build all of those yourself, the CLI should start working nicely (and I'd love to hear updates about progress there, in case any other users need this).

[...]

PR #852, opened 2018-04-09
Explicitly depend on tar-stream

pimterry:
I'm not 100% sure how, but I think it's very likely that this is causing #851, and it's definitely wrong regardless. Right now it normally works because various subdeps do depend on tar-stream, but if we're using it ([we are](https://github.com/resin-io/resin-cli/blob/ba82b1fa2719c5db4c836557299ad7171f3e18a9/lib/utils/compose.coffee#L106)) we should depend on it explicitly.

Change-Type: patch


thgreasi:
Also, just spotted that `removedrive` is an optional dependency, should we better handle [the case that it's missing](https://github.com/resin-io/resin-cli/blob/d6e0616c58ce090fe0ff8d001e86c17f540479d6/lib/actions/local/flash.coffee#L116)?

pimterry:
I don't think we need to check for removedrive. It's never been an issue (according to sentry). The only time when it won't be present is if you're not on Windows (in which case, we don't use it), or if the package somehow gets broken and becomes uninstallable, and I'm fine with getting a bunch of sentry errors if that happens. There's not much we can do about it in either case anyway.

pimterry:
(We are talking about adding a bunch of new optional dependencies as a solution for #845 though, and we should definitely check for those, since they'll be unpredictably available or not, depending on your install process)

PR #853, opened 2018-04-09
Correct documented node version requirement to 6+

pimterry:
Super tiny fix - @alisondavis17 just spotted that the docs on this aren't up to date.

Change-Type: patch


pimterry:
Merging this one manually - seems like an unnecessary release for the tiniest possible docs change.

PR #854, opened 2018-04-10
Add 'api-key generate' command

### LINKED ISSUE #191: Expose a command to generate API Keys
This command will allow the user to generate ~~application/device specific API keys will can later be used in their applications.~~ per-user api keys (for now, we can extend to different scopes later, as that becomes possible)

pimterry:
Fixes #191
Change-Type: minor


lib/actions/api-key.ts
```diff
@@ -0,0 +1,27 @@
...
+Examples:
+
+    $ resin api-key generate "Jenkins Key"
+`,
+	async action(params, _options, done) {
+		const resin = (await import('resin-sdk')).fromSharedOptions();
+
+		resin.models.apiKey.create(params.name).then(key => {
+			console.log(`Registered api key '${params.name}':\n${key}`);
+			done();
```
thgreasi: Can we move `done()` into a `finally()` so that it get's called even if there is a network error?

lib/actions/api-key.ts
```diff
@@ -0,0 +1,27 @@
...
+
+Examples:
+
+    $ resin api-key generate "Jenkins Key"
+`,
+	async action(params, _options, done) {
+		const resin = (await import('resin-sdk')).fromSharedOptions();
+
+		resin.models.apiKey.create(params.name).then(key => {
+			console.log(`Registered api key '${params.name}':\n${key}`);
```
thgreasi: We could also suggest to take a note of this since the user will not be able to find this string at a later point of time.

lib/actions/api-key.ts
```diff
@@ -0,0 +1,36 @@
+import { CommandDefinition } from 'capitano';
+import { stripIndent } from 'common-tags';
+
+export const generate: CommandDefinition<{
+	name: string;
+}> = {
+	signature: 'api-key generate <name>',
+	description: 'Generate a new API key with the given name',
+	help: stripIndent`
```
thgreasi: Woohoo! Now all that is left is to find a transformer to run this on compile.

PR #861, opened 2018-04-16
727 sentry improvements

### LINKED ISSUE #727: Update Sentry
Lots of our Sentry errors ([e.g.](https://sentry.io/resinio/cli/issues/416554065/)) have a footer that looks like:

screenshot from 2017-12-12 12-31-07

I'm reasonably sure this means we need to update our Sentry dependency, it'd be nice to quickly do that when we have a sec.

pimterry:
Fixes #727 

As part of trying to debug some recent Sentry errors, I spotted that they have an auto breadcrumb feature, which automatically detects context (HTTP requests & console.* calls the CLI has made during this command) and includes it in any reported errors. This'd be super useful, so I've enabled it. Required a small update to pkg config to get it working for the standalone build too, since they use some dynamic requires internally.

I've also updated Sentry to the latest version, which has [a bunch of improvements](https://github.com/getsentry/raven-node/blob/master/History.md), and moved lots of user input errors to `expectedError` (which ensures they don't get reported to Sentry).


lib/actions/help.coffee
```diff
@@ -78,7 +79,7 @@ command = (params, options, done) ->
 		return done(error) if error?
 
 		if not command? or command.isWildcard()
-			return done(new Error("Command not found: #{params.command}"))
+			expectedError("Command not found: #{params.command}")
```
thgreasi: I was tricked by this for a bit, not knowing that `expectedError` calls `process.exit` :)

pimterry: @thgreasi Hmm, interesting, and a good point. We should definitely have a verb in here. How do you feel about `failWithExpectedError`?

pimterry: I've gone for `exitWithExpectedError` - I'd rather this be a little verbose but clearer I think.

Thanks for bringing this up! Good improvement :+1:

PR #862, opened 2018-04-16
Update Dockerode to fix local push issue in standalone builds

### LINKED ISSUE #824: Local push fails with no such file or directory, lstat 'C:\<path>\snapshot'
```
C:\BlahDocker\node-red>resin local push -f 192.
168.127.4 --source .
Reporting discovered devices
? select a device resin.local (192.168.127.4)
Building..
Stopping and removing any previous 'local-app' container
Uploading build context & starting build...
{ Error: ENOENT: no such file or directory, lstat 'C:\BlahDocker\node-red\snapshot'
at Error (native)
errno: -4058,
code: 'ENOENT',
syscall: 'lstat',
path: 'C:\blah\snapshot
' }
```

Definitely fails for pushes to iot2000 2.7.6+rev1 (resinos) from Windows, probably other cases too. Looks likely (based on `snapshot`) that this is a standalone build bug only, tbc.

Reported by @ruudboon

pimterry:
Dockerode patch (https://github.com/apocas/dockerode/pull/445) has now been merged and released, which should let us fix #824 once and for all.

Change-Type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "docker-progress": "^3.0.1",
     "docker-qemu-transpose": "^0.5.1",
     "docker-toolbelt": "^3.1.0",
-    "dockerode": "^2.5.0",
+    "dockerode": "^2.5.5",
     "dockerode-options": "^0.2.1",
     "drivelist": "^5.0.22",
     "ejs": "^2.5.7",


thgreasi:
I totally agree bumping package version to explicitelly tell that we need that version to fix all known issues, but do we technically need this?
Since the CLI isn't using lock files of any kind, isn't the `^` going to pick the new version on install? (but obviously still have to do a fresh one before building the standalone version)

pimterry:
> Since the CLI isn't using lock files of any kind, isn't the ^ going to pick the new version on install? (but obviously still have to do a fresh one before building the standalone version)

Yep, 100%. But this is a bug that specifically affects only the standalone build, so we need a fresh release to fix it, and merging this PR will trigger that nicely. Imo it's nicer to have a separate PR doing that (and thereby have a specific version & release) that fixes this, rather than having the fix bundled into whatever the next PR is that gets merged.

It's also useful to be explicit about the versions we need anyway. If we had a subdependency with a conflicting version for some reason then we could get an older version of Dockerode. With this change, we guarantee we're definitely always going to get a fixed version in future.

Issue #843, opened 2018-04-04
The CLI should be able to trigger builds on resin's build server infrastructure


CameronDiver:
Essentially, using the CLI, a user should be able to forgo the entire git push workflow, aiding development - especially for windows users.

HQ: https://github.com/resin-io/hq/issues/265

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_b33f)

lvaylet:
I got redirected to this issue following [a post on the forum](https://forums.resin.io/t/multicontainer-and-multiple-repos/2924) and I am having a hard time figuring out the relationship. Could you elaborate a bit on how this feature is supposed to help? Are you planning to give us full control over the command that is executed on resin's build servers?

CameronDiver:
Hey @lvaylet I'd expect the way that this solves the issue that you're having would be that you can explicitly clone your submodules before a `resin push`, e.g.

```
git clone myrepo --recurse-submodules
cd myrepo
resin push myapp
`

This would ensure that all of your submodules are deployed correctly.

lvaylet:
That would be amazing indeed. Looking forward to it!

cdgraff:
+1

CameronDiver:
Forgot to close the issue, implemented with https://github.com/resin-io/resin-cli/pull/868.

PR #868, opened 2018-04-25
Add push command which starts a build on remote resin servers

CameronDiver:
This PR adds a module which coordinates building projects using the resin remote servers. Currently this is meant to have feature parity with `git push resin master`.

Also added a function in patterns.ts which provides an interface to ask the user a question, and give several options. 

Change-type: minor
Connects-to: #843
---- Autogenerated Waffleboard Connection: Connects to #843


lib/utils/remote-build.ts
```diff
@@ -0,0 +1,213 @@
...
+		url: await getBuilderEndpoint(build.baseUrl, build.owner, build.app),
+		auth: {
+			bearer: build.auth,
+		},
+	});
+	if (DEBUG_MODE) {
+		console.log('[debug] Opened builder connection. Streaming data...');
+	}
+
+	tarStream.pipe(post);
```
pimterry: Right now, there's no output until the context directory is tarred and fully uploaded. That's not great if you have a large local directory. There was quite a bit of confusion from local push where people didn't understand what it was doing and why it took so long, because they accidentally had enormous local folders they were pushing over the network each time.

Can we add some initial output and a nice spinner while we do the upload?

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,213 @@
...
+				} else {
+					console.log(message);
+				}
+			}
+		}
+	};
+}
+
+async function cancelBuildIfNecessary(build: RemoteBuild): Promise<void> {
+	if (build.running && build.releaseId != null) {
```
pimterry: Why do we need the separate `running` flag? Can't we just check if we've got a release id?

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,213 @@
...
+
+import * as JSONStream from 'JSONStream';
+import * as request from 'request';
+import { ResinSDK } from 'resin-sdk';
+import * as Stream from 'stream';
+
+import { tarDirectory } from './compose';
+
+const DEBUG_MODE = process.env.DEBUG === '1';
+const SUPPRESS_OUTPUT = process.env.SUPPRESS_BUILDER === '1';
```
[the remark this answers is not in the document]
CameronDiver: It was mainly for debugging the builder output, and now you mention it I'm not 100% on it being needed elsewhere - I'll just remove it.

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,213 @@
...
+		// This is necessary because the `exit-hook` module is used by several
+		// dependencies, and will exit without calling the following handler.
+		// Once https://github.com/resin-io/resin-cli/issues/867 has been solved,
+		// we are free to (and definitely should) remove the below line
+		process.removeAllListeners('SIGINT');
+		process.on('SIGINT', () => {
+			process.stderr.write('Received SIGINT, cleaning up. Please wait.\n');
+			cancelBuildIfNecessary(build).then(() => {
+				stream.end();
+				process.exit(0);
```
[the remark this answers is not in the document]
CameronDiver: Given the workflow, I don't think it matters too much, e.g. if we're here, we know we can exit straight away. That being said, `130` is more correct, so I'll change it.

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,213 @@
...
+		// ctrl+c
+
+		// This is necessary because the `exit-hook` module is used by several
+		// dependencies, and will exit without calling the following handler.
+		// Once https://github.com/resin-io/resin-cli/issues/867 has been solved,
+		// we are free to (and definitely should) remove the below line
+		process.removeAllListeners('SIGINT');
+		process.on('SIGINT', () => {
+			process.stderr.write('Received SIGINT, cleaning up. Please wait.\n');
+			cancelBuildIfNecessary(build).then(() => {
```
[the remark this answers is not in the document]
CameronDiver: Due to the ELB, this doesn't always work. We have extra code in the git server to handle this too, and tbh I think redundancy on something like this is completely fine. If a user ctrl+c's a `resin push`, the last thing we want is for the build to still get deployed.

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,213 @@
...
+				status: 'cancelled',
+				end_timestamp: Date.now(),
+			},
+		});
+	}
+}
+
+async function getRequestStream(build: RemoteBuild): Promise<Stream.Duplex> {
+	// Tar the directory so that we can send it to the builder
+	const tarStream = await tarDirectory(build.source);
```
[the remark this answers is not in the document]
CameronDiver: Ah, good to know, thanks!

lib/actions/push.ts
```diff
@@ -0,0 +1,128 @@
...
+*/
+
+import { CommandDefinition } from 'capitano';
+import { ResinSDK } from 'resin-sdk';
+
+const PUSH_HELP_TEXT =
+`This command can be used to start a build on the remote resin.io cloud
+builders. The given source directory will be added to a tarfile, which
+will be sent to the resin.io builder, and the build will proceed. This
+can be used as a drop-in replacement for git push to deploy.`;
```
pimterry: Have you pulled this out just because indentation with backticks is messy? There's a nicer solution to that: `stripIndent` (e.g. https://github.com/resin-io/resin-cli/blob/b7214a306c6b0a5e45387f93c9142219ff6d7b60/lib/actions/api-key.ts)

```ts

...

export const push: CommandDefinition<...> = {
	help: stripIndent`
		This command can be used to start a build on the remote resin.io cloud
		builders. The given source directory will be added to a tarfile, which
		will be sent to the resin.io builder, and the build will proceed. This
		can be used as a drop-in replacement for git push to deploy.
	`,
```

lib/actions/push.ts
```diff
@@ -0,0 +1,128 @@
...
+limitations under the License.
+*/
+
+import { CommandDefinition } from 'capitano';
+import { ResinSDK } from 'resin-sdk';
+
+const PUSH_HELP_TEXT =
+`This command can be used to start a build on the remote resin.io cloud
+builders. The given source directory will be added to a tarfile, which
+will be sent to the resin.io builder, and the build will proceed. This
```
pimterry: I'd skip the `tarfile` bit - I think a few users won't be familiar with tar files (especially on Windows etc), and it's not strictly necessary. Just `will be send to the resin.io builder in a compressed form` is fine imo.

lib/actions/push.ts
```diff
@@ -0,0 +1,128 @@
...
+			app_name: appName,
+		},
+		$select: ['id'],
+	});
+
+	if (applications == null || applications.length === 0) {
+		exitWithExpectedError(
+			`No applications found with name: ${appName}.\n\n` +
+			'This could mean that the application does not exist, or you do \n' +
+			'not have the permissions required to access it.'
```
pimterry: You probably want `stripIndent` here and a multiline string here too.

lib/actions/push.ts
pimterry: We also normally include an example command in the help text. This is what becomes the CLI docs on https://docs.resin.io/reference/cli/, and they get nicely rendered there (and they're useful here too).

https://github.com/resin-io/resin-cli/blob/b7214a306c6b0a5e45387f93c9142219ff6d7b60/lib/actions/device.coffee#L133-L143 is a nice example imo.

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,219 @@
...
+				stream.end();
+				process.exit(130);
+			});
+		});
+
+		const dataHandler = getBuilderMessageHandler(build);
+		stream.on('data', (data: BuilderMessage) => {
+			if (!spinnerStopped) {
+				uploadSpinner.stop();
+				spinnerStopped = true;
```
pimterry: This spinner actually has its own check to make `stop()` safe: https://github.com/resin-io-modules/resin-cli-visuals/blob/edeb543a5c83cb67b684564f7835e8a69dfeaf6a/lib/widgets/spinner.coffee#L78

You could drop `spinnerStopped` entirely, and just call `uploadSpinner.stop()` every time.

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,219 @@
...
+				return;
+			}
+
+			const match = obj.value.match(CURSOR_METADATA_REGEX);
+
+			if (!match) {
+				// FIXME: Make this error nicer.
+				console.log(
+					'Warning: ignoring unknown builder command. You may experience odd build output. ' +
+						'Maybe you need to update resin-cli?',
```
pimterry: stripIndent

lib/utils/remote-build.ts
```diff
@@ -0,0 +1,219 @@
...
+		url: await getBuilderEndpoint(build.baseUrl, build.owner, build.app),
+		auth: {
+			bearer: build.auth,
+		},
+	});
+
+	const uploadSpinner = new visuals.Spinner('Uploading source package to resin cloud');
+	uploadSpinner.start();
+
+	tarStream.pipe(post);
```
pimterry: In fact, imo we should add a `data` handler here that's tied to `uploadSpinner.stop()`, and not return the spinner at all. I think that should work fine, and lets the code everywhere else ignore the spinners entirely.

lib/actions/push.ts
```diff
@@ -0,0 +1,139 @@
...
+	{
+		application: string;
+	},
+	{
+		source: string;
+	}
+> = {
+	signature: 'push <application>',
+	description: 'Start a remote build on the resin.io cloud build servers',
+	help: stripIndent`This command can be used to start a build on the remote
```
pimterry: Sorry, if you test this help text, you'll see this doesn't work properly.

`stripIndent` strips the common indent, ignoring empty lines, so you'll need to start with an empty line. If you don't, as here, then the common indent is zero (the first line has none) and text all goes wrong.

Same for the other cases here too. You want things like:

```ts
	help: stripIndent`
		This command can be used to start a build on the remote
		resin.io cloud builders. The given source directory will be sent to the
		resin.io builder, and the build will proceed. This can be used as a drop-in
		replacement for git push to deploy.
	 
		Examples:
	 
			$ resin push myApp
			$ resin push myApp --source <source directory>
			$ resin push myApp -s <source directory>
	`,
```

That should do what you expect.

PR #874, opened 2018-05-03
Handle failed requires & missing bindings

### LINKED ISSUE #872: "Could not locate the bindings file" for ext2fs
https://sentry.io/resinio/cli/issues/540067477/

```
Error: Could not locate the bindings file. Tried:
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/Debug/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/Release/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/out/Debug/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/Debug/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/out/Release/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/Release/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/default/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/compiled/8.11.1/linux/x64/bindings.node
  File "/usr/local/lib/node_modules/resin-cli/node_modules/bindings/bindings.js", line 93, in bindings
    err = new Error('Could not locate the bindings file. Tried:\n'
  File "/usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/lib/ext2fs.js", line 3, in Object.<anonymous>
    const bindings = require('bindings')('bindings');
  File "module.js", line 652, in Module._compile
  File "module.js", line 663, in Module._extensions..js
  File "module.js", line 565, in Module.load
...
(57 additional frame(s) were not displayed)

Error: Could not locate the bindings file. Tried:
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/Debug/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/Release/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/out/Debug/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/Debug/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/out/Release/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/Release/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/build/default/bindings.node
 → /usr/local/lib/node_modules/resin-cli/node_modules/resin-image-fs/node_modules/ext2fs/compiled/8.11.1/linux/x64/bindings.node
```

pimterry:
This PR does two separate but closely related things. Check the commits individually if you'd like to separate them.

First: it moves [resin-cli-errors](https://github.com/resin-io-modules/resin-cli-errors) into the CLI itself, and converts it to TS en route. This is a separate module that isn't really useful outside the CLI, doesn't need to be separate, and would be better inlined here. I'll deprecate the separate npm lib version later.

Second: it adds error handling & a nice message for require & missing binding errors. This will fix #872 (by providing a helpful message so users can do a complete uninstall & reinstall to recover for themselves), which lets us rerelease #870.

I've also included a nice message in the 2nd part for non-x64 users. Right now this shouldn't come up much, if at all, but it'll be super useful when we look at #845.

screenshot from 2018-05-03 15-51-28


lib/errors.ts
```diff
@@ -14,18 +14,98 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import errors = require('resin-cli-errors');
-import patterns = require('./utils/patterns');
+import _ = require('lodash');
```
dimitrisnl: Why not use `import { isEmpty } from 'lodash'` in order to reduce bundle size?

dimitrisnl: Just wondering since we do this in various places while using only 1-2 lodash functions.

lib/errors.ts
```diff
@@ -14,18 +14,76 @@ See the License for the specific language governing permissions and
...
+import os = require('os');
```
zvin: Why not
```typescript
import * as os from 'os';
```
?

lib/errors.ts
pimterry: Mainly because we don't do any bundling in the CLI :smiley:. The normal build is just raw JS, and although the standalone build bundles into a single executable, I believe it just works file by file. It's not clever enough for tree shaking, so requiring individual methods won't make any difference.

We have talked about using webpack here too, but it's not the top priority, so there hasn't been any movement there yet. Download size for the CLI would be nice to improve (mainly for workshops), but outside the web isn't not as important, and the standalone build provides a lot of similar benefits, and lots of others.

We could also potentially move to the tinier standalone lodash modules like https://www.npmjs.com/package/lodash.isempty, which would help overall install size, but I think they'd actually hurt startup time (I believe, not 100% sure, that the IO time reading modules from disk is the expensive bit, and 1 large file is typically cheaper than many small ones), and we've got lots of subdependencies that use lodash anyway, so it wouldn't help much.

pimterry:
Tests fail because the API is down, sigh.

lib/errors.ts
```diff
@@ -14,18 +14,76 @@ See the License for the specific language governing permissions and
...
+
+		if (message) {
+			return message;
+		}
+
+		if (!_.isEmpty(error.message)) {
+			return `${error.code}: ${error.message}`;
+		}
+
+		return;
```
zvin: This `return` isn't needed.

lib/errors.ts
```diff
@@ -14,18 +14,76 @@ See the License for the specific language governing permissions and
...
+import { stripIndent } from 'common-tags';
+
+import patterns = require('./utils/patterns');
 
 const captureException = Promise.promisify<string, Error>(
 	Raven.captureException,
 	{ context: Raven },
 );
 
+function hasCode(error: any): error is Error & { code: string } {
```
zvin: Shouldn't the return type be `boolean`?
I don't really understand this return type.

lib/errors.ts
dimitrisnl: Thanks for the explanation. I forgot that CLI doesn't use Webpack and rushed to edit my comment but you beat me to it!

lib/errors.ts
pimterry: This is a [user-defined type guard](http://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards).

It's basically a more specific type of boolean: not only does it return a boolean value, the value also gives TypeScript some information about types, which it can use infer the type of the variable in case that uses this test.

As a specific example, in this case this means that inside the `if (hasCode(error)) { ... }` block, we're allowed to look at `error.code`, because TS can infer that this method has returned true, and this type signatures says that this means the object has a `.code` property. If this just returns boolean, that won't work, because it's just an `Error`, and `.code` isn't a property that all errors have.

lib/errors.ts
pimterry: No real reason, it's just that that's what the existing imports did. I've now updated all of them.

lib/errors.ts
pimterry: True, removed.

PR #883, opened 2018-05-29
Update resin-compose-parse dependency version

flesler:
#882


thgreasi:
@flesler you should probably do a fresh `npm i` and then `npm run build` once again.
This will probably install an updated `prettier` version and update the source formatting so that `catch-uncommitted` will not find anything different when building on travis.
@pimterry Maybe we should pin prettier dependency with a `~` so that we don't have to pollute unrelated PR diffs whenever prettier changes its mind on something.

flesler:
@thgreasi thanks, I did it but still failing. I had to commit a styling change, which I'm not a fan of doing but it's not a big deal because it's a small one.

This is still failing (https://concourse.ci.resin.io/builds/19097) it doesn't seem to be due to this PR per-se. It's only the dependency change and the version is indeed published on NPM. cc @pimterry

flesler:
Seems like AppVeyor simply takes forever to build, will have to wait for it to finish then

flesler:
Tested on staging, results here: https://github.com/resin-io/resin-api/issues/1011#issuecomment-393985376

PR #887, opened 2018-05-31
Add a multicontainer caveat to the env var commands

pimterry:
Based on @martenjacobs submission to the docs: https://github.com/resin-io/docs/pull/819

Change-Type: patch


pimterry:
(Failure is from the prettier bump again - https://github.com/resin-io/resin-cli/pull/883 is going to fix that anyway, so we'll just wait & rebase)

Issue #888, opened 2018-06-04
Updating Resin CLI failed


svenschwermer:
- **resin-cli version:** latest (7.5.1)
- **Operating system and architecture:** Mac OSX amd64

I just ran `npm install resin-cli -g --production --unsafe-perm` to update my existing installation. This is the error I'm getting:

```
[...]
  LIBTOOL-STATIC Release/libext2fs.a
Traceback (most recent call last):
  File "./gyp-mac-tool", line 611, in <module>
    sys.exit(main(sys.argv[1:]))
  File "./gyp-mac-tool", line 28, in main
    exit_code = executor.Dispatch(args)
  File "./gyp-mac-tool", line 43, in Dispatch
    return getattr(self, method)(*args[1:])
  File "./gyp-mac-tool", line 246, in ExecFilterLibtool
    if not libtool_re.match(line) and not libtool_re5.match(line):
TypeError: cannot use a string pattern on a bytes-like object
make: *** [Release/libext2fs.a] Error 1
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:258:23)
gyp ERR! stack     at ChildProcess.emit (events.js:182:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:235:12)
gyp ERR! System Darwin 17.5.0
gyp ERR! command "/usr/local/Cellar/node/10.1.0/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /usr/local/lib/node_modules/resin-cli/node_modules/ext2fs
gyp ERR! node -v v10.1.0
gyp ERR! node-gyp -v v3.6.2
gyp ERR! not ok
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! ext2fs@1.0.7 install: `prebuild-install || node-gyp rebuild`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the ext2fs@1.0.7 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_c5rf)

[...]

PR #893, opened 2018-06-12
Document Python native build dependency

pimterry:
Connects to #888 (caused by using Python 3, instead of 2.7)

I'm also adding a recommendation for the standalone mode here, since that's now reasonably stable, and likely to be the easier option for most users here, rather than fighting native build environments.

Change-Type: patch


package.json
```diff
@@ -79,7 +79,7 @@
     "gulp-shell": "^0.5.2",
     "mochainon": "^2.0.0",
     "pkg": "^4.3.0-beta.1",
-    "prettier": "^1.9.2",
+    "prettier": "1.13.5",
```
thgreasi: How about a `~`?

PR #907, opened 2018-06-26
Make sure 'resin push' is included in the CLI docs

### LINKED ISSUE #906: Resin push does not appear in the docs

pimterry:
Fixes #906
Change-Type: patch


zwalchuk:
@pimterry is this feature stable enough to include in our getting started guides? It would be awesome to have an alternative for people who don't want to bother with SSH keys

pimterry:
@zwalchuk We should check with @shaunmulligan, but personally I think yes. I've been using it in lots of the workshops recently, and it's got some small rough edges but overall it makes life much easier, especially for users without SSH keys.

shaunmulligan:
@zwalchuk and @pimterry yeah I definitely think we should start pushing it, its pretty good as it is.

PR #911, opened 2018-06-28
Device api keys are no longer used in the registration process

telphan:


Page-:
The commit message is a little off btw, since you can still pass a deviceApiKey to use whilst registering, just the SDK doesn't support it

pimterry:
@telphan is there a reason this hasn't been merged?

Issue #915, opened 2018-07-04
Malformed Token on os configure 


joshughes:
- **resin-cli version: 7.6.2**
- **Operating system and architecture: Ubuntu 16.04**

```
 DEBUG=true resin os configure jetson-v2.13.0+rev1-prod.img --app Fooo
Configuring operating system image
ResinMalformedToken: Malformed token: VALID_API_TOKEN
    at /snapshot/resin-cli/node_modules/resin-token/build/token.js:54:23
    at tryCatcher (/snapshot/resin-cli/node_modules/bluebird/js/release/util.js:16:23)
    at Promise.module.exports.Promise._settlePromiseFromHandler (/snapshot/resin-cli/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise.module.exports.Promise._settlePromise (/snapshot/resin-cli/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise.module.exports.Promise._settlePromiseCtx (/snapshot/resin-cli/node_modules/bluebird/js/release/promise.js:606:10)
    at Async._drainQueue (/snapshot/resin-cli/node_modules/bluebird/js/release/async.js:138:12)
    at Async._drainQueues (/snapshot/resin-cli/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.e.Async.drainQueues (/snapshot/resin-cli/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:672:20)
    at tryOnImmediate (timers.js:645:5)
    at processImmediate [as _immediateCallback] (timers.js:617:5)
```

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_cjgb)

pimterry:
Resolved in CLI 7.7.4

PR #923, opened 2018-07-11
Update OS & config actions to the MC SDK, and add a --version option

pimterry:
Updating this in advance of the full MC update as there's a specific user request for it: #915. Sadly that alone doesn't actually work with named api keys, as it tries to configure by generating a (deprecated) user api key by default, and named api keys don't have permissions for that.

Thus, this PR also adds a `--version` parameter, and ensures that the correct keys are used for all modern versions, which are available when using the CLI with api keys. Some discussion: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/L0654dpqwkQP_6aR1QerYRO47RO.

In future, we'll start autodetecting the OS version where possible (i.e. at least for ResinOS >= 2.9.7), and erroring if we can't autodetect and no `--version` is provided. For now though, existing behaviour remains, and `--version` is only strongly recommended.

Change-Type: patch


doc/cli.markdown
```diff
@@ -1011,6 +1014,10 @@ device uuid
 
 custom device key - note that this is only supported on ResinOS 2.0.3+
 
+#### --version &#60;version&#62;
+
+a ResinOS version
```
Page-: Should `a` be capitalised?

doc/cli.markdown
```diff
@@ -1129,22 +1136,29 @@ show advanced commands
...
+	$ resin config generate --device 7cf02a6 --version 2.12.7 --output config.json
+	$ resin config generate --app MyApp --version 2.12.7
+	$ resin config generate --app MyApp --version 2.12.7 --output config.json
+	$ resin config generate --app MyApp --version 2.12.7 --network wifi --wifiSsid mySsid --wifiKey abcdefgh --appUpdatePollInterval 1
 
 ### Options
 
+#### --version &#60;version&#62;
+
+a ResinOS version
```
Page-: Should `a` be capitalised?

doc/cli.markdown
Page-: `ResinOS` should be `resinOS`

doc/cli.markdown
Page-: `ResinOS` should be `resinOS`

lib/actions/command-options.coffee
```diff
@@ -44,6 +44,11 @@ exports.optionalDeviceApiKey =
 	parameter: 'device-api-key'
 	alias: 'k'
 
+exports.optionalOsVersion =
+	signature: 'version'
+	description: 'a ResinOS version'
```
Page-: `ResinOS` should be `resinOS`

lib/utils/config.ts
```diff
@@ -17,16 +17,17 @@ import Promise = require('bluebird');
...
 
 export function generateBaseConfig(
 	application: ResinSdk.Application,
-	options: { appUpdatePollInterval?: number },
+	options: { version?: string; appUpdatePollInterval?: number },
 ) {
 	options = _.mapValues(options, function(value, key) {
 		if (key === 'appUpdatePollInterval') {
-			return value! * 60 * 1000;
+			return <number>value * 60 * 1000;
```
Page-: I think this can use `typeof options.appUpdatePollInterval` to automatically match the type

pimterry: Sure, done (though it also needs an `!`, as `appUpdatePollInterval?: number` means the type might be undefined)

lib/actions/command-options.coffee
pimterry: True, fixed.

doc/cli.markdown
pimterry: I've left this as is - for whatever reason, all of the option descriptions everywhere start lowercase, and I'd rather leave it consistent for now.

Page-: Ok

lib/utils/config.ts
```diff
@@ -66,27 +67,30 @@ export function generateBaseConfig(
...
 export function generateDeviceConfig(
-	device: ResinSdk.Device & { application_name: string },
+	device: ResinSdk.Device & { belongs_to__application: ResinSdk.PineDeferred },
 	deviceApiKey: string | true | null,
-	options: {},
+	options: { version?: string },
 ) {
 	return resin.models.application
-		.get(device.application_name)
+		.get(device.belongs_to__application.__id)
```
thgreasi: I guess that we should eventually move those pine helpers in the SDK (and later in pine itself).

lib/utils/config.ts
```diff
@@ -17,16 +17,17 @@ import Promise = require('bluebird');
...
 
 export function generateBaseConfig(
 	application: ResinSdk.Application,
-	options: { appUpdatePollInterval?: number },
+	options: { version?: string; appUpdatePollInterval?: number },
 ) {
 	options = _.mapValues(options, function(value, key) {
 		if (key === 'appUpdatePollInterval') {
-			return value! * 60 * 1000;
+			return (<typeof options.appUpdatePollInterval>value)! * 60 * 1000;
```
thgreasi: This seems a bit over-complicated. How about `options[key]!`?

lib/utils/config.ts
```diff
@@ -66,27 +67,30 @@ export function generateBaseConfig(
...
+			return addProvisioningKey(config, application.id);
+		} else {
+			return addApplicationKey(config, application.id);
+		}
+	});
 }
 
 export function generateDeviceConfig(
-	device: ResinSdk.Device & { application_name: string },
+	device: ResinSdk.Device & { belongs_to__application: ResinSdk.PineDeferred },
```
thgreasi: We might need to change this once its consumers get rewritten to TS.
It would be great to check both `__id` & `[0].id` like what `getId` does.

lib/utils/config.ts
pimterry: Oooh, great idea, done.

lib/utils/config.ts
pimterry: Yeah, once we've got more TS in here and consistent usage of the same SDK Version everywhere, there's a lot of improvements like this (and your other comment) I'd like to do.

For now I think this is ok though - even if we need to change it later, it'll be something the type system finds and complains about for us, not a bug.

lib/utils/config.ts
```diff
@@ -17,16 +17,17 @@ import Promise = require('bluebird');
...
 
 export function generateBaseConfig(
 	application: ResinSdk.Application,
-	options: { appUpdatePollInterval?: number },
+	options: { version?: string; appUpdatePollInterval?: number },
 ) {
 	options = _.mapValues(options, function(value, key) {
 		if (key === 'appUpdatePollInterval') {
-			return value! * 60 * 1000;
+			return options[key]! * 60 * 1000;
```
Page-: I actually think this could be simplified more with:
```typescript
if (options.appUpdatePollInterval) {
	options.appUpdatePollInterval = options.appUpdatePollInterval * 60 * 1000
}
```
and just drop the `mapValues` altogether (could use `_.clone` if the cloning was a desired side-effect)

pimterry: I do indeed want the clone, rather than mutating my arguments. I think the latter is a nice way of combining them both, I've gone with that (but also, I feel like we're bike shedding a bit here, let's just ship it).

thgreasi:
@pimterry should I resume the TS 2.8.1 work on the SDK?
Conditional mapped typed would help us properly type patch & post requests.

pimterry:
@thgreasi yes, that'd be great! It'd be good to push to 2.9 really in fact, if possible. The only reason I haven't done that here is that it's not strictly necessary to fix these immediate errors, and 2.9 fails to build right now due to some of the SDK's pine types.

PR #930, opened 2018-07-20
Be explicit about how much initial history log tailing includes

pimterry:
Change-Type: patch


lib/actions/logs.coffee
```diff
@@ -51,7 +51,7 @@ module.exports =
 			console.log("#{timestamp} #{line.message}")
 
 		if options.tail
-			resin.logs.subscribe(params.uuid).then (logs) ->
+			resin.logs.subscribe(params.uuid, { count: 100 }).then (logs) ->
```
flesler: NIT: move the `100` to a constant for visibility

PR #934, opened 2018-07-31
Add an env vars example config to the local push docs

### LINKED ISSUE #933: Local push doesn't include env vars in its example resin-sync.yml

pimterry:
Fixes #933 

Following on from a support conversation asking about the format for env vars here (https://app.frontapp.com/open/cnv_xupwbv), I checked the example. Beforehand this was showing a pure resin sync example, which isn't the generally useful case (we're aiming towards deprecating it really), and doesn't show most of the features people want. I suspect we copied it from the resin sync docs originally.

This now uses a proper local push example highlighting the important bits from there, notably including env vars.

For some reason I also hit an issue where TSC has stopped building the CLI correctly, because it's compiling `node_modules/typed-error/src/typed-error.ts` in as part of the source. I'm not totally sure what's changed, but the new version of typed-error moves to proper type definitions and solves this anyway, so I've moved to that too.


lib/actions/local/push.coffee
```diff
@@ -43,12 +43,14 @@ resinPushHelp = '''
...
-		before: 'echo Hello'
-		after: 'echo Done'
-		ignore:
-			- .git
-			- node_modules/
+		local_resinos:
+			app-name: local-app
+			build-triggers:
+				- Dockerfile: abcdefabcdefabcdefabcdefabcdefabcdefabcdef
+				- package.json: abcdefabcdefabcdefabcdefabcdefabcdefabcdef
```
thgreasi: Could these `abcdefabcdefabcdefabcdefabcdefabcdefabcdef` be a bit more descriptive on what the expected value should be?

PR #943, opened 2018-08-15
Don't resolve symlinks for resin push - instead push them as-is

Page-:
This might fix resin push on linux/mac, it needs testing as the O_SYMLINK flag isn't always available

### DIFF
--- a/lib/utils/compose.coffee
+++ b/lib/utils/compose.coffee
@@ -110,7 +110,7 @@ exports.tarDirectory = tarDirectory = (dir) ->
 	streamToPromise = require('stream-to-promise')
 
 	getFiles = ->
-		streamToPromise(klaw(dir))
+		streamToPromise(klaw(dir, { preserveSymlinks: true }))
 		.filter((item) -> not item.stats.isDirectory())
 		.map((item) -> item.path)
 
@@ -118,7 +118,7 @@ exports.tarDirectory = tarDirectory = (dir) ->
 	getFiles(dir)
 	.map (file) ->
 		relPath = path.relative(path.resolve(dir), file)
-		Promise.join relPath, fs.stat(file), fs.readFile(file),
+		Promise.join relPath, fs.lstat(file), fs.readFile(file, { flag: fs.constants.O_SYMLINK | fs.constants.O_RDONLY }),
 			(filename, stats, data) ->
 				pack.entry({ name: toPosixPath(filename), size: stats.size, mode: stats.mode }, data)
 	.then ->


Page-:
@pimterry when you commit a symlink to git you commit the symlink and not the file it points to, and I would expect `resin push` to match `git push` (at least for now), which is my number 1 reason.
Beyond that there are problems with copying the resolved file rather than the symlink itself:
* if you have circular symlinks you'll never be able to complete the push
* symlinks that go outside of the build directory would end up copying files from outside the build directory which I would definitely *not* expect if I'm pushing a specific folder, and I would be very annoyed if my private data from another folder was pushed because of it
* you might want to have the symlink itself pointing to another file in the built docker container (eg some config files)
* if you have a symlink to a binary that you then copy it might fail because it expects its working directory to contain other files that were not copied

pimterry:
Ok, all fair points.

Do you know what `docker build` does here?

If that does follow links (might not) then this ties into my conversation with Cameron on #945 - do we want to follow git conventions and be consistent with out current approach, or follow docker and aim for that going forwards.

That's assuming docker does follow links. If they both don't then I'm fine with this, though we do need to make it work nicely for windows etc too.

pimterry:
I tested `docker build` - it too preserves symlinks, I'm sold in theory.

The implementation still needs work though, and then more testing. On my machine at least (Ubuntu, Node 8.11.1) `fs.constants.O_SYMLINK` is undefined, so the read here will throw. I suspect we might also need some handling for Windows to Linux link conversion.

Page-:
Yeah it seems like `O_SYMLINK` is basically just not defined anywhere so I'm not sure how to handle this.  I tested on my win10, WSL ubuntu, and VM ubuntu and all 3 were missing it, and Shaun's osx failed with an error and so was presumably missing it as well.
I do think we should definitely either use the symlink directly or fail rather than giving surprises but I'm not sure how to make that happen in nodejs right now - at the very least the terrible UX of silently doing nothing was fixed in #942

pimterry:
Ah, I just discovered fs.readlink: https://nodejs.org/api/fs.html#fs_fs_readlink_path_options_callback

In theory we should be able to use that to read & then recreate links, right?

pdcastro:
I found this old PR today. Earlier this year, I had created issue #1169 _(`balena push` should preserve symlinks),_ but some time later I closed it as I learned that `docker build` does not preserve symlinks: https://github.com/balena-io/balena-cli/issues/1169#issuecomment-512978943

Given:
* `docker build` compatibility;
* The fact that this change (this PR) would be a breaking change with the CLI's own behaviour for years (unless made optional, with a flag);
* Some users feel strongly in defence of the existing behaviour: https://github.com/balena-io/balena-cli/issues/1169#issuecomment-508287187

... I am not currently convinced that it's worth it to change it.

Issue #889, opened 2018-06-05
Resin push should ignore `.git` directories, and files pointed to by .gitignore, .dockerignore


CameronDiver:
Often these files are completely useless and add weight (and therefore time) to the source package upload. Ignoring these (maybe behind a flag in case somebody wants their git history deployed) would make the majority of repos much faster to upload and start to build.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_d8o3)

richbayliss:
+1 this makes a lot of sense to me.

pimterry:
Totally agree, it definitely should. Right now I'm not going to find the time for this any time soon, but I'd happily review a PR if somebody else wants to take a look at it :+1:.

zrzka:
Just to add some numbers with simple [rust application](https://github.com/zrzka/rust-thermometer):

```bash
 2018-07-03 17:01:55 ⌚  Zrzka in ~/Work/rust-thermometer
± |master → origin {1} U:4 ✗| → du -sh .
564M	.

 2018-07-03 17:09:15 ⌚  Zrzka in ~/Work/rust-thermometer
± |master → origin {1} U:3 ✗| → du -sh target
562M	target
```

Removing `target` (build artefacts) or move it aside for a while is not an option, because [RLS](https://github.com/rust-lang-nursery/rls) starts immediately building it again when these files are gone. As a workaround for now, one have to clone workspace, move it aside, \[local\] push, move it back and open workspace again. Or wait ... And this is quite simple app, these numbers are much bigger for other projects.

pdcastro:
Implemented in PR #950. Further work required in issue #1032.

PR #945, opened 2018-08-15
fix: Respect ignore files, and ignore git dir when tarring sources

CameronDiver:
This commit pulls in the minimatch library, and uses it to remove any
globs present in .gitignore or .dockerignore files. The ignore files
themselves are also not forwarded to the tar archive (which is used in
both resin push and resin deploy).

Change-type: major
Closes: #889


lib/utils/compose.coffee
```diff
@@ -103,24 +103,51 @@ toPosixPath = (systemPath) ->
...
-		Promise.join relPath, fs.stat(file), fs.readFile(file),
-			(filename, stats, data) ->
-				pack.entry({ name: toPosixPath(filename), size: stats.size, mode: stats.mode }, data)
+	.each (file) ->
+		# Find any files that should be ignored
+		if path.relative(path.resolve(dir), file) in ['.gitignore', '.dockerignore' ]
+			return extractIgnored(file)
+			.then (ignoreGlobs) ->
+				# add the ignored files, and also add the ignore file itself
+				ignoreGlobs.push(file)
```
pimterry: Why do we also include the ignore file? That's not standard behaviour for Docker builds, afaict, and you might want it (maybe if you use git inside your build).

lib/utils/compose.coffee
```diff
@@ -103,24 +103,51 @@ toPosixPath = (systemPath) ->
...
 
 	getFiles = ->
 		streamToPromise(klaw(dir))
 		.filter((item) -> not item.stats.isDirectory())
 		.map((item) -> item.path)
 
+	extractIgnored = (file) ->
+		fs.readFile(file, 'utf8').then (content) ->
+			_(content.split(/\r?\n/))
+			.filter((f) -> !_.isEmpty(f))
```
pimterry: We should also drop any lines starting with `#`

lib/utils/compose.coffee
```diff
@@ -103,24 +103,51 @@ toPosixPath = (systemPath) ->
...
+			return extractIgnored(file)
+			.then (ignoreGlobs) ->
+				# add the ignored files, and also add the ignore file itself
+				ignoreGlobs.push(file)
+				toIgnore = toIgnore.concat(ignoreGlobs)
+		return []
+	.then (files) ->
+		_(toIgnore).uniq().each (glob) ->
+			# Negate the glob so that we keep everything except the matches
+			files = files.filter(minimatch.filter("!#{glob}"))
```
pimterry: This doesn't work, as rules can override each other. Taking an example from the Docker docs:

```
*.md
!README.md
```

This should mean 'ignore every .md except README.md'. With this though, we'll filter out all the md's first, so README.md will already be gone before we get to the 2nd rule.

lib/utils/compose.coffee
```diff
@@ -103,24 +103,51 @@ toPosixPath = (systemPath) ->
...
+		# Also remove any git directory files - we don't need to worry about the .git
+		# directory itself as without any entries the directory will never be created
+		files = files.filter(minimatch.filter("!#{dir}#{path.sep}.git/**/*"))
```
pimterry: I don't think we want to do this - you really might want to use git in your build, and this behaviour doesn't match `docker build`, which will catch people out.

lib/utils/compose.coffee
```diff
@@ -103,24 +103,51 @@ toPosixPath = (systemPath) ->
...
 	pack = tar.pack()
 	getFiles(dir)
-	.map (file) ->
-		relPath = path.relative(path.resolve(dir), file)
-		Promise.join relPath, fs.stat(file), fs.readFile(file),
-			(filename, stats, data) ->
-				pack.entry({ name: toPosixPath(filename), size: stats.size, mode: stats.mode }, data)
+	.each (file) ->
+		# Find any files that should be ignored
+		if path.relative(path.resolve(dir), file) in ['.gitignore', '.dockerignore' ]
```
pimterry: You're not going to like this: gitignore and dockerignore aren't the same format.

https://zzz.buzz/2018/05/23/differences-of-rules-between-gitignore-and-dockerignore/ has some details, and https://gist.github.com/wassname/b25471b0f3bb2f9ff81f might be interesting.

lib/utils/compose.coffee
```diff
@@ -103,24 +103,51 @@ toPosixPath = (systemPath) ->
...
+		_(toIgnore).uniq().each (glob) ->
+			# Negate the glob so that we keep everything except the matches
+			files = files.filter(minimatch.filter("!#{glob}"))
+		# Also remove any git directory files - we don't need to worry about the .git
+		# directory itself as without any entries the directory will never be created
+		files = files.filter(minimatch.filter("!#{dir}#{path.sep}.git/**/*"))
+
+		Promise.map files, (file) ->
+			relPath = path.relative(path.resolve(dir), file)
+			Promise.join relPath, fs.stat(file), fs.readFile(file),
```
pimterry: As in #943, we might need to lstat and handle symlinks here (tbc depending on the resolution of that PR)

lib/utils/compose.coffee
CameronDiver: you're right, I don't like this :/

CameronDiver:
Closing to rebase onto v8 branch.

Issue #947, opened 2018-08-16
resin push <application> does not return error code on build failure


jarrodsinclair:
* Resin CLI v7.9.0, running on CircleCI Linux

When building on remote resin.io cloud builders, the push command does not return an error code when the build fails. Therefore, CircleCI thinks the step has completed successfully.

Page-:
Hey, I think this is already fixed in v7.9.1 @jarrodsinclair, could you update and try again please

jarrodsinclair:
Hi Page,

I just updated to v7.9.2 (standalone) and added a bug to my Dockerfile, but the ```resin push``` command did not return the error code. See the output from CircleCI below. The last line is the return code (using ```echo $?``` after ```resin push```).

```
...
[platform]  tar: TESTING_BROKEN_FILE.tar.gz: Cannot open: No such file or directory
[platform]  tar: Error is not recoverable: exiting now
[platform]  tar: Child returned status 2
[platform]  tar: Error is not recoverable: exiting now
[platform]  
[platform]  Removing intermediate container 135a89a7fc95
[platform]  The command '/bin/sh -c cd /root &&     tar -xzf TESTING_BROKEN_FILE.tar.gz' returned a non-zero code: 2
[Info]      Uploading images
[================>                                  ] 33%[================>                                  ] 33%[================>                                  ] 33%[=================================>                 ] 66%[=================================>                 ] 66%[Success]   Successfully uploaded images

[Error]     Some services failed to build:
[Error]       Service: platform
[Error]         Error: The command '/bin/sh -c cd /root &&     tar -xzf TESTING_BROKEN_FILE.tar.gz' returned a non-zero code: 2
[Error]     Not deploying release.
0
```

Maybe something related to the standalone install?

Page-:
No, it's probably just a different thing than the bug I fixed.  cc @pimterry @CameronDiver

CameronDiver:
Hey @jarrodsinclair thanks for bringing this to our attention! I've made a fix here: https://github.com/resin-io/resin-cli/pull/948

PR #948, opened 2018-08-17
fix: Correctly error out on failed remote build

CameronDiver:
The push command was relying on the output from the builder to indicate
the build status, but this isn't helpful for CI. This commit makes the
remote build module respect the `isError` flag which the builder sends
in any errors. Any errors which come from the builder indicate the
release will not be deployed.

Change-type: patch
Closes: #947 

### DIFF
--- a/lib/utils/remote-build.ts
+++ b/lib/utils/remote-build.ts
@@ -18,6 +18,7 @@ import * as JSONStream from 'JSONStream';
 import * as request from 'request';
 import { ResinSDK } from 'resin-sdk';
 import * as Stream from 'stream';
+import { TypedError } from 'typed-error';
 
 import { tarDirectory } from './compose';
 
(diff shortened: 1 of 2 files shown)


lib/utils/remote-build.ts
```diff
@@ -105,7 +107,11 @@ export async function startRemoteBuild(build: RemoteBuild): Promise<void> {
 		stream.on('data', getBuilderMessageHandler(build));
 		stream.on('end', resolve);
 		stream.on('error', reject);
-	}).return();
+	}).then(() => {
+		if (build.hadError) {
+			throw new Error('Remote build failed');
```
pimterry: This seems good, but I'd like to also ensure we don't send all these errors to Sentry as well. There's an `expectedError` helper for exactly that, which prints the normal exception handling user error text, then kills the process without reporting anything.

We should probably do that in the action though, not here... Sound sensible? If so, I think you want to:

* Throw a custom error class here, like `BuildFailedError`
* Catch that error in the action, and consider it expected instead

CameronDiver: Ah, I forgot about the `exitWithExpectedError` function! Yeah that sounds good, will do.

CameronDiver:
Closing to rebase onto the v8 branch.

PR #950, opened 2018-08-17
fix: Respect ignore files when tarring sources

CameronDiver:
Change-type: major
Closes: #889


lib/utils/ignore.ts
```diff
@@ -0,0 +1,151 @@
...
+	/**
+	 * Given two paths, check whether the first contains the second
+	 * @param path1 The potentially containing path
+	 * @param path2 The potentially contained path
+	 * @return A boolean indicating whether `path1` contains `path2`
+	 */
+	private static contains(path1: string, path2: string): boolean {
+		// First normalise the input, to remove any path weirdness
+		path1 = path.posix.normalize(path1);
+		path2 = path.posix.normalize(path2);
```
dfunckt: How does this behave with symlinks? Are they expanded to their target path?

lib/utils/ignore.ts
```diff
@@ -0,0 +1,151 @@
...
+
+		// Now test if any part of the relative path contains a .. ,
+		// which would tell us that path1 is not part of path2
+		return !/^\.\.$|\.\.\//.test(path.posix.relative(path1, path2));
```
dfunckt: I would expect that testing only whether the string starts with “..” is enough. If the paths are relative, they can’t have “..” in inner path components, can they?

lib/utils/compose.coffee
```diff
@@ -108,14 +108,20 @@ exports.tarDirectory = tarDirectory = (dir) ->
...
 	getFiles = ->
 		streamToPromise(klaw(dir))
 		.filter((item) -> not item.stats.isDirectory())
 		.map((item) -> item.path)
 
+	ignore = new FileIgnorer(dir)
 	pack = tar.pack()
 	getFiles(dir)
+	.each (file) ->
+		if (type = ignore.isIgnoreFile(path.relative(dir, file)))?
```
Page-: I'd prefer
```coffee
type = ignore.isIgnoreFile(path.relative(dir, file))
if type?
```
as I just find it clearer

lib/utils/ignore.ts
```diff
@@ -0,0 +1,151 @@
...
+	 * @returns Promise
+	 */
+	public async addIgnoreFile(
+		fullPath: string,
+		type: IgnoreFileType,
+	): Promise<void> {
+		const contents = await fs.readFile(fullPath, 'utf8');
+
+		contents.split('\n').forEach(line => {
+			const trimmed = line.trimLeft();
```
Page-: Why do you only trim the left side?  I know you do the left side at a minimum for comment purposes but it seems like either both sides or neither would be necessary (how do you ignore a file with leading/trailing spaces in the name?)

lib/utils/ignore.ts
Page-: I'm pretty sure it can contain `..` in inner components by default, but if you use `path.normalize` then it should remove those?

lib/utils/remote-build.ts
```diff
@@ -43,17 +44,25 @@ export interface RemoteBuild {
...
 	type?: string;
 	replace?: boolean;
+	isError?: boolean;
 	// These will be set when the type === 'metadata'
 	resource?: string;
 	value?: string;
 }
 
+export class RemoteBuildFailedError extends TypedError {
+	public constructor() {
```
Page-: You can do:
```typescript
public constructor(message = 'Remote build failed') {
	super(message)
}
```
to keep the ability to specify custom messages

lib/utils/ignore.ts
```diff
@@ -0,0 +1,151 @@
...
+	/**
+	 * @param  {string} relativePath
+	 * 	The relative pathname from the build context, for example a root level .gitignore should be
+	 * 		./.gitignore
+	 * @returns IgnoreFileType
+	 * 	The type of ignore file, or null
+	 */
+	public isIgnoreFile(relativePath: string): IgnoreFileType | null {
+		for (const { pattern, type, allowSubdirs } of FileIgnorer.ignoreFiles) {
+			if (relativePath === pattern || relativePath === `./${pattern}`) {
```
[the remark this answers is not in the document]
CameronDiver: It's because of the fact that dockerignore files can only appear in the root, so if I only checked the basename, it would allow them in every directory. I also can't remove the `./${pattern}` check, because then a root level dockerignore file wouldn't be detected. Keen to hear any other ideas you have though, I could be missing something obvious.

lib/utils/ignore.ts
```diff
@@ -0,0 +1,151 @@
...
+
+		_.each(this.gitIgnoreEntries, ({ pattern, filePath }) => {
+			if (FileIgnorer.contains(path.posix.dirname(filePath), filename)) {
+				gitIgnoreHandle.add(pattern);
+			}
+		});
+
+		if (gitIgnoreHandle.ignores(relFile)) {
+			return false;
+		}
```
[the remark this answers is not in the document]
CameronDiver: I'm not sure what you mean? We need to store the different patterns seperately as they are worked on differently, unless that's not what you meant?

lib/utils/ignore.ts
CameronDiver: They aren't, it doesn't care about the contents, only the path. The symlinks are then handled by tar-stream (or rather not handled, and just added as symlinks).

lib/utils/ignore.ts
CameronDiver: Ah I think you're right yeah.

lib/utils/compose.coffee
CameronDiver: Done

lib/utils/ignore.ts
CameronDiver: Yeah you're right, the ignore files do take into account leading and trailing whitespace, so I just need to use the trimmed version to check for comments, and the actual version for the ignore pattern - good catch!

lib/utils/remote-build.ts
CameronDiver: Good idea, I like it.

lib/utils/ignore.ts
Page-: In that case it's probably simpler to do
```typescript
if (/\s*#/.test(line)) {
	return;
}
```
and avoid needing to create a copy of the string

lib/utils/ignore.ts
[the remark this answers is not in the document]
pimterry: I think the point here is that the above repeats exactly the same structure twice. The only difference is the `git` vs `docker` for the `XIgnoreEntries` `XIgnoreHandle`. You should be able to factor those out, which would make this a little more generic, and shorter too:

```ts
const ignorers = [ // You can probably find a better name for this
	{ handle: gitIgnoreHandle, entries: this.gitIgnoreEntries },
	{ handle: dockerIgnoreHandle, entries: this.dockerIgnoreEntries }
];
_.each(ignorers, ({ handle, entries }) => 
	_.each(entries, ({ pattern, filePath }) => {
		if (FileIgnorer.contains(path.posix.dirname(filePath), filename)) {
			handle.add(pattern);
		}
	});
});

return !_.any(ignorers, ({ handle }) => handle.ignores(relFile));
```

(I haven't tested this, but I think that's a likely outline, you get the idea)

PR #954, opened 2018-08-20
preload: Check for correct architecture instead of correct device type

zvin:


lib/actions/preload.coffee
```diff
@@ -18,24 +18,52 @@ dockerUtils = require('../utils/docker')
...
+	.then (dt) ->
+		deviceTypes = dt
+		return deviceTypes
+
+getDeviceTypesWithSameArch = (deviceType) ->
+	_ = require('lodash')
+
+	getDeviceTypes()
+	.then (deviceTypes) ->
+		deviceType = _.find deviceTypes, (dt) ->
```
pimterry: deviceType changes type here, which is pretty confusing. Can we call the parameter `deviceTypeSlug` instead, and have two separate variables?

lib/actions/preload.coffee
```diff
@@ -143,9 +171,9 @@ module.exports =
 			alias: 's'
 		}
 		{
-			signature: 'dont-check-device-type'
+			signature: 'dont-check-arch'
```
pimterry: Is this a breaking change? I suspect so, since any users using the previous argument will suddenly find that checks _are_ happening (though the checks themselves have changed).

Are there any possible cases where they could be doing something that would work, that will now start failing?

If not, this is ok I think. If so though, we'll need to retarget this to the `v8-meta-branch` branch, which has a couple of other breaking changes @CameronDiver is planning to release, and merge this PR manually, and then merge that branch later to release these breaking changes together.

lib/actions/preload.coffee
```diff
@@ -190,12 +218,12 @@ module.exports =
...
-		options.dontCheckDeviceType = options['dont-check-device-type']
-		delete options['dont-check-device-type']
-		if options.dontCheckDeviceType and not options.appId
+		options.dontCheckArch = options['dont-check-arch'] || false
+		delete options['dont-check-arch']
+		if options.dontCheckArch and not options.appId
 			exitWithExpectedError('You need to specify an app id if you disable the device type check.')
 
-		options.pinDevice = options['pin-device-to-release']
+		options.pinDevice = options['pin-device-to-release'] || false
```
pimterry: I think doing this is sensible, but is it just to make things tidier, or is this going to actually change the resulting behaviour from what was happening before?

lib/actions/preload.coffee
pimterry: Ok. We should retarget, merge manually (with no-checks), and release along with Cameron's changes then.

pimterry:
Just to be clear @zvin, I'm happy to approve this code-wise now, we just need to change it to a major change, rebase onto the v8 branch, and edit this PR to point there instead first.

zvin:
@pimterry by `v8` branch you mean `v8-meta-branch` ?

lib/actions/preload.coffee
```diff
@@ -190,12 +212,12 @@ module.exports =
 		options.splashImage = options['splash-image']
 		delete options['splash-image']
 
-		options.dontCheckDeviceType = options['dont-check-device-type']
-		delete options['dont-check-device-type']
-		if options.dontCheckDeviceType and not options.appId
+		options.dontCheckArch = options['dont-check-arch'] || false
+		delete options['dont-check-arch']
+		if options.dontCheckArch and not options.appId
 			exitWithExpectedError('You need to specify an app id if you disable the device type check.')
```
thgreasi: NIT: should this be `if you disable the device architecture type check`?

lib/actions/preload.coffee
thgreasi: Just-an-idea: If we prefer having this as a minor release, then we could just add `dont-check-arch` and since it sounds like it would have the same effect as `dont-check-device-type`, since in both cases, we don't use the device type at all. In that case I guess we could also add a deprecation warning in the docs for `dont-check-device-type`.

lib/actions/preload.coffee
zvin: good catch, updated

pimterry:
> by v8 branch you mean v8-meta-branch ?

@zvin yes, exactly. I explained in a bit more detail here: https://github.com/resin-io/resin-cli/pull/954#discussion_r211578861.

Alternatively, @thgreasi's idea of keeping the existing option for compatibility is a good one. We could add the new option, but also keep the existing option as a deprecated alias, and then we can delay the breaking change until later and merge this to master right away. Up to you.

zvin:
@pimterry @CameronDiver rebased on `v8-meta-branch`

lib/actions/preload.coffee
```diff
@@ -190,12 +212,12 @@ module.exports =
...
 		delete options['splash-image']
 
-		options.dontCheckDeviceType = options['dont-check-device-type']
-		delete options['dont-check-device-type']
-		if options.dontCheckDeviceType and not options.appId
-			exitWithExpectedError('You need to specify an app id if you disable the device type check.')
+		options.dontCheckArch = options['dont-check-arch'] || false
+		delete options['dont-check-arch']
+		if options.dontCheckArch and not options.appId
+			exitWithExpectedError('You need to specify an app id if you disable the architechture check.')
```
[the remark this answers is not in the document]
zvin: good catch, fixed.

zvin:
Merged into `v8-meta-branch`

Issue #978, opened 2018-10-15
Should support the oss flow


thgreasi:
Depends-on: https://github.com/resin-io/resin-sdk/issues/591
Depends-on: https://github.com/resin-io/resin-sdk/pull/594
HQ: https://github.com/resin-io/balena/pull/1140

dfunckt:
@thgreasi @pimterry should we close this and create new issues going forward?

pimterry:
Sounds sensible to me.

PR #979, opened 2018-10-15
Add support for the Opensource provisioning flow

thgreasi:
Resolves: #978
Change-type: major
Depends-on: https://github.com/resin-io/resin-sdk/pull/591
Depends-on: https://github.com/resin-io/resin-sdk/pull/594
Depends-on: https://github.com/resin-io/resin-sdk/pull/595
HQ: https://github.com/resin-io/balena/pull/1140
Signed-off-by: Thodoris Greasidis <thodoris@resin.io>
---- Autogenerated Waffleboard Connection: Connects to #978

### DIFF
--- a/lib/actions/os.coffee
+++ b/lib/actions/os.coffee
@@ -96,7 +96,7 @@ exports.download =
 			alias: 'o'
 			required: 'You have to specify the output location'
 		}
-		commandOptions.osVersion
+		commandOptions.osVersionOrSemver
 	]
 	action: (params, options, done) ->
 		Promise = require('bluebird')
@@ -197,14 +197,13 @@ exports.buildConfig =
 		.nodeify(done)
 
 exports.configure =
-	signature: 'os configure <image> [uuid] [deviceApiKey]'
+	signature: 'os configure <image>'
 	description: 'configure an os image'
 	help: '''
 		Use this command to configure a previously downloaded operating system image for
 		the specific device or for an application generally.
 
-		Calling this command without --version is not recommended, and may fail in
-		future releases if the OS version cannot be inferred.
+		Calling this command with the exact version number of the targeted image is required.
 
 		Note that device api keys are only supported on ResinOS 2.0.3+.
 
@@ -224,7 +223,7 @@ exports.configure =
 		commandOptions.optionalApplication
 		commandOptions.optionalDevice
 		commandOptions.optionalDeviceApiKey
-		commandOptions.optionalOsVersion
+		commandOptions.osVersion
 		{
 			signature: 'config'
 			description: 'path to the config JSON file, see `resin os build-config`'
@@ -232,7 +231,6 @@ exports.configure =
 		}
 	]
 	action: (params, options, done) ->
-		normalizeUuidProp(params)
 		normalizeUuidProp(options, 'device')
 		fs = require('fs')
 		Promise = require('bluebird')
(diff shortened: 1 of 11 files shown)


lib/utils/config.ts
```diff
@@ -13,77 +13,60 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-	if (!caFile) {
-		return Promise.resolve();
-	}
-	return Promise.fromCallback(cb =>
-		fs.readFile(caFile, { encoding: 'utf8' }, cb),
-	)
-		.then(pem => Buffer.from(pem).toString('base64'))
-		.catch({ code: 'ENOENT' }, () => {});
-}
+type ImgConfig = {
```
dfunckt: You're stop importing `device-config` but aren't removing it as a dependency so I think you should use the type from that module instead of duplicating it here.

thgreasi: @dfunckt I might be missing something, but I think we didn't have typings for `resin-device-config` and it was just `any`.

thgreasi: @dfunckt  I could of course try moving this in this custom repo-wise typings for `resin-device-config` and see whether anything else breaks if you prefer that.

dfunckt: That sounds like a good idea. That module is also used in the api. I would not expect anything to break, adding types means TS will just not compile and we’ll fix it and CS code will just ignore it.

lib/utils/config.ts
```diff
@@ -13,77 +13,60 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-				pubnub: results.apiConfig.pubnub,
-				mixpanel: {
-					token: results.apiConfig.mixpanelToken,
-				},
-				balenaRootCA: results.rootCA,
-			},
-			options,
-		);
-	});
+	return resin.models.os.getConfig(application.app_name, options) as Promise<
```
pimterry: This changes how api keys get inserted, because `getConfig` already includes them, and then we do our own injection on top of that using the below methods.

Have you tested all the cases? Using `--device <existing UUID>` at least I end up with a config.json that includes both `apiKey` and `deviceApiKey`. Definitely doesn't do that on master, I'm not sure if that will cause problems.

lib/actions/os.coffee
```diff
@@ -246,30 +245,20 @@ exports.configure =
 		if _.filter([
 			options.device
 			options.application
-			params.uuid
```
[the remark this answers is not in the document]
thgreasi: Nice catch :+1:

dfunckt:
@thgreasi can you please change this branch to use https://github.com/resin-io/resin-sdk/tree/balena-sdk? Unless there's no benefit, that is, but I'd expect that branch is needed to get all needed changes into the CLI.

thgreasi:
The flow that we tried so far is:
```
bin/resin login
bin/resin whoami

bin/resin app create ossApp4
bin/resin apps
bin/resin app ossApp4

bin/resin os configure ~/Downloads/Resin.io/resinOS\ 2.15.1+rev2.dev/resin.img --app ossApp4 --version 2.15.1

// flash with Etcher and power up the device
bin/resin devices
bin/resin device DEVICE_UUID

// point cli to resinstaging
bin/resin login
bin/resin whoami
bin/resin join
```

PR #988, opened 2018-10-18
Update to qemu version 3.0.0 and support aarch64 emulation

edkeeble:
Disclaimer: I've never written anything for node before, so if I'm not following some best practices, please correct me.

I found that when trying to build a Jetson TX2 project locally, using the `--emulated` flag, the builds always failed with the error "Invalid ELF image for this architecture". This is because the version of qemu downloaded by Resin is built for arm, rather than aarch64 (which Jetson uses).

This PR updates the qemu version to 3.0.0, both because it's the latest version for which Resin has a release and because there are binaries built for both arm and aarch64. I've updated the qemu installation functions to accept an `arch` parameter and use that to download and install the correct binary. I've updated the extraction code because the archive format has changed.

While this change works for my current purposes, it introduces a potential design problem if a single user is building projects for both arm and aarch64. Since the binary is only downloaded once, if the user has previously built an aarch64 project and now wants to build an arm project, they will need to manually remove the `~/.resin/bin/qemu-execve` file before running the build command, in order to force the download of the correct binary.

I can think of two ways to resolve this:

1) Instead of simply checking for the binary's existence, check the binary version and initiate a download if it doesn't match the desired architecture; or

2) Store each binary separately (as `~/.resin/bin/qemu-execve-aarch64`, for instance) and copy the correct version into the Docker container when building.

I think 2 is preferable because it avoids unnecessary downloads and works with concurrent builds with different architectures, but I am not sure where to make that change.

### DIFF
--- a/lib/utils/qemu.coffee
+++ b/lib/utils/qemu.coffee
@@ -80,21 +80,19 @@ installQemu = (arch) ->
 			downloadArchiveName = "qemu-3.0.0+resin-#{arch}.tar.gz"
 			qemuUrl = "https://github.com/balena-io/qemu/releases/download/#{QEMU_VERSION}/#{downloadArchiveName}"
 			extract = tar.extract()
-			extract.on('entry', (header, stream, next) ->
+			extract.on 'entry', (header, stream, next) ->
 				stream.on('end', next)
 				if header.name.includes("qemu-#{arch}-static")
 					stream.pipe(installStream)
 				else
 					stream.resume()
-			)
 			request(qemuUrl)
 			.on('error', reject)
 			.pipe(zlib.createGunzip())
 			.on('error', reject)
 			.pipe(extract)
 			.on('error', reject)
-			.on('finish', ->
+			.on 'finish', ->
 				# make qemu binary executable
 				fs.chmodSync(qemuPath, '755')
 				resolve()
-			)
--- a/lib/utils/qemu.coffee
+++ b/lib/utils/qemu.coffee
@@ -1,16 +1,16 @@
 Promise = require('bluebird')
 
-exports.QEMU_VERSION = QEMU_VERSION = 'v2.5.50-resin-execve'
+exports.QEMU_VERSION = QEMU_VERSION = 'v3.0.0+resin'
 exports.QEMU_BIN_NAME = QEMU_BIN_NAME = 'qemu-execve'
 
-exports.installQemuIfNeeded = Promise.method (emulated, logger) ->
+exports.installQemuIfNeeded = Promise.method (emulated, logger, arch) ->
 	return false if not (emulated and platformNeedsQemu())
 
 	hasQemu()
 	.then (present) ->
 		if !present
-			logger.logInfo('Installing qemu for ARM emulation...')
-			installQemu()
+			logger.logInfo("Installing qemu for #{arch} emulation...")
+			installQemu(arch)
 	.return(true)
 
 exports.qemuPathInContext = (context) ->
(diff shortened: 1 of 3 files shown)


pimterry:
@edkeeble Awesome! This looks great. We're at our annual summit at the moment, so a detailed review will need to wait until next week, but in theory this sounds excellent. @CameronDiver will definitely be interested.

I do think we'll need the last step though, so that we do correctly use the right qemu binary for both architectures even if you're switching between them. I'll look into that more thoroughly next week, but it definitely sounds like the right direction to me. Nice work, thanks!

lib/utils/qemu.coffee
```diff
@@ -67,20 +67,43 @@ platformNeedsQemu = ->
...
-			.pipe(zlib.createGunzip())
-			.on('error', reject)
-			.pipe(installStream)
-			.on('error', reject)
-			.on('finish', resolve)
+			downloadArchiveName = "qemu-3.0.0+resin-#{arch}.tar.gz"
+			fs.mkdtemp(path.join(os.tmpdir(), 'resinqemu-'), (err, folder) ->
+				if err
+					reject(err)
+				qemuUrl = "https://github.com/resin-io/qemu/releases/download/#{QEMU_VERSION}/#{downloadArchiveName}"
```
CameronDiver: This should now be `balena-io`, i.e. 
```
qemuUrl = "https://github.com/balena-io/qemu/releases/download/#{QEMU_VERSION}/#{downloadArchiveName}"

File: lib/utils/qemu.coffee [hunk 2]
```diff
@@ -67,20 +67,43 @@ platformNeedsQemu = ->
...
+				request(qemuUrl)
+				.on('error', reject)
+				.pipe(zlib.createGunzip())
+				.on('error', reject)
+				.pipe(tar.extract({
+					C: folder,
+					filter: (path, entry) -> return path.includes("qemu-#{arch}-static"),
+					strip: 1
+				})).on('finish', ->
```
CameronDiver: NIT: `.on` on a newline.

lib/utils/qemu.coffee
```diff
@@ -67,20 +67,43 @@ platformNeedsQemu = ->
 	os = require('os')
 	os.platform() == 'linux'
 
-installQemu = ->
+installQemu = (arch) ->
 	request = require('request')
 	fs = require('fs')
 	zlib = require('zlib')
+	tar = require('tar')
```
CameronDiver: We have `tar-stream` as a dependency already, would it be possible to use this instead? From the usage below it looks like it would be.

edkeeble: Yeah, I initially tried using `tar-stream`, as I didn't want to add unnecessary dependencies, but was having trouble getting it to work with the existing in-memory approach of extracting the file to `installStream` (can't remember what the exact issue was). I'll circle back to it and try again.

edkeeble:
Thanks for the reviews @CameronDiver and @pimterry. I think I've addressed your feedback. I've also updated the PR to store different local binaries for aarch64 and arm. The correct binary is then copied into the local build dir when balena starts a build.

lib/utils/qemu.coffee
```diff
@@ -61,26 +61,40 @@ getQemuPath = ->
...
-	getQemuPath()
+	getQemuPath(arch)
 	.then (qemuPath) ->
 		new Promise (resolve, reject) ->
 			installStream = fs.createWriteStream(qemuPath)
-			qemuUrl = "https://github.com/balena-io/qemu/releases/download/#{QEMU_VERSION}/#{QEMU_BIN_NAME}.gz"
+			downloadArchiveName = "qemu-3.0.0+resin-#{arch}.tar.gz"
+			qemuUrl = "https://github.com/balena-io/qemu/releases/download/#{QEMU_VERSION}/#{downloadArchiveName}"
+			extract = tar.extract()
+			extract.on('entry', (header, stream, next) ->
```
CameronDiver: you can drop the parens on this call if you like

lib/utils/qemu.coffee
```diff
@@ -61,26 +61,40 @@ getQemuPath = ->
...
+			)
 			request(qemuUrl)
 			.on('error', reject)
 			.pipe(zlib.createGunzip())
 			.on('error', reject)
-			.pipe(installStream)
+			.pipe(extract)
 			.on('error', reject)
-			.on('finish', resolve)
+			.on('finish', ->
```
CameronDiver: Same here.

lib/utils/qemu.coffee
edkeeble: Ah, coffeescript. Thanks!

pimterry:
@resin-ci retest

PR #995, opened 2018-10-21
chore: Add on the fly transpiled bin

thgreasi:
Loved @dfunckt 's suggestion that saves a huge amount of time by not having to build after each change.
Adds an alternative bin file that does not require building the project but loads the source files directly.
Changed the package.json so that the published module doesn't contain the new file and stays the same as before.

Change-type: minor

### DIFF
--- a/bin/resin-dev
+++ b/bin/resin-dev
@@ -1,0 +1,15 @@
+#!/usr/bin/env node
+
+// ****************************************************************************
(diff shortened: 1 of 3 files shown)


bin/resin-dev
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+
```
pimterry: Can we put a loud comment in here documenting that this is for dev use _only_? Might be nice to mention it in the dev section of the readme too.

PR #997, opened 2018-10-22
Rename everything from 'resin' to 'balena'

### LINKED ISSUE #984: Rename the CLI

pimterry:
There's a couple more changes, and definitely manual testing required here, but I think this is 95% of it, and it'd definitely be useful to get a first review as is.

Fixes #984
Change-type: major


capitanodoc.ts
```diff
@@ -53,8 +53,8 @@ The CLI does support HTTP(S) proxies.
 
 You can configure the proxy using several methods (in order of their precedence):
 
-* set the \`RESINRC_PROXY\` environment variable in the URL format (with protocol, host, port, and optionally the basic auth),
-* use the [resin config file](https://www.npmjs.com/package/resin-settings-client#documentation) (project-specific or user-level)
+* set the \`BALENARC_PROXY\` environment variable in the URL format (with protocol, host, port, and optionally the basic auth),
```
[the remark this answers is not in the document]
pimterry: This is a very good point. I started testing some of this, and I'd missed it. The actual change is outside this PR, but it does still need fixing. Good point!

I've opened a PR to fix this here: https://github.com/resin-io-modules/balena-settings-client/pull/57

resin-completion.bash
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
```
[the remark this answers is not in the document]
pimterry: True, fixed

typings/resin-sdk-preconfigured.d.ts
```diff
@@ -1,801 +0,0 @@
...
-		(options?: SdkOptions): ResinSdk.ResinSDK;
-
-		setSharedOptions(options: SdkOptions): void;
-		fromSharedOptions: () => ResinSdk.ResinSDK;
-	}
-
-	const ResinSdk: ResinSdk.ResinSDK;
-
-	export = ResinSdk;
-}
```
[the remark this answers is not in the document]
pimterry: Yes, this is intentional. The preconfigured SDK is deprecated, so we're removing it entirely (there's actually a little more to do there, but removing this helps make it easy to find the last bits)

thgreasi:
In `bin/balena-dev` there is also now a comment `using bin/resin` that we should also change.

thgreasi:
Should we also change this to be compared against the `v9-meta-branch`?

lib/utils/patterns.ts
```diff
@@ -276,11 +276,11 @@ export function printErrorMessage(message: string) {
...
 
-export function exitWithExpectedError(message: string | Error) {
+export function exitWithExpectedError(message: string | Error): never {
 	if (message instanceof Error) {
 		({ message } = message);
 	}
 
 	printErrorMessage(message);
-	process.exit(1);
+	return process.exit(1);
```
thgreasi: I'm not sure about the `return` but don't have a strong opinion either :+1:

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+
+At the moment the CLI does not support per-service variables,
+so the following commands will only show service-wide
+environment variables.
+
+Example:
+
+	$ balena envs --application MyApp
+	$ balena envs --application MyApp --config
+	$ balena envs --device 7cf02a6\
```
thgreasi: Would an extra line break look awkward?

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+`,
+	options: [
+		commandOptions.optionalApplication,
+		commandOptions.optionalDevice,
+
+		{
+			signature: 'config',
+			description: 'show config variables',
+			boolean: true,
+			alias: ['v', 'verbose'],
```
thgreasi: Should we now drop these (and/)or use `-c`?
If that's for compatibility reasons then :+1:

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+		},
+	],
+	permission: 'user',
+	async action(_params, options, done) {
+		normalizeUuidProp(options, 'device');
+		const Bluebird = await import('bluebird');
+		const _ = await import('lodash');
+		const balena = (await import('balena-sdk')).fromSharedOptions();
```
thgreasi: Optional-NIT: Would love this to be named as `sdk` like we do in other modules.
Should be easy to do in here, now that it's in TS :)

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+		const { exitWithExpectedError } = await import('../utils/patterns');
+
+		return Bluebird.try(function(): Promise<
+			DeviceVariable[] | ApplicationVariable[]
+		> {
+			if (options.application != null) {
+				return balena.models.application[
+					options.config ? 'configVar' : 'envVar'
+				].getAllByApplication(options.application);
+			} else if (options.device != null) {
```
thgreasi: NIT: Could we drop the `!= null`?

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+				return balena.models.device[
+					options.config ? 'configVar' : 'envVar'
```
thgreasi: I'm impressed that the typings can figure this out!
I would expect that we would first had to assign the functions to a variable with a ternary before invoking it.

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+				].getAllByDevice(options.device);
+			} else {
+				return exitWithExpectedError(
+					'You must specify an application or device',
+				);
+			}
+		})
+			.tap(function(environmentVariables) {
```
thgreasi: I guess these could be arrow functions, but I'm also fine with these as is.

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+import { normalizeUuidProp } from '../utils/normalization';
+import { DeviceVariable, ApplicationVariable } from 'balena-sdk';
+
+const getReservedPrefixes = async (): Promise<string[]> => {
+	const balena = (await import('balena-sdk')).fromSharedOptions();
+	const settings = await balena.settings.getAll();
+
+	const response = await balena.request.send({
+		baseUrl: settings.apiUrl,
+		url: '/config/vars',
```
thgreasi: Atm this isn't part of the opensource api.
I would prefer a `--config` option in the respective `add` action.
We will have to add a way to specify service vars as well in the future, so imo it makes sense to prepare for that.
See: https://github.com/resin-io/resin-api/blob/master/src/core/routes/index.ts

We could maybe try catch 404's (and possibly filter the `Cannot GET /config/vars` body as well) so that this at least works out of the box for non-oss users, but it might make the code more complex and have different results based on the api used (sounds bad).

lib/actions/environment-variables.ts
```diff
@@ -0,0 +1,299 @@
...
+Notice this command asks for confirmation interactively.
+You can avoid this by passing the \`--yes\` boolean option.
+
+If you want to eliminate a device environment variable, pass the \`--device\` boolean option.
+
+Examples:
+
+	$ balena env rm 215
+	$ balena env rm 215 --yes
+	$ balena env rm 215 --device\
```
thgreasi: How about adding a `--config` option as well, so that users can delete config vars (since we provide a way to create config vars).

thgreasi:
Forgot to mention that this is awesome work :+1:

lib/actions/environment-variables.ts
pimterry: Yes, this is going to be a problem. We definitely need to support it without `--config` for now though, as this will break some users workflows (I've looked in mixpanel and checked, there's people who have automation that depends on this working).

In future, I'd like to do a big breaking change that rewrite all this in a way that matches the SDK - no ids exposed, just get/set/delete by key. Even doing that without the vars config endpoint will be tricky though. We basically can't do any client-side validation or understand anything clever about env vars or anything unless this info is exposed somehow.

Is there a good reason why we can't just add this endpoint to open-source?

lib/actions/environment-variables.ts
pimterry: We should, but for now I'd rather burn this to the ground and write it fresh :smiley:. The UX here could be way nicer, pretty easily, the only reason I'm not doing that immediately is that it's a huge breaking change.

For the moment, my goal here is to make sure it doesn't break existing workflows, and then make it powerful and nice to use in a later major bump.

Previously I think this command did actually work for config vars, but nobody appears to be using that (based on the last couple of months of mixpanel data), and the docs had a big warning telling you not to, so I think breaking that is ok.

lib/actions/environment-variables.ts
pimterry: Yeah, I didn't think this would work either, but it totally does. TS is cool :sunglasses:

lib/actions/environment-variables.ts
thgreasi: I'm not sure about that.
Let's bring @dfunckt in the discussion!
Maybe all we need is a new versioned endpoint that returns the same results.

thgreasi: Last resort would still be to try fetch the `/config/vars` (catching 404's) and if that's available and the name matches, infer that the user really mean to be using `--config`.

pimterry: This is now part of the open-source API: https://github.com/resin-io/open-balena-api/pull/2 :smile:

lib/actions/environment-variables.ts
pimterry: Yep, this is for compatibility reasons, we need these to keep working.

I've added the `c` alias as well though, good idea.

lib/actions/environment-variables.ts
pimterry: By itself, it adds an extra line to the output, but you're right that we shouldn't need this. I've move to `stripIndent` instead, which helpfully handles this as well.

lib/utils/patterns.ts
pimterry: It's required to make the `never` typings work properly, otherwise TS doesn't know what this function will never return.

lib/actions/environment-variables.ts
pimterry: Yes we could, done.

PR #1009, opened 2018-11-01
Dont ignore balena metadata directories when balena pushing

CameronDiver:
Change-type: patch

### DIFF
--- a/lib/utils/ignore.ts
+++ b/lib/utils/ignore.ts
@@ -77,6 +77,7 @@ export class FileIgnorer {
 			if (/\s*#/.test(line) || _.isEmpty(line)) {
 				return;
 			}
+
 			this.addEntry(line, fullPath, type);
 		});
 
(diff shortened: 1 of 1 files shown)


lib/utils/ignore.ts
```diff
@@ -77,6 +77,19 @@ export class FileIgnorer {
...
+			// We want to be able to use files in the .balena directory,
+			// even if they're part of the gitignore file (which they should be)
+			if (type === IgnoreFileType.GitIgnore) {
+				const trimmed = path.normalize(line.trimLeft());
+				if (
+					_.startsWith(trimmed, '.balena') ||
+					_.startsWith(trimmed, '.resin')
+				) {
+					return;
+				}
```
pimterry: I think we need to do this when matching the files, not building the matchers.

If you have a match like `.*` for example (ignore all dot files) or various other cases, then it won't hit this check, and this folder will still be ignored.

PR #1012, opened 2018-11-05
Make specifying the version during configuration optional

dfunckt:
`version` used to be optional but it seems we recently had to make it a required parameter. However it really feels redundant when all it’s used for is to determine whether the command should issue a legacy user API key or a provisioning key.

This makes `version` optional but tries to figure it out by itself by reading `os-release` from the image's boot partition. This is not foul-proof however, and while it'll work with most recent images (see https://github.com/balena-io-modules/resin-device-init/pull/37) it won't work with all and in that case it'll bail out and only then warn the user to specify it via the `--version` argument.

Change-type: minor
Depends-on: https://github.com/balena-io-modules/resin-device-init/pull/37


dfunckt:
@thgreasi no, version will only be optional for the user, it will still be required in the SDK and API. This PR takes care of figuring the version automatically if the user didn't provide one.

lib/actions/os.coffee
```diff
@@ -273,6 +273,18 @@ exports.configure =
 			.then (answers) ->
 				answers.version = options.version
 
+				if not answers.version?
+					answers.version = helpers.getOsVersion(params.image).tap (version) ->
+						console.log(version)
```
thgreasi: `console.log` :police_car:

lib/actions/os.coffee
```diff
@@ -213,17 +213,17 @@ exports.configure =
 
 		Examples:
 
-			$ balena os configure ../path/rpi.img --device 7cf02a6 --version 2.12.7
-			$ balena os configure ../path/rpi.img --device 7cf02a6 --version 2.12.7 --device-api-key <existingDeviceKey>
-			$ balena os configure ../path/rpi.img --app MyApp  --version 2.12.7
+			$ balena os configure ../path/rpi.img --device 7cf02a6
+			$ balena os configure ../path/rpi.img --device 7cf02a6 --device-api-key <existingDeviceKey>
+			$ balena os configure ../path/rpi.img --app MyApp
```
thgreasi: Could we also have an example that uses the `--version` param?

dfunckt:
@thgreasi updated, can you please another look?

lib/utils/helpers.ts
```diff
@@ -84,18 +85,21 @@ export function getManifest(
...
-	// we encounter any errors along the way.
-	return imagefs
-		.readFile({
-			image,
-			partition: 1,
-			path: '/device-type.json',
-		})
-		.then(JSON.parse)
-		.catch(() => balena.models.device.getManifestBySlug(deviceType));
+	const init: typeof _deviceInit = require('balena-device-init');
```
pimterry: Nit: If you make this function `async`, then this can become:

```typescript
const init = await import('balena-device-init');
```

and that way it'll automatically pick up the types by itself, so you can drop the `_deviceInit` import.

That would also let you use `await` below to flatten the promise chain there, which would make that a little clearer too.

The one constraint is that you can't import `Bluebird` as `Promise` any more in that case. Elsewhere we've just been renaming the import to `Bluebird`.

lib/utils/helpers.ts
```diff
@@ -84,18 +85,21 @@ export function getManifest(
...
+		}
+		return balena.models.device.getManifestBySlug(deviceType);
+	});
+}
+
+export function getOsVersion(
+	image: string,
+	manifest: BalenaSdk.DeviceType,
+): Promise<string | null> {
+	const init: typeof _deviceInit = require('balena-device-init');
```
pimterry: Ditto

dfunckt:
@resin-ci retest

nazrhom:
@resin-ci retest

dfunckt:
@resin-ci retest

PR #1016, opened 2018-11-07
Qemu3 multiple architecture support

pimterry:
Versionist kept fighting me on the external PR from the fork, so I've merged to this branch instead, and I believe that creating a PR from here should now all work totally effortlessly, as normal :crossed_fingers:


CameronDiver:
@pimterry Maybe it's worth squashing the commits?

Issue #1019, opened 2018-11-12
Build & deploy failing on v9.1.0 for armv7hf/raspberrypi3


thgreasi:
- **Cli version:** 9.1.0
- **node -v:** v8.12.0
- **npm -v:** 6.4.1
- **Operating system and architecture:** Ubuntu 18.04 x86-64

This does work on v9.0.x but starting failing in v9.1.0, possibly related w/ the qemu v3 bump in #988 .
```
$ DEBUG=1 balena build --application testProjectDay1 --build --emulated
[Debug]   Parsing input...
[Debug]   Loading project...
[Debug]   Resolving project...
[Debug]   Failed to resolve project: 
[Debug]   AggregateError of:
[Debug]       Error: ENOENT: no such file or directory, open '/home/teo/balena.io/sample projects/simple-server-node/docker-compose.yml'
[Debug]       Error: ENOENT: no such file or directory, open '/home/teo/balena.io/sample projects/simple-server-node/docker-compose.yaml'
[Info]    Creating default composition with source: /home/teo/balena.io/sample projects/simple-server-node
[Debug]   Creating project...
[Info]    Building for armv7hf/raspberrypi3
[Info]    Emulation is enabled
[Debug]   Prepared tasks; building...
[Build]   Built 1 service in 0:03
[Build]   main [============>        ]  62% Step 5/8: standard_init_linux.go:190: exec user process caused "exec format error"
[Error]   Build failed
Error: The command '/tmp/qemu-execve -execve /bin/sh -c JOBS=MAX npm install --production --unsafe-perm && npm cache clean && rm -rf /tmp/*' returned a non-zero code: 1
    at Stream.<anonymous> (/usr/lib/node_modules/balena-cli/node_modules/resin-docker-build/build/builder.js:71:33)
    at Stream.stream.write (/usr/lib/node_modules/balena-cli/node_modules/through/index.js:26:11)
    at Stream.ondata (internal/streams/legacy.js:16:26)
    at emitOne (events.js:116:13)
    at Stream.emit (events.js:211:7)
    at drain (/usr/lib/node_modules/balena-cli/node_modules/through/index.js:36:16)
    at Stream.stream.queue.stream.push (/usr/lib/node_modules/balena-cli/node_modules/through/index.js:45:5)
    at Parser.parser.onToken (/usr/lib/node_modules/balena-cli/node_modules/JSONStream/index.js:132:18)
[...]

PR #1020, opened 2018-11-12
Fix the architecture string used when downloading qemu versions

CameronDiver:
Closes: #1019
Change-type: patch


lib/utils/qemu.coffee
```diff
@@ -34,7 +34,9 @@ exports.copyQemu = (context, arch) ->
 		new Promise (resolve, reject) ->
 			read = fs.createReadStream(qemu)
 			write = fs.createWriteStream(binPath)
+
 			read
+			.on('error', reject)
```
thgreasi: Do we really need this one as well?
Wouldn't `reject` end-up being called twice?

PR #1025, opened 2018-11-14
Include `join` and `leave` commands in API documentation

dfunckt:
Change-type: patch

### DIFF
--- a/capitanodoc.ts
+++ b/capitanodoc.ts
@@ -144,6 +144,10 @@ environment variable (in the same standard URL format).\
 			files: ['build/actions/build.js', 'build/actions/deploy.js'],
 		},
 		{
+			title: 'Platform',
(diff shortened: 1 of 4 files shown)


dfunckt:
I can't get the doc generator to include the blurb about each command. See `join` for example: https://github.com/balena-io/balena-cli/blob/master/lib/actions/join.ts#L31 -- `description` is missing from the generated docs. I tried a couple of obvious things like changing the ES6 backticks for plain quotes, `module.exports = ...` instead of `exports.join` but didn't work. Any idea what might be happening?

dfunckt:
Thanks @pimterry, that did it :) I also took the chance and wrote what these commands do in somewhat less obscure terms.

dfunckt:
@thgreasi @pimterry can you have another look?

PR #1026, opened 2018-11-15
Add case-insensitive checking for application names in balena push

CameronDiver:
The filter is added with an `as any`, as the typings dont yet support
using $eq and $ne.

Change-type: patch


CameronDiver:
@resin-ci retest

thgreasi:
@resin-ci retest

PR #1030, opened 2018-11-21
Don't show list of devices when UUID is not found when doing an SSH

CameronDiver:
If the user provides a UUID which is not found or not accessible, the CLI will proceed to show a list of accessible devices. This doesn't make sense from both a UX and a automation POV.

Change-type: patch

---
##### Contributor checklist

### DIFF
--- a/lib/actions/ssh.coffee
+++ b/lib/actions/ssh.coffee
@@ -99,10 +99,13 @@ module.exports =
 			return "-o #{bash.args({ ProxyCommand: proxyCommand }, '', '=')}"
 
 		Promise.try ->
-			return false if not params.uuid
-			return balena.models.device.has(params.uuid)
-		.then (uuidExists) ->
-			return params.uuid if uuidExists
+			return { provided: false } if not params.uuid
+			return { provided: true, uuid: balena.models.device.has(params.uuid) }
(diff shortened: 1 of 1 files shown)


lib/actions/ssh.coffee
```diff
@@ -99,10 +99,13 @@ module.exports =
 			return "-o #{bash.args({ ProxyCommand: proxyCommand }, '', '=')}"
 
 		Promise.try ->
-			return false if not params.uuid
-			return balena.models.device.has(params.uuid)
-		.then (uuidExists) ->
-			return params.uuid if uuidExists
+			return { provided: false } if not params.uuid
+			return { provided: true, uuid: balena.models.device.has(params.uuid) }
```
lekkas: Probably need something like

```
return Promise.props({ provided: true, uuid: balena.models.device.has(params.uuid) })
```

lib/actions/ssh.coffee
```diff
@@ -99,10 +99,13 @@ module.exports =
...
+		.then ({ provided, uuidExists }) ->
```
lekkas: This should be

```
.then ({ provided, uuid }) ->
```

ditto in line 106, `s/uuidExists/uuid/`

Issue #1007, opened 2018-10-31
'You have to specify an exact os version'


BorisBokmans:
Dear people,

I'm trying to provision a VM (WIN10 Enterprise LTSB) to the Balena dashboard.
I installed the standalone Balena-CLI on the VM.
After specifying my hardware (I chose Intel NUC because of the VM usage) it started to download the DeviceOS image. This had success.
After that I received an error: 'You have to specify an exact os version'.

Somebody knows a fix for this?

Thanks!

https://ibb.co/ddBVdL

- **Cli version:**
- **Operating system and architecture:**

pimterry:
Hi @BorisBokmans thanks for the report, that definitely sounds like a bug! We'll take a look at it now, watch this space.

pimterry:
Just to keep you updated @BorisBokmans: this is caused by changes in CLI v9.0.0 to make a `version` argument required when configuring images (which quickstart does for you, under the hood). We're fixing this, but doing it properly is part of some larger changes to how version detection works in configuration, so it'll take a little longer.

In the meantime, you can either:

* Temporarily downgrade to a CLI version before v9 (note that before v9 the npm package was called `resin-cli`, not `balena-cli`)
* Run the setup commands for your device manually, instead of via quickstart:
	- Download a fresh OS: `balena os download intel-nuc --output <path> --version <version>` (the latest version for the NUC is `v2.13.6`)
	- Configure it for your app: `balena os configure <path> --application <appname> --version <version>`
	- Write it to disk: `balena os initialize <path> --type intel-nuc` (since you're setting up a VM, there may be a different step required to create a bootable VM image, that'll depend on the VM software you're using)

I'll update this ticket once we've resolved this more permanently, but hopefully one of the workarounds above will work for you in the meantime.

pimterry:
Just an update: this should be fully fixed by #1012, once that's merged & released.

dfunckt:
@BorisBokmans, the PR Tim mentioned above has been merged and released as part of v9.2.2 -- can you please check whether it solves the issue?

thgreasi:
Hi we have identified the issue and working on a fix.
Thanks for bringing this up.

srlowe:
Closing as fixes have been merged.  Please re-open if you still encounter problems.

PR #1035, opened 2018-11-23
actions/config: Fix examples to include --version as required

thgreasi:
This is currently required by the SDK's `os.os.getConfig()` which throws `An OS version is required when calling os.getConfig` when it isn't provided.

Connects-to: #1007
Change-type: patch

---
##### Contributor checklist

### DIFF
--- a/lib/actions/config.coffee
+++ b/lib/actions/config.coffee
@@ -231,17 +231,17 @@ exports.generate =
 
 		Examples:
 
-			$ balena config generate --device 7cf02a6
-			$ balena config generate --device 7cf02a6 --generate-device-api-key
-			$ balena config generate --device 7cf02a6 --device-api-key <existingDeviceKey>
-			$ balena config generate --device 7cf02a6 --output config.json
-			$ balena config generate --app MyApp
-			$ balena config generate --app MyApp --output config.json
-			$ balena config generate --app MyApp \
+			$ balena config generate --device 7cf02a6 --version 2.12.7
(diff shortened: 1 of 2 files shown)


lib/actions/config.coffee
```diff
@@ -231,17 +231,17 @@ exports.generate =
...
+			$ balena config generate --device 7cf02a6 --version 2.12.7 --device-api-key <existingDeviceKey>
+			$ balena config generate --device 7cf02a6 --version 2.12.7 --output config.json
+			$ balena config generate --app MyApp --version 2.12.7
+			$ balena config generate --app MyApp --version 2.12.7 --output config.json
+			$ balena config generate --app MyApp --version 2.12.7 \
 				--network wifi --wifiSsid mySsid --wifiKey abcdefgh --appUpdatePollInterval 1
 	'''
 	options: [
-		commandOptions.optionalOsVersion
+		commandOptions.osVersion
```
dfunckt: I don't understand -- why do you make it a required argument again?

dfunckt: Ignore me, this is another command.

thgreasi: Sorry I didn't include a description in the commit :disappointed: 
Just for tracking reasons: This is currently required by the SDK's `os.os.getConfig()` which throws `An OS version is required when calling os.getConfig` when it isn't provided. `balena os configure` tries to find the versions from the image provided, but `balena configure` doesn't work on a specific image.

Issue #1032, opened 2018-11-22
Revisit .gitignore file treatment by balena build / balena push


pdcastro:
[PR #950](https://github.com/balena-io/balena-cli/pull/950) implemented the feature described in [issue #889](https://github.com/balena-io/balena-cli/issues/889) but this has led to users reporting valid use cases where files should be uploaded for the build but not git-committed, hence listed in .gitnore files but not in .dockerignore files:

https://forums.balena.io/t/ignoring-dockerignore/2805/17

A user reported a situation where, for example, `.gitgnore` files present in subfolders like the Golang `vendor` folder (often consisting of github repo clones) cause the listed files to be excluded from the image build, when they shouldn't.

Consider alternative behaviours such as:

* Look at `.dockerignore` files only, not `.gitignore` (for both `balena build` and `balena push`). This would however be inconsistent with `git push` when used to trigger a build.
* Have `balena build` look at `.dockerignore` only, while `balena push` looks at both `.dockerignore` and `.gitignore`. (Inconsistency within the balena cli itself?)
* Having a command-line option to select which .ignore files to consider. By default, perhaps consider .dockerignore only.
* Having a `.balenaignore` file which could contain "import directives" to refer to the `.dockerignore` and `.gitignore` files that the user is interested in.
* Something else.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_h503)

magiconair:
My suggestion would be to leave the behavior as is since this is what most users are expecting but provide a flag to use only the one or the other or neither.

matijse:
First of all, sorry for the long comment...

I ended up here because I had noticed unpredictable files being in/excluded with the `balena push` command. The documentation of this command doesn't mention what ignore files it looks at...

My setup:

In the main folder from where I run the `balena push` command, I have the following `docker-compose.yml` file. This file builds an application from a subfolder, which is a **git submodule**.

```
version: "2.1"

services:
  app:
    build:
      context: ./core-application/

  database:
    image: timescale/timescaledb:1.0.0-pg10
```
(relevant pieces shown)

I noticed the following cases, where I think the behaviour might be wrong:

Case 1: The `.gitignore` in the root of `core-application` contained the `vendor` folder like this: (with leading slash).

```
/vendor
```

[...]

PR #1036, opened 2018-11-26
Filter files relative to the .gitignore file to ensure the parent pat…

matijse:
All `.gitignore` files in the file tree are used to filter files. Files deeper in the file tree specify rules that only apply to files and subfiles from that point onwards. This patch makes sure only the relative path from the specific `.gitignore` file is compared to the rules inside the `.gitignore` file.

This ensures that parts from the parent path cannot trigger the rules.

Resolves: # 
See: https://github.com/balena-io/balena-cli/issues/1032 (issue not resolved, as .gitignore files may stil need to be treated differently
Depends-on: <url> 
Change-type: patch 

---
##### Contributor checklist

### DIFF
--- a/lib/utils/ignore.ts
+++ b/lib/utils/ignore.ts
@@ -87,7 +87,7 @@ export class FileIgnorer {
 	// Pass this function as a predicate to a filter function, and it will filter
 	// any ignored files
 	public filter = (filename: string): boolean => {
-		const relFile = path.relative(this.basePath, filename);
+		let relFile = path.relative(this.basePath, filename);
 
 		// Don't ignore any metadata files
 		const baseDir = path.dirname(relFile).split(path.sep)[0];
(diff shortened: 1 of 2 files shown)


matijse:
I created this by accident, and to soon. There is an error in this code, as it only works when there is only one ignore file, because on every ignore the `relFile` gets updated.

I will update this PR with a failing test of two `.gitignore` files to indicate the problem. I don't have any time to continue to work on this today, so if anybody wants to help...

thgreasi:
@matijse thanks for opening this!
We started an internal discussion on how should this be working so that it doesn't feel inconsistent when compared with docker ignore files.
I comment back once we end up with the desired approach, but thank you already for providing those test cases, documenting the case that you faced issues with.

matijse:
Depending on when you expect to make a decision, it would be nice if this can be fixed in the mean time.... 

This is because now the cli parses `.gitignore` files wrong. Especially for my third case, this cannot be fixed by removing my own `.gitignore` files, as it triggers on `.gitignore` files in dependencies. See this comment: https://github.com/balena-io/balena-cli/issues/1032#issuecomment-441574594 for more details.

Do you have a timeline on your discussion? So I can know if it is useful to spend any time on this, as now it is keeping me from setting up a deployment pipeline... Thanks.

pdcastro:
Starting with balena CLI v12 (2020-06-16), `.gitignore` files are not taken into account by default, and starting with balena CLI v13 (2021-12-23), `.gitignore` files are not taken into account at all.

PR #1037, opened 2018-11-26
Update dependencies and pin event-stream

CameronDiver:
Change-type: patch

---
##### Contributor checklist


package.json
```diff
@@ -121,13 +121,14 @@
...
-    "docker-qemu-transpose": "^0.5.1",
-    "docker-toolbelt": "^3.1.0",
+    "docker-qemu-transpose": "^0.5.3",
+    "docker-toolbelt": "^3.3.5",
     "dockerode": "^2.5.5",
     "dockerode-options": "^0.2.1",
     "drivelist": "^5.0.22",
     "ejs": "^2.5.7",
     "etcher-image-write": "^9.0.3",
+    "event-stream": "^3.3.5",
```
thgreasi: @CameronDiver this doesn't pin event-stream as in the PR title, it just adds an explicit dependency.
Is `^3.3.5` what you intended to do?

CameronDiver: Oh man, that's the exact opposite of what I want to do. This still works, because of some downstream dependencies pinning, but will change.

Issue #1038, opened 2018-11-27
os initialize fails b/c of non-bluebird promise


thgreasi:
- **Cli version:** 9.3.2
- **Standalone install:** no
- **Node version:** v8.12.0
- **Npm version:** 6.4.1
- **Operating system and architecture:** Ubuntu 18.04

PR #1039, opened 2018-11-27
actions/os-initialize: Convert Promise to a Bluebird one

thgreasi:
Resolves: #1038
Change-type: patch

---
##### Contributor checklist

### DIFF
--- a/lib/actions/os.coffee
+++ b/lib/actions/os.coffee
@@ -339,7 +339,7 @@ exports.initialize =
 
 			#{INIT_WARNING_MESSAGE}
 		""")
-		helpers.getManifest(params.image, options.type)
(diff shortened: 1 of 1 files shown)


dfunckt:
Yeah, I saw the `.tap` but I was confused because `form.run` should return a Bluebird promise and thought the chain will go on from there.

thgreasi:
@dfunckt The `.then()` chaining internally is done by using `Promise.resolve()` to handle both `then`ables & plain values. As a result, since `getManifest` returns a native Promise, that implementation of `Promise.resolve()` it is also used for unwrapping and chaining the returned value. You can observe that on the typings as well, that correctly type the return type of a promise chain based on the promise on the head of the chain

thgreasi:
@resin-ci retest

Issue #1015, opened 2018-11-07
Add --registry-secrets option for balena push (private registry authentication)


pdcastro:
References and related work:
* [Roadmap card](https://trello.com/c/ucTQpzRE/44-build-secrets-and-variables#)
* Backend spec: https://github.com/balena-io/balena/pull/1287
* HQ: https://github.com/balena-io/balena/issues/12 (Users want Build time secrets that don't reach the running container)
* HQ: https://github.com/balena-io/balena/issues/196 (Customers want to base their dockerfiles on a base from a private docker hub registry)
* HQ: https://github.com/balena-io/balena/issues/1456 (User wants to use private Docker registry)
* balena-cli: https://github.com/balena-io/balena-cli/pull/1041 (Add --registry-secrets option for balena push (private registry auth))
* resin-builder: https://github.com/balena-io/resin-builder/issues/485 (The builder should support build secrets being sent to the build)
* resin-builder: https://github.com/balena-io/resin-builder/issues/515 (The builder should support build variables from the secrets directory)
* resin-builder: https://github.com/balena-io/resin-builder/issues/561 (Add builder support for private docker registry authentication)
* [7 Nov 2018 Flowdock thread](https://www.flowdock.com/app/rulemotion/resin-tech/threads/SR2QszHOuITWG6XfS83m-LxzwB3)
* [7 Nov 2018 Google Doc notes/comments](https://docs.google.com/document/d/1jfeV1rGU3sXZtolb5LfV2o8WO4v7aSIzSoryhyNf_ek)

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_g4gb)

PR #1041, opened 2018-12-02
Add --registry-secrets option for balena push (private registry auth)

pdcastro:
Resolves: #1015
Depends-on: https://github.com/balena-io/resin-builder/pull/562
Change-type: minor

---
##### Contributor checklist

- [x] Introduces security considerations


lib/actions/push.ts
```diff
@@ -98,30 +113,53 @@ async function getAppOwner(sdk: BalenaSDK, appName: string) {
...
-	{
-		applicationOrDevice: string;
-	},
-	{
-		source: string;
-		emulated: boolean;
-		nocache: boolean;
+function parseRegistrySecrets(secretsFilename: string): RegistrySecrets {
+	let registrySecrets: RegistrySecrets = {};
+	if (secretsFilename) {
```
CameronDiver: I'd prefer this typed correctly above, e.g.
```
function parseRegistrySecrets(secretsFilename?: string): RegistrySecrets
```
and then a test like so below:
```
if (secretsFilename == null)
```

Note the `==`, rather than `===` which will test for both `undefined` and `null`

lib/actions/push.ts
```diff
@@ -98,30 +113,53 @@ async function getAppOwner(sdk: BalenaSDK, appName: string) {
...
+		let isYaml = false;
+		if (/.+\.ya?ml$/i.test(secretsFilename)) isYaml = true;
+		else if (!/.+\.json$/i.test(secretsFilename))
+			throw new Error(
+				'The registry-secrets filename must end with .json, .yml or .yaml',
+			);
+
+		const raw = require('fs')
+			.readFileSync(secretsFilename)
```
CameronDiver: We shouldn't be doing synchronous file I/O. You can do
```
import { fs } from 'mz';
```

To get a nice promise based interface to fs, and using `async/await` there would also be preferred.

lib/utils/compose.coffee
```diff
@@ -100,7 +100,7 @@ toPosixPath = (systemPath) ->
 	path = require('path')
 	systemPath.replace(new RegExp('\\' + path.sep, 'g'), '/')
 
-exports.tarDirectory = tarDirectory = (dir) ->
+exports.tarDirectory = tarDirectory = (dir, preFinalizeCallback = null) ->
```
CameronDiver: nice :+1:

lib/utils/compose.d.ts
```diff
@@ -29,4 +30,7 @@ export function loadProject(
 	image?: string,
 ): Bluebird<ComposeProject>;
 
-export function tarDirectory(source: string): Promise<Stream.Readable>;
+export function tarDirectory(
+	source: string,
+	preFinalizeCallback?: { (pack: Pack): void },
```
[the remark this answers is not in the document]
CameronDiver: Generally we would specify it:
```
preFinalizeCallback?: (pack: Pack) => void,
```

lib/utils/registry-secrets.ts
```diff
@@ -0,0 +1,65 @@
...
+
+	/**
+	 * Call JSON.parse() on the given string, then validate the result against
+	 * the registry secrets schema.
+	 * @param json String containing a JSON representation of registry secrets
+	 */
+	public parseRegistrySecrets(json: string): RegistrySecrets {
+		return this.validateRegistrySecrets(JSON.parse(json));
+	}
+}
```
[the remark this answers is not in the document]
CameronDiver: Replied on the builder PR, but for the sake of continuity, yes I think that's a good idea.

lib/utils/remote-build.ts
```diff
@@ -210,6 +213,28 @@ async function cancelBuildIfNecessary(build: RemoteBuild): Promise<void> {
 	}
 }
 
+/**
```
CameronDiver: I'm thinking that this doesn't necessarily only apply to remote-builds, right? Especially if it's added to `resin-multibuild`, we can get this on local mode devices for free. I definitely think that we should move this to a less specific location and use it in both remote-build and local-build.

lib/utils/remote-build.ts
```diff
@@ -210,6 +213,28 @@ async function cancelBuildIfNecessary(build: RemoteBuild): Promise<void> {
...
+ * contains usernames and passwords to private docker registries. The builder
+ * will remove the file from the tar stream and use the secrets to pull base
+ * images from users' private registries.
+ * @param registrySecrets JS object containing registry usernames and passwords
+ * @returns A callback function, or undefined if registrySecrets is empty
+ */
+function getTarStreamCallbackForRegistrySecrets(
+	registrySecrets: RegistrySecrets,
+): { (pack: Pack): void } | undefined {
+	if (Object.keys(registrySecrets).length) {
```
CameronDiver: Can we get a test against `> 0` please? I know it's pretty much equivalent, but it makes it easier to parse quickly.

lib/utils/remote-build.ts
```diff
@@ -210,6 +213,28 @@ async function cancelBuildIfNecessary(build: RemoteBuild): Promise<void> {
...
+ * and uses it to add the '.balena/registry-secrets.json' metadata file that
+ * contains usernames and passwords to private docker registries. The builder
+ * will remove the file from the tar stream and use the secrets to pull base
+ * images from users' private registries.
+ * @param registrySecrets JS object containing registry usernames and passwords
+ * @returns A callback function, or undefined if registrySecrets is empty
+ */
+function getTarStreamCallbackForRegistrySecrets(
+	registrySecrets: RegistrySecrets,
+): { (pack: Pack): void } | undefined {
```
CameronDiver: Different lambda type style

lib/utils/remote-build.ts
```diff
@@ -218,19 +243,24 @@ async function getRequestStream(build: RemoteBuild): Promise<Stream.Duplex> {
...
+		getTarStreamCallbackForRegistrySecrets(build.opts.registrySecrets),
+	);
 	tarSpinner.stop();
 
+	const url = await getBuilderEndpoint(
+		build.baseUrl,
+		build.owner,
+		build.app,
+		build.opts,
+	);
```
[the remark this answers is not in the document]
CameronDiver: The builder on balena at least is always https. This is not always true for openBalena and on-prem - I'd consider adding a check, and a warning if they're sending data plaintext.

lib/actions/push.ts
```diff
@@ -98,30 +113,53 @@ async function getAppOwner(sdk: BalenaSDK, appName: string) {
...
-	},
-	{
-		source: string;
-		emulated: boolean;
-		nocache: boolean;
+function parseRegistrySecrets(secretsFilename: string): RegistrySecrets {
+	let registrySecrets: RegistrySecrets = {};
+	if (secretsFilename) {
+		let isYaml = false;
+		if (/.+\.ya?ml$/i.test(secretsFilename)) isYaml = true;
```
hedss: Don't use single line conditionals. This should be formatted:
```typescript
if (<yaml>) {
  isYaml = true;
} else if (<!jsonFile>) {
  ...
}
```

lib/actions/push.ts
```diff
@@ -98,30 +113,53 @@ async function getAppOwner(sdk: BalenaSDK, appName: string) {
...
+		else if (!/.+\.json$/i.test(secretsFilename))
+			throw new Error(
+				'The registry-secrets filename must end with .json, .yml or .yaml',
+			);
+
+		const raw = require('fs')
+			.readFileSync(secretsFilename)
+			.toString();
+		registrySecrets = new RegistrySecretValidator().validateRegistrySecrets(
+			isYaml ? require('js-yaml').safeLoad(raw) : JSON.parse(raw),
```
hedss: Both `safeLoad` and `parse` can potentially throw. I'm guessing if this is the case we're just letting it bubble up to the top and stop the command, as I can't see any `try/catch` block where this is called from (assuming `await` and not Promises). Same applies for other instances of these methods.

lib/actions/push.ts
```diff
@@ -154,6 +192,13 @@ export const push: CommandDefinition<
 			description: "Don't use cache when building this project",
 			boolean: true,
 		},
+		{
+			signature: 'registry-secrets',
+			alias: 'R',
+			parameter: 'secrets.yml|.json',
+			description: stripIndent`
+				Path to a local YAML or JSON file containing Docker registry passwords used to pull base images`,
```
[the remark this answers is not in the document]
hedss: I personally feel putting it in the help text/docs is enough.

lib/utils/registry-secrets.ts
```diff
@@ -0,0 +1,65 @@
+import * as ajv from 'ajv';
```
hedss: This should include the Balena boilerplate (technically it all should, but I'm aware the other files you're just appending to).

lib/utils/registry-secrets.ts
```diff
@@ -0,0 +1,65 @@
...
+				properties: {
+					username: { type: 'string' },
+					password: { type: 'string' },
+				},
+				additionalProperties: false,
+			},
+		},
+		additionalProperties: false,
+	};
+	private _ajv: ajv.Ajv;
```
hedss: @CameronDiver @Page- Any idea when we started allowing leading underscores? I thought we were avoiding them for private members and just using them for unused function args (certainly I've always tried to follow this based on the MS TS guidelines).

CameronDiver: It seems I replied this on the builder PR instead of this one, but yeah we don't use leading underscores.

lib/actions/push.ts
pdcastro: Agreed. I've ended up moving the test to the caller, so the function must be called with a truthy filename.

lib/actions/push.ts
pdcastro: Done as suggested, thanks. In this particular case it wasn't much of an advantage -- I think the CLI has nothing else to do while waiting for the file to be read asynchronously. I imagine that the async (hidden promise) handling makes the code run marginally slower.

lib/utils/remote-build.ts
pdcastro: Done

lib/utils/remote-build.ts
pdcastro: Done

package.json
```diff
@@ -112,7 +112,7 @@
     "bash": "0.0.1",
     "bluebird": "^3.3.3",
     "body-parser": "^1.14.1",
-    "capitano": "^1.7.0",
+    "capitano": "^1.8.2",
```
pdcastro: Version bumped to take advantage of a bug fix: https://github.com/balena-io/capitano/pull/68

lib/utils/registry-secrets.ts
pdcastro: Renamed.

lib/utils/registry-secrets.ts
pdcastro: Added, thanks.

package.json
pdcastro: Version bumped to take advantage of a bug fix: https://github.com/balena-io/capitano/pull/68

pdcastro:
@CameronDiver, I think you were rebasing some work on this PR and had asked me not rebase or squash commits before checking with you. Is it OK for me to squash them now and rebase on master?

lib/actions/config.coffee
```diff
@@ -306,7 +306,7 @@ exports.generate =
 			'''
 
 		if !options.application and options.deviceType
-			patterns.exitWithExpectedError '''
+			exitWithExpectedError '''
```
pdcastro: Trivial bug fix, unrelated to this PR.

lib/utils/remote-build.ts
pdcastro: You mean for example the following command, right?
```
balena push 192.168.2.3 --registry-secrets sample_secrets.json -s secrets_app
```
I was looking at the implementation of push-to-IP-address and, if I got it right, the CLI opens a connection to the docker daemon (balena engine) of a local-mode device and uploads the tar stream directly to the daemon. In this case, I think we don't want a `.balena` metadata folder to be inserted in the stream. Instead, it might be enough for the CLI itself to add the `registryconfig` option to Dockerode -- which is what the cloud builder does _after_ extracting the `.balena` metadata folder from the tar stream. If this is right, then actually this `getTarStreamCallbackForRegistrySecrets()` function applies only to remote builds.

lib/utils/device/deploy.ts
```diff
@@ -233,6 +254,7 @@ async function assignDockerBuildOpts(
 				'io.resin.local.image': '1',
 				'io.resin.local.service': task.serviceName,
 			},
+			registryconfig: opts.registrySecrets,
```
pdcastro: This is the key change that implements the local-mode push:
```
balena push 192.168.2.3 --registry-secrets sample_secrets.json -s secrets_app
```

lib/utils/device/deploy.ts
```diff
@@ -301,6 +323,6 @@ async function inspectBuildResults(images: LocalImage[]): Promise<void> {
 	});
 
 	if (failures.length > 0) {
-		exitWithExpectedError(new LocalPushErrors.BuildError(failures));
+		exitWithExpectedError(new LocalPushErrors.BuildError(failures).toString());
```
pdcastro: Bug fix. Without this change, the message printed was a rather unelaborated "Build error":
```
[Build]   [main] Step 1/5 : FROM 192.168.2.1:5000/busybox
Build error
```
After this change it is a bit better:
```
[Build]   [main] Step 1/5 : FROM 192.168.2.1:5000/busybox
Some services failed to build:
	main: Get https://192.168.2.1:5000/v2/: http: server gave HTTP response to HTTPS client
```

lib/utils/registry-secrets.ts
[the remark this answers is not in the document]
pdcastro: Done

lib/utils/remote-build.ts
pdcastro: I've now put together a private registry with a self-signed SSL certificate for testing and I can confirm that pushing to a device IP address works with the implementation as it is!
(The cloud builder is OK with pulling from a private registry over plain HTTP, but the balena-engine daemon on a device insists on HTTPS.)

lib/actions/push.ts
hedss: Ug, this is part of the reason I don't like `prettier` :(

lib/actions/push.ts
hedss: Yep, this is usually the reason I catch stuff and then 'translate' it into something that's obvious. :D

Issue #866, opened 2018-04-18
Add resource tag methods


thgreasi:
Just adding this so we can track the respective PR.
HQ: https://github.com/resin-io/hq/issues/150
HQ: https://github.com/resin-io/hq/pull/281
Depends-On: https://github.com/resin-io/resin-sdk/issues/491

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_b3uj)

PR #1046, opened 2018-12-07
actions: Add resource tag operations

thgreasi:
Resolves: #866
HQ: https://github.com/resin-io/hq/issues/150
HQ: https://github.com/resin-io/hq/pull/281
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -106,6 +106,12 @@ environment variable (in the same standard URL format).
 	- [env add &#60;key&#62; [value]](#env-add-key-value-)
 	- [env rename &#60;id&#62; &#60;value&#62;](#env-rename-id-value-)
 
+- Tags
+
+	- [tags](#tags)
+	- [tag set &#60;tagKey&#62; [value]](#tag-set-tagkey-value-)
+	- [tag rm &#60;tagKey&#62;](#tag-rm-tagkey-)
+
 - Help
 
 	- [help [command...]](#help-command-)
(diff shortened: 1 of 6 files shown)


doc/cli.markdown
```diff
@@ -672,6 +678,84 @@ Examples:
...
+
+Example:
+
+	$ balena tags --application MyApp
+	$ balena tags --device 7cf02a6
+	$ balena tags --release 1234
+
+### Options
+
+#### --application, -a, --app &#60;application&#62;
```
dfunckt: What is the docs tool we're using? These encoded entities make output pretty much unreadable.

doc/cli.markdown
```diff
@@ -672,6 +678,84 @@ Examples:
...
+release id
+
+## tag set &#60;tagKey&#62; [value]
+
+Use this command to set a tag to an application, device or release.
+
+Examples:
+
+	$ balena tag set beta --application MyApp
+	$ balena tag set status tested --application MyApp
```
shaunmulligan: @thgreasi is this "key" and "value", from the docs its unclear if this is setting key value pairs or multiple keys. We should make it more clear that this is to set the key value pair.

thgreasi: I guess that all my brain needed to see was `tag set <tagKey> [value]`, but makes sense @shaunmulligan :+1: . Pushing a commit in a moment.

Issue #1013, opened 2018-11-05
`os configure` should allow specifying a device type


pcarranzav:
Otherwise, the command will always specify the device type from the application (or uuid, if it's a specific device - but I'm referring to the `--app` case).

Since applications can now have multiple device types, a user could, for instance, want to provision a Fin on a Raspberry Pi 3 app, and they currently can't do it using `os configure` because the resulting deviceType will always be raspberrypi3.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_fjyz)

PR #1047, opened 2018-12-07
Add explicit device type option to `os configure` & `config generate`

thgreasi:
Resolves: #1013
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/lib/actions/config.coffee
+++ b/lib/actions/config.coffee
@@ -229,6 +229,9 @@ exports.generate =
 		by specifying an option for each question on the command line, if you know the questions
 		that will be asked for the relevant device type.
 
+		In case that you want to configure an image for an application with mixed device types,
+		you can pass the --device-type argument along with --app to specify the target device type.
+
 		Examples:
 
 			$ balena config generate --device 7cf02a6 --version 2.12.7
@@ -236,6 +239,7 @@ exports.generate =
 			$ balena config generate --device 7cf02a6 --version 2.12.7 --device-api-key <existingDeviceKey>
 			$ balena config generate --device 7cf02a6 --version 2.12.7 --output config.json
 			$ balena config generate --app MyApp --version 2.12.7
+			$ balena config generate --app MyApp --version 2.12.7 --device-type fincm3
 			$ balena config generate --app MyApp --version 2.12.7 --output config.json
 			$ balena config generate --app MyApp --version 2.12.7 \
 				--network wifi --wifiSsid mySsid --wifiKey abcdefgh --appUpdatePollInterval 1
@@ -245,6 +249,7 @@ exports.generate =
 		commandOptions.optionalApplication
 		commandOptions.optionalDevice
 		commandOptions.optionalDeviceApiKey
+		commandOptions.optionalDeviceType
 		{
 			signature: 'generate-device-api-key'
 			description: 'generate a fresh device key for the device'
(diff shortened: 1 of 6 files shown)


lib/actions/config.coffee
```diff
@@ -299,12 +304,26 @@ exports.generate =
...
+			'''
+
 		Promise.try ->
 			if options.device?
 				return balena.models.device.get(options.device)
 			return balena.models.application.get(options.application)
 		.then (resource) ->
-			balena.models.device.getManifestBySlug(resource.device_type)
+			deviceType = options.deviceType || resource.device_type
+			balena.models.device.getManifestBySlug(deviceType)
```
dfunckt: Ideally you'd validate that the given device type is of the same arch as the application's. This is actually the definition of mixed device-type apps -- i.e. different device-types but common architecture. Same applies to the other commands touched by this PR

thgreasi: Oh yea that totally makes sense.
This should probably also be in place in the `os configure` action as well I guess.
The API doesn't have such checks atm and the UI & SDK try to enforce this rule on their relevant methods.

Will try to add this as part of this PR, to avoid having users in hard to debug situations.

lib/utils/helpers.ts
```diff
@@ -92,6 +92,13 @@ export async function getManifest(
 	return balena.models.device.getManifestBySlug(deviceType);
 }
 
+export const areDeviceTypesCompatible = (
+	deviceTypeA: BalenaSdk.DeviceType,
+	deviceTypeB: BalenaSdk.DeviceType,
+) =>
+	deviceTypeA.arch === deviceTypeB.arch &&
+	!!deviceTypeA.isDependent === !!deviceTypeB.isDependent;
```
dfunckt: Why do you need to check for "isDependent"?

thgreasi: Because you can't/shouldn't configure a dependent device type for a top level app, even if they are of the same arch.

dfunckt: Then the check you need should be `!deviceTypeA.isDependent === !deviceTypeB.isDependent` right?

thgreasi: The typings have it as nullable, and it's sometimes missing depending on the config.json on s3, so I decided it to normalize the boolean values, like we ended up doing in the UI.

dfunckt: `!!deviceTypeA.isDependent === !!deviceTypeB.isDependent` indeed ensures you'll compare 2 booleans but also means that when both are `true` the function will return `true` (i.e. "yes, the device types are compatible") which isn't what you describe about dependent device types above.

thgreasi: Oh, I just realized that you are just skipping the duplicate inverse checks :)
I would say let's keep it as is since `!!x` is an easier patter to spot conversion to boolean, and js engines optimize that anyway.

dfunckt: That's not my point :) I'm saying that with `!!` this function will return "true" both when the device types are of dependent devices *and* when they're not, but if I understand your comment above correctly the latter shouldn't be allowed, in which case you need to ensure *both* are *not dependent*. Makes sense?

thgreasi: Discussed on FD.
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/7rPD_QLlNzK9uobomhHPgruDgba

PR #1050, opened 2018-12-13
Fix deploy action on node 6

zvin:
Change-type: patch

### DIFF
--- a/lib/utils/patterns.ts
+++ b/lib/utils/patterns.ts
@@ -283,5 +283,7 @@ export function exitWithExpectedError(message: string | Error): never {
 	}
 
 	printErrorMessage(message);
-	return process.exit(1);
+	process.exit(1);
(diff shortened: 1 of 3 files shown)


lib/utils/patterns.ts
```diff
@@ -283,5 +283,6 @@ export function exitWithExpectedError(message: string | Error): never {
 	}
 
 	printErrorMessage(message);
-	return process.exit(1);
+	process.exit(1);
+	throw new Error('exit'); // make tsc happy
```
thgreasi: I didn't get what was the reason for this, but after a second read I realized that the node 6 typings didn't use `never` as a return type. Maybe we can explain that in the comment.

thgreasi:
Actually this could be better titled as 
> Fix deploy action on node 6

Since downgrading is just a side-action to catch all node 8+ interfaces that we were using.

PR #1052, opened 2018-12-20
Update "local flash" and "util available-drives" actions

zvin:
* switch from coffeescript to typescript;
 * replace etcher-image-write and drivelist with etcher-sdk


zvin:
@pdcastro this PR does not add any new functionality.
Not sure about the tests, non of the actions is tested in balena-cli.
What would we test? Everything would be mocked.

pdcastro:
> _Not sure about the tests, none of the actions is tested in balena-cli.
> What would we test? Everything would be mocked._

@zvin, I suppose the change from Coffeescript to Typescript in itself improves confidence, as at least it catches typo and type/interface bugs. Thanks! But if unit tests would be redundant, maybe we could add an end-to-end flashing test? Perhaps we could mount a file as a disk (block device), then maybe that "virtual disk" could be "flashed"?
* [Linux loop device](https://en.wikipedia.org/wiki/Loop_device) - `losetup /dev/loop0 example.img`
* [Windows Mount-DiskImage](https://docs.microsoft.com/en-us/powershell/module/storage/mount-diskimage?view=win10-ps)
* [MacOS hdiutil](https://ss64.com/osx/hdiutil.html) - `hdiutil attach /path/to/diskimage.dmg`

So the test might be to mount a file as a disk, run `balena local flash` in a child process to flash that disk, and then assert that the disk contents match what we expect. I'm not sure how to do this separately for each platform (Windows / Mac / Linux), but there's probably a way... Appveyor runs Windows, right? I've found this [CircleCI page about MacOS testing](https://circleci.com/build-environments/xcode/osx/), but I have never tried it.

Is it overkill to write such end to end tests for the CLI for each platform? The CLI is of course an essential tool for many customers, and I have been assigned as its maintainer. I start from the optimistic assumption that all the existing CLI functionality works fine on all platforms (apart from the 167 open issues on GitHub), but after this pull request, how do I know that image flashing still works? :-) If you think that manual testing is enough for this PR, then please reassure me that you have manually tested it in each platform...

zvin:
@pdcastro 
`etcher-sdk` already has tests for this.
balena-cli is only a frontend here.

I agree e2e tests would be nice but it's a bit out of the scope of this PR.
I have only tested on Linux so far but to reassure you I'll test on Windows and macOS asap (tomorrow most probably).

pdcastro:
@zvin, cool, thank you. I was also planning to test this PR manually but couldn't find the time yet (I've been working on private registry authentication). This PR is of course a nice welcome improvement and we want to merge it as soon as possible. I never meant to delay it, and multi-platform testing will be a thorny matter for any substantial CLI changes. I'm not sure of the right approach/answer and maybe I'm being overzealous, but a principle that I generally follow is "if it's not tested, it's broken". :-)

zvin:
@pdcastro I just finished testing this on win32 and osx. It works fine for me. Sorry for the delay, I encountered unexpected issues with the latest xcode on osx.

PR #1055, opened 2018-12-27
Adding information about dependency installation

pcgeek86:
Added documentation about the dependencies required to build balena-cli

Moved from https://github.com/balena-io/docs/pull/995

Resolves: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -6,6 +6,23 @@ Please make sure your system meets the requirements as specified in the [README]
 
 ## Install the CLI
 
+### Dependencies
+
+Before installing the Balena CLI from npm, make sure you have the following dependencies installed:
+
+* make
+* g++ compiler
+* Python 2.7
+* git
+
+For example, to install these packages on a Debian-based Linux operating systems:
+
+```
+$ sudo apt-get install g++ make python git --yes
+```
+
+**NOTE**: If you are installing the stand-alone binary CLI, you will not need to install these dependencies.
+
 ### Npm install
 
 The best supported way to install the CLI is from npm:


pcgeek86:
@shaunmulligan Thanks, how do we get these automated checks to pass? Are there any issues we need to resolve elsewhere?

shaunmulligan:
@pcgeek86 I believe its failing because of the commit format, you need it to match something like this: https://github.com/balena-io/balena-cli/commit/0f302d30eccfaa9ea72f1c114c615794b3ba26dc with:
```
fix: blabla 

Change-type: minor

pcgeek86:
@shaunmulligan I amended the commit message using the template you provided. Let me know if this looks okay.

lurch:
I noticed that the Travis builds for this PR were failing. Out of curiosity (and boredom :wink: ) I decided to dig into it, and _eventually_ figured out that the `doc/cli.markdown` that this PR modifies is actually an output file, and I think the file you **should** have edited is `capitanodoc.ts` ? (See https://github.com/balena-io/balena-cli/blob/master/package.json#L38 )

Suggestion: Perhaps capitanodoc (whatever that is) could automatically insert a comment into the markdown files it creates, warning that they shouldn't be edited directly? (i.e. similar to https://github.com/balena-io-modules/drivelist/pull/164#pullrequestreview-37358006 and https://raw.githubusercontent.com/balena-io-modules/drivelist/master/README.md )

pcgeek86:
@lurch Thanks for digging into this. I'm learning about this repo myself. Based on a little more research:

- [capitano](https://www.npmjs.com/package/capitano) is an npm dependency (package.json) [for this project](https://github.com/balena-io/balena-cli/blob/master/package.json), which offers an interface to build command line tools (similar to commander.js)
- It would appear that [capitanodoc.ts](https://github.com/balena-io/balena-cli/blob/master/capitanodoc.ts) is a file containing documentation for the Balena CLI tool
- This file is used in the `npm run build:doc` command that you referenced in `package.json`

Looks like you are spot on ... I need to go back and update `capitanodoc.ts` instead. Cheers for the help.

Issue #1045, opened 2018-12-06
Improve UX with `balena deploy` when pushing an identical image


xginn8:
Create a new release object rather than re-uploading the image entirely

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_fuxf)

dfunckt:
To flesh this out some more, the CLI does create a new release but *also* re-uploads the complete image, and that is the issue here really. #1057 improved this so repeated invocations of `balena deploy` benefit from cross-repo mounts and no blobs are uploaded.

What still does not work however, is doing the equivalent of `balena deploy; docker rmi <built image>; balena deploy`. In this case the full image is pushed again, because of the requirement for the engine to explicitly instruct the registry which repo it should mount layers from -- it simply doesn't know of the previous images because they've been removed locally.

The first case is typically the workflow of a developer doing work on their laptop and deploying. The latter may be the case if the CLI is naively integrated in a CI pipeline, where there's build, deploy and finally cleanup.

To fix the latter case, `balena deploy` should pull the previous release images first thing, which can also serve as a build cache, like on the cloud builder. This will add a tiny bit of redundant overhead in the first case (for the CLI to get a token and pull the image, which should result in zero bytes downloaded), which is acceptable.

cc @pdcastro @srlowe

pdcastro:
Closing this issue as I understand that the feature was delivered by @xginn8 's PR #1057 of Dec 2018 - CLI v9.9.0. Furthermore, PR #1848 proved to be of no additional benefit.

PR #1057, opened 2018-12-28
Request access to previously pushed release via `balena deploy`

xginn8:
This access is used to cross mount the old layers, rather than
reuploading the layers each time.

Connects-to: #1045
Change-type: minor
Signed-off-by: Matthew McGinn <mamcgi@gmail.com>

- [x] Affects the development, build or deployment processes of the component

### DIFF
--- a/lib/utils/compose.coffee
+++ b/lib/utils/compose.coffee
@@ -349,13 +349,54 @@ tagServiceImages = (docker, images, serviceImages) ->
 			logs: d.logs
 			props: d.props
 
-authorizePush = (tokenAuthEndpoint, registry, images) ->
+
+getPreviousRepos = (sdk, docker, logger, appID) ->
+	sdk.pine.get(
+		resource: 'release'
+		options:
+			$filter:
+				belongs_to__application: appID
+				status: 'success'
+			$select:
+				[ 'id' ]
+			$expand:
+				contains__image:
+					$expand: 'image'
+			$orderby: 'id desc'
+			$top: 1
+	)
+	.then (release) ->
+		# grab all images from the latest release, return all image locations in the registry
+		if release?.length > 0
+			images = release[0].contains__image
+			Promise.map images, (d) ->
+				imageName = d.image[0].is_stored_at__image_location
+				docker.getRegistryAndName(imageName)
+				.then ( registry ) ->
+					logger.logDebug("Requesting access to previously pushed image repo (#{registry.imageName})")
+					return registry.imageName
+	.catch (e) ->
+		logger.logDebug("Failed to access previously pushed image repo: #{e}")
+
+authorizePush = (sdk, logger, tokenAuthEndpoint, registry, images, previousRepos) ->
 	_ = require('lodash')
-	sdk = require('balena-sdk').fromSharedOptions()
+
+	# TODO: https://github.com/balena-io/balena-cli/issues/1070
+	maxRepos = 20
 
 	if not _.isArray(images)
 		images = [ images ]
 
+	if images.length > maxRepos
+		throw new Error (
+			"More than #{maxRepos} containers is currently not supported, see " +
+			'https://github.com/balena-io/balena-cli/issues/1070 for more information'
+		)
+	images.push previousRepos...
+	if images.length + previousRepos?.length > maxRepos
+		logger.logDebug("Truncating requested repositories to #{maxRepos} by limiting previously pushed repo access")
+		# at this point, we know we're only truncating access to previously pushed repos
+		images = images[0...maxRepos]
 	sdk.request.send
 		baseUrl: tokenAuthEndpoint
 		url: '/auth/v1/token'
@@ -423,7 +464,10 @@ exports.deployProject = (
 		tagServiceImages(docker, images, serviceImages)
 		.tap (images) ->
 			logger.logDebug('Authorizing push...')
-			authorizePush(apiEndpoint, images[0].registry, _.map(images, 'repo'))
+			sdk = require('balena-sdk').fromSharedOptions()
+			getPreviousRepos(sdk, docker, logger, appId)
+			.then (previousRepos) ->
+				authorizePush(sdk, logger, apiEndpoint, images[0].registry, _.map(images, 'repo'), previousRepos)
 			.then (token) ->
 				logger.logInfo('Pushing images to registry...')
 				pushAndUpdateServiceImages docker, token, images, (serviceImage) ->


lib/utils/compose.coffee
```diff
@@ -349,13 +349,40 @@ tagServiceImages = (docker, images, serviceImages) ->
...
+					$expand: 'image'
+	)
+
+	.then (release) ->
+            # grab all images from the latest release, return all image locations in the registry
+		images = release[0].contains__image
+		repos = []
+		for i in images
+			imageName = i.image[0].is_stored_at__image_location
+			[ _, _, repoName, _ = 'latest' ] = /(.*?)\/(.*?)(?::([^/]*))?$/.exec(imageName)
```
pcarranzav: Might be better to use a tighter regex, as in https://github.com/balena-io-modules/docker-toolbelt/blob/master/lib/docker-toolbelt.coffee#L338 (you could even use that docker-toolbelt module)

xginn8: Admittedly I stole this regex from https://github.com/balena-io/balena-cli/blob/master/lib/utils/compose.coffee#L338, but I think making use of docker-toolbelt sounds like a much better long-term plan. Will update that, thank you!

xginn8:
I've restructured this PR to use `docker-toolbelt` and a Promise.map rather than a for loop, PTAL!

xginn8:
I believe the build failures in travis are related to https://github.com/balena-io/balena-cli/pull/1059

xginn8:
@pdcastro or @thgreasi, mind taking a look at this PR? I'm relatively new to promises and would love a second look.

lib/utils/compose.coffee
```diff
@@ -349,13 +349,42 @@ tagServiceImages = (docker, images, serviceImages) ->
 			logs: d.logs
 			props: d.props
 
-authorizePush = (tokenAuthEndpoint, registry, images) ->
+
+getPreviousRepos = (logger, appID) ->
+	sdk = require('balena-sdk').fromSharedOptions()
```
pdcastro: Looking at the implementation of `fromSharedOptions()`, it seems a non-trivial amount setting up, unclear if actually establishing network connections. It might be better to reuse the same sdk object between `getPreviousRepos` and `authorizePush`, but moving its creation to the common parent function (`deployProject`) and passing it down as argument.

lib/utils/compose.coffee
```diff
@@ -349,13 +349,42 @@ tagServiceImages = (docker, images, serviceImages) ->
...
+		# grab all images from the latest release, return all image locations in the registry
+		if release.length > 0
+			images = release[0].contains__image
+			Promise.map images, (d) ->
+				imageName = d.image[0].is_stored_at__image_location
+				dockerToolbelt.getRegistryAndName(imageName)
+				.then ( registry ) ->
+					logger.logDebug("Requesting access to previously pushed image repo (#{registry.imageName})")
+					return registry.imageName
+		return
```
pdcastro: NIT: redundant `return` statement

lib/utils/compose.coffee
```diff
@@ -423,18 +452,20 @@ exports.deployProject = (
...
+					logger.logInfo('Pushing images to registry...')
+					pushAndUpdateServiceImages docker, token, images, (serviceImage) ->
+						logger.logDebug("Saving image #{serviceImage.is_stored_at__image_location}")
+						if skipLogUpload
+							delete serviceImage.build_log
+						releaseMod.updateImage(client, serviceImage.id, serviceImage)
+				.finally ->
+					logger.logDebug('Untagging images...')
+					Promise.map images, ({ localImage }) ->
+						localImage.remove()
```
pdcastro: This `.finally` block seems to have the purpose of undoing the earlier `tagServiceImages` call, "no matter what". It needs to be indented one level less (to the left) in order to cover the case of `getPreviousRepos` failing (rejected promise). Actually, I think the whole block from `.then (token)` to `.finally` could/should be indented one level less:
```coffeescript
			getPreviousRepos(logger, appId)
			.then (previousRepos) ->
				authorizePush(apiEndpoint, images[0].registry, _.map(images, 'repo'), previousRepos)
			.then (token) ->
				logger.logInfo('Pushing images to registry...')
				pushAndUpdateServiceImages docker, token, images, (serviceImage) ->
					logger.logDebug("Saving image #{serviceImage.is_stored_at__image_location}")
					if skipLogUpload
						delete serviceImage.build_log
					releaseMod.updateImage(client, serviceImage.id, serviceImage)
			.finally ->
				logger.logDebug('Untagging images...')
				Promise.map images, ({ localImage }) ->
					localImage.remove()
```

pdcastro:
> _I believe the build failures in travis are related to #1059_

@xginn8, I was told before that we can/should ignore tests that don't have the REQUIRED label next to them. The travis tests don't have the REQUIRED label, so to my trained brain it's like they don't exist. :-)

lib/utils/compose.coffee
```diff
@@ -349,13 +349,42 @@ tagServiceImages = (docker, images, serviceImages) ->
 			logs: d.logs
 			props: d.props
 
-authorizePush = (tokenAuthEndpoint, registry, images) ->
+
+getPreviousRepos = (logger, appID) ->
+	sdk = require('balena-sdk').fromSharedOptions()
+	DockerToolbelt = require('docker-toolbelt')
+	dockerToolbelt = new DockerToolbelt()
+	sdk.pine.get(
```
dfunckt: This request pulls all releases on the app which can be fairly expensive — it could be hundreds of rows. Might be better to sort by id descending and limit to the latest 10 or so results. The Pine options you’re looking for to do this are iirc “$order” and “$top”.

dfunckt: You also don’t need any fields on the release so you could also limit what is fetched with “$select: [ “id” ]”

xginn8:
I am still sorting the case where pushing more than 10 images as part of a docker-compose causes the json to be truncated.

lib/utils/compose.coffee
xginn8: Nice, thanks!

lib/utils/compose.coffee
xginn8: Good catch, thanks!

lib/utils/compose.coffee
```diff
@@ -349,13 +349,42 @@ tagServiceImages = (docker, images, serviceImages) ->
 			logs: d.logs
 			props: d.props
 
-authorizePush = (tokenAuthEndpoint, registry, images) ->
+
+getPreviousRepos = (logger, appID) ->
+	sdk = require('balena-sdk').fromSharedOptions()
+	DockerToolbelt = require('docker-toolbelt')
+	dockerToolbelt = new DockerToolbelt()
```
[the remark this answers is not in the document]
xginn8: Yep, this reuse makes sense to me!

PR #1071, opened 2019-01-11
Improve startup time by adding fast-boot

shaunmulligan:
Change-type: minor
Signed-off-by: Shaun Mulligan <shaun@balena.io>

This seems to halve the time for executing the cli calls. I only tested `balena apps` and `balena whoami` without this change those both take ~3.5 seconds to run, with fast-boot, they take about ~1.3


package.json
```diff
@@ -131,6 +131,7 @@
     "etcher-image-write": "^9.0.3",
     "event-stream": "3.3.4",
     "express": "^4.13.3",
+    "fast-boot": "^1.0.7",
```
Page-:
```suggestion
    "fast-boot2": "^1.0.7",
```

lib/actions/index.coffee
```diff
@@ -14,6 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+
+# Use fast-boot to cache require lookups, speeding up startup
+require('fast-boot2').start({
+	cacheFile: '.fast-boot.json'
```
CameronDiver: Is this added in the directory where the cli is executed? I feel like it should be in the install location, or the home directory or something.

Page-:
@pdcastro for me it drops the time for `balena help` from ~1800ms to ~1400ms - the savings are proportional to the amount of `require` calls you have.  Also what's most noticeable is that it affects the time to first response, which should change the numbers from ~1400ms to ~1000ms (from other tests I did previously there was ~400ms spent on things that happen after the output), overall it's definitely very noticeable on my machine which has an SSD, for a HDD it'll be massive

lib/actions/index.coffee
Page-: Changed it to use __dirname to be consistently in the same place as the binary

PR #1076, opened 2019-01-14
Add livepush to balena push

CameronDiver:
Depends-on: https://github.com/balena-io/balena-supervisor/pull/828

---
##### Contributor checklist

### DIFF
--- a/lib/actions/push.ts
+++ b/lib/actions/push.ts
@@ -114,8 +114,17 @@ export const push: CommandDefinition<
 	description:
 		'Start a remote build on the balena cloud build servers or a local mode device',
 	help: stripIndent`
-		This command can be used to start an image build on the remote balenaCloud build
-		servers, or on a local-mode balena device.
+		This command can be used to start a build on the remote balena cloud builders,
+		or a local mode balena device.
+
+		When building on the balena cloud the given source directory will be sent to the
+		balena builder, and the build will proceed. This can be used as a drop-in
+		replacement for git push to deploy.
+
+		When building on a local mode device, the given source directory will be built
+		on the device, and the resulting containers will be run on the device. Logs will
+		be streamed back from the device as part of the same invocation. This requires
+		a device with a supervisor version of at least v7.21.0.
 
 		When building on the balenaCloud servers, the given source directory will be
 		sent to the remote server. This can be used as a drop-in replacement for the
--- a/lib/actions/push.ts
+++ b/lib/actions/push.ts
@@ -106,6 +106,7 @@ export const push: CommandDefinition<
 		emulated: boolean;
 		nocache: boolean;
 		'registry-secrets': string;
+		live: boolean;
 	}
 > = {
 	signature: 'push <applicationOrDevice>',
@@ -165,6 +166,14 @@ export const push: CommandDefinition<
 			description: stripIndent`
 				Path to a local YAML or JSON file containing Docker registry passwords used to pull base images`,
 		},
+		{
+			signature: 'live',
+			alias: 'l',
+			boolean: true,
+			description: stripIndent`
+				Start a live session after the push, which will wait for code changes, and synchronise them with
+				running containers. This mode is only valid when pushing to a local device.`,
+		},
 	],
 	async action(params, options, done) {
 		const sdk = (await import('balena-sdk')).fromSharedOptions();
@@ -194,6 +203,13 @@ export const push: CommandDefinition<
 		const buildTarget = getBuildTarget(appOrDevice);
 		switch (buildTarget) {
 			case BuildTarget.Cloud:
+				// Ensure that the live argument has not been passed to a cloud build
+				if (options.live) {
+					exitWithExpectedError(
+						'The --live flag is only valid when pushing to a local device.',
+					);
+				}
+
 				const app = appOrDevice;
 				await exitIfNotLoggedIn();
 				await Bluebird.join(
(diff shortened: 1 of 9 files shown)


lib/actions/push.ts
```diff
@@ -166,6 +176,14 @@ export const push: CommandDefinition<
 			description: stripIndent`
 				Path to a local YAML or JSON file containing Docker registry passwords used to pull base images`,
 		},
+		{
+			signature: 'live',
+			alias: 'l',
+			boolean: true,
+			description: stripIndent`
+				Start a live session after the push, which will wait for code changes, and synchronise them with
+				running containers. This mode is only valid when pushing to a local device.`,
```
pdcastro: I suggest a short description here, with a reference, so it formats more nicely on the screen:
```
Start a live session after the push. See live push description above.
```

lib/utils/device/deploy.ts
```diff
@@ -82,9 +85,18 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
...
 	try {
 		const version = await api.getVersion();
+		globalLogger.logDebug(`Checking device version: ${version}`);
 		if (!semver.satisfies(version, '>=7.21.4')) {
 			exitWithExpectedError(versionError);
 		}
+		// FIXME: DO NOT MERGE until this version number has been updated
+		// with the version which the following PR ends up in the supervisor
+		// https://github.com/balena-io/balena-supervisor/pull/828
+		if (opts.live && !semver.satisfies(version, '>=1.0.0')) {
```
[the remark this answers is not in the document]
CameronDiver: It should be, but I'd like to do more testing before locking this in, in case the supervisor needs any more patches.

pdcastro:
Just a note, before I forget: the Travis builds are not labeled 'Required' but they are actually (still) required. They are (still) the ones that create the MacOS and Linux standalone release zips files in the [balena-cli repo's releases page](https://github.com/balena-io/balena-cli/releases). I think the reason why they are currently failing for this PR is that `npm run build` was not run on your laptop to generate the markdown docs before git-pushing. (Yes, maybe it would make sense for the CI to generate that even if `npm run build` wasn't run on the laptop...)

lib/actions/push.ts
```diff
@@ -107,15 +107,25 @@ export const push: CommandDefinition<
...
+		or a local mode balena device.
+
+		When building on the balena cloud the given source directory will be sent to the
+		balena builder, and the build will proceed. This can be used as a drop-in
+		replacement for git push to deploy.
+
+		When building on a local mode device, the given source directory will be built
+		on the device, and the resulting containers will be run on the device. Logs will
+		be streamed back from the device as part of the same invocation. This requires
+		a device with a supervisor version of at least v7.21.0.
```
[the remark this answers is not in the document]
hedss: I'd personally move this description to the `live` signature description below, and move a small two line sentence here that says 'A live session can be started after the push by passing the `--live` flag', or similar. Else we're starting a precedence for any major sub-feature.

lib/utils/device/live.ts
```diff
@@ -0,0 +1,244 @@
...
+	buildContext: string;
+	composition: Composition;
+	buildTasks: BuildTask[];
+	docker: Dockerode;
+	api: DeviceAPI;
+	logger: Logger;
+	buildLogs: BuildLogs;
+}
+
+export class LivepushManager {
```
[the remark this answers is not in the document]
hedss: I think the reason we haven't is that we've tried to port a lot over to TS relatively quickly. The entire codebase could probably do with going over and auditing. But, time...

lib/utils/device/live.ts
```diff
@@ -0,0 +1,244 @@
...
+			`Device state not settled, retrying in ${DEVICE_STATUS_SETTLE_CHECK_INTERVAL}ms`,
+		);
+		await Bluebird.delay(DEVICE_STATUS_SETTLE_CHECK_INTERVAL);
+		await this.awaitDeviceStateSettle();
+	}
+
+	private async handleFSEvent(fsEvent: ContextEvent): Promise<void> {
+		// TODO: If there's a dockerfile event, we must perform a rebuild
+		this.logger.logDebug(
+			`Got an filesystem event for service: ${
```
hedss: `Got a filesystem`

lib/utils/device/live.ts
```diff
@@ -0,0 +1,244 @@
...
+				fsEvent.serviceName
+			}. Event: ${JSON.stringify(fsEvent)}`,
+		);
+
+		let updates: string[] = [];
+		let deletes: string[] = [];
+		switch (fsEvent.type) {
+			case 'add':
```
hedss:
```suggestion
			case 'add':
			case 'change':
				updates = [fsEvent.filename];
				break;
```

typings/global.d.ts
```diff
@@ -0,0 +1,3 @@
+interface Dictionary<T> {
```
hedss: I find this amusing. :)

lib/utils/device/live.ts
```diff
@@ -0,0 +1,409 @@
...
+		process.removeAllListeners('SIGINT');
+		process.on('SIGINT', async () => {
+			this.logger.logLivepush('Cleaning up device...');
+			await Promise.all(
+				_.map(this.containers, container => {
+					container.livepush.cleanupIntermediateContainers();
+				}),
+			);
+
+			process.exit(0);
```
hedss: `process.exitCode = 0` instead maybe?

CameronDiver: We need to actually force an exit, as we caught the interrupt.

hedss: Ah, good point.

PR #1078, opened 2019-01-16
Support compressed images in `balena local flash`

zvin:
Change-type: minor


zvin:
This supports zip archives and raw images bz2, gz or lzma copmpressed.

PR #1082, opened 2019-01-21
actions/ssh: Establish SSH connections directly on OS >= 2.44.0

wrboyce:
Change-type: minor
Signed-off-by: Will Boyce \<will@balena.io\>

---
##### Contributor checklist


richbayliss:
@wrboyce @dfunckt did we make a decision to continue with this one? If so then I will make sure it is reviewed today.

lib/actions/ssh-direct.ts
```diff
@@ -0,0 +1,245 @@
...
+
+const createSSHClient = (socket: stream.Duplex): Bluebird<Ssh2ClientAsync> =>
+	new Bluebird((resolve, reject) => {
+		const client = new ssh2.Client() as Ssh2ClientAsync;
+		return client
+			.on('ready', () => {
+				resolve(client);
+			})
+			.on('error', err => {
+				const errSource = err && err.level ? 'ssh client socket' : 'ssh client';
```
pdcastro: Trust your testing rather than my reading, but my reading of the docs suggests that `err.level` would contain one two values: `'client-socket'` or `'client-ssh'`, rather than being set for the former and unset for the latter, in which case the code doesn't look right.
> error(< Error >err) - An error occurred. A 'level' property indicates 'client-socket' for socket-level errors and 'client-ssh' for SSH disconnection messages. In the case of 'client-ssh' messages, there may be a 'description' property that provides more detail.

lib/actions/ssh-direct.ts
```diff
@@ -0,0 +1,245 @@
...
+				let errMsg = `${errSource} error while initiating SSH connection`;
+				if (err && err.description) {
+					errMsg += `: ${err.description}`;
```
pdcastro: Could the following be ever useful? The docs say that _" In the case of 'client-ssh' messages, there may be a 'description' property that provides more detail"_ (leaving room for other types of messages).
```
if (err && (err.description || err.message)) {
	errMsg += `: ${err.description || err.message}`;
```

lib/actions/ssh-direct.ts
```diff
@@ -0,0 +1,245 @@
...
+				}
+				console.error(errMsg);
+				reject(new Error(errMsg));
+			})
+			.connect({
+				sock: socket,
+				username: 'root',
+				port: 22222,
+				agent: process.env.SSH_AUTH_SOCK,
```
pdcastro: ooh SSH_AUTH_SOCK 👍

lib/actions/ssh-direct.ts
```diff
@@ -0,0 +1,245 @@
...
+	const dsWrite = (data: Buffer) => deviceStream.write(data);
+
+	if (process.stdin != null && process.stdout.isTTY) {
+		if (typeof process.stdin.setRawMode === 'function') {
+			process.stdin.setRawMode(true);
+		}
+
+		// Set initial window size and handle resize events
+		dsResize();
+		process.stdout.on('resize', dsResize);
```
pdcastro: So we need to deal with window resizing in our code... :-/ Not sure if a blessing or a curse:
[Broken line wrapping / cursor behavior with balena ssh](https://github.com/balena-io/balena-cli/blob/master/TROUBLESHOOTING.md#broken-line-wrapping--cursor-behavior-with-balena-ssh)

I guess a new troubleshooting item will be: _"try the `-x` option"..._ :-)

package.json
```diff
@@ -227,6 +229,7 @@
     "semver": "^5.7.0",
     "shell-escape": "^0.2.0",
     "split": "^1.0.1",
+    "ssh2": "^0.8.5",
```
pdcastro: Not sure how important, but this PR was parked for a little while and there are newer versions of ssh2 (0.8.7) and types/ssh2 (0.5.39).

lib/actions/ssh.ts
```diff
@@ -13,13 +13,15 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 */
+import { BalenaDeviceNotFound } from 'balena-errors';
 import * as BalenaSdk from 'balena-sdk';
 import { CommandDefinition } from 'capitano';
 import { stripIndent } from 'common-tags';
 
-import { BalenaDeviceNotFound } from 'balena-errors';
 import { validateDotLocalUrl, validateIPAddress } from '../utils/validation';
 
+const MIN_DIRECT_SSH_VERSION = '2.44.0';
```
pdcastro: Perhaps I should know the answer, but if you would indulge me, :-) what is it that 2.44 has to allow "direct ssh"?

lib/actions/ssh-direct.ts
```diff
@@ -0,0 +1,245 @@
...
+};
+
+const createCommandPipe = (param: {
+	cmd: string;
+	socket: net.Socket;
+}): Bluebird<ExecResponse> =>
+	Bluebird.using(createSSHClientDisposer(param.socket), client =>
+		client
+			.execAsync(param.cmd, { pty: process.stdout.isTTY })
+			.then(pipeDeviceToStdout),
```
pdcastro: Not sure where exactly where the issue is, but testing the recent Charlie's Tips 'n' Tricks from [balena's newsletter](https://www.balena.io/blog/balena-monthly-roundup-january-2020/):

Charlie's Tips 'n' Tricks

I have found an incompatibility with the `-x` alternative. Note how, without `-x`, the commands and the shell prompt get printed to the console, whereas with `-x` they don't:

```
$ echo "uptime; exit;" | ./bin/balena-dev ssh da4da236b4f76820d69941e315fd215a
root@da4da23:~# uptime; exit;
 02:52:56  up  12:05,  1 user,  load average: 0.27, 0.24, 0.18
logout
```
```
$ echo "uptime; exit;" | ./bin/balena-dev ssh da4da236b4f76820d69941e315fd215a -x
=============================================================
    Welcome to balenaOS
=============================================================
 02:45:36  up  11:57,  0 users,  load average: 0.12, 0.15, 0.14
```

lib/actions/ssh.ts
wrboyce: As of 2.44.0 BalenaOS is running opensshd which allows us to authenticate against ssh keys in balena-api, meaning there is no necessity to go via balena-proxy (gateway).

lib/actions/ssh-direct.ts
wrboyce: Good point, I will add this in. Should I add a `-s`, `-q`, or similar option to suppress the header also?

lib/actions/ssh.ts
```diff
@@ -207,18 +210,24 @@ export const ssh: CommandDefinition<
...
 			signature: 'verbose',
 			boolean: true,
 			description: 'Increase verbosity',
 			alias: 'v',
 		},
+		{
+			signature: 'gateway',
+			boolean: true,
+			description: `Use SSH gateway (required on BalenaOS < ${MIN_DIRECT_SSH_VERSION})`,
+			alias: 'x',
```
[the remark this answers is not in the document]
pdcastro: I think I found where the welcome message comes from:
https://github.com/balena-io/balena-proxy/blob/v5.21.1/src/common/host-scripts.ts#L77

You probably already knew it, as your name is all over that file. :-)

lib/actions/ssh-direct.ts
```diff
@@ -0,0 +1,245 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import * as sdk from 'balena-sdk';
+import * as Bluebird from 'bluebird';
+import * as globalTunnel from 'global-tunnel-ng';
```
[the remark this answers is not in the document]
pdcastro: Related new finding: proxy is currently not working with Node.js versions >= 10.16.0 (so not just Node 11, 12). Nothing specific to this PR -- the master branch is affected. Ive created this issue: https://github.com/balena-io/balena-cli/issues/1579

lib/actions/ssh-direct.ts
```diff
@@ -0,0 +1,245 @@
...
+}): Bluebird<ExecResponse> =>
+	Bluebird.using(createSSHClientDisposer(param.socket), client =>
+		client
+			.execAsync(param.cmd, { pty: process.stdout.isTTY })
+			.then(pipeDeviceToStdout),
+	);
+
+const getCommand = (containerId?: string): string => {
+	if (containerId != null) {
+		return `exec "$(find /usr/bin/{balena{-machine,},docker,rce} 2>/dev/null | head -1)" exec -it ${containerId} /bin/sh -c 'test -x /bin/bash && exec /bin/bash -l || exec /bin/sh -l'`;
```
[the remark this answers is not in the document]
pdcastro: Something else possibly also related to this exec logic. I have now tested `balena ssh` to a service running Alpine Linux, and I found that the login shell is invoked differently depending on whether the `-x` option is used. Running `ps`, I see that the shell is executed as `/bin/bash -l` in the case of ssh direct, and just `/bin/bash` when via the gateway:
```
$ ./bin/balena ssh da4da236b4f76820d69941e315fd215a foo
27bcdbce1ccc:/# ps
PID   USER     TIME  COMMAND
    1 root      0:00 /bin/sh -c i=1; while :; do echo "foo 5-$i $(uname -a)"; s
  148 root      0:00 /bin/bash
  181 root      0:00 /bin/bash -l
  713 root      0:00 sleep 10
  714 root      0:00 ps
27bcdbce1ccc:/#
```
```
./bin/balena ssh da4da236b4f76820d69941e315fd215a foo -x
bash-5.0# ps
PID   USER     TIME  COMMAND
    1 root      0:00 /bin/sh -c i=1; while :; do echo "foo 5-$i $(uname -a)"; s
  148 root      0:00 /bin/bash
  722 root      0:00 /bin/bash
  731 root      0:00 sleep 10
  732 root      0:00 ps
bash-5.0#
```
A "visual" difference is how the command prompt looks: `27bcdbce1ccc:/#` _vs._ `bash-5.0#`, but users could come across more "serious" consequences. I understand that the `-l` flag causes several scripts to be sourced: `/etc/profile`, `~/.bash_profile`, `~/.bash_login`, `~/.profile` (ref: https://linux.die.net/man/1/bash). If we introduce this change in the login shell behaviour, it is conceivable that we'd get some support queries from users saying that their services "got broken" on their ssh sessions, or that they started getting some error messages, and the underlying reason is that those scripts started being sourced, when "they were never sourced before".

Having the `-l` flag actually sounds like a "more correct" behaviour, but perhaps the strongest argument is that **the web dashboard shell matches the `-x` behaviour.** We would avoid support queries by matching the web dashboard and the present `-x` behaviour.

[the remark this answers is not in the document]
pdcastro: An even stronger argument to match the `-x` behaviour in relation to the login shell not having the `-l` option: support agents could fail to reproduce issues or even cause "damage" to users' services, by simply opening a ssh shell with `balena ssh`. The scenario is that many users will only ever have used the web dashboard to open a shell to their services, but a support agent might use `balena ssh` (as advised in the [debugging masterclass](https://github.com/balena-io/debugging-masterclass)), which could cause "a few scripts" to run for the first time ever.

lib/actions/ssh-direct.ts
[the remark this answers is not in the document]
pdcastro: I have created PR #1582 that uses `global-agent` for Node.js >= 10.16.0, and `global-tunnel-ng` for older versions, addressing issues #1579, #1580 and #1335.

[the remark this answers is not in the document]
pdcastro: PR #1582 was merged, but something that remains unexplained is that I couldn't get `balena ssh` to work behind a proxy (with balenaCloud, not openBalena) when `proxytunnel` is involved. And that was the case both before and after the PR was merged, testing on both Linux and macOS. Therefore, if it would be great if you could test that `balena ssh` works behind a proxy with balenaCloud, even if just on your own machine (Linux?), to check whether it was just me (maybe I was setting something up incorrectly).

lib/actions/ssh-direct.ts
pdcastro: Oh - "today I learned" :-) that the CLI already had some code to listen to the stdout 'resize' event, though probably not executed for the `balena ssh` command:
https://github.com/balena-io/balena-cli/blob/v11.25.12/lib/utils/tty.coffee
It is currently being converted to Typescript in PR #1602 (becoming `tty.ts`).
Maybe there's some room for code reuse, though not necessary -- I simply thought I would mention it.

lib/actions/ssh-direct.ts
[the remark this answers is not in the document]
pdcastro: > I couldn't get `balena ssh` to work behind a proxy (with balenaCloud, not openBalena) with `proxytunnel`

Now I managed to get it working, including basic authentication (username and password). What was happening is that my proxy server was not correctly configured to allow proxying to port 22 + SSL. PR #1685 adds a note to the README file about proxy configuration for ssh access, and makes some changes to address issue #1681.

pdcastro:
To however takes on the task of how to take this PR forward: Consider also my [handover notes](https://docs.google.com/document/d/1NxYwQrmSakQvwf9v8hXhCzuJzoQOjKZWrWZOZM4edI4/edit#) and the [ssh-uuid](https://github.com/pdcastro/ssh-uuid) proof-of-concept implementation.

Issue #1108, opened 2019-02-21
Balena device not working when device UUID has a 0 as first character


t16n:
- 9.12.1
- no
- v10.15.0
- 6.4.1
- macOS 10.14.3

When the device UUID has a 0 as the first character the balena device cli command is not working.
For example:

> balena device 0733307
> 
> returns:
> BalenaDeviceNotFound: Device not found: 733307

Putting the device id in String quotes did not help.
(balena device "0733307" / balena device '0733307')

pdcastro:
@t16n, thanks for reporting this issue! It has been fixed in release 9.12.7. You can check the version you are running with the `balena version` command, and the CLI can be updated by re-running the installation steps: https://www.balena.io/docs/reference/cli/

PR #1112, opened 2019-02-23
Fix parsing of not-really-numeric device UUID parameters

pdcastro:
Resolves: https://github.com/balena-io/balena-cli/issues/1108
Depends-on: https://github.com/balena-io/capitano/pull/70
Change-type: patch

---
##### Contributor checklist


lib/utils/device/api.ts
```diff
@@ -165,7 +165,7 @@ export class DeviceAPI {
 			}
 		}
 
-		return Bluebird.fromCallback(
+		return Bluebird.fromCallback<[request.Response, { message: string }]>(
```
pdcastro: Required as a result of adding `"@types/bluebird": "3.5.25"` to `package.json`.

lib/utils/normalization.ts
```diff
@@ -21,6 +21,7 @@ export function normalizeUuidProp(
 	propName = 'uuid',
 ) {
 	if (_.isNumber(params[propName])) {
-		params[propName] = _.toString(params[propName]);
+		params[propName] =
+			params[propName + '_raw'] || _.toString(params[propName]);
```
pdcastro: This is the change that matters in this PR. It builds on this other Capitano PR:
https://github.com/balena-io/capitano/pull/70/files

lib/utils/patterns.ts
```diff
@@ -217,7 +217,7 @@ export function awaitDevice(uuid: string) {
 			`Waiting for ${deviceName} to come online`,
 		);
 
-		const poll = (): Promise<void> => {
+		const poll = (): any => {
```
pdcastro: Required as a result of adding `"@types/bluebird": "3.5.25"` to `package.json`. Well, `any` is to some extent an admission of defeat and the reviewer is challenged to come up with a type expression that works (I've tried several alternatives without success). Note that, before I added `"@types/bluebird": "3.5.25"` to `package.json`, the typescript compiler was simply not checking / not trying to match the `Promise<void>` type against the implementation of the `poll()` function. Once it started checking it, I think it actually revealed that the implementation is type-wise internally inconsistent and would require some refactoring. However, in absence of evidence to the contrary, I am adequately satisfied that the type inconsistency is sufficiently harmless to be brushed aside.

package.json
```diff
@@ -95,8 +95,10 @@
   },
   "dependencies": {
     "@resin.io/valid-email": "^0.1.0",
+    "@types/bluebird": "3.5.25",
     "@types/dockerode": "2.5.5",
     "@types/mixpanel": "2.14.0",
+    "@types/request": "2.48.1",
```
pdcastro: These `@types` additions are merely incidental to this PR. My VS Code was nagging me with red underlines that the typescript code was missing out on type definitions and I agreed.

package.json
```diff
@@ -114,7 +116,7 @@
     "bash": "0.0.1",
     "bluebird": "^3.3.3",
     "body-parser": "^1.14.1",
-    "capitano": "^1.8.2",
+    "capitano": "^1.9.0-60-preserve-raw-string-values-7051fdce11705b5506434084b99821f955715a18",
```
pdcastro: This needs to be updated after the Capitano PR is merged. Meanwhile, it should allow the CI tests to pass.

lib/utils/normalization.ts
lurch: LOL, so Capitano converts it from a string to a number, and balena-cli turns it back from a number to a string :slightly_smiling_face:   (but I agree that given the complexity, it's probably worth doing the minimum change possible!)

package.json
```diff
@@ -95,8 +95,10 @@
   },
   "dependencies": {
     "@resin.io/valid-email": "^0.1.0",
+    "@types/bluebird": "3.5.25",
```
thgreasi: How about bumping the devDependency instead of adding this one and having both?

lib/utils/patterns.ts
thgreasi: I think that the issue is that the sdk is using 3.5.21 and 3.5.25 seems to contain a weird change.
Would 3.5.21 work for you in this?
If you indeed need 3.5.21 for this (let me know to also change that in the next major version of the SDK), you can cast the result of the `isOnline()` method before chaining the `.then`
```
return (balena.models.device.isOnline(uuid) as Promise<boolean>).then(function(isOnline) {
```
IMO this is definitely preferable instead of `Promise<any>`

package.json
pdcastro: Oh, I had not noticed that `@types/bluebird` was already listed in the devDependencies section. `@types/bluebird` is relevant for typescript code, and we use typescript in production not just in development (e.g. not just for running tests), so I assume that we want to have it in dependencies section. We should not duplicate it in both sections, of course, so my action was to move it devDependencies to dependencies. Is there a reason not to do this?

pdcastro:
> _[...] Would 3.5.21 work for you in this?_

@thgreasi, thanks for the review! I can confirm that the `Promise<void>` → `any` type change is not needed (I have reverted it) after the `@types/bluebird` downgrade from 3.5.25 to 3.5.21. (I had also removed the duplication between devDependencies and dependencies to rule out interference from that.)

package.json
pdcastro: I can answer my own question... `@types/bluebird` is not needed for a "production" installation to be installed and work correctly, and the fewer required packages in production, the better. So I have moved it back to the devDependencies section. :-)

Issue #1116, opened 2019-02-26
Extend private registry auth support to the balena build and deploy commands


pdcastro:
Private docker registry support is currently limited to the `balena push` command:
```
$ balena help push
...
The --registry-secrets option specifies a JSON or YAML file containing private
Docker registry usernames and passwords to be used when pulling base images.
```
Extend support to the `balena build` and `balena deploy` commands.

PR #1118, opened 2019-02-27
Extend private registry support to balena build and deploy commands

pdcastro:
Resolves: #1116
Change-type: minor

### DIFF
--- a/lib/actions/push.ts
+++ b/lib/actions/push.ts
@@ -18,8 +18,6 @@ import { CommandDefinition } from 'capitano';
 import { stripIndent } from 'common-tags';
 import { BalenaSDK } from 'balena-sdk';
 
-import { BuildError } from '../utils/device/errors';
-
 // An regex to detect an IP address, from https://www.regular-expressions.info/ip.html
 const IP_REGEX = new RegExp(
 	/\b(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\b/,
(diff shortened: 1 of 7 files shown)


lib/actions/push.ts
```diff
@@ -98,42 +96,6 @@ async function getAppOwner(sdk: BalenaSDK, appName: string) {
...
-interface RegistrySecrets {
-	[registryAddress: string]: {
-		username: string;
-		password: string;
-	};
-}
-
-async function parseRegistrySecrets(
-	secretsFilename: string,
-): Promise<RegistrySecrets> {
```
pdcastro: This function was moved from `push.ts` to `compose_ts.ts` because it is now shared and used by the three commands: push, build and deploy.

lib/utils/compose.coffee
```diff
@@ -41,6 +41,12 @@ exports.appendOptions = (opts) ->
 			description: 'Display full log output'
 			boolean: true
 		},
+		{
+			signature: 'registry-secrets',
+			alias: 'R',
+			parameter: 'secrets.yml|.json',
+			description: 'Path to a YAML or JSON file with passwords for a private Docker registry',
+		},
```
pdcastro: FYI, this option is shared by the `build` and `deploy` commands. In case it occurs to you, I realise that the `registry-secrets` spelling with a dash does not match the camelCase spelling of the other `build` and `deploy` commands, but it is the spelling used by the `push` command and there is also the goal of having the same options across all relevant CLI commands. There've been separate discussions about the homogenisation of CLI command options and it is part of the longer-term CLI redesign plans. It would be nice to accept both variants (reduces user friction, I suppose), but it's something for a future PR.

lib/utils/docker.coffee
```diff
@@ -147,6 +147,8 @@ exports.generateBuildOpts = (options) ->
 		opts.squash = true
 	if options.buildArg?
 		opts.buildargs = parseBuildArgs(options.buildArg)
+	if not _.isEmpty(options['registry-secrets'])
+		opts.registryconfig = options['registry-secrets']
```
pdcastro: This change is key to pass the user-provided secrets to the docker daemon.

PR #1124, opened 2019-03-04
Minor doc updates (add DEBUG=1 hint)

pdcastro:
Change-type: patch

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -589,17 +589,14 @@ path to the config JSON file, see `balena os build-config`
 
 ## envs
 
-Use this command to list all environment variables for
-a particular application or device.
+Use this command to list the environment variables of an application
+or device.
 
-This command lists all application/device environment variables.
+The --config option is used to list "config" variables that configure
+balena features.
 
-If you want to see config variables, used to configure
-balena features, use the --config option.
-
-At the moment the CLI does not support per-service variables,
-so the following commands will only show service-wide
-environment variables.
+Service-specific variables are not currently supported. The following
+examples list variables that apply to all services in an app or device.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


pdcastro:
@thgreasi, @nazrhom, this small pull request also has the purpose of testing the change to the Travis CI [build pushed branches](https://travis-ci.org/balena-io/balena-cli/settings) setting, which we hope will restore the standalone binaries in the GitHub releases download page.

lib/actions/environment-variables.ts
```diff
@@ -53,7 +53,7 @@ export const list: CommandDefinition<
 		balena features, use the --config option.
 
 		At the moment the CLI does not support per-service variables,
-		so the following commands will only show service-wide
+		so the following commands will only show application-wide
```
pdcastro: As suggested by @nazrhom (kind of) :-)

doc/cli.markdown
```diff
@@ -598,7 +598,7 @@ If you want to see config variables, used to configure
 balena features, use the --config option.
 
 At the moment the CLI does not support per-service variables,
-so the following commands will only show service-wide
+so the following commands will only show application-wide
```
thgreasi: How about ?
```suggestion
so the following commands will only show application/device-wide
```
```suggestion
so the following commands will only show application-wide or device-wide
```

lib/utils/messages.ts
```diff
@@ -1,15 +1,22 @@
...
 If you need help, or just want to say hi, don't hesitate in reaching out at:
 
-  GitHub: https://github.com/balena-io/balena-cli/issues/new
-  Forums: https://forums.balena.io\
+  Forums: https://forums.balena.io
+  GitHub: https://github.com/balena-io/balena-cli/issues/new\
+`;
+
+const debugHint = `\
+Additional information may be available in debug mode by prefixing the
```
thgreasi: How about ?
```suggestion
Additional information may be available if you run the CLI in debug mode. To do so prefix the
```

doc/cli.markdown
pdcastro: Thanks, I've reworded it to avoid '-wide' altogether. :-)

lib/utils/messages.ts
pdcastro: Thanks Thodoris. I am not a fan of using _'you'_ in help messages, but I've reworded it to avoid _'by prefixing'_ which I think is the bit you wanted to avoid.

Issue #1010, opened 2018-11-02
`balena deploy` should retry image push a few times before giving up


dfunckt:
There are transient errors that can cause the push to fail and it's worth retrying a few times before giving up, especially when pushing large images, because re-running `balena deploy` starts the push from scratch again (which is another issue that needs fixing).

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_f2bv)

PR #1125, opened 2019-03-04
Handle 'unknown blob' errors and retry image pushing

pdcastro:
This PR bumps `docker-progress` to improved detection of 'unknown blob' errors, add image push retry (including in case of an 'unknown blob' error) for the `balena deploy` command as requested in issue #1010, and in a separate commit also hardens error handling for the `balena push` command (`remote-build.ts`).

Resolves: #1010 
HQ: balena-io/balena#1531
Depends-on: balena-io-modules/docker-progress/pull/46
Change-type: patch

### DIFF
--- a/lib/utils/compose.coffee
+++ b/lib/utils/compose.coffee
@@ -409,6 +409,7 @@ authorizePush = (sdk, logger, tokenAuthEndpoint, registry, images, previousRepos
 pushAndUpdateServiceImages = (docker, token, images, afterEach) ->
 	chalk = require('chalk')
 	{ DockerProgress } = require('docker-progress')
+	{ retry } = require('./helpers')
 	tty = require('./tty')(process.stdout)
 
 	opts = { authconfig: registrytoken: token }
@@ -421,7 +422,13 @@ pushAndUpdateServiceImages = (docker, token, images, afterEach) ->
 		Promise.map images, ({ serviceImage, localImage, props, logs }, index) ->
 			Promise.join(
 				localImage.inspect().get('Size')
-				progress.push(localImage.name, reporters[index], opts).finally(renderer.end)
+				retry(
+					-> progress.push(localImage.name, reporters[index], opts)
+					3  # `times` - retry 3 times
+					localImage.name  # `label` included in retry log messages
+					2000  # `delayMs` - wait 2 seconds before the 1st retry
+					1.4  # `backoffScaler` - wait multiplier for each retry
+				).finally(renderer.end)
 				(size, digest) ->
 					serviceImage.image_size = size
 					serviceImage.content_hash = digest
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -35,7 +35,8 @@ export interface RegistrySecrets {
 export async function parseRegistrySecrets(
 	secretsFilename: string,
 ): Promise<RegistrySecrets> {
-	const { fs } = require('mz');
+	const { fs } = await import('mz');
+	const { exitWithExpectedError } = await import('../utils/patterns');
 	try {
 		let isYaml = false;
 		if (/.+\.ya?ml$/i.test(secretsFilename)) {
@@ -50,10 +51,11 @@ export async function parseRegistrySecrets(
 		MultiBuild.addCanonicalDockerHubEntry(registrySecrets);
 		return registrySecrets;
 	} catch (error) {
-		error.message =
-			`Error validating registry secrets file "${secretsFilename}":\n` +
-			error.message;
-		throw error;
+		return exitWithExpectedError(
+			`Error validating registry secrets file "${secretsFilename}":\n${
+				error.message
+			}`,
+		);
 	}
 }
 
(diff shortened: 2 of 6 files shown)


lib/utils/remote-build.ts
```diff
@@ -216,72 +213,120 @@ async function cancelBuildIfNecessary(build: RemoteBuild): Promise<void> {
...
+			exitOnError,
+		);
+		return buildRequest
+			.pipe(JSONStream.parse('*'))
+			.once('data', () => uploadSpinner.stop())
+			.once('end', () => uploadSpinner.stop())
+			.once('close', () => uploadSpinner.stop())
+			.once('error', () => uploadSpinner.stop());
+	} catch (error) {
+		return exitOnError(error);
```
pdcastro: Previously, in some cases if the builder request returned a HTTP error code, the CLI would not print any error message and would get stuck showing the spinner forever...

lib/utils/compose_ts.ts
```diff
@@ -50,10 +51,10 @@ export async function parseRegistrySecrets(
 		MultiBuild.addCanonicalDockerHubEntry(registrySecrets);
 		return registrySecrets;
 	} catch (error) {
-		error.message =
+		return exitWithExpectedError(
 			`Error validating registry secrets file "${secretsFilename}":\n` +
-			error.message;
-		throw error;
+				error.message,
```
hedss: Seems a bit strange to use a template literal and then concat?
```typescript
return exitWithExpectedError(
	`Error validating registry secrets file "${secretsFilename}":\n${error.message}`
);
```

lib/utils/helpers.ts
```diff
@@ -187,3 +187,27 @@ export function getSubShellCommand(command: string) {
...
+	times: number,
+	label: string,
+	delayMs = 1000,
+	factor = 2,
+	count = 0,
+): Bluebird<T> {
+	let promise = Bluebird.try(func);
+	if (count < times) {
+		promise = promise.catch((err: Error) => {
+			const delay = factor ** count * delayMs;
```
hedss: Associativity for the operators used in this statement are not consistent. It's a lot nicer to make this explicit:
```typescript
const delay = (factor ** count) * delayMS;
```
That said, it'd also be nice to make it explicit that `factor` is a backoff scaler, perhaps `backoffScaler` rather than `factor`?

lib/utils/helpers.ts
```diff
@@ -187,3 +187,27 @@ export function getSubShellCommand(command: string) {
...
+			console.log(
+				`Retrying "${label}" after ${(delay / 1000).toFixed(2)}s (${count +
+					1} of ${times}) due to: ${err}`,
+			);
+			return Bluebird.delay(delay).then(() =>
+				retry(func, times, label, delayMs, factor, count + 1),
```
hedss: I'm missing something here, why doesn't it `return` the retry promise?

lib/utils/compose.coffee
```diff
@@ -421,7 +422,10 @@ pushAndUpdateServiceImages = (docker, token, images, afterEach) ->
 		Promise.map images, ({ serviceImage, localImage, props, logs }, index) ->
 			Promise.join(
 				localImage.inspect().get('Size')
-				progress.push(localImage.name, reporters[index], opts).finally(renderer.end)
+				retry(
+					-> progress.push(localImage.name, reporters[index], opts)
+					3, localImage.name, 2000, 1.4)
+						.finally(renderer.end)
```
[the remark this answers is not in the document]
pdcastro: Reformatted

lib/utils/compose_ts.ts
pdcastro: Reformatted

lib/utils/helpers.ts
pdcastro: Renamed. The parentheses are automatically removed by `prettier`.

lib/utils/helpers.ts
pdcastro: Do you mean `return retry()` as opposed to `retry()`? Note there are no curly brackets, so the `return` statement is not needed. Compare:
* `() => retry()`
* `() => { return retry(); }`

lib/utils/compose.coffee
```diff
@@ -421,7 +422,13 @@ pushAndUpdateServiceImages = (docker, token, images, afterEach) ->
 		Promise.map images, ({ serviceImage, localImage, props, logs }, index) ->
 			Promise.join(
 				localImage.inspect().get('Size')
-				progress.push(localImage.name, reporters[index], opts).finally(renderer.end)
+				retry(
+					-> progress.push(localImage.name, reporters[index], opts)
+					3
+					localImage.name
+					2000
+					1.4
```
dfunckt: Worth adding a comment beside the value as to what these are for.

pdcastro: Done

lib/utils/helpers.ts
hedss: Ah yes, I'd missed the lack of curly braces. My fault. :)

lib/utils/helpers.ts
hedss: Thanks. Prettier strikes again. :(

Issue #1102, opened 2019-02-15
balena deploy fails for additional members of legacy apps


pdcastro:
As reported in the forums:
https://forums.balena.io/t/ssh-agent-build-forwarding/5300
```
user@server:~/Github# DEBUG=1 balena deploy legacy/app d9541f4cab41
[Debug]   Parsing input...
[Debug]   Loading project...
[Info]    Creating default composition with image: d9541f4cab41
[Debug]   Creating project...
[Info]    Everything is up to date (use --build to force a rebuild)
[Warn]    Target application requires legacy deploy method.
[Info]    Initializing deploy...
[Info]    Uploading [========================] 100% eta 0s            
[Debug]   Received data: {"type":"error","message":"Unauthorized. Cannot access application"}
[Error]   Deploy failed
Error: Remote error: undefined
    at Request.handleMessage (/snapshot/balena-cli/build/utils/deploy-legacy.js:67:25)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
    at IncomingMessage.<anonymous> (/snapshot/balena-cli/node_modules/request/request.js:1080:12)
    at emitOne (events.js:96:13)
    at IncomingMessage.emit (events.js:188:7)
    at IncomingMessage.Readable.read (_stream_readable.js:381:10)
    at flow (_stream_readable.js:761:34)
    at resume_ (_stream_readable.js:743:3)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
```

Known workarounds:
* Use CLI version 6.10.3
* Upgrade to a new (non-legacy) app

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_hc5v)

pdcastro:
On investigation, I've gathered that since a refactoring about a year ago, the CLI was discarding the 'owner' part of 'owner/appName' for legacy applications when creating the builder endpoint URL.

PR #1131, opened 2019-03-07
Fix 'unauthorized' error for additional members of legacy apps

pdcastro:
Please see description of issue #1102 for what this PR aims at. On investigation, I've gathered that since a refactoring about a year ago, the CLI was discarding the 'owner' part of 'owner/appName' for legacy applications when creating the builder endpoint URL.

Resolves: #1102
Change-type: patch

### DIFF
--- a/lib/actions/deploy.coffee
+++ b/lib/actions/deploy.coffee
@@ -84,7 +84,8 @@ deployProject = (docker, logger, composeOpts, opts) ->
 					sdk.auth.whoami()
 					sdk.settings.get('balenaUrl')
 					{
-						appName: opts.app.app_name
+						# opts.appName may be prefixed by 'owner/', unlike opts.app.app_name
+						appName: opts.appName
 						imageName: images[0].name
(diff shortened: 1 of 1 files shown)


lib/actions/deploy.coffee
```diff
@@ -184,6 +185,7 @@ module.exports =
 
 		logger.logDebug('Parsing input...')
 
+		appName = undefined
```
CameronDiver: This is unnecesary

pdcastro: In this case this statement is necessary, as otherwise the coffeescript compiler generates `var appName` in the wrong scope. Without it, the linter says:
```
  ⚡ lib/actions/deploy.coffee
     ⚡ #226: Undefined identifier "appName".
```
And when the CLI is executed I get:
```
ReferenceError: appName is not defined
```

CameronDiver: Ah the scope is a good point, sorry for the noise.

PR #1135, opened 2019-03-11
[WIP] ssh: option to send command through SSH to the host OS and return exit code

imrehg:
* Have an option to run commands over ssh on the device. Due to
  Capitano limitations, it seems to need to be a flag, so a new flag
  of `-c` is introduced.
* Exit the CLI with the exit code from the SSH connection, so that
  we can correctly catch errors
* Due to the limitations of connecting to the device, this command
  flag only seem to work with the host OS

Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -966,7 +966,7 @@ installed in your shell environment. For more information (including Windows
 support) please check the README here: https://github.com/balena-io/balena-cli
 
 Use this command to get a shell into the running application container of
-your device.
+your device, and you can also run commands in the host OS
 
 Examples:
 
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -975,6 +975,7 @@ Examples:
 	$ balena ssh 7cf02a6 --port 8080
 	$ balena ssh 7cf02a6 -v
 	$ balena ssh 7cf02a6 -s
+	$ balena ssh 7cf02a6 -s -c 'date'
 
 ### Options
 
@@ -993,6 +994,10 @@ access host OS (for devices with balenaOS >= 2.0.0+rev1)
 #### --noproxy
 
 don't use the proxy configuration for this connection. Only makes sense if you've configured proxy globally.
+
+#### --command, -c &#60;command&#62;
+
+command to run after connecting to the device (only in the host OS)
 
 # Notes
 
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -154,7 +154,7 @@ environment variable (in the same standard URL format).
 
 - SSH
 
-	- [ssh [uuid]](#ssh-uuid)
+	- [ssh [uuid] [command]](#ssh-uuid-command)
 
 - Notes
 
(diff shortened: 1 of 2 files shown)


doc/cli.markdown
```diff
@@ -975,6 +975,7 @@ Examples:
 	$ balena ssh 7cf02a6 --port 8080
 	$ balena ssh 7cf02a6 -v
 	$ balena ssh 7cf02a6 -s
+	$ balena ssh 7cf02a6 -s -c 'date'
```
CameronDiver: Can we get away with not needing the `-c`?

lib/actions/ssh.coffee
```diff
@@ -53,6 +54,11 @@ module.exports =
 			boolean: true
 			description: "don't use the proxy configuration for this connection.
 				Only makes sense if you've configured proxy globally."
+	,
+			signature: 'command'
+			parameter: 'command'
+			description: 'command to run after connecting to the device (only in the host OS)'
```
CameronDiver: How come it has to be in the host OS?

doc/cli.markdown
imrehg: @CameronDiver don't think we can do multiple `COMMAND` in Capitano? And the UUID is already the command? I'd love to get rid of that too :/ Will try some more

lib/actions/ssh.coffee
imrehg: @CameronDiver connecting to the container with the current "exec" method does not seem to run the command passed. Will have to check with @wrboyce whether that's really the case, but right now it doesn't seem to work. I hope later we can expand on this.

CameronDiver: There should be a method, but as you say Will is the one to ask.

doc/cli.markdown
pdcastro: iirc, Capitano had the "concepts" of parameters and options, and there could be multiple parameters, though I'm not 100% sure. So in `balena ssh 7cf02a6 'date' -s`, there's be two parameters (`7cf02a6` and `'date'`) and one option (`-s`). I could well be wrong, though!

imrehg: @pdcastro @CameronDiver updated this, it's now an optional command

lib/actions/ssh.coffee
imrehg: Checking with Will, but there doesn't seem to be a way currently as we are connecting inside the container.

imrehg: Nope, it won't work with the container.

imrehg:
@CameronDiver @pdcastro added the changes from your comments (and discussion in [Flowdock]), it's now a command, and only works in the host OS still

pdcastro:
@imrehg, when the PR is ready to receive approval, please remove `[WIP]` from its title.

imrehg:
@pdcastro yes, that's what I will do, just checking whether to actually drop this PR for the time being, hence the `[WIP]` for a bit longer.

pdcastro:
Closing because both the client side (balena CLI) and the server side (balena-proxy) evolved such that the proposed solution no longer applies. Some of the issues remain unresolved though. See:

* #1482 
* #2342 
* https://github.com/balena-io/balena-proxy/issues/465
* https://github.com/balena-io/balena-proxy/issues/466
* https://github.com/pdcastro/ssh-uuid

PR #1136, opened 2019-03-11
ssh: add flag not to suggest devices to connect to

imrehg:
The suggestion happens if the UUID supplied is not found. Because of that function, it's impossible to do an atomic connect to a device in non-interactive mode. The auto-suggestion results connecting to the first available device, which is likely not the intended action. The current workaround is running a `balena device UUID` and check its exit code before running `balena ssh UUID`, but since these are independent steps, still can connect to another device, if between the two commands anything changes. With this flag used, one could never connect accidentally to the wrong device due to suggestions.

The `--nosuggest` flag name chosen similarly to the `--noproxy` flag in the same command.

Change-type: minor
Signed-off-by: Gergely Imreh <gergely@balena.io>

---
##### Contributor checklist

### DIFF
--- a/lib/actions/ssh.coffee
+++ b/lib/actions/ssh.coffee
@@ -35,6 +35,7 @@ module.exports =
 			$ balena ssh 7cf02a6 --port 8080
 			$ balena ssh 7cf02a6 -v
 			$ balena ssh 7cf02a6 -s
+			$ balena ssh 7cf02a6 --noninteractive
 	'''
 	permission: 'user'
 	primary: true
@@ -53,6 +54,10 @@ module.exports =
 			boolean: true
 			description: "don't use the proxy configuration for this connection.
 				Only makes sense if you've configured proxy globally."
+	,
+			signature: 'noninteractive'
+			boolean: true
+			description: 'run command non-interactively, do not automatically suggest devices to connect to if UUID not found'
 	]
 	action: (params, options, done) ->
 		normalizeUuidProp(params)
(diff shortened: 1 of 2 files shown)


lib/actions/ssh.coffee
```diff
@@ -53,6 +54,10 @@ module.exports =
 			boolean: true
 			description: "don't use the proxy configuration for this connection.
 				Only makes sense if you've configured proxy globally."
+	,
+			signature: 'nosuggest'
+			boolean: true
+			description: 'do not automatically suggest devices to connect to'
```
pdcastro: The suggestion of devices to connect to -- is this the only interactive action taken by balena ssh? Does it perhaps "allocate a tty" (for the CLI process, if this makes sense) when this is unnecessary in the context of a script?
The reason for these questions is that, as you know, there is the idea/wish of having a "scriptable" or "non interactive" mode across all CLI commands... Here it's only for `balena ssh`, of course, but perhaps the name could be a bit more generic, like '--interactive off' or '--noninteractive' or '--scriptable' or something in these lines. But 'nosuggest' is OK too. :-)

imrehg: @pdcastro not sure what else is using interactive mode (maybe configure?) but this one is a big pain for me.

You are right, let's change the flag, will have something more generic in a bit.

pdcastro: Sorry, I meant whether any other aspect or option of `balena ssh` is interactive, not whether other CLI commands are interactive. Focusing only on `balena ssh` here... I don't expect you to fix the whole CLI. :-)

PR #1137, opened 2019-03-11
Add maintainer as code owner

imrehg:
Change-type: patch
Signed-off-by: Gergely Imreh <gergely@balena.io>

---
##### Contributor checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*	@pdcastro @thgreasi @CameronDiver @hedss


thgreasi:
@imrehg It would be nice to add an extra reviewer, so that Paulo can also benefit for the reviewer list pre-population. It probably makes sense to add me as well and maybe we can re-iterate later and add Aki & Alexis when affecting the build & deploy related files

imrehg:
@thgreasi added you, and indeed I was wondering who else to add, I should have asked explicitly :)

CameronDiver:
Can you also add me too please @imrehg and perhaps @hedss, because we're both in the devexp project.

Issue #1130, opened 2019-03-07
[push] When pushing to a local device, it shouldn't be necessary to be logged into balenaCloud


shaunmulligan:
It should be able to do the `balena push <IP>` workflow with the unmanaged version of the OS without needing to login with the CLI.
```
shaun-mbp-2:balena-rust-blinkt hobochild$ balena push 192.168.1.41
You have to log in to continue

Run the following command to go through the login wizard:

  $ balena login

Additional information may be available in debug mode. Prefix the command
line with DEBUG=1, i.e.:  DEBUG=1 balena ...
If you need help, don't hesitate in contacting us at:

  Forums: https://forums.balena.io
  GitHub: https://github.com/balena-io/balena-cli/issues/new

```

shaunmulligan:
Part of the reason this is currently the case is due to a limitation in the CLI capitano apparently.

PR #1145, opened 2019-03-25
Allow 'balena push <deviceIpAddress>' when not logged in to balenaCloud.

pdcastro:
Resolves: #1130 
Change-type: patch

### DIFF
--- a/lib/utils/patterns.ts
+++ b/lib/utils/patterns.ts
@@ -14,8 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import BalenaSdk = require('balena-sdk');
-import Promise = require('bluebird');
+import Bluebird = require('bluebird');
 import chalk from 'chalk';
+import { stripIndent } from 'common-tags';
 import _ = require('lodash');
 import _form = require('resin-cli-form');
 import _visuals = require('resin-cli-visuals');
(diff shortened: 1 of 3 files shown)


lib/utils/patterns.ts
```diff
@@ -75,6 +75,22 @@ export function authenticate(options: {}): Promise<void> {
...
+
+Run the following command to go through the login wizard:
+
+	$ balena login\
+`);
+		}
+		return isLoggedIn;
+	});
+}
+
```
zvin:
```typescript

export async function exitIfNotLoggedIn(): Promise<void> {
	const balena = BalenaSdk.fromSharedOptions();
	if (!(await balena.auth.isLoggedIn())) {
		exitWithExpectedError(stripIndent`
			You have to log in to continue
			Run the following command to go through the login wizard:
				$ balena login\
		`);
	}
}
```

pdcastro:
Thanks for the review @zvin. I've pushed a new version that fixes an issue found by @shaunmulligan -- `balena whoami` was broken! The fix was to replace:
```
const balena = BalenaSdk.fromSharedOptions();
```
with
```
```
in `patterns.ts`.

lib/utils/patterns.ts
pdcastro: Done, thanks!

PR #1186, opened 2019-04-24
Add per-service filtering to logs and push

CameronDiver:
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -865,18 +865,26 @@ If an IP address is passed to this command, logs are displayed from
 a local mode device with that address. Note that --tail is implied
 when this command is provided an IP address.
 
+Logs from a single service can be displayed with the --service flag.
+
 Examples:
 
 	$ balena logs 23c73a1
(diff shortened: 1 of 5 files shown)


doc/cli.markdown
```diff
@@ -1475,6 +1485,11 @@ This feature requires a device running supervisor version v9.7.0 or greater.
 
 Don't tail application logs when pushing to a local mode device
 
+#### --service &#60;service&#62;
+
+Only show logs from a single service.
+Only valid when pushing to a local mode device.
```
zvin: Doesn't this also work for non local devices?
If it doesn't the example above is wrong:
```
balena logs 23c73a1 --service my-service
```

CameronDiver: Oh yeah, forgot to take this out... Good catch!

CameronDiver: Wait, no, both the documentation and example are correct, logs are only displayed when pushing to a local mode device. The section you are referencing is for push, not logs.

PR #1190, opened 2019-04-24
Allow specifying a .local address for logs and push

CameronDiver:
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/lib/actions/logs.ts
+++ b/lib/actions/logs.ts
@@ -18,6 +18,7 @@ import { CommandDefinition } from 'capitano';
 import { stripIndent } from 'common-tags';
 
 import { normalizeUuidProp } from '../utils/normalization';
+import { validateDotLocalUrl } from '../utils/validation';
 
 type CloudLog =
 	| {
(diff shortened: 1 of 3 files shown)


lib/actions/push.ts
```diff
@@ -152,6 +153,12 @@ export const push: CommandDefinition<
 			$ balena push 10.0.0.1 --service my-service
 			$ balena push 10.0.0.1 --system
 			$ balena push 10.0.0.1 --system --service my-service
+
+			$ balena push 23c73a1.local
+			$ balena push 23c73a1.local --source <source directory>
+			$ balena push 23c73a1.local -s <source directory>
+			$ balena push 23c73a1.local --service my-service
+			$ balena push 23c73a1.local --system --service my-service
```
pdcastro: These 14 lines of examples between myApp / 10.0.0.1 / 23c73a1.local are a bit too many, no? I suggest spreading the different examples of flags between the IP address and domain name examples, so having 3 examples for myApp, 3 for IP address and 3 for local domain.

lib/actions/logs.ts
```diff
@@ -67,7 +68,12 @@ export const logs: CommandDefinition<
 			$ balena logs 192.168.0.31
 			$ balena logs 192.168.0.31 --service my-service
 			$ balena logs 192.168.0.31 --system
-			$ balena logs 192.168.0.31 --system --service my-service`,
+			$ balena logs 192.168.0.31 --system --service my-service
+
+			$ balena logs 23c73a1.local
+			$ balena logs 23c73a1.local --tail
+			$ balena logs 23c73a1.local --service my-service
+			$ balena logs 23c73a1.local --system --service my-service`,
```
pdcastro: Similar to the other comment, maybe 3 examples for each are enough, covering different flags between them.

lib/actions/push.ts
CameronDiver: Done

PR #1201, opened 2019-04-29
Remove --dockerPort's -p alias for `balena preload`

zvin:
Remove --dockerPort's -p alias for `balena preload`

It was conflicting with --pin-device-to-release -p alias

Changelog-entry: Remove --dockerPort's -p alias for `balena preload`
Change-type: major

### DIFF
--- a/lib/actions/preload.coffee
+++ b/lib/actions/preload.coffee
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+_ = require('lodash')
+
 dockerUtils = require('../utils/docker')
 
 LATEST = 'latest'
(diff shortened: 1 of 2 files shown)


lib/actions/preload.coffee
```diff
@@ -179,7 +179,7 @@ module.exports =
 			signature: 'pin-device-to-release'
 			boolean: true
 			description: 'Pin the preloaded device to the preloaded release on provision'
-			alias: 'p'
```
pdcastro: Cool -- does that work?! :-) I didn't know that the alias could have more than one character.

pdcastro:
And of course we need to hear from @shaunmulligan too...

lib/actions/preload.coffee
zvin: Affirmative, just checked to be sure :)

pdcastro:
Flowdock discussion: https://www.flowdock.com/app/rulemotion/r-product/threads/CvnlARvv_5b71EhBBGBqgblAuaH

shaunmulligan:
@pdcastro we should write a summary of the flowdock discussion above, since this is an open source repo and people can't see our internal convo

pdcastro:
> LGTM, but I've not been part of the conversation. I think it makes a tonne of sense, though.

The [conversation](https://www.flowdock.com/app/rulemotion/r-product/threads/CvnlARvv_5b71EhBBGBqgblAuaH) so far resulted in the "new tentative plan" summarised in my previous comment above, meaning that this PR needs to be changed (no `--pin` option).

hedss:
> > LGTM, but I've not been part of the conversation. I think it makes a tonne of sense, though.
> 
> The [conversation](https://www.flowdock.com/app/rulemotion/r-product/threads/CvnlARvv_5b71EhBBGBqgblAuaH) so far resulted in the "new tentative plan" summarised in my previous comment above, meaning that this PR needs to be changed (no `--pin` option).

Fair enough!

lib/actions/preload.coffee
```diff
@@ -130,6 +132,43 @@ offerToDisableAutomaticUpdates = (application, commit, pinDevice) ->
...
+	}
+	{
+		signature: 'pin-device-to-release'
+		boolean: true
+		description: 'Pin the preloaded device to the preloaded release on provision'
+		alias: 'p'
+	}
+]
+# Remove dockerPort `-p` alias as it conflicts with pin-device-to-release
+delete _.find(preloadOptions, signature: 'dockerPort').alias
```
pdcastro: That's cheating! Or clever.  :-)

Issue #842, opened 2018-04-03
Update notifier doesn't make much sense in the standalone build


pimterry:
For a few reasons:
* If you have npm + standalone installed, you can get cached notifications from the npm version when you start the standalone version
* If there isn't a cached notification from npm, I'm not even sure if it works?
* The instructions tell you to go to npm to update, which definitely isn't right

PR #1203, opened 2019-04-30
Update daily upgrade notifier message for non-npm installations

pdcastro:
Resolves: #842
Change-type: patch

### DIFF
--- a/lib/utils/update.ts
+++ b/lib/utils/update.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2016-2017 Balena
+Copyright 2016-2019 Balena
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lib/utils/update.ts
```diff
@@ -25,26 +25,26 @@ const balenaUpdateInterval = 1000 * 60 * 60 * 24 * 1;
...
-	if (notifier.update != null) {
-		console.log(
-			'Notice that you might need administrator privileges depending on your setup\n',
-		);
+	const up = notifier.update;
+	if (up) {
+		notifier.notify({
+			defer: false,
+			message: `Update available ${up.current} → ${up.latest}\n
+https://github.com/balena-io/balena-cli/blob/master/INSTALL.md`,
```
pdcastro: What URL do we want to use? -
1. https://github.com/balena-io/balena-cli/blob/master/INSTALL.md
2. https://www.balena.io/docs/reference/cli/
3. https://www.balena.io/docs/reference/cli/#installation
4. A tinyurl to any of the above, e.g. https://tinyurl.com/yymwhcaj

Option 3 would be good if it wasn't for the fact that the anchor link results in the **Installation** section title to get under the page header, leaving the **Getting Started** section prominent. Users would need some scrolling to figure it out, then click on a link to takes them to GitHub. All things considered, I think my preference is for option 1, without a tinyurl link. The CLI is usually used on large enough screens to accommodate the long URL, and then it's straight to the point and clear that the URL will land on INSTALL.md on GitHub.

hedss: Concur.

pdcastro:
force push = rebase on HEAD

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

PR #1206, opened 2019-05-01
qemu: use v4.0.0

imrehg:
Also append the QEMU version to the locally cached copy, so the CLI can correctly bump version whenever QEMU_VERSION is bumped in the future.

Change-type: minor
Signed-off-by: Gergely Imreh <imrehg@gmail.com>

---


imrehg:
@balena-ci retest please

Issue #977, opened 2018-10-13
Getting Error Message No such file or directory: /snapshot/resin-cli/node_modules/denymount/bin/denymount


13bmartens:
- **resin-cli version:** 7.10.6
- **Operating system and architecture:** MAC-OSX 10.14.1 Beta

Following the tutorial at https://resinos.io/docs/iot2000/gettingstarted/
sudo resin local configure ~/Downloads/resin.img shows error:

No such file or directory: /snapshot/resin-cli/node_modules/denymount/bin/denymount

Help would be greatly appreciated :)

gpulido:
Hello,
I'm having the same issue with resin-cli version 10.0.1 on a Mac-osx Darwin Kernel Version 18.5.0: Mon Mar 11 20:40:32 PDT 2019; root:xnu-4903.251.3~3/RELEASE_X86_64 x86_64
Using Debug=1 I have the following output:
````
Error: spawn /snapshot/balena-cli/node_modules/denymount/bin/denymount ENOENT
    at exports._errnoException (util.js:1020:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:197:32)
    at onErrorNT (internal/child_process.js:376:16)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
```
Thanks in advance

pdcastro:
@gpulido, what command line were you running? Was it `sudo balena local configure`? I am using the exact same version of MacOS as yours, and also using CLI version 10.0.1, and the following works for me without errors:
```
$ pwd
/Users/paulo

$ sudo balena local configure ~/Downloads/test-raspberrypi3-2.31.5+rev1-dev-v9.11.3.img
Password:
? Network SSID My_Wifi_Ssid
? Network Key super_secret_wifi_password
? Do you want to set advanced settings? No
Done!
```
Did you use the [standalone install](https://www.balena.io/docs/reference/cli/#standalone-install) zip file, or installed via npm?
I wonder what I would have to do in order to reproduce the error. Could it be the image file?!

gpulido:
Hello,
I'm using the standalone balena-cli and using the macos terminal.
And the command is the one you mentioned using a fresh raspberrypi3 balena.img downloaded from the balena repositories. I added balena to the path using the suggestions from the readme:
```
sudo balena local configure ~/Downloads/balena.img
Password:
[...]

Issue #1212, opened 2019-05-07
Facing issue in balena local configure 


Sharvin26:
I have downloaded the balena-cli from this [Link](https://github.com/balena-io/balena-cli/releases) I have unzipped the Balena image and set the balena-cli path using the following command:

`export PATH=$PATH:/usr/local/lib/balena-cli`

Now when I run 

`sudo balena local configure resin.img`

I get following error:

```
➜ sudo balena local configure resin.img
No such file or directory: /snapshot/balena-cli/node_modules/denymount/bin/denymount

Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, or "export DEBUG=1"
on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```

I have downloaded all the Requirements stated in this [documentation](https://github.com/balena-io/balena-cli/blob/master/INSTALL.md). 

```
➜ balena version
10.10.1
```

Note: I am using mac os for this process.

pdcastro:
@Sharvin26, thank you for reporting this issue. I was able to reproduce it and I have a pull request that fixes it, so the fix should be released soon: https://github.com/balena-io/balena-cli/pull/1213

I believe this issue is also a duplicate of #977 [Getting Error Message No such file or directory: /snapshot/resin-cli/node_modules/denymount/bin/denymount](https://github.com/balena-io/balena-cli/issues/977)

Sharvin26:
Okay.

Sharvin26:
I am still facing the issue. Has the issue been fixed?

pdcastro:
@Sharvin26, the pull request I linked is still open, as the reviewer requested some changes, and addressing those changes required the creation of a second pull request in another repository:
* https://github.com/balena-io/balena-cli/pull/1213
* https://github.com/balena-io-modules/denymount/pull/8

I am working on them at the moment. When the PRs are approved, they will be merged and the automated CLI release process will then produce a new CLI version. I will update issue #977 with the new CLI version number when it is available, and the issue will then be closed. You can watch that issue and/or the PRs for updates. Thanks again for reporting this issue.

pdcastro:
[...]

Issue #875, opened 2018-05-04
Local configure does not work on Windows


pimterry:
Local configure throws:

```
NOENT: No such file or directory
```

when run on Windows (repro against ResinOS Artik 710 2.12.5+rev1.dev). The same CLI & image work perfectly on Linux. Seems to be an exception from within `fatfs`. Fails the same way in both the standalone & npm releases, and in git bash/cmd.

See https://sentry.io/resinio/cli/issues/543307865/events/19470038594/ for more details.

Erebus25:
Same thing with BeagleBone Black

PR #1213, opened 2019-05-08
Fix 'local configure' on macOS standalone installation

pdcastro:
Resolves: #977
Change-type: patch

### DIFF
--- a/patches/denymount+2.2.0.patch
+++ b/patches/denymount+2.2.0.patch
@@ -1,0 +1,38 @@
+diff --git a/node_modules/denymount/lib/index.js b/node_modules/denymount/lib/index.js
+index 93b8e59..86d53dc 100644
+--- a/node_modules/denymount/lib/index.js
++++ b/node_modules/denymount/lib/index.js
+@@ -24,7 +24,9 @@ var utils = require('./utils');
+  * @module denymount
+  */
+ 
+-var EXECUTABLE_PATH = path.join(__dirname, '..', 'bin', 'denymount');
++var EXECUTABLE_PATH = process.pkg
++  ? path.join(path.dirname(process.execPath), 'denymount')
++  : path.join(__dirname, '..', 'bin', 'denymount');
+ 
+ /**
+  * @summary Prevent automatic mounting of an OS X disk
+diff --git a/node_modules/denymount/lib/index.js.rej b/node_modules/denymount/lib/index.js.rej
+new file mode 100644
+index 0000000..a2c0516
+--- /dev/null
++++ b/node_modules/denymount/lib/index.js.rej
+@@ -0,0 +1,17 @@
++***************
++*** 24,30 ****
++   * @module denymount
++   */
++  
++- var EXECUTABLE_PATH = path.join(__dirname, '..', 'bin', 'denymount');
++  
++  /**
++   * @summary Prevent automatic mounting of an OS X disk
++--- 24,30 ----
++   * @module denymount
++   */
++  
+++ var EXECUTABLE_PATH = path.join(path.dirname(process.execPath), 'denymount', 'bin', 'denymount');
++  
++  /**
++   * @summary Prevent automatic mounting of an OS X disk
--- a/automation/build-bin.ts
+++ b/automation/build-bin.ts
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import * as Bluebird from 'bluebird';
 import * as filehound from 'filehound';
 import * as fs from 'fs-extra';
 import * as path from 'path';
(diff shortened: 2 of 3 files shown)


patches/denymount+2.2.0.patch
```diff
@@ -0,0 +1,38 @@
...
+--- a/node_modules/denymount/lib/index.js
++++ b/node_modules/denymount/lib/index.js
+@@ -24,7 +24,9 @@ var utils = require('./utils');
+  * @module denymount
+  */
+ 
+-var EXECUTABLE_PATH = path.join(__dirname, '..', 'bin', 'denymount');
++var EXECUTABLE_PATH = process.pkg
++  ? path.join(path.dirname(process.execPath), 'denymount')
++  : path.join(__dirname, '..', 'bin', 'denymount');
```
pdcastro: As far as my investigation went... The pkg standalone installer bundles together the node.js executable with the node_modules folder, all in a single large executable file. But it seems that a pkg limitation means that the OS cannot run executable binaries inside the large executable binary. (I guess this is the same limitation that requires native/binary modules to be packaged outside the large executable.) I had success copying the `denymount` executable to the `build-bin` folder, but the `node_modules/denymount/bin/denymount` location was hardcoded in the denymount `index.js` file. So I patched the file location using the `patch-package` npm package.

pdcastro:
This PR fixes `'local configure'` for the macOS standalone installation, but does not change the status under Windows. I've tested all of the following:

Platform | Install type | Before this PR | After this PR
--- | --- | --- | ---
Linux | npm | OK | OK
Linux | ts-node | OK | OK
Linux | standalone | OK | OK
macOS | npm | OK | OK
macOS | ts-node | OK | OK
**macOS** | **standalone** | **Fail** | **OK**
Windows | npm | Fail | Fail
Windows | ts-node | Fail | Fail
Windows | standalone | Fail | Fail

Issue #875 suggests that 'local configure' may have never worked on Windows, although I don't get the error reported in that issue. The error I get is that the CLI gets stuck, not printing anything at all until CTRL-C is pressed (I've probably waited half an hour at one point). Nothing is printed when CTRL-C is pressed either.

patches/denymount+2.2.0.patch
dfunckt: @pdcastro denymount is our own (in fact, if memory serves, I think I did the initial implementation for Etcher :) — can’t this change be integrated directly? Repo is here: https://github.com/balena-io-modules/denymount

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

patches/denymount+2.2.0.patch
pdcastro: @dfunckt, thanks for the review! Indeed I had missed the fact that 'denymount' is our own, and had initially used `patch-package` exactly because I wanted to avoid having to wait for the package's maintainer to accept a PR. Following your review, I created a `denymount` PR that would probably solve the problem more cleanly by avoiding the need for `patch-package`: https://github.com/balena-io-modules/denymount/pull/8
However, what I then found in the `denymount` repo/PR is that even the head of master (before any change of mine) would not pass the existing tests on Linux (they pass on macOS, but not on Linux), and the balena CI tests would thus fail. I think the tests require some changes to Sinon mocking. Unfortunately, following a recent priority discussion with Shaun, I must focus on CLI native installers and I cannot justify spending more time on the `denymount` PR. Some options now:
* We could merge this CLI PR as it is, as it solves the problem and users are waiting for the fix (#1212, #977).
* If you are interested, as the original author of the `denymount` code (?), you could take over that PR (https://github.com/balena-io-modules/denymount/pull/8). If you did that, I already have an alternative CLI PR that passes the `executablePath` option in the call to `denymount`:
```
# Coffeescript code
    dmOpts = {}
    if process.pkg
        dmOpts.executablePath = path.join(path.dirname(process.execPath), 'denymount')
    denymount params.target, dmHandler, dmOpts
```
* We could do both things: merge this PR now, and improve the solution later.

dfunckt: @pdcastro I think the best option is to merge this one as-is, I don't see myself picking that up soon unfortunately.

pdcastro: @dfunckt that's fine, then all I need is your review approval so I can click the green merge button. :-)

dfunckt: @pdcastro those Linux checks are a relic of resin-ci going wild, they don't actually have to pass because they don't even make sense (it's a project specific to the Mac afterall, it doesn't compile anywhere else) and they're not required. All you care about are those travis-ci checks. So it still seems to be an option to update denymount and refer to it, but it's really up to you.

dfunckt: I didn't see your last comment. I've approved the other PR, and I'll also approve this one so you can go either way at will.

pdcastro: Ha - I hadn't seen your comment either :-) (about the `denymount` resin-ci tests not being required), only saw it after I merged this PR. I've now merged the `denymount` PR as well, as it's not a breaking change. I will update the CLI in a future PR.

PR #1215, opened 2019-05-08
Update balena ssh command to support local devices and multicontainer

CameronDiver:
Change-type: major

---
##### Contributor checklist


doc/cli.markdown
```diff
@@ -982,45 +982,46 @@ increase verbosity
...
 
-	$ balena ssh MyApp
-	$ balena ssh 7cf02a6
-	$ balena ssh 7cf02a6 --port 8080
-	$ balena ssh 7cf02a6 -v
-	$ balena ssh 7cf02a6 -s
-	$ balena ssh 7cf02a6 --noninteractive
+Warning: 'balena ssh' requires an openssh-compatible client to be correctly
+installed in your shell environment. For more information (including Windows
+support) please check the README here: https://github.com/balena-io/balena-cli
```
pdcastro: Not sure if it's better to point straight to the Additional Dependencies sections in INSTALL.md, as the instructions were moved there:
https://github.com/balena-io/balena-cli/blob/master/INSTALL.md#additional-dependencies

And if not, perhaps point directly to README.md instead of the repo's landing page. People who are not used to GitHub get confused about all the source code folders being displayed at the top of the page, and they don't realise that the README contents are displayed by scrolling down.

By the way, talking of "including Windows support", if you have a chance please test the ssh commands on [Windows](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/). Just yesterday Thodoris was raising it in support and this forum thread:
https://forums.balena.io/t/ssh-to-container-fails-with-sh-if-f-usr-bin-balena/7344

typings/bash.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'bash';
```
pdcastro: Does this work with `ts-node` v8.x (version introduced in v11-meta), when running `./bin/balena-dev`? My previous experience was that it was necessary to create a folder named after the module, something like `typings/bash/index.d.ts`. (According to changes introduced by the first PR to v11-meta.)

doc/cli.markdown
CameronDiver: I've moved the documentation link :+1:

Regarding the windows failure, I reproduced, and it suffers from the same problem. The issue as I see it is that we execute commands in a system-specific subshell, and in that subshell the bash escaping is used - which obviously doesn't work in windows. Working on a fix now. All of the other features work by default.

typings/bash.d.ts
CameronDiver: Done.

lib/utils/validation.ts
```diff
@@ -16,12 +16,13 @@ limitations under the License.
 
 import validEmail = require('@resin.io/valid-email');
 
-const APPNAME_REGEX = new RegExp(/^[a-zA-Z0-9_-]+$/);
+const APPNAME_REGEX = new RegExp(/^[a-za-z0-9_-]+$/);
```
hedss: Either this is mistakenly lowercased or ... it's mistakenly lowercased? :)

CameronDiver: What on earth.... No idea how that happened.

pdcastro:
@CameronDiver, in case of v11 rebasing / force pushing, heads up: something got merged to v11-meta: https://github.com/balena-io/balena-cli/pull/1201

doc/cli.markdown
```diff
@@ -982,45 +982,47 @@ increase verbosity
...
+	balena ssh f49cefd my-service
+	balena ssh f49cefd --port <port>
+
+	balena ssh 192.168.0.1 --verbose
+	balena ssh f49cefd.local my-service
+
+Warning: 'balena ssh' requires an openssh-compatible client to be correctly
+installed in your shell environment. For more information (including Windows
+support) please check the information here:
+	https://github.com/balena-io/balena-cli/blob/master/INSTALL.md#additional-dependencies
```
pdcastro: NIT:
From: `For more information (including Windows support) please check the information here:`
To: `For more information (including Windows support) please check:`

No need to change and re-push this PR, this can be changed in any other future PR (if it was to be changed at all).

CameronDiver: Changed.

PR #1220, opened 2019-05-13
Use TCP keepalive probes to detect local log stream closing

CameronDiver:
Change-type: patch
Closes: #1219

---
##### Contributor checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "@types/lodash": "4.14.112",
     "@types/mixpanel": "2.14.0",
     "@types/mkdirp": "0.5.2",
+    "@types/net-keepalive": "^0.4.0",
     "@types/node": "6.14.2",
     "@types/prettyjson": "0.0.28",
     "@types/raven": "2.5.1",
@@ -186,6 +187,7 @@
     "window-size": "^1.1.0"
   },
   "optionalDependencies": {
+    "net-keepalive": "^1.2.1",
     "removedrive": "^1.0.0",
     "windosu": "^0.2.0"
   }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -153,6 +154,7 @@
     "moment": "^2.24.0",
     "moment-duration-format": "^2.2.1",
     "mz": "^2.6.0",
+    "net-keepalive": "^1.2.1",
```
pdcastro: From `https://www.npmjs.com/package/net-keepalive`:
> The Missing (TCP_KEEPINTVL and TCP_KEEPCNT) SO_KEEPALIVE socket option setters and getters for Node using ffi module. Tested on linux, should work on osx and freebsd.

I think it is OK for this feature/fix to apply to Linux only (maybe macOS). The question is: does it raise any errors when ran on Windows? Because raising errors would be a problem. :-)

hedss: So, in Windows 10, this should work as expected as the `setsockopt()` includes both `TCP_KEEPINTVL` and `TCP_KEEPCNT`. However, in prior versions, these were set as registry keys and will simply be ignored.

I don't think this'll throw an error, but it's probably worth checking.

CameronDiver: Yeah I initially misread it to support the 3 big ones, testing on a windows vm now.

CameronDiver: This module isn't supported on windows. Will merge this try to find a method of achieving the same thing on Windows.

pdcastro: @CameronDiver, your laptop runs Linux, right? I offer to test it on macOS, just in case. I've tried it now, but it's failing for other reasons. I'll try again when the build tests are passing.

PR #1221, opened 2019-05-13
Debounce livepush invocations to collect changes together

CameronDiver:
Change-type: patch
Signed-off-by: Cameron Diver <cameron@balena.io

---
##### Contributor checklist

### DIFF
--- a/lib/utils/device/live.ts
+++ b/lib/utils/device/live.ts
@@ -1,7 +1,10 @@
 import * as Bluebird from 'bluebird';
 import * as chokidar from 'chokidar';
 import * as Dockerode from 'dockerode';
-import Livepush from 'livepush';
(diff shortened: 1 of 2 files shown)


lib/utils/device/live.ts
```diff
@@ -428,6 +406,14 @@ export class LivepushManager {
 				return [];
 		}
 	}
+
+	// For each service, get a debounced function
+	private getDebouncedEventHandle = _.memoize((serviceName: string) => {
```
hedss:
```suggestion
	private getDebouncedEventHandleFunc = _.memoize((serviceName: string) => {
```
Slight name change to make it obvious it's a function (I looked at where it was called first, just a but more explicit).

PR #1224, opened 2019-05-14
Update instructions for adding folder to path in MacOS

Jazzagi:
Change-type: patch

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,7 +17,7 @@ for installing and getting started with the balena CLI on Windows!
 3. Add the `balena-cli` folder to the system's `PATH` environment variable. See instructions for:
    [Windows](https://www.computerhope.com/issues/ch000549.htm) |
    [Linux](https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix) |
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -17,7 +17,7 @@ for installing and getting started with the balena CLI on Windows!
 3. Add the `balena-cli` folder to the system's `PATH` environment variable. See instructions for:
    [Windows](https://www.computerhope.com/issues/ch000549.htm) |
    [Linux](https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix) |
-   [macOS](https://stackoverflow.com/questions/22465332/setting-path-environment-variable-in-osx-permanently).
+   [macOS](https://www.architectryan.com/2012/10/02/add-to-the-path-on-mac-os-x-mountain-lion/#.Uydjga1dXDg)
```
pdcastro: Yeah, I think this is an improvement over the previous URL. Developers or some advanced users might prefer to modify the `~/.profile` or `~/.bash_profile` or `~/.bashrc` files, but chances are that those users would not need to refer to those instructions anyway...

Issue #1177, opened 2019-04-22
Should use the open-balena-api endpoints for device type & version info


thgreasi:
Depends-on: https://github.com/balena-io/balena-sdk/issues/664
Depends-on: https://github.com/balena-io/balena-sdk/pull/669
HQ: https://github.com/balena-io/balena/issues/1744

PR #1228, opened 2019-05-15
Use the open-balena-api endpoints for device type & version info

thgreasi:
New CLI installs will actually get the latest SDK version, but older CLI upgrades might not.
This just allows us to be sure that CLI installations > vx.x.x will not be using the img maker for fetching device types & OS versions.

Resolves: #1177
HQ: https://github.com/balena-io/balena/issues/1744
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "balena-device-init": "^5.0.0",
     "balena-image-manager": "^6.0.0",
     "balena-preload": "^8.1.3",
-    "balena-sdk": "^11.15.0",
+    "balena-sdk": "^11.17.0",
     "balena-settings-client": "^4.0.0",
     "balena-sync": "^10.0.3",
     "bash": "0.0.1",


pdcastro:
@thgreasi, is the `Change-type` minor rather than patch? hmm

thgreasi:
@pdcastro that's a weird question that opinions can vary.
Imo patches should be used for actual bug fixes and chores (eg updating a dev dependency, readme, etc).
On the other hand this is a notable underlying change, since the CLI will now start using a completely different backend for device types & versions.
I have a specific example in mind from a past all hands, where such a notable change was added to a component in v2.0.3, where the actual change was in a dependency of it. It sounded weird to say that the behavior changed on v2.0.3 and ppl were asking why that wasn't a minor change.
That's the rule of thumb that I have now.

Issue #1241, opened 2019-05-20
balena apps: TypeError: Cannot read property 'length' of undefined


pdcastro:
https://forums.balena.io/t/facing-issue-in-openbalena-configuration/7927/67
```
$ balena version
10.13.3
$ balena apps
TypeError: Cannot read property 'length' of undefined
    at /snapshot/balena-cli/build/actions/app.js:75:60
    at Array.forEach (native)
    at /snapshot/balena-cli/build/actions/app.js:74:20
    at tryCatcher (/snapshot/balena-cli/node_modules/bluebird/js/release/util.js:16:23)
    at Promise.module.exports.Promise._settlePromiseFromHandler (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise.module.exports.Promise._settlePromise (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise.module.exports.Promise._settlePromise0 (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise.module.exports.Promise._settlePromises (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:694:18)
    at Promise.module.exports.Promise._fulfill (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:638:18)
    at MappingPromiseArray.module.exports.PromiseArray._resolve (/snapshot/balena-cli/node_modules/bluebird/js/release/promise_array.js:126:19)
    at MappingPromiseArray.module.exports.MappingPromiseArray._promiseFulfilled (/snapshot/balena-cli/node_modules/bluebird/js/release/map.js:101:18)
    at MappingPromiseArray.module.exports.PromiseArray._iterate (/snapshot/balena-cli/node_modules/bluebird/js/release/promise_array.js:114:31)
    at MappingPromiseArray.init (/snapshot/balena-cli/node_modules/bluebird/js/release/promise_array.js:78:10)
    at Promise.module.exports.Promise._settlePromise (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:566:21)
    at Promise.module.exports.Promise._settlePromise0 (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise.module.exports.Promise._settlePromises (/snapshot/balena-cli/node_modules/bluebird/js/release/promise.js:694:18)
[...]

PR #1242, opened 2019-05-20
Fix TypeError when running 'balena apps'

pdcastro:
Resolves: #1241
Change-type: patch


pdcastro:
> How about using `_.size` and `_.filter` which do check against null & undefined?

Done

Issue #1243, opened 2019-05-20
balena preload: "Release not found" when the commit value is "latest"


pdcastro:
CLI counterpart to the balena-preload issue: https://github.com/balena-io/balena-preload/issues/204

`balena preload --commit latest` fails with "Release not found"

Known workaround: specify a commit hash value instead of the word "latest", e.g.:

`balena preload --commit 267408b`

This is believed to have been introduced by the following PR merged around 10 May 2019, and would affect several CLI versions that were installed via "npm install" since that date:
https://github.com/balena-io/balena-preload/pull/201/files

Page-:
@pdcastro the correct way to do it would be to use `balena preload` since if you don't specify a commit then it'll default to the application commit which will be the current release (and also the latest unless using staged-releases to delay roll outs)

pdcastro:
Thanks @Page-. I've tested omitting the `--commit` option and in this case the CLI presents an interactive menu with releases to choose from, including an option labeled "latest". If I select the "latest" option then indeed it works, but it's a different behaviour than `--commit latest` because the latter does not invoke an interactive menu and thus can be scripted.

If I edit the CLI’s `package.json` to pin `balena-preload` to version 8.1.1, then `--commit latest` does not raise any errors, nor raises an interactive menu. That’s how I narrowed the issue down to that PR. @zvin has pointed out that `--commit latest` selects the "application commit", not necessarily really the "latest" release:
https://github.com/balena-io/balena-cli/blob/master/lib/actions/preload.coffee#L289

Page-:
Yeah, I was basing it on if you don't pass a commit to balena-preload, but the cli interacts heavily with the balena-preload internals whilst mid-process so it doesn't act in the same way when going via the cli

pdcastro:
@zvin, reopening the issue because I am still getting "`Release not found`" with CLI v10.13.5 and `balena-preload` v8.1.4 if I use `--commit current`:
```
$ ./bin/balena-dev preload test.img --app test-rpi --commit current
Building Docker preloader image. [========================] 100%

| Creating preloader container
- Starting preloader container
| Fetching application test-rpi
| Reading image information
Release not found: fbeae9ac25cbf6c808f7d1905c26a38c
```
However it works if I use `--commit latest`:
```
$ ./bin/balena-dev preload ~/Downloads/dev-test-rpi-raspberrypi3-2.32.0+rev1-dev-v9.14.0.img --app test-rpi --commit latest
[...]

PR #1244, opened 2019-05-21
Update balena preload

zvin:
Resolves: #1243

### DIFF
--- a/lib/actions/preload.coffee
+++ b/lib/actions/preload.coffee
@@ -16,9 +16,10 @@ limitations under the License.
 
 dockerUtils = require('../utils/docker')
 
-LATEST = 'latest'
-
 allDeviceTypes = undefined
+
+isCurrent = (commit) ->
+	return commit == 'latest' or commit == 'current'
 
 getDeviceTypes = ->
 	Bluebird = require('bluebird')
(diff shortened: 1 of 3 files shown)


lib/actions/preload.coffee
```diff
@@ -159,8 +160,8 @@ module.exports =
 			signature: 'commit'
 			parameter: 'hash'
 			description: '''
-				the commit hash for a specific application release to preload, use "latest" to specify the latest release
-				(ignored if no appId is given)
+				the commit hash for a specific application release to preload, use "current" (or "latest") to specify
+				the current release (ignored if no appId is given)
```
pdcastro: I'm debating with myself whether mentioning "latest" does more harm than good... I expect that the vast majority of users will wonder what "current" means exactly, and mentioning "or latest" could hint that "current" means "most recent" (but it doesn't). I've thought of the following description:
```
the commit hash for a specific application release to preload. Use "current"
to refer to the current value of the API Application object's commit field.
(ignored if no appId is given)
```
Maybe we could have more feedback from people like... @imrehg and @dfunckt

dfunckt: I agree, the term "current" is more accurate and it just happens that usually also means "latest". This may be a good clarification to make. For example:

> [...] the commit hash for a specific application release to preload, use "current" to specify the current release (ignored if no appId is given). The current release is usually also the latest, but can be manually pinned using the CLI to set the `commit` field on the application.

Issue #1245, opened 2019-05-22
Add Windows and macOS native installers (beta)


pdcastro:
The native installers are built on [oclif's release capabilities](https://oclif.io/docs/releasing), patched to support native/binary Node.js modules on Windows and to fit our continuous integration setup.

The native installers will at least initially be uploaded to the GitHub releases page as usual, alongside the existing "standalone zip releases". To avoid confusion, the existing zip files are being renamed to include the "standalone" suffix, as follows:

* `balena-cli-v10.13.3-macOS-x64-installer.pkg`
* `balena-cli-v10.13.3-macOS-x64-standalone.zip`
* `balena-cli-v10.13.3-windows-x64-installer.exe`
* `balena-cli-v10.13.3-windows-x64-standalone.zip`

We're also using the opportunity to rename "darwin" to "macOS", as the former was confusing to many users. This covers issue #1080.

pdcastro:
Implemented, with the following additional issues for the installers to be signed:
* #1250
* #1251

Issue #1080, opened 2019-01-18
Standalone executables and documentation: rename 'darwin' or 'OSX' to MacOS


pdcastro:
Standalone executables are currently named 'balena-cli-version-darwin-x64.zip', and the standalone documentation refers to OSX. Rename both to 'MacOS', which is Apple's official name for the OS and more readily identifiable to many users.

jarrah31:
I agree, it shouldn't be called Darwin if it's for MacOS as it certainly confused me. Had to google which led me to this Issue.

pdcastro:
Resolved in CLI v11: https://github.com/balena-io/balena-cli/releases

PR #1249, opened 2019-05-22
Add native installers for Windows and macOS

pdcastro:
This PR provides the ability for the CLI CI build setup (Appveyor for Windows, Travis for macOS) to generate native installers based on a patched `oclif-dev` tool. See issue #1245 for a few more details. I have tested the builds and installers using my own GitHub fork of `balena-cli`. Check my releases page: https://github.com/pdcastro/balena-cli/releases

Connects-to: #1245
Connects-to: #1080
Change-type: minor


patches/denymount+2.2.0.patch
```diff
@@ -1,38 +0,0 @@
-diff --git a/node_modules/denymount/lib/index.js b/node_modules/denymount/lib/index.js
```
pdcastro: This patch was no longer necessary with the `denymount` version bump from 2.2.0 to 2.3.0.

lib/actions/local/configure.coffee
```diff
@@ -217,14 +218,20 @@ module.exports =
...
+				dmOpts.executablePath = path.join(path.dirname(process.execPath), 'denymount')
+			dmHandler = (cb) ->
 				reconfix.readConfiguration(configurationSchema, params.target)
 				.then(getConfiguration)
 				.then (answers) ->
 					if not answers.hostname
 						removeHostname(configurationSchema)
 					reconfix.writeConfiguration(configurationSchema, answers, params.target)
 				.asCallback(cb)
+			denymount params.target, dmHandler, dmOpts
```
pdcastro: These changes to `configure.coffee` were incidental to this PR. There was a `denymount` patch file in the `patches` folder that could be deleted with these changes.

patches/@oclif+dev-cli+1.22.0.patch
```diff
@@ -0,0 +1,146 @@
+diff --git a/node_modules/@oclif/dev-cli/lib/commands/pack/win.js b/node_modules/@oclif/dev-cli/lib/commands/pack/win.js
```
pdcastro: This patch files patches the `@oclif/dev-cli` package so that the `oclif-dev pack:win` command can run on Windows. (`oclif` was originally designed so `pack:win` "cross builds" on Linux with a bash shell, and would fail to run on Windows.) It was necessary or desirable for `pack:win` to run on Windows in order for "npm install" to build native Node.js modules for Windows.

.gitignore
```diff
@@ -40,6 +40,7 @@ balenarc.yml
 build/
 build-bin/
 build-zip/
+dist/
```
pdcastro: `oclif` places the native installers under the `dist` folder. I've changed the build scripts so that the standalone zip packages are placed in the dist folder as well. The `build-zip` folder is no longer used, but keeping it in `.gitignore` for now as people will still have that folder around as they update the CLI.

appveyor.yml
```diff
@@ -1,6 +1,8 @@
 # appveyor file
 # http://www.appveyor.com/docs/appveyor-yml
 
+image: Visual Studio 2017
```
pdcastro: Previously we were using the default image, "Visual Studio 2015", which was getting outdated. The old image ran on MS Windows Server 2012, whereas the new image runs on MS Windows Server 2016, which has better support for RDP (remote desktop) that was useful for debugging / evaluating the shell solutions (MSYS2).

appveyor.yml
```diff
@@ -14,21 +16,25 @@ matrix:
 # what combinations to test
 environment:
   matrix:
-    - nodejs_version: 8
+    - nodejs_version: 10
```
pdcastro: The CLI is also using balena CI's own VMs to run tests other versions of Node. While we rely on Appveyor to build the native installers using Node 10, it makes sense to make sure that the tests pass on Appveyor for Node 10.

appveyor.yml
```diff
@@ -14,21 +16,25 @@ matrix:
...
   - npm --version
-  - cmd: npm test
+  - npm test
 
 deploy_script:
-  - IF "%APPVEYOR_REPO_TAG%" == "true" (npm run release)
-  - IF NOT "%APPVEYOR_REPO_TAG%" == "true" (echo 'Not tagged, skipping deploy')
+  - node --version
+  - npm --version
+  - npm run build:installer
```
pdcastro: `npm run build:installer` generates the native installer for the current platform (in the case of an Appveyor VM, the Windows `'.exe'` installer is created).

.travis.yml
```diff
@@ -3,15 +3,20 @@ os:
...
 notifications:
   email: false
 deploy:
   - provider: script
-    script: npm run release
+    script:
+      - node --version
+      - npm --version
+      - npm run build:installer
+      - npm run release
```
pdcastro: * `npm run build:installer` generates the native installer for the current platform (in the case of a Travis macOS VM, a macOS `'.pkg'` installer is created; in the case of an Appveyor Windows VM, a Windows `'.exe'` installer is created).
* `npm run release` creates or updates a GitHub release, uploading the installer files.

package.json
```diff
@@ -31,21 +31,22 @@
     ]
   },
   "scripts": {
-    "prebuild": "rimraf build/ build-bin/ build-zip/ && patch-package",
-    "build": "npm run build:src && npm run build:bin",
+    "postinstall": "patch-package",
+    "prebuild": "rimraf build/ build-bin/",
+    "build": "npm run build:src",
```
pdcastro: * I've renamed `build:bin` to `build:standalone` to make it clearer that what it does is to create the standalone `pkg` bundles (which are later zipped by `npm release` -- hmm thinking about it, it would be even clearer if the zip files were also created by `build:standalone`).
* I've removed `build:bin` (now `build:standalone`) from `npm run build` (which is called by `npm test`). The bottom line is that neither `npm test` nor `npm publish` will create the standalone pkg packages (or the oclif native installers). The motivation was to save build time and make it consistent with native oclif installers.

Am I missing something? I hope not!

package.json
```diff
@@ -31,21 +31,22 @@
...
     "build:src": "npm run prettify && npm run lint && npm run build:fast && npm run build:doc",
     "build:fast": "gulp build && tsc",
     "build:doc": "mkdirp doc/ && ts-node --type-check -P automation/tsconfig.json automation/capitanodoc/index.ts > doc/cli.markdown",
-    "build:bin": "ts-node --type-check -P automation/tsconfig.json automation/build-bin.ts",
-    "release": "npm run build && ts-node --type-check -P automation/tsconfig.json automation/deploy-bin.ts",
+    "build:standalone": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:standalone",
+    "build:installer": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:installer",
```
pdcastro: `build:installer` builds the native executable oclif installers. Note that the "standalone zip packages" are not technically installers, they're just zip files that get extracted by the user to some folder. We do occasionally refer to the "standalone zip packages" as "standalone installation", though.

INSTALL.md
```diff
@@ -1,30 +1,72 @@
...
 
 * `balena` - should print the balena CLI help
 * `balena version` - should print the installed CLI version
 
+> Note: If you had previously installed the CLI using a standalone zip package, it may be a good
+> idea to check your system's `PATH` environment variable for duplicate entries, as the terminal
+> will use the entry that comes first. Check the [Standalone Zip Package](#standalone-zip-package)
+> instructions for how to modify the PATH variable.
+
+For your information only, by default the balena CLI is installed to the following folders:
```
hedss:
```suggestion
By default the balena CLI is installed to the following folders:  
```

automation/deploy-bin.ts
```diff
@@ -15,70 +15,131 @@
...
-import * as os from 'os';
 import * as path from 'path';
 import * as publishRelease from 'publish-release';
 
-import * as packageJSON from '../package.json';
-
-const publishReleaseAsync = Promise.promisify(publishRelease);
-const mkdirpAsync = Promise.promisify<string | null, string>(mkdirp);
-
+const mkdirpAsync = Bluebird.promisify<string | null, string>(mkdirp);
```
hedss: `mkdirp-promise` also exists.

automation/deploy-bin.ts
```diff
@@ -15,70 +15,131 @@
...
-	'build-zip',
-	`balena-cli-${version}-${os.platform()}-${os.arch()}.zip`,
-);
+const arch = process.arch;
+
+function dPath(...paths: string[]) {
+	return path.join(ROOT, 'dist', ...paths);
+}
+
+const standaloneZips: { [platform: string]: string } = {
```
hedss: I know they don't explicitly get used anywhere, but given this and the next three all end up as the same essential keyed type, might be nice to explicitly define that type.

automation/deploy-bin.ts
```diff
@@ -15,70 +15,131 @@
...
+				renamedOclifInstallers[process.platform],
+			);
+		}
+	}
+	try {
+		await createGitHubRelease();
+	} catch (err) {
 		console.error('Release failed');
 		console.error(err);
 		process.exit(1);
```
hedss: Possibly nicer as `process.exitCode = 1` unless there are other code paths that may require short-circuiting.

INSTALL.md
```diff
@@ -1,30 +1,72 @@
...
+> will use the entry that comes first. Check the [Standalone Zip Package](#standalone-zip-package)
+> instructions for how to modify the PATH variable.
+
+For your information only, by default the balena CLI is installed to the following folders:  
+
+OS  | Folders
+--- | ---
+Windows: | `C:\Program Files\balena-cli\`  
+macOS:   | `/usr/local/lib/balena-cli/`
+"        | `/usr/local/bin/balena`
```
shaunmulligan: what is the third row? linux?

pdcastro: Right -- if you had to ask the question, then clearly I need to change this! :-D
Markdown doesn't seem to allow line breaks inside a table cell, nor a cell to span multiple rows, but both paths applied to macOS. So I used two rows for macOS, and used the double quote as symbol to mean "idem" or "ditto". Now that's a funny thing: thinking about it, the double quote to represent "idem" comes from handwriting -- when I was student, it was common for teachers / lecturers to use a symbol that looks like a double quote when writing rows / tables on a blackboard with chalk. But I guess we don't see it much in the digital world, and I don't know if it's also a regional thing, like more common in some countries than others... :-)
Anyway, I'll reformat the text to avoid a Markdown table, I think.

INSTALL.md
pdcastro: Done

INSTALL.md
pdcastro: I was able to use the `<br>` tag inside a markdown table cell to achieve the line break.

INSTALL.md
```diff
@@ -1,30 +1,70 @@
...
-3. Add the `balena-cli` folder to the system's `PATH` environment variable. See instructions for:
-   [Windows](https://www.computerhope.com/issues/ch000549.htm) |
-   [Linux](https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix) |
-   [macOS](https://www.architectryan.com/2012/10/02/add-to-the-path-on-mac-os-x-mountain-lion/#.Uydjga1dXDg)
+_Please note: the executable installers are in **beta** status (recently introduced)._
+
+1. Download the latest installer from the [releases page](https://github.com/balena-io/balena-cli/releases).
+   Look for a file name that ends with "installer-BETA", for example:  
+   `balena-cli-v10.13.6-windows-x64-installer-BETA.exe`  
+   `balena-cli-v10.13.6-macOS-x64-installer-BETA.pkg`
```
pdcastro: I've made a change here: added the `-BETA` suffix to the native installer file names (the change was to the docs only, as the code was already generating files with the `-BETA` suffix). This is debatable though, maybe we don't want it. I was thinking that the we could remove the `-BETA` suffix when the installers get digitally signed and no longer raise Windows Defender and Apple Gatekeeper warnings of "unknown publisher" or "unidentified developer".

automation/deploy-bin.ts
```diff
@@ -15,70 +15,135 @@
...
+ */
+export async function release() {
+	console.log(`Creating release assets for CLI ${version}`);
+	try {
+		await zipStandaloneInstaller();
+	} catch (error) {
+		console.log(`Error creating standalone installer zip file: ${error}`);
+		process.exit(1);
+	}
+	if (process.platform === 'win32' || process.platform === 'darwin') {
```
CameronDiver: Why is this synchronous?

pdcastro: Done (I didn't even need to import `mz` for promises, as `deploy-bin` itself was already importing `fs-extra`!)

automation/deploy-bin.ts
pdcastro: All right Hedley, you win! I didn't even need to import `mkdirp-promise`, as I realised that `fs-extra` also provides `fs.mkdirp` with typing information and all, and it's already being imported in this same module... :-)

INSTALL.md
pdcastro: ha, Wikipedia knows about the "ditto mark"! https://en.wikipedia.org/wiki/Ditto_mark
:-)

PR #1265, opened 2019-05-29
Make livepush the default when pushing to a local device

CameronDiver:
Change-type: major

---
##### Contributor checklist

### DIFF
--- a/lib/actions/push.ts
+++ b/lib/actions/push.ts
@@ -110,7 +110,7 @@ export const push: CommandDefinition<
 		dockerfile?: string; // DeviceDeployOptions.dockerfilePath (alternative Dockerfile)
 		nocache?: boolean;
 		'registry-secrets'?: string;
-		live?: boolean;
+		nolive?: boolean;
 		detached?: boolean;
 		service?: string | string[];
 		system?: boolean;
@@ -139,9 +139,12 @@ export const push: CommandDefinition<
 		showing only the system logs can be achieved with --system. Note that these
 		flags can be used together.
 
-		It is also possible to run a push to a local mode device in live mode.
-		This will watch for changes in the source directory and perform an
-		in-place build in the running containers [BETA].
+		When pushing to a local device a live session will be started.
+		The project source folder is watched for filesystem events, and changes
+		to files and folders are automatically synchronized to the running
+		containers. The synchronisation is only in one direction, from this machine to
+		the device, and changes made on the device itself may be overwritten.
+		This feature requires a device running supervisor version v9.7.0 or greater.
 
 		${registrySecretsHelp.split('\n').join('\n\t\t')}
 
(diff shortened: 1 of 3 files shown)


lib/actions/push.ts
```diff
@@ -193,18 +197,12 @@ export const push: CommandDefinition<
...
 			description: stripIndent`
-				Note this feature is in beta.
-
-				Start a live session with the containers pushed to a local mode device.
-				The project source folder is watched for filesystem events, and changes
-				to files and folders are automatically synchronized to the running
-				containers. The synchronisation is only in one direction, from this machine to
-				the device, and changes made on the device itself may be overwritten.
-				This feature requires a device running supervisor version v9.7.0 or greater.`,
+				Don't run a live instance on this push. The filesystem will not be monitored, and changes
```
pdcastro: "live session" instead of "live instance"?

lib/utils/device/deploy.ts
```diff
@@ -46,7 +46,7 @@ export interface DeviceDeployOptions {
 	dockerfilePath?: string;
 	registrySecrets: RegistrySecrets;
 	nocache: boolean;
-	live: boolean;
+	nolive: boolean;
```
pdcastro: Minor: it would nice to accept and ignore a `--live` flag for users that have already been using it...

lib/utils/device/deploy.ts
```diff
@@ -149,10 +149,11 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 		if (!semver.satisfies(version, '>=7.21.4')) {
 			exitWithExpectedError(versionError);
 		}
-		if (opts.live && !semver.satisfies(version, '>=9.7.0')) {
-			exitWithExpectedError(
-				new Error('Using livepush requires a supervisor >= v9.7.0'),
+		if (!opts.nolive && !semver.satisfies(version, '>=9.7.0')) {
+			globalLogger.logWarn(
+				`Using livepush requires a supervisor >= 9.7.0. A live session will not be started.`,
```
pdcastro: Suggestion:
"Using livepush requires a **balena** supervisor **version** >= 9.7.0. A live session will not be started."

lib/utils/device/deploy.ts
CameronDiver: `--live` is already ignored by capitano

pdcastro: ha, good point! Capitano could use this example to claim that ignoring flags is a feature rather than a bug. :-D

lib/actions/push.ts
CameronDiver: Done

lib/utils/device/deploy.ts
CameronDiver: Done

lib/utils/device/deploy.ts
CameronDiver: Let's not go there... :P

PR #1272, opened 2019-05-30
Add warning notices for replaced 'local' commands in v11

pdcastro:
Change-type: major
Signed-off-by: Paulo Castro paulo@balena.io


pdcastro:
> CameronDiver: I'd definitely prefer the reverting of local flash and the warning methods to be in separate commits.

Done 👍

PR #1277, opened 2019-05-31
.gitignore: Add fast-boot.json generated by balena-dev command

thgreasi:
Change-type: patch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,4 +42,4 @@ build-bin/
 build-zip/
 
 # Ignore fast-boot cache file
-/bin/.fast-boot.json
+**/.fast-boot.json


.gitignore
```diff
@@ -43,3 +43,4 @@ build-zip/
 
 # Ignore fast-boot cache file
 /bin/.fast-boot.json
+.fast-boot.json
```
pdcastro: Is the `.fast-boot.json` file created on the "current working directory" of the `balena-dev` process? Perhaps a wildcard syntax like `**/.fast-boot.json` (to replace the existing `/bin/.fast-boot.json`) would be suitable.

thgreasi: Cool @pdcastro I will go for that :+1:

Issue #1281, opened 2019-06-04
ReferenceError: window is not defined (moment-duration-format)


pdcastro:
A dependency of the balena CLI, [moment-duration-format](https://www.npmjs.com/package/moment-duration-format) was udpated from v2.2.2 to v2.3.0 and this is causing the CLI to break. I've first noticed it during the CLI builds:

```
C:\Users\Administrator\concourse\work\volumes\live\52cab583-cff9-4400-44e1-0e7b3683d1d4\volume\node_modules\moment-duration-format\lib\moment-duration-format.js:1723
        if (window.Intl && window.Intl.NumberFormat) {
        ^
ReferenceError: window is not defined
...
```

I've also opened an issue in their repo: https://github.com/jsmreese/moment-duration-format/issues/131

Affects fresh **npm** installations of balena CLI version 10.17.4 and "all others before that".
I believe it does not affect installations with the standalone zip file.

pdcastro:
Anyone affected may also need to delete the `.fast-boot.json` file from the balena-cli's folder and/or the `./bin/` subfolder (i.e. `./bin/.fast-boot.json` subfolder). Otherwise, the CLI may fail with an unhelpful error message:
```
$ ./bin/balena version
null
```
Btw, @shaunmulligan, @thgreasi, that's a downside I have noticed with the `.fast-boot.json` file: when I switch branches I got in habit of deleting the `.fast-boot.json` file to avoid those `null` errors. But so far I am not aware of users reporting this error, I've only noticed it myself.

Martii:
> ... causing the CLI to break ...

And the API usage of *moment-duration-format*. Thanks for reporting this upstream as we just encountered this in a different project not too long ago. :) \*watching that issue\*

thgreasi:
@Martii this should now be resolved by https://github.com/balena-io/balena-cli/pull/1282 .
Can you try updating your balena-cli installation and give us a note if that's still an issue for you?

PR #1282, opened 2019-06-04
Pin moment-duration-format package (ReferenceError: window is not defined)

pdcastro:
Connects-to: #1281
Change-type: patch


pdcastro:
I'll be afk until late UK morning, so feel free to take over this PR or hit the merge button sooner than later (to reduce the number of affected users).

Issue #1283, opened 2019-06-04
Remove 'signup' command


pdcastro:
`resin-api` PR balena-io/resin-api/pull/1933 adds a signup captcha that prevents signup via the CLI. Use the major v11 release opportunity to remove the `signup` command. It is understood that a tiny teensy fraction of users ever used it ([flowdock thread](https://www.flowdock.com/app/rulemotion/resin-tech/threads/pVNnsJ4748R0e2TUwi9GBZKf10_)).

The `login` command still displays an option `"I don't have a balena account!"`, but it can open the signup URL on a browser (like it already does for the login page) instead of redirecting to the CLI's `signup` command:
```
$ ./bin/balena-dev login
 _            _
| |__   __ _ | |  ____  _ __    __ _
| '_ \ / _` || | / __ \| '_ \  / _` |
| |_) | (_) || ||  ___/| | | || (_) |
|_.__/ \__,_||_| \____/|_| |_| \__,_|

Logging in to balena-cloud.com
? How would you like to login?
  Web authorization (recommended)
  Credentials
  Authentication token
❯ I don't have a balena account!

Please sign up at https://dashboard.balena-cloud.com/signup
```

PR #1284, opened 2019-06-04
Remove 'signup' command

pdcastro:
Resolves: #1283 
Change-type: major


pdcastro:
force push: remove 'signup' (and other commands removed in previous pull requests) from the forgotten `balena-completion.bash` file.

Issue #1289, opened 2019-06-05
"npm install --production" fails on v11.0.1 with "sh: 1: patch-package: not found"


pdcastro:
```
$ npm i -g --production --unsafe-perm
...
sh: 1: patch-package: not found
...
```
`package.json` has `"postinstall": "patch-package"`, but the `patch-package` dependency is only listed in the `devDependencies` section. Currently the only packages that require patching are development packages, but the `postinstall` script runs on production installs too.

PR #1290, opened 2019-06-05
Fix "npm install --production" installation (missing patch-package dependency)

pdcastro:
Resolves: #1289 
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "preferGlobal": true,
   "files": [
-    "bin/run",
+    "bin/",
     "build/",
     "doc/",
     "lib/"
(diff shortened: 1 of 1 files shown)


pdcastro:
> This will also publish balena-dev, but I don't think that's a big deal

Indeed. It's related to the fact that we used to have a "./bin/balena" entrypoint, then it was renamed "./bin/run" to please oclif, then renamed back for "npm install" compatibility... If I individually list both 'run' and 'balena' in the files section:
```
  "files": [
    "bin/balena",
    "bin/run",
```
Then I fear error messages when 'run' is not there -- because 'run' is dynamically created (renamed) during the oclif native installer generation. If I list only "bin/balena", then I think oclif will fail to copy the required "bin/run". The common denominator seemed to be to list simply "bin/", for whatever is in there for whoever is interested (npm, oclif).

PR #1294, opened 2019-06-05
Add 'scan' command to the website reference documentation

pdcastro:
The 'local scan' command was renamed 'scan' in v11, and in this process it was unintentionally dropped from the CLI reference documentation.

Change-type: patch

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -137,8 +137,9 @@ If you come across any problems or would like to get in touch:
 
 	- [logs &#60;uuidOrDevice&#62;](#logs-uuidordevice)
 
-- SSH
+- Network
 
+	- [scan](#scan)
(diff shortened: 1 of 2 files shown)


doc/cli.markdown
```diff
@@ -137,8 +137,9 @@ If you come across any problems or would like to get in touch:
 
 	- [logs &#60;uuidOrDevice&#62;](#logs-uuidordevice)
 
-- SSH
+- Network
```
pdcastro: Documentation section renamed from 'SSH' to 'Network' to better fit the collection of these three commands: scan, ssh and tunnel. Previously, the 'local scan' command was in the 'Local' section. I suppose the 'scan' command could still be placed under the 'Local' section despite the rename, but I think that section is going to disappear in the not-too-distant future.

PR #1296, opened 2019-06-06
Add single code path to get full, online-only device UUIDs

richbayliss:
Both the tunnel and SSH commands require a full UUID for an online
device. A single code path was added to provide this, taking either
an application name or a partial UUID as a search parameter.

In the event of an application name being provided, a device select
form is presented to the user to pick from the online devices at that
time.

Change-type: patch

### DIFF
--- a/lib/actions/ssh.ts
+++ b/lib/actions/ssh.ts
@@ -17,84 +17,8 @@ import * as BalenaSdk from 'balena-sdk';
 import { CommandDefinition } from 'capitano';
 import { stripIndent } from 'common-tags';
 
-import { BalenaApplicationNotFound, BalenaDeviceNotFound } from 'balena-errors';
-import {
-	validateApplicationName,
-	validateDotLocalUrl,
-	validateIPAddress,
-	validateShortUuid,
-	validateUuid,
-} from '../utils/validation';
-
-enum SSHTarget {
-	APPLICATION,
-	DEVICE,
-	LOCAL_DEVICE,
-}
-
-async function getSSHTarget(
-	sdk: BalenaSdk.BalenaSDK,
-	applicationOrDevice: string,
-): Promise<{
-	target: SSHTarget;
-	deviceChecked?: boolean;
-	applicationChecked?: boolean;
-	device?: BalenaSdk.Device;
-} | null> {
-	if (
-		validateDotLocalUrl(applicationOrDevice) ||
-		validateIPAddress(applicationOrDevice)
-	) {
-		return { target: SSHTarget.LOCAL_DEVICE };
-	}
-
-	const appTest = validateApplicationName(applicationOrDevice);
-	const uuidTest = validateUuid(applicationOrDevice);
-	if (appTest || uuidTest) {
-		// Do some further processing to work out which it is
-		if (appTest && !uuidTest) {
-			return {
-				target: SSHTarget.APPLICATION,
-				applicationChecked: false,
-			};
-		}
-		if (uuidTest && !appTest) {
-			return {
-				target: SSHTarget.DEVICE,
-				deviceChecked: false,
-			};
-		}
-
-		// This is the harder part, we have a string that
-		// fulfills both the uuid and application name
-		// requirements. We should go away and test for both a
-		// device with that uuid, and an application with that
-		// name, and choose the appropriate one
-		try {
-			await sdk.models.application.get(applicationOrDevice);
-			return { target: SSHTarget.APPLICATION, applicationChecked: true };
-		} catch (e) {
-			if (e instanceof BalenaApplicationNotFound) {
-				// Here we want to check for a device with that UUID
-				try {
-					const device = await sdk.models.device.get(applicationOrDevice, {
-						$select: ['id', 'uuid', 'supervisor_version', 'is_online'],
-					});
-					return { target: SSHTarget.DEVICE, deviceChecked: true, device };
-				} catch (err) {
-					if (err instanceof BalenaDeviceNotFound) {
-						throw new Error(
-							`Device or application not found: ${applicationOrDevice}`,
-						);
-					}
-					throw err;
-				}
-			}
-			throw e;
-		}
-	}
-	return null;
-}
+import { BalenaDeviceNotFound } from 'balena-errors';
+import { validateDotLocalUrl, validateIPAddress } from '../utils/validation';
 
 async function getContainerId(
 	sdk: BalenaSdk.BalenaSDK,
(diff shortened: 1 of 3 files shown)


lib/actions/tunnel.ts
```diff
@@ -129,9 +127,11 @@ export const tunnel: CommandDefinition<Args, Options> = {
 				? (options.port as string[])
 				: [options.port as string];
 
-		return inferOrSelectDevice(params.uuid)
+		return getTargetUuid(sdk, params.deviceOrApplication)
 			.then(sdk.models.device.get)
 			.then(device => {
+				logger.logInfo(`Tunnel to ${device.uuid}`);
```
CameronDiver: I know this isn't technically a change, but maybe something like `Opening a tunnel to ${device.uuid}...` would be better?

lib/utils/patterns.ts
```diff
@@ -282,6 +283,69 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+	const uuidTest = validation.validateUuid(applicationOrDevice);
+
+	if (!appTest && !uuidTest) {
+		throw new Error(`Device or application not found: ${applicationOrDevice}`);
+	}
+
+	const isBalenaApplicationNotFound = (e: any) =>
+		e instanceof BalenaApplicationNotFound;
+
+	// if we have a definate device UUID...
```
CameronDiver:
```suggestion
	// if we have a definite device UUID...
```

lib/utils/patterns.ts
```diff
@@ -282,6 +283,69 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+	if (uuidTest && !appTest) {
+		return sdk.models.device
```
CameronDiver: We can potentially skip hitting the api, if we first test for it being a long uuid

lib/utils/patterns.ts
```diff
@@ -282,6 +283,69 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+	return sdk.models.application
+		.get(applicationOrDevice)
+		.then(app => {
+			return sdk.models.device.getAllByApplication(app.id, {
+				$filter: { is_online: true },
+			});
+		})
+		.then(devices => {
+			if (_.isEmpty(devices)) {
+				throw new Error('No target devices are online');
```
CameronDiver:
```suggestion
				throw new Error('No accessible devices are online');
```

lib/utils/patterns.ts
```diff
@@ -282,6 +283,69 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+				type: 'list',
+				default: devices[0].uuid,
+				choices: _.map(devices, device => ({
+					name: `${device.device_name || 'Untitled'} (${device.uuid.slice(
+						0,
+						7,
+					)})`,
+					value: device.uuid,
+				})),
+			}) as string;
```
CameronDiver: This cast is slightly weird, it's actually a promise that this function returns.

lib/utils/patterns.ts
```diff
@@ -282,6 +283,69 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+		e instanceof BalenaApplicationNotFound;
+
+	// if we have a definate device UUID...
+	if (uuidTest && !appTest) {
+		return sdk.models.device
+			.get(applicationOrDevice, { $select: ['uuid'] })
+			.then(d => d.uuid);
+	}
+
+	// otherwise, it may be a device OR and application...
```
CameronDiver:
```suggestion
	// otherwise, it may be a device OR an application...
```

lib/utils/patterns.ts
```diff
@@ -282,6 +283,69 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+): Bluebird<string> {
+	const appTest = validation.validateApplicationName(applicationOrDevice);
+	const uuidTest = validation.validateUuid(applicationOrDevice);
+
+	if (!appTest && !uuidTest) {
+		throw new Error(`Device or application not found: ${applicationOrDevice}`);
+	}
+
+	const isBalenaApplicationNotFound = (e: any) =>
+		e instanceof BalenaApplicationNotFound;
```
CameronDiver: This isn't necessary, you can pass the `BalenaApplicationNotFound` constructor directly the the first parameter of `.catch`

lib/utils/patterns.ts
hedss: If you used an `await` this would happily resolve itself.

lib/utils/patterns.ts
richbayliss: @CameronDiver Since this is getting a target UUID, the resulting UUID should be presumed to be online... if this is wrong, I think we should remove the `is_online` filter from each API lookup in this method, so it just returns valid/existing UUIDs irrespective of their online status. Otherwise the caller will not know if they should also check for the `is_online` status once getting the result.

CameronDiver: Hm, yeah I know what you mean, I tried to reconcile this in ssh by keeping track of if we've done an online check by that point (for example, when going from short uuid -> long uuid, also do an online check and keep the result). This worked pretty well but it made the code a little confusing. Perhaps you're right, and we should just always do the online check after determining the target.

richbayliss: I renamed the method to `getOnlineTargetUuid` so it will only return a UUID for an online device. This makes it very clear what you will get, so I think that makes it better.

richbayliss:
> it'd be nice to drop the promise chains and `async/await` if possible.

@hedss what do we have against promise chains? I genuinely think that are really nice to read, but if I am missing something then please let me know 👍 

> Also, the PR title.... sigh. :)

This started out as a quick fix to 1 issue, but it's escalated to cover more things. I will be re-wording it 😉

lib/actions/ssh.ts
```diff
@@ -350,155 +349,95 @@ export const ssh: CommandDefinition<
...
-				}
+		// this will be a tunnelled SSH connection...
+		exitIfNotLoggedIn();
+		const uuid = await getOnlineTargetUuid(sdk, params.applicationOrDevice);
 
-				if (!isOnline) {
-					throw new Error(`Device ${uuid} is not online.`);
-				}
+		// We want to do two things here; firstly, check
+		// that the device exists and is accessible
```
CameronDiver: Comment went funky here

lib/actions/ssh.ts
```diff
@@ -350,155 +349,95 @@ export const ssh: CommandDefinition<
...
-				]);
-				const proxyCommand = getSshProxyCommand(hasTunnelBin);
-
-				if (username == null) {
-					exitWithExpectedError(
-						`Opening an SSH connection to a remote device requires you to be logged in.`,
-					);
-				}
+		// We also want to avoid checking for a device if we
+		// know it exists
```
CameronDiver: This doesn't seem to match the code below anymore

richbayliss: removed it

lib/actions/ssh.ts
richbayliss: removed it, the code is clearer now so no need to be explicit

Issue #1251, opened 2019-05-22
macOS CLI installer can't be opened because it is from an unidentified developer


pdcastro:
The native / executable CLI installer for macOS is interrupted by a dialog that says the installer _"can't be opened because it is from an unidentified developer"._ We are investigating how to get the installer digitally signed in our continuous integration builds. Meanwhile, the following screenshots show how to get past that pop-up dialog:

![gatekeeper1](https://user-images.githubusercontent.com/15091591/58212295-3aa93200-7ce6-11e9-8a84-640a10aa61ba.png)
Open System Preferences by clicking on the Apple menu at the top left corner of the screen:

<img width="235px" alt="apple_menu" src="https://user-images.githubusercontent.com/15091591/58212302-3e3cb900-7ce6-11e9-9fae-d531d23d67b6.jpg"/>
<img width="672" alt="system_preferences" src="https://user-images.githubusercontent.com/15091591/58212308-439a0380-7ce6-11e9-8113-ae3e7620e318.png"/>

![security_and_privacy](https://user-images.githubusercontent.com/15091591/58212316-472d8a80-7ce6-11e9-8b6a-a87857c88e05.jpg)
![gatekeeper2](https://user-images.githubusercontent.com/15091591/58212321-4ac11180-7ce6-11e9-9d6c-a04eabaa06f1.png)

pdcastro:
"Signing Your Apps for Gatekeeper": https://developer.apple.com/developer-id/

zrzka:
Related issue https://github.com/balena-io/balena/issues/1605

Issue #1250, opened 2019-05-22
Windows Defender SmartScreen and User Account Control interrupt CLI installation


pdcastro:
The native / executable CLI installer for Windows is interrupted twice when installing on Windows 10, first by "Windows Defender SmartScreen" and then again by the Windows User Account Control (UAC). We are investigating how to get the installer digitally signed in our continuous integration builds, to prevent this from happening. Meanwhile, the following screenshots show how to get past the pop-up dialogs:

![defender1](https://user-images.githubusercontent.com/15091591/58206159-18101c80-7cd8-11e9-8de0-f63ece543304.png)
![defender2](https://user-images.githubusercontent.com/15091591/58206167-1b0b0d00-7cd8-11e9-9003-4efe615a03c1.png)
![access_control](https://user-images.githubusercontent.com/15091591/58206178-1d6d6700-7cd8-11e9-8772-d5bd4a62216b.png)

pdcastro:
Apparently this MS Visual Studio "Sign Tool" will be part of the solution:
https://docs.microsoft.com/en-us/dotnet/framework/tools/signtool-exe

PR #1298, opened 2019-06-06
Integrate with balena CI (installer signing)

pdcastro:
Moving away from Travis / Appveyor, towards balena CI.
Also provide digital signing for Windows and macOS installers.

Connects-to: #1250
Connects-to: #1251
Change-type: minor


nazrhom:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

nazrhom:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

package.json
```diff
@@ -67,7 +68,8 @@
     "bin": "balena",
     "commands": "./build/actions-oclif",
     "macos": {
-      "identifier": "io.balena.cli"
+      "identifier": "io.balena.cli",
+      "sign": "Developer ID Installer: Rulemotion Ltd (66H43P8FRG)"
```
pdcastro: This matches the signing certificate obtained from Apple a few days ago.

package.json
```diff
@@ -39,6 +39,7 @@
     "build:doc": "mkdirp doc/ && ts-node --type-check -P automation/tsconfig.json automation/capitanodoc/index.ts > doc/cli.markdown",
     "build:standalone": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:standalone",
     "build:installer": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:installer",
+    "package": "npm run build:fast && npm run build:standalone && npm run build:installer",
```
pdcastro: balena CI calls `npm run package` as one of the CI steps, following `npm test`.

automation/run.ts
```diff
@@ -89,6 +61,15 @@ export async function run(args?: string[]) {
 	// the current working dir becomes the MSYS2 homedir, so we change back.
 	process.chdir(ROOT);
 
+	if (process.platform === 'win32' && !process.env.BUILD_TMP) {
+		const randID = require('crypto')
+			.randomBytes(6)
+			.toString('base64')
+			.replace(/\+/g, '-')
+			.replace(/\//g, '_'); // base64url (RFC 4648)
+		process.env.BUILD_TMP = `C:\\tmp\\${randID}`;
```
pdcastro: We did this to get around the 260-character limit of paths under Windows.

automation/build-bin.ts
```diff
@@ -104,9 +264,13 @@ export async function buildOclifInstaller() {
 			await Bluebird.fromCallback(cb => rimraf(dir, cb));
 		}
 		console.log('=======================================================');
-		console.log(`oclif-dev "${packCmd}" [${packOpts}]`);
+		console.log(`oclif-dev "${packCmd}" "${packOpts.join('" "')}"`);
+		console.log(`cwd="${process.cwd()}" ROOT="${ROOT}"`);
```
pdcastro: This was added for debugging but I'd just leave it there. The build output has thousands and thousands of output lines... Having these few extra `console.log` won't hurt and can be quite useful to understand and reproduce some issues in `fly intercept` shell sessions.

automation/run.ts
nazrhom: Lets add this as a comment in the code as well, so we don't risk breaking it in the future once we forget :)

automation/build-bin.ts
```diff
@@ -77,14 +149,101 @@ export async function buildPkg() {
...
+			'-d',
+			`balena-cli ${version}`,
+		]);
+	} else {
+		console.log(
+			'Skipping installer signing step because CSC_* env vars are not set',
+		);
+	}
+}
+
```
nazrhom: We already discussed to add a comment here explaining the differences between what oclif does in osx vs win. Just adding this as a reminder

automation/deploy-bin.ts
```diff
@@ -122,21 +49,6 @@ export async function createGitHubRelease() {
...
-	if (process.platform === 'win32' || process.platform === 'darwin') {
-		if (await fs.pathExists(oclifInstallers[process.platform])) {
-			await fs.rename(
-				oclifInstallers[process.platform],
-				renamedOclifInstallers[process.platform],
-			);
-		}
-	}
 	try {
 		await createGitHubRelease();
```
nazrhom: Aren't we creating the release from the CI in https://github.com/balena-io/resin-concourse/blob/node-cli/pipelines/github-events/template.yml#L912-L916 ?

repo.yml
```diff
@@ -0,0 +1 @@
+type: node-cli
```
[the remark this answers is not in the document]
nazrhom: More of a comment for another PR; but since we have the file, we could start recording upstream dependencies in here, this would enable nested changelogs in the CLI. https://github.com/balena-io/pinejs/blob/master/repo.yml shows how these can be setup (requires some setup on the upstream repos as well)

automation/build-bin.ts
pdcastro: Good that you raised this, because on review I see that setting the OSX_KEYCHAIN env var at that point was useless (at that point, `pkgbuild` had already run). I've deleted that code and added a comment to explain that the Windows installer is explicitly signed by a script invoked by balena-cli, whereas the macOS installer is signed by oclif which runs the `pkbuild` utility.

automation/deploy-bin.ts
pdcastro: Yes! This function is only called by Travis or Appveyor via the `npm release` command. Balena CI calls `npm package`, which does not call `createGitHubRelease()`. It should be OK to delete the whole `deploy-bin.ts` file (as it currently stands) when we permanently switch off Travis and Appveyor.

repo.yml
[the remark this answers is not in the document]
pdcastro: I see, thanks! It sounds like this would complement an effort by @imrehg in adding the `.versionbot/CHANGELOG.yml` file. I think it will be great to have more informative changelogs.

pdcastro:
@balena-ci retest

automation/build-bin.ts
```diff
@@ -24,6 +25,43 @@ import { exec as execPkg } from 'pkg';
 import * as rimraf from 'rimraf';
 
 export const ROOT = path.join(__dirname, '..');
+// Note: the following 'tslint disable' line was only required to
+// satisfy ts-node under Appveyor's MSYS2 on Windows -- oddly specific.
+// Maybe something to do with '/' vs '\' in paths in some tslint file.
+// tslint:disable-next-line:no-var-requires
+export const packageJSON = require(path.join(ROOT, 'package.json'));
```
zvin: You can import json files in typescript now https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#new---resolvejsonmodule

automation/build-bin.ts
```diff
@@ -77,14 +120,67 @@ export async function buildPkg() {
...
+ */
+async function zipPkg() {
+	const outputFile = standaloneZips[process.platform];
+	if (!outputFile) {
+		throw new Error(
+			`Standalone installer unavailable for platform "${process.platform}"`,
+		);
+	}
+	await fs.mkdirp(path.dirname(outputFile));
+	await new Bluebird((resolve, reject) => {
```
zvin: A regular Promise should work.

pdcastro:
@balena-ci retest

automation/build-bin.ts
```diff
@@ -174,6 +203,29 @@ async function renameInstallerFiles() {
 	}
 }
 
+/**
+ * If the CSC_LINK and CSC_KEY_PASSWORD env vars are set, digitally sign the
+ * executable installer by running the balena-io/scripts/shared/sign-exe.sh
+ * script (which must be in the PATH) using a MSYS2 bash shell.
+ */
+async function signWindowsInstaller() {
```
zvin: Why not:

```typescript
import { execFile } from 'child_process';
import { promisify } from 'util';

const execFileAsync = promisify(execFile);

await execFileAsync(
    'C:\\msys64\\usr\\bin\\bash.exe',
    ['sign-exe.sh, '-f', exeName, '-d', `balena-cli ${version}`],
)
```
?
You don't need to escape anything this way.

nazrhom:
@balena-ci retest

automation/build-bin.ts
pdcastro: Yes, the feature is even enabled in `automation/tsconfig.json`, I tried it before... Tried again now, today I got:
```
Module '"balena-io/balena-cli/package"' can only be default-imported using the 'esModuleInterop' flagts(1259)
package.json(1, 1): This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```
... and I didn't want to spend more time on this line. :-)

automation/build-bin.ts
pdcastro: Done

automation/build-bin.ts
pdcastro: Thanks. I understand that the main advantage of this suggestion is to avoid the `bash -c` argument, which runs commands from a string and creates the need for escaping arguments with the `shell-escape` package.  This suggestion may work for running shell scripts like `sign-exe.sh`, but the `runUnderMsys()` function also runs other non-shell executables, like re-running `npm run build:installer` in a MSYS2 bash "environment" (`runUnderMsys()` is also called by the `run()` function in `run.ts`). In that case, `bash -c` is required, and escaping becomes necessary.
So yes, the suggestion would probably work well for `signWindowsInstaller()`, but we would still need to use `shell-escape` in `runUnderMsys()` for other cases. For this reason, I'd just leave `signWindowsInstaller()` as it is.

.travis.yml
```diff
@@ -4,12 +4,15 @@ os:
   - osx
 node_js:
   - "10"
+matrix:
+  exclude:
+    node_js: "10"
```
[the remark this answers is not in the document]
pdcastro: hmm sounds like the main benefit would be to have artefacts produced by both balena CI and Travis / Appveyor, but currently they are only uploaded to the GitHub releases page (not saved elsewhere), so some changes would be needed. I think that having a separate branch would be extra work... If balena CI fails to produce working assets at any point, I think it's easy enough for me to revert the changes to the yaml files and quickly create a new CLI release.

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

automation/build-bin.ts
pdcastro: @zvin, I am convinced that you jinxed it! :-) Calling `runUnderMsys()` to sign the Windows installer had been working fine for a while (2 weeks?), until about the time when posted this comment. Then, otherwise inexplicably, it started failing: apparently the PATH env var was somehow between modified in the child process, so that the `sign-exe.sh` could no longer find `signtool.exe`. But invoking bash via `execFile`, as you suggested, prevented the PATH modification and it's now working again. So I've adopted your suggestion because it is actually fixing a problem!

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

nazrhom:
@balena-ci retest

nazrhom:
@balena-ci retest

PR #1305, opened 2019-06-12
livepush: Ignore the .git directory when performing a livepush

CameronDiver:
Change-type: patch

---
##### Contributor checklist


lib/utils/device/live.ts
```diff
@@ -159,6 +159,7 @@ export class LivepushManager {
 				const monitor = chokidar.watch('.', {
 					cwd: context,
 					ignoreInitial: true,
+					ignored: '.git'
```
pdcastro: The build is failing -- I think that `npm run prettify` fixes it.

Issue #1164, opened 2019-04-12
support build time secrets with balena push <deviceIP>


shaunmulligan:
Currently `balena push <deviceIP>` doesn't support the `.balena` folder and the build time secrets and variables as documented here: https://github.com/balena-io/docs/pull/1079/files#diff-d8ad4020e4b4355ea278e3ea3c424f7fR1

CameronDiver:
Looks like this was closed in error, I'm not sure why. Re-opening.

CameronDiver:
The groundwork for this has been done in this PR: https://github.com/balena-io-modules/resin-multibuild/pull/53

PR #1314, opened 2019-06-18
If a secrets file is not specified, read it from the data directory

CameronDiver:
Change-type: minor
Closes: #1164

---
##### Contributor checklist

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -21,6 +21,7 @@ import * as MultiBuild from 'resin-multibuild';
 import { Readable } from 'stream';
 import * as tar from 'tar-stream';
 
+import { BalenaSDK } from 'balena-sdk';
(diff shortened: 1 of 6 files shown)


lib/utils/compose_ts.ts
```diff
@@ -31,7 +32,34 @@ export interface RegistrySecrets {
...
+
+	const directory = await sdk.settings.get('dataDirectory');
+	const potentialPaths = [
+		Path.join(directory, 'secrets.yml'),
+		Path.join(directory, 'secrets.yaml'),
+		Path.join(directory, 'secrets.json'),
+	];
+
+	for (const path of potentialPaths) {
+		if (await fs.exists(path)) {
```
zvin: NIT: It would be safer to `parseRegistry` secrets without checking for `fs.exists` and catch `ENOENT`.

Issue #1293, opened 2019-06-05
Commit a 'npm-shrinkwrap.json' file to control dependency updates


pdcastro:
Often enough, the CLI breaks because someone, somewhere out there, merges a pull request in a dependency package that has some issue on some relevant version of Node. So far, the argument for not committing a [package-lock.json](https://docs.npmjs.com/files/package-lock.json) file was that the npm repository discards it, so that when CLI users run "npm install -g" (without a cloned balena-cli repo), they're installing the CLI without a `package-lock.json` file, regardless of whether that file was committed to the repo. If we committed `package-lock.json`, the CLI's developers and the continuous integration tests would be effectively running the CLI in a different environment than end users. Recently, we even committed a `.npmrc` file with `package-lock=false` so that `package-lock.json` is never ever created (issue #1275). :-)

> https://docs.npmjs.com/files/package-lock.json
> _One key detail about package-lock.json is that it cannot be published [...]_

But the npm repo does recognise and recommend the use of a `npm-shrinkwrap.json` file:

> https://docs.npmjs.com/files/shrinkwrap.json
> _npm-shrinkwrap.json is a file created by npm-shrinkwrap. It is identical to package-lock.json, with one major caveat: Unlike package-lock.json, npm-shrinkwrap.json may be included when publishing a package._
> _The recommended use-case for npm-shrinkwrap.json is applications deployed through the publishing process on the registry: for example, daemons and command-line tools intended as global installs or devDependencies._

pdcastro:
When this is done, review pinned dependencies (including patch pinning with `'~'`) such as `moment-duration-format` (CC: @Page-, @horia-delicoti).

pdcastro:
Having an `npm-shrinkwrap.json` file would have probably also avoided issue #1322 _(Email prompt hangs on `balena login` credentials menu):_ `balena login` broke on CLI v11.3.2 and got fixed on CLI v11.3.4 "all by itself", so likely a case of dependencies getting broken and fixed outside of our control. But fear not: I now have a pull request for this issue: #1326

pdcastro:
Additional pull request: #1337

PR #1326, opened 2019-06-26
Add npm-shrinkwrap.json to control dependency updates

pdcastro:
Resolves: #1293
Change-type: patch


package.json
```diff
@@ -167,7 +167,7 @@
     "mixpanel": "^0.10.1",
     "mkdirp": "^0.5.1",
     "moment": "^2.24.0",
-    "moment-duration-format": "~2.2.2",
+    "moment-duration-format": "^2.2.2",
```
thgreasi: I think that 2.3 broke the typings but for sure we now don't have any reason to not change it to a `^`.
After all, whoever tries to change the shrinkwrap will see whether it breaks the build.

pdcastro:
@balena-ci retest

package.json
pdcastro: I didn't know about the typings, but I knew about 2 issues that I raised myself in the upstream repo:
* [ReferenceError: window is not defined in v2.3.0](https://github.com/jsmreese/moment-duration-format/issues/131)
* [ReferenceError: window is not defined in v2.3.1](https://github.com/jsmreese/moment-duration-format/issues/133)

... which had broken balena CLI and required us to pin it to `~2.2.2`. Upstream claims to have fixed it but I have not tested it myself. I was careful to add `npm-shrinkwrap.json` in a separate previous commit to the unpinning, so the CLI should still use `moment-duration-format` v2.2.2 anyway.

.npmrc
```diff
@@ -1 +0,0 @@
-package-lock=false
```
pdcastro: I've found that `package-lock=false` also disables `npm-shrinkwrap.json`, hence deleting this line and the resulting blank file.

thgreasi:
It seems that we will have to cut a release for https://github.com/balena-io-modules/node-web-streams as `@balena/node-web-streams` and update the respective line in balena request (we did something like this before for [grid-styled](https://github.com/balena-io-modules/grid/blob/ebc0bc3303679841ff9d4eed0e7896ba93b37e2d/package.json#L2)).
See: https://github.com/balena-io-modules/balena-request/blob/4f33d99336a0c697d25980aa3132cd2de18bbc65/package.json#L59

pdcastro:
> _It seems that we will have to cut a release for https://github.com/balena-io-modules/node-web-streams as @balena/node-web-streams_

@thgreasi, although I couldn't reproduce the issue on Friday as in the [Flowdock discussion](https://www.flowdock.com/app/rulemotion/resin-frontend/threads/aPVQ3rIDGMGUN7amkz8iaQRa2Bx), I think there was indeed a problem that even affected users: see issues #1331 and #1332. Please review pull request #1334 that hopefully fixes the issue.

PR #1337, opened 2019-07-02
Add npm-shrinkwrap in package.json so that it gets published to the npm registry

nazrhom:
Change-type: patch
Signed-off-by: Giovanni Garufi <giovanni@balena.io>

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "bin/",
     "build/",
     "doc/",
-    "lib/"
+    "lib/",
+    "npm-shrinkwrap.json"
   ],
   "bin": {
     "balena": "./bin/balena"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,6 +14,8 @@
     "build/",
     "doc/",
     "lib/",
+    "patches/",
+    "*.md",
```
pdcastro: The `patches` folder was also missing, and I've added `*.md` too. Without `*.md`, by default npm publishes the CHANGELOG, README and LICENSE files, and with `*.md` we further publish the INSTALL and TROUBLESHOOTING files, which I thought could help someone with installation issues (if they are browsing that folder, chances are they are trying to solve some issue...).

Issue #1340, opened 2019-07-08
Include sshKeys in os configure


mwohlert:
When supplying a config.json with sshKeys in it, they are not written onto the ssh partition, since they are omitted while generating the config, that is written onto the sd card.

@CameronDiver Since you pointed me to the correct spots on the pr with balena-supervisor. I would like to help here aswell, since we also have the need to have this working.

My investigation so far:
os configure, calls generateApplicationConfig or generateDeviceConfig in config.ts and they call generateBaseConfig, which calls balena.sdk model.os.getConfig, this config does of course not have any ssh keys in it.
My idea is to add the sshKeys to the config outputted by generateBaseConfig, when they are specified in the options.

pdcastro:
Thanks for PR @mwohlert ! Merged and [released as CLI v11.6.0](https://github.com/balena-io/balena-cli/releases).

MBerka:
This one absolutely saved production deployability by making it possible to stop manually adding the keys to the disk image. Thank you.

ashwinvandijk:
I'm not sure if I'm missing something here, or what I want to do isn't possible. I'd like to issue a command to create a config, with a specified ssh-key already in there. Is there a way to do this?

PR #1341, opened 2019-07-08
Add support for sshKeys in config file with os configure

mwohlert:
This applies specified os.sshKeys inside a config.json supplyed during os configure.

Resolves: #1340 
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/lib/utils/config.ts
+++ b/lib/utils/config.ts
@@ -46,6 +46,10 @@ interface ImgConfig {
 	deviceId?: number;
 	uuid?: string;
 	registered_at?: number;
+
+	os?: {
+		sshKeys?: string[];
+	};
 }
 
 export function generateBaseConfig(
@@ -54,6 +58,9 @@ export function generateBaseConfig(
 		version: string;
 		appUpdatePollInterval?: number;
 		deviceType?: string;
+		os?: {
+			sshKeys?: string[];
+		}
 	},
 ): Promise<ImgConfig> {
 	options = {
@@ -68,6 +75,15 @@ export function generateBaseConfig(
 	return promise.tap(config => {
 		// os.getConfig always returns a config for an app
 		delete config.apiKey;
+
+		// merge sshKeys to config, when they have been specified
+		if(options.os && options.os.sshKeys) {
+			// Create config.os object if it does not exist
+			config.os = config.os ? config.os : {};
+			config.os.sshKeys = config.os.sshKeys
+				? [...config.os.sshKeys, ...options.os.sshKeys]
+				: options.os.sshKeys;
+		}
 	});
 }
 


pdcastro:
@mwohlert, thanks for this PR, it looks promising! If I recall correctly, balena CI has a limitation that makes it unable to process pull requests where the source branch is in another repo, in this case the `mwohlert:balena-cli` repo. This would be why we see _"Some checks haven't completed yet"_ in this page. @nazrhom, does this sound right?

I don't remember if any user has permission to create a branch in the `balena-io/balena-cli` repo... @mwohlert, are you able to? I guess something like:
```
cd mwohlert/balena-cli
git checkout feature/include_sshKeys_in_os_configure
git remote add balena git@github.com:balena-io/balena-cli.git
git push balena mwohlert_include_sshKeys_in_os_configure
```
If you find that the above succeeds, you could create a new PR from such a branch, and then hopefully balena CI would be able to run the tests and allow the PR to be merged.

lib/utils/config.ts
```diff
@@ -68,6 +75,13 @@ export function generateBaseConfig(
 	return promise.tap(config => {
 		// os.getConfig always returns a config for an app
 		delete config.apiKey;
+
+		// merge sshKeys to config, when they have been specified
+		if(options.os && options.os.sshKeys) {
+			// Create config.os object if it does not exist
+			config.os = config.os ? config.os : {};
+			config.os.sshKeys = options.os.sshKeys;
```
pdcastro: Just an observation: currently, the `config` obect returned by `getConfig` does not include an `os.sshKeys` field, but if did / when it  does, this code would override (replace) it with the value provided in the file specified in the command line. This is probably a reasonable behaviour. An alternative behaviour would be to add extra ssh keys, given that it is a list.

mwohlert: You're right. When os.sshKeys are specified we should merge instead of override. Thanks for the catch, I'll change that.

On another note. What about the other params, that live inside the os object? 
Should I strive for a more general approach? Or do you think it is fine for now and extend this approach whenever it is needed?

mwohlert:
> @mwohlert, thanks for this PR, it looks promising! If I recall correctly, balena CI has a limitation that makes it unable to process pull requests where the source branch is in another repo, in this case the `mwohlert:balena-cli` repo. This would be why we see _"Some checks haven't completed yet"_ in this page. @nazrhom, does this sound right?
> 
> I don't remember if any user has permission to create a branch in the `balena-io/balena-cli` repo... @mwohlert, are you able to? I guess something like:
> 
> ```
> cd mwohlert/balena-cli
> git checkout feature/include_sshKeys_in_os_configure
> git remote add balena git@github.com:balena-io/balena-cli.git
> git push balena mwohlert_include_sshKeys_in_os_configure
> ```
> 
> If you find that the above succeeds, you could create a new PR from such a branch, and then hopefully balena CI would be able to run the tests and allow the PR to be merged.

Nope I don't have permission to do that:
`ERROR: Permission to balena-io/balena-cli.git denied to mwohlert.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.`

lib/utils/config.ts
pdcastro: You mean, you could also merge or override other values like `os.network.*` or `os.udevRules.*`, right? I think it would be nice to have a generic merging approach where the `config.json` file given on the command line takes precedence (in case of value clashes), but not "too generic" because I think the CLI should only copy / merge "known" values, ignoring any "typos" or invalid fields. I suspect there is `config.json` format validation somewhere, if not on `balena-cli` then in other repos like maybe [balena-sdk](https://github.com/balena-io/balena-sdk) or [balena-supervisor](https://github.com/balena-io/balena-supervisor), which could be used in combination with a "generic merge" approach.

By the way, I've remembered something else, to keep in mind: except for the first time the device boots, the balena supervisor may overwrite `config.json` on the device to reflect the backend / API database. (I'm not sure about `os.sshKeys` specifically.) But I think your purpose with `balena os configure` is to configure the device before its first boot, so it should be OK.

pdcastro:
> ERROR: Permission to balena-io/balena-cli.git denied to mwohlert.

That's a shame. Let's see if @nazrhom knows of a way around this. If it came to it, I could create a PR myself, but I would rather you took credit (and blame!) for it. :-)

mwohlert:
> > ERROR: Permission to balena-io/balena-cli.git denied to mwohlert.
> 
> That's a shame. Let's see if @nazrhom knows of a way around this. If it came to it, I could create a PR myself, but I would rather you took credit (and blame!) for it. :-)

I guess there is a reasoning to this, since you don't want random people spamming your repo with branches.

lib/utils/config.ts
mwohlert: I took a look at the supervisor and the sdk, but I cannot seem to find a config.json format validation, so I think we would have to do it in the supervisor ourselves.

The sshKeys are copied from the config.json on first boot, see: https://github.com/balena-os/meta-balena/pull/1227/commits/6d2e43d4693208c8df9436ea1f96c844b4faeb75

CameronDiver: @pdcastro @mwolhert there is only validation of certain keys from config.json in the supervisor, we have no need to concern ourselves with the others. The supervisor also never touches `.os.sshKeys` (but as aside will touch `os.udevRules` very soon, just something to keep in mind), so it doesn't contain a schema for this.

My advice would be to not make this PR too complicated by trying to consider all of the future updates that may come, as these should be thoroughly tested by both the implementer and maintainer. I'm happy with this PR as it is, as it solves the problem nicely, and shouldn't have any side effects as far as I can see. This could definitely be a rabbit hole to fall down, so let's avoid it entirely. 

Thanks again for the submission Michel!

nazrhom:
@balena-ci retest

lib/utils/config.ts
mwohlert: @CameronDiver Thanks for sharing your view on this. I'll keep this pr as it is then and do not develop some kind of generic merging and validation approach.

pdcastro:
> If I recall correctly, balena CI has a limitation that makes it unable to process pull requests where the source branch is in another repo, in this case the mwohlert:balena-cli repo.

@mwohlert, @nazrhom explained that the above is **not** the case. balena CI can run builds based off forks, the only issue is that it can not detect those automatically (opening the PR, updating it by pushing more commits), so it must be triggered manually by adding a PR comment: `(at)balenaci retest` (like was already done above).

> This branch is out-of-date with the base branch

It looks like the PR needs rebasing on the master branch.

mwohlert:
@pdcastro done

pdcastro:
@balena-ci retest

PR #1351, opened 2019-07-15
Remove BETA suffix from Windows and macOS installers, and update INSTALL.md

pdcastro:
Resolves: #1250
Resolves: #1251
Change-type: patch

### DIFF
--- a/lib/utils/config.ts
+++ b/lib/utils/config.ts
@@ -60,7 +60,7 @@ export function generateBaseConfig(
 		deviceType?: string;
 		os?: {
 			sshKeys?: string[];
-		}
+		};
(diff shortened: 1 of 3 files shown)


lib/utils/config.ts
```diff
@@ -60,7 +60,7 @@ export function generateBaseConfig(
 		deviceType?: string;
 		os?: {
 			sshKeys?: string[];
-		}
```
pdcastro: These cosmetic amendments to `config.ts` were done by `npm run prettify`, which is automatically called by `npm test`.

pdcastro:
@balena-ci retest

PR #1361, opened 2019-07-22
Introduce new treatment of .gitignore and .dockerignore files with -I flag

pdcastro:
Connects-to: #1032
Change-type: minor


pdcastro:
Draft PR: implemented behaviour to be discussed in the CLI check-in, especially with regard to automatically creating and initialising a `.dockerignore` file if one doesn't exist.

lib/utils/ignore.ts
```diff
@@ -174,3 +190,136 @@ export class FileIgnorer {
...
+	}
+	const through2 = await import('through2');
+	const dockerIgnore = require('@zeit/dockerignore')();
+	if (dockerIgnoreStr) {
+		dockerIgnore.add(dockerIgnoreStr);
+	}
+	// Don't ignore any metadata files
+	// The pattern below matches `.balena/qemu` and `myservice/.balena/qemu`
+	// but not `some.dir.for.balena/qemu`.
+	dockerIgnore.add(['!**/.balena', '!**/.resin']);
```
pdcastro: A test is failing on Windows and I have traced it to a bug in the `@zeit/dockerignore` library:
https://github.com/zeit/dockerignore/issues/15

The library bug seems to be actually worse than I reported on that issue: it seems that any pattern including a slash (forward slash or backslash) fails to match on Windows. This also affects the existing dockerignore implementation in the CLI, but could become a more acute problem if we remove support for gitignore.

LucianBuzzo:
@pdcastro there has been no movement on this PR since August 2019 - is it still relevant? Should it be closed?

pdcastro:
@LucianBuzzo, I had paused this PR because of an issue with a dependency package: https://github.com/zeit/dockerignore/issues/15
Coincidentally, this week I intend to work on that issue, which would then unblock this PR. Let's leave it around a bit longer...

pdcastro:
Closing this PR in favour of #1813, as the implementation changed significantly.

PR #1365, opened 2019-07-24
Update mixpanel tracking (v8)

pdcastro:
Change-type: patch


lib/app.coffee
```diff
@@ -14,6 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+# Assign bluebird as the global promise library
+# stream-to-promise will produce native promises if not
+# for this module, which could wreak havoc in this
+# bluebird-only codebase.
+require('any-promise/register/bluebird')
```
pdcastro: This statement relocation (from line 86 to line 21) solved an error during testing:
```
Error: any-promise already defined as "global.Promise".  You can only register an implementation before the first  call to require("any-promise") and an implementation cannot be changed
```
I didn't really understand why it necessary for the v8 branch, and not for the v9 / v10 / v11 branches. I assume it's to do with dependency versions (older CLI versions did not have a npm-shrinkwrap file). But the change seems harmless, and passed some basic tests.

package.json
```diff
@@ -79,7 +79,7 @@
     "gulp-mocha": "^2.0.0",
     "gulp-shell": "^0.5.2",
     "mochainon": "^2.0.0",
-    "pkg": "^4.3.0-beta.1",
+    "pkg": "4.3.8",
```
pdcastro: Because older CLI versions are tested with Node 6, and 4.3.8 is the highest `pkg` version that maintains Node 6 compatibility.

pdcastro:
Note: this PR was merged, but then superseded by PR #1366 because I had to recreate the v8 backport because of a resin vs balena name issue that prevented Travis from creating the standalone zip packages when a v8.1.1 tag was created. (The v8.1.1 tag was then amended.)

PR #1366, opened 2019-07-24
Update mixpanel tracking (v8)

pdcastro:
Change-type: patch


pdcastro:
Redoing the v8 backport because the Travis tag build was skipped because of a resin vs balena repo name change.

lib/app.coffee
```diff
@@ -14,6 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+# Assign bluebird as the global promise library
+# stream-to-promise will produce native promises if not
+# for this module, which could wreak havoc in this
+# bluebird-only codebase.
+require('any-promise/register/bluebird')
```
pdcastro: This statement relocation (from line 86 to line 21) solved an error during testing:
```
Error: any-promise already defined as "global.Promise".  You can only register an implementation before the first  call to require("any-promise") and an implementation cannot be changed
```
I didn't really understand why it was necessary for the v7 and v8 branches, and not for the v9 / v10 / v11 branches. I assume it's to do with dependency versions (older CLI versions did not have an `npm-shrinkwrap` file). But the change seems harmless, and passed some basic tests.

package.json
```diff
@@ -79,7 +79,7 @@
     "gulp-mocha": "^2.0.0",
     "gulp-shell": "^0.5.2",
     "mochainon": "^2.0.0",
-    "pkg": "^4.3.0-beta.1",
+    "pkg": "4.3.8",
```
pdcastro: Because older CLI versions are tested with Node 6, and 4.3.8 is the highest `pkg` version that maintains Node 6 compatibility.

pdcastro:
For the record, this was discussed in Flowdock:
https://www.flowdock.com/app/rulemotion/resin-frontend/threads/pf9k9O4tCeW5JM-P1dk2fXQzxQQ

> pdcastro at 15:09
> @CameronDiver, @thgreasi, @dfunckt, @hedss, you may notice that you were added as reviewers on a couple of PRs (so far) that I have already merged. But they were only merged to the v7 and v8 branches that I created today, have not yet been published, and can still be amended (indeed I've already amended v8). I merged them before the review because I was testing triggering Travis and Appveyor to build the standalone zip packages when tags are manually added to the repo, and to automatically upload them to draft GitHub releases. Seems to work well, and avoids the need for manually creating the assets. So please review the PRs despite them having been merged: I can still action any required changes:
> * v8 branch: https://github.com/balena-io/balena-cli/pull/1366
> * v7 branch: https://github.com/balena-io/balena-cli/pull/1367
>
> The PRs are fairly similar to PRs already reviewed and published for the v9 and v10 branches. Thank you!

> CameronDiver at 15:14
> can't approve because they're already merged but lgtm and if it's tested then im happy

Thanks!

PR #1368, opened 2019-07-24
Update mixpanel tracking (v6)

pdcastro:
Change-type: patch


automation/build-bin.ts
```diff
@@ -24,7 +24,7 @@ execPkg(['--target', 'host', '--output', 'build-bin/resin', 'package.json'])
 	.then(nativeExtensions => {
 		console.log(`\nCopying to build-bin:\n${nativeExtensions.join('\n')}`);
 
-		return nativeExtensions.map(extPath => {
+		return nativeExtensions.map((extPath: string) => {
```
dfunckt: It strikes me as odd that the type of `extPath` can't be inferred here. Are you sure the type annotation is necessary? Isn't the type of `nativeExtensions` `string[]`?

pdcastro:
I'm deciding to close this PR for a few reasons:
* As it is, the CLI fails to build. Getting the v6 code to build (and the tests to pass) is difficult because of several unpinned dependencies that have been updated (by their maintainers) out of sync with each other, and some no longer support Node 6. The balena CLI didn't have a `npm-shrinkwrap.json` file back then.
* We've advised concerned users of available CLI releases for major versions 7, 8, 9, 10 and 11 (about a year of major releases).
* I left this PR aside for 2 weeks (annual leave) while keeping an eye on any user requests for an updated version of CLI v6. I am not aware of any such request being made.

The PR can be reopened / revisited if a request comes through.

PR #1369, opened 2019-07-25
Support headless remote builds

CameronDiver:
Change-type: patch

---
##### Contributor checklist

### DIFF
--- a/lib/utils/remote-build.ts
+++ b/lib/utils/remote-build.ts
@@ -20,6 +20,7 @@ import * as readline from 'readline';
 import * as request from 'request';
 import { RegistrySecrets } from 'resin-multibuild';
 import * as Stream from 'stream';
+import streamToPromise = require('stream-to-promise');
 import { Pack } from 'tar-stream';
 import { TypedError } from 'typed-error';
 
@@ -36,6 +37,7 @@ export interface BuildOpts {
 	emulated: boolean;
 	nocache: boolean;
 	registrySecrets: RegistrySecrets;
+	headless: boolean;
 }
 
 export interface RemoteBuild {
@@ -63,6 +65,13 @@ interface BuilderMessage {
 	value?: string;
 }
 
+interface HeadlessBuilderMessage {
+	started: boolean;
+	error?: string;
+	message?: string;
+	releaseId?: number;
+}
+
 export class RemoteBuildFailedError extends TypedError {
 	public constructor(message = 'Remote build failed') {
 		super(message);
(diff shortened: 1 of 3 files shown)


lib/utils/remote-build.ts
```diff
@@ -101,31 +111,63 @@ export async function startRemoteBuild(build: RemoteBuild): Promise<void> {
...
 
-		stream.on('data', getBuilderMessageHandler(build));
-		stream.on('end', resolve);
-		stream.on('error', reject);
-	}).then(() => {
-		if (build.hadError) {
-			throw new RemoteBuildFailedError();
-		}
-	});
+	// We're running a non-headless build, which means we'll
```
hedss: A headless build, not non-headless, surely?

lib/utils/remote-build.ts
```diff
@@ -101,31 +111,63 @@ export async function startRemoteBuild(build: RemoteBuild): Promise<void> {
...
+		result = JSON.parse(response.toString());
+	} catch (e) {
+		throw new Error(
+			'There was an error reading the response from the remote builder',
+		);
+	}
+	handleHeadlessBuildMessage(result);
+}
+
+function handleHeadlessBuildMessage(message: HeadlessBuilderMessage) {
```
hedss: Not sure it's worth having this as a separate function, given it's not called from anywhere else and doesn't really make the calling function that much bigger.

lib/utils/remote-build.ts
```diff
@@ -101,31 +111,63 @@ export async function startRemoteBuild(build: RemoteBuild): Promise<void> {
...
+	// We're running a non-headless build, which means we'll
+	// get a single object back, detailing if the build has
+	// been started
+	let result: HeadlessBuilderMessage;
+	try {
+		const response = await streamToPromise(stream);
+		result = JSON.parse(response.toString());
+	} catch (e) {
+		throw new Error(
+			'There was an error reading the response from the remote builder',
```
zvin: Maybe we can show the error itself here?

lib/utils/remote-build.ts
```diff
@@ -300,38 +353,56 @@ function createRemoteBuildRequest(
...
+	let uploadSpinner = {
+		stop: () => {
+			/* noop */
+		},
+	};
+	let exitOnError = (error: Error) => {
 		return exitWithExpectedError(error);
 	};
+	// We only show the spinner when outputting to a tty
+	if (process.stdout.isTTY) {
```
memory: Note that sniffing for TTYs is not always sufficient for determining if you are in a CI environment -- Jenkins builds, as a very pertinent example, allow users to turn tty emulation on/off in the Jenkinsfile.  It would be nice to be able to explicitly disable spinners and other visuals with a flag or environment variable.

CameronDiver: Thanks for the heads up, I agree that this would certainly be better, but it's a much larger PR, as we'll need to handle all commands which can be started with `balena push`. I'd rather tackle this seperately. Given that it's possible to force non-tty execution even within a tty, this can be worked around until that PR is merged. I've created an issue here to track it: https://github.com/balena-io/balena-cli/issues/1378

PR #1383, opened 2019-08-07
Improvements to the "env rm" command and oclif migration

LucianBuzzo:
Resolves: #1380
Change-type: patch

---
##### Contributor checklist

- [x] Affects the development, build or deployment processes of the component

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -601,12 +601,12 @@ device uuid
 
 show config variables
 
-## env rm &#60;id&#62;
+## env rm ID
 
-Use this command to remove an environment variable from an application
-or device.
+Remove an environment variable from an application or device, as selected
+by command-line options.
 
-Notice this command asks for confirmation interactively.
+Note that this command asks for confirmation interactively.
 You can avoid this by passing the `--yes` boolean option.
 
 The --device option selects a device instead of an application.
@@ -620,15 +620,21 @@ Examples:
 	$ balena env rm 215 --yes
 	$ balena env rm 215 --device
 
+### Arguments
+
+#### ID
+
+environment variable id
+
 ### Options
 
-#### --yes, -y
+#### -d, --device
 
-confirm non interactively
+Selects a device environment variable instead of an application environment variable
 
-#### --device, -d
+#### -y, --yes
 
-device
+Run in non-interactive mode
 
 ## env add NAME [VALUE]
 
--- a/lib/actions-oclif/env/rm.ts
+++ b/lib/actions-oclif/env/rm.ts
@@ -1,0 +1,111 @@
+/**
+ * @license
+ * Copyright 2019 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Command, flags } from '@oclif/command';
+import { stripIndent } from 'common-tags';
+
+import { CommandHelp } from '../../utils/oclif-utils';
+
+interface FlagsDef {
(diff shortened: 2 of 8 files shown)


doc/cli.markdown
```diff
@@ -601,9 +601,8 @@ device uuid
 
 show config variables
 
-## env rm &#60;id&#62;
+## env rm ID
 
-Use this command to remove an environment variable from an application
 or device.
```
pdcastro: This "or device" line seems to be left on its own. See other comment in the respective source code file.

lib/actions-oclif/env/rm.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+interface ArgsDef {
+	id: number;
+}
+
+export default class EnvRmCmd extends Command {
+	public static description = stripIndent`
+		Use this command to remove an environment variable from an application
+		or device.
```
pdcastro: The `cli.markdown` file is printing only "or device", skipping the first line (see the other review comment). If you check the similar `add.ts` file that implements `env add` with oclif, you'll see a duplicate first line:
```
		Add an environment or config variable to an application or device.

		Add an environment or config variable to an application or device, as selected
		by the respective command-line options.
```
It's on purpose, because the oclif help output will print the first line at the very top, and the following lines under "Description".
```
$ balena help env add
Add an environment or config variable to an application or device.

USAGE
...
DESCRIPTION
  Add an environment or config variable to an application or device, as selected
  by the respective command-line options.
```

In addition to the markdown, please check that the following produces the expected output:
```
# 'help -v' should print lines for env add, env rename, env rm, envs
balena help -v
balena help env rm
```

lib/actions/help_ts.ts
```diff
@@ -24,8 +24,9 @@ export function getOclifHelpLinePairs(): Array<[string, string]> {
 	// improvement would probably be to automatically scan the actions-oclif
 	// folder.
 	const EnvAddCmd = require('../actions-oclif/env/add').default;
+	const EnvRmCmd = require('../actions-oclif/env/rm').default;
 	const VersionCmd = require('../actions-oclif/version').default;
-	return [EnvAddCmd, VersionCmd].map(getCmdUsageDescriptionLinePair);
+	return [EnvAddCmd, EnvRmCmd, VersionCmd].map(getCmdUsageDescriptionLinePair);
```
pdcastro: Just FYI, an improvement I thought about in the past (not necessary to implement in this PR, unless you felt like doing it...) is to "centralise" the changes needed to add an oclif command in a single place. I mean, as you have noticed, you've had to add/modify lines in `help_ts.ts`, `app.ts`, `oclif-utils.ts` and `capitanodoc.ts`, so it feels a bit like "all over the place". I wonder if it could be all in an array or object in, say, `oclif-utils.ts`, then imported by the other files. But it's just a thought.

LucianBuzzo: @pdcastro yeah I noticed this, and I have some ideas of how we can improve the process. I will probably send a follow up PR soon :smile:

LucianBuzzo: see https://github.com/balena-io/balena-cli/issues/1391

lib/actions-oclif/env/rm.ts
```diff
@@ -0,0 +1,111 @@
...
+interface ArgsDef {
+	id: number;
+}
+
+export default class EnvRmCmd extends Command {
+	public static description = stripIndent`
+		Remove an environment variable from an application or device.
+
+		Remove an environment variable from an application or device, as selected
+		by the respective command-line options.
```
pdcastro: Minor: the word "respective" above applies to "env add" but not so much to "env rm", because "env add" has the "--application" and "--device" options, whereas "env rm" only has "--device". Perhaps:
```
		Remove an environment variable from an application or device, as selected
		by command-line options.
```

lib/actions-oclif/env/rm.ts
```diff
@@ -0,0 +1,111 @@
...
+
+		Notice this command asks for confirmation interactively.
```
pdcastro: While at it, as I think you're a native English speaker, does _"Note that"_ sound better than _"Notice"_ ?

LucianBuzzo: I think "note that" sounds a little more natural here

lib/actions-oclif/env/rm.ts
LucianBuzzo: Yep, ideal

PR #1385, opened 2019-08-07
Tests part 1: Remove mochainon dependency and replace with direct testing dependencies

LucianBuzzo:
Change-type: patch

---
##### Contributor checklist

- [X] Affects the development, build or deployment processes of the component


LucianBuzzo:
@balena-ci retest

LucianBuzzo:
@balena-ci retest

npm-shrinkwrap.json
```diff
@@ -321,6 +321,42 @@
       "resolved": "https://registry.npmjs.org/@resin.io/valid-email/-/valid-email-0.1.0.tgz",
       "integrity": "sha1-DnUwmoQ8AUqAqhSC+WmQYvL6UV0="
     },
+    "@sinonjs/commons": {
```
pdcastro: @LucianBuzzo, nice that you're improving the test files! 👍
The `npm-shrinkwrap.json` file is being significantly modified in this PR, so I wanted to check that you were using `npm` version `6.9.0` or later when running `npm install`. This is because of a bug in npm (fixed in 6.9.0) that causes it to incorrectly update the `npm-shrinkwrap.json` file.

LucianBuzzo: @pdcastro I'll double check. We should add this information to the Contributing doc I think.

LucianBuzzo: Looks like I ran it with npm v6.4.1

LucianBuzzo: I've reset `npm-shrinkwrap.json` to master and then re-run `npm install` using npm v6.10.3

pdcastro: Agreed, it would be good to add this to CONTRIBUTING.md, or even to add a test that runs with "npm run build" to ensure that the npm version is at least 6.9.0. Until a few months ago, there was one such test that checked for npm version >= 4.0.0.

FYI, I checked out your branch (`tests`) and ran `npm install` with npm v6.9.0 (rather than v6.10.3), only because it's the default npm version that ships with Node 10 (10.16.2), and thus the version more likely to be used by CLI contributors (assuming they are using Node 10, which is the latest supported Node version according to the INSTALL.md file). Starting with a `npm-shrinkwrap.json` file checked out from `master`, the resulting `npm-shrinkwrap.json` file was identical to the one you pushed to this PR. 👍

LucianBuzzo: Ok great, thats good to know

PR #1390, opened 2019-08-08
Add homebrew installation method for macOS

chrisys:
Change-type: patch
Signed-off-by: Chris Crocker-White <chriscw@balena.io>

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -94,6 +94,13 @@ errors when using the system Node / npm packages installed via "apt-get". We sug
 [nvm](https://github.com/creationix/nvm) instead. Check this sample Dockerfile for installing the
 CLI on an Ubuntu Docker image: https://gist.github.com/pdcastro/5d4d96652181e7da685a32caf629dd44
 
+## Other package managers
+
+If you're using macOS, there is a `balena-cli` package available on [homebrew](https://brew.sh/) which can be installed using:
+```
+$ brew install balena-cli
+```
+
 ## Additional Dependencies
 
 * The `balena ssh` command requires a recent version of the `ssh` command-line tool to be available:


INSTALL.md
```diff
@@ -94,6 +94,13 @@ errors when using the system Node / npm packages installed via "apt-get". We sug
 [nvm](https://github.com/creationix/nvm) instead. Check this sample Dockerfile for installing the
 CLI on an Ubuntu Docker image: https://gist.github.com/pdcastro/5d4d96652181e7da685a32caf629dd44
 
+## Other package managers
+
+If you're using macOS, there is a `balena-cli` package available on [homebrew](https://brew.sh/) which can be installed using:
```
pdcastro: Hang on... The last I heard / knew about it, it was something created by some balena users, outside of balena's control. For a recent example of why it matters, there was a recent mixpanel tracking issue that required us to promptly create a new CLI release and backport the fix to older major versions (v7, 8, 9, 10). New releases were created for the npm installation, standalone zip packages and executable installers, for Mac, Windows and Linux, but the `brew` installation was not touched because "it's not ours". Indeed, brew installations are still affected by that issue, as `brew info` says it's currently in version 11.3.1 while the fix was released to 11.7.3, and the CLI is currently at 11.7.8.
* If we wanted to take control of brew installation, we could kindly ask the relevant users to transfer it to us, but I think we discussed before that brew installation does not add much value when users can already choose the native MacOS installer or a standalone zip package, and it would mean yet another channel for us to maintain.
* And if we don't have control over it, perhaps we should kindly ask the relevant users to modify that line that says "The official balena CLI tool" to "This is not the official balena CLI tool", in the output of `brew info`.

```
12:46 $ brew info balena-cli
balena-cli: stable 11.3.1 (bottled)
The official balena CLI tool
https://www.balena.io/docs/reference/cli/
Not installed
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/balena-cli.rb
==> Dependencies
Required: node ✘
==> Analytics
install: 205 (30 days), 371 (90 days), 595 (365 days)
install_on_request: 205 (30 days), 371 (90 days), 595 (365 days)
build_error: 0 (30 days)
```

chrisys:
@drjasonharrison-vp-eio ^ see above for feedback re your original docs PR

INSTALL.md
hedss: I'm of the opinion we shouldn't point users at our official tools if they've not been released by us. Whilst the likelihood is low, there's always a possibility of malicious control which could end up with us in a very negative light. I'm in agreement with asking either to take it over (though I could understand why someone would decline), or dropping this.

chrisys: @pdcastro @hedss feel free to close this - we've been cleaning up the PRs on the `docs` repo, where this was an external contribution and transferred from there. Hopefully now the original author can see our comments. 

I agree with you both, however 205 installs in the last 30 days would seem significant enough to warrant trying to make it an 'officially supported' package.

pdcastro: @shaunmulligan, thoughts? We could discuss this further on Monday's CLI check-in.

_"205 installs in the last 30 days"_ sounds like we should do something about it, but not necessarily making it "officially supported"... We could ask them to change the "brew info" description to say something like:

> Please note that 'brew' is **not** an official balena-cli release channel.
> For official balena-cli releases and installation options, please visit:
> https://github.com/balena-io/balena-cli/blob/master/INSTALL.md

hedss: In agreement, we should discuss this at the CLI checkin.

drjasonharrison-vp-eio:
Thank you for continuing the discussion about the balena CLI and the homebrew version.

I had previously installed the CLI on macos using npm, and when I tried to update it I ran into trouble. Brew as a package manager is a lot more familiar to me than npm. Especially when it comes to installing global command line tools, npm has created more frustration than brew.

If you drop or delay support for homebrew, I will try the other distribution methods and let you know if I have problems keeping them up to date.

drjasonharrison-vp-eio:
Also, I don't know if the CLI already checks for updates when run, it might have and given the instructions on how to update which did not work.

pdcastro:
Related issue: #1406 [Investigate adopting homebrew as an official release channel](https://github.com/balena-io/balena-cli/issues/1406)

pdcastro:
@chrisys, I suggest bringing this to a Product Call (whether balena wishes to support Homebrew as a distribution channel).

PR #1392, opened 2019-08-09
balena-ssh: add info about remote vs local connections

xginn8:
small formatting fixups
    
Closes: https://github.com/balena-io/docs/issues/974

Change-type: patch
Signed-off-by: Matthew McGinn <matthew@balena.io>

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -944,7 +944,11 @@ If an application name is provided, all online devices in the application
 will be presented, and the chosen device will then have a shell opened on
 in it's service container or host OS.
 
-For local devices, the ip address and .local domain name are supported.
+For local devices, the IP address and .local domain name are supported.
+If the device is referenced by IP or `.local` address, the connection
+is initiated directly to balenaOS on port `22222` via an
+openssh-compatible client. Otherwise, any connection initiated remotely
+traverses the balenaCloud VPN.
 
 Examples:
 	balena ssh MyApp
@@ -956,7 +960,7 @@ Examples:
 	balena ssh 192.168.0.1 --verbose
 	balena ssh f49cefd.local my-service
 
-Warning: 'balena ssh' requires an openssh-compatible client to be correctly
+Warning: `balena ssh` requires an openssh-compatible client to be correctly
 installed in your shell environment. For more information (including Windows
 support) please check:
 	https://github.com/balena-io/balena-cli/blob/master/INSTALL.md#additional-dependencies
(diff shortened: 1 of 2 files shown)


doc/cli.markdown
```diff
@@ -950,7 +951,7 @@ Examples:
 	balena ssh 192.168.0.1 --verbose
 	balena ssh f49cefd.local my-service
 
-Warning: 'balena ssh' requires an openssh-compatible client to be correctly
+Warning: `balena ssh` requires an openssh-compatible client to be correctly
```
pdcastro: `cli.markdown` is auto generated, so these changes would have to go in, I think, `lib/actions/ssh.ts`, then run `npm run build`.

doc/cli.markdown
```diff
@@ -959,7 +960,7 @@ support) please check:
 
 #### --port, -p &#60;port&#62;
 
-SSH gateway port
+SSH gateway port (by default balenaOS listens on `22222`)
```
pdcastro: hmm "gateway port" is ***not*** the port that a ssh server listens on a device, right? I think it refers to the port that some cloud service or openBalena service listens on, maybe a VPN service (?), but in any case "balenaOS" would not be relevant here, because the whole thing goes over the cloud/VPN. Am I misunderstanding it? I could well be wrong, but I think that "balena ssh" is fundamentally different to plain ssh as in:
```
ssh -p 22222 root@192.168.2.5
```

xginn8: I was under the impression that ssh'ing to a device in local mode would initiate a direct ssh connection (rather than going over the VPN), which is what I intended to document here.

doc/cli.markdown
xginn8: noted, thanks

doc/cli.markdown
hedss: It does, the assertion that it goes over the VPN is incorrect. balenaOS is listening for SSH traffic on port 22222, regardless of development or production edition or local mode activation. In production mode, the port is not open and a specific device key is required (and in fact users can register these keys using the `sshKeys` property in `config.json` on-device. I agree with the modified text.

hedss: So, there's some misinformation here:

`balena ssh <uuid>` will always use the proxy/VPN
`balena ssh <x.local>` and `balena ssh x.x.x.x` will use a direct SSH connection to port `22222` on the local device

Because of this I think it's probably *really important* to make that distinction in the documentation. But I think @xginn8 is right in documenting that balenaOS is listening on port 22222, but it's not really a gateway port as it's not doing any forwarding. It's just the device's local SSH port.

pdcastro: > `balena ssh x.x.x.x` will use a direct SSH connection

So it does! I stand corrected. Sorry for the noise @xginn8, @hedss. Documenting it as Hedley suggested would be nice.

xginn8: I have updated the documentation accordingly, thanks for the feedback both! I'm glad to clear this up

PR #1398, opened 2019-08-15
Fix incorrect arguments passed to oclif bootstrap function

LucianBuzzo:
Adds tests for existing oclif commands to avoid regressions

Change-type: patch

---
##### Contributor checklist


lib/app.ts
```diff
@@ -57,22 +57,23 @@ function routeCliFramework(argv: string[], options: AppOptions): void {
...
+					argv[0],
+					argv[1],
+					...oclifArgs,
+				]}] length=${oclifArgs.length + 2}`,
+			);
 		}
-		return require('./app-oclif').run(cmdSlice, options);
+		return require('./app-oclif').run(oclifArgs, options);
 	} else {
 		return require('./app-capitano').run(cmdSlice);
```
pdcastro: Regarding "balena help -v" not printing the full list of commands, possibly replacing `cmdSlice` with `argv` for capitano fixes it:
```
return require('./app-capitano').run(argv);
```

LucianBuzzo: @pdcastro yep, thats the case, I didn't account for the cmdSlice modification to `help`

LucianBuzzo: I've also added a regression test to ensure that `balena help -v` works as expected

PR #1399, opened 2019-08-15
Add NYC code coverage report

LucianBuzzo:
After running tests, a code coverage report can be found at ./coverage/lcov-report/index.html

Change-type: patch

Resolves: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist


package.json
```diff
@@ -186,6 +186,7 @@
     "moment-duration-format": "^2.2.2",
     "mz": "^2.7.0",
     "node-cleanup": "^2.1.2",
+    "nyc": "^14.1.1",
```
pdcastro: Should this be under "devDependencies" instead of "dependencies"? (In which case, `npm-shrinkwrap.json` would also need to be regenerated.)

LucianBuzzo: Ah yes, good catch @pdcastro

LucianBuzzo:
@balena-ci retest

LucianBuzzo:
@balena-ci retest

LucianBuzzo:
@balena-ci retest

pdcastro:
> By the way, another form of testing that we need to look at is the testing of the "built assets", especially the pkg / standalone binary (which bundles Node.js with node_modules in a single binary file). I find myself manually testing built packages quite often...

Coincidentally, I've just fixed a pkg-related bug (reported by a CLI user in #1419) and I've added one ad-hoc test based on the `balena version -j` command: PR #1423.

Issue #1397, opened 2019-08-14
"balena build": registry secrets bypassed when -a option not provided


pdcastro:
The `--registry-secrets` option works as expected in a command line like:
```
balena build -a test-nuc --registry-secrets sample_secrets.yml .
```
But seems to be ignored in a command line like:
```
balena build --deviceType intel-nuc --arch amd64 --registry-secrets sample_secrets.yml .
```
Leading to authentication errors:
```
Error: unauthorized: You don't have the needed permissions to perform this operation,
and you may have invalid credentials.
```

pdcastro:
Support thread: https://www.flowdock.com/app/rulemotion/s-sales/threads/eMnbQeRuqCdCzNX0q1MzR-v5ZyA

PR #1400, opened 2019-08-16
Fix failing registry secrets authentication under certain conditions

pdcastro:
Resolves: #1397 
Change-type: patch

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -1447,8 +1447,12 @@ Sample registry-secrets YAML file:
 		username: '_json_key'
 		password: '{escaped contents of the GCR keyfile.json file}'
 
-If an option is not specified, and a secrets.yml or secrets.json file exists in
-the balena directory (usually $HOME/.balena), this file will be used instead.
+For a sample project using registry secrets with the Google Container Registry,
+check: https://github.com/balena-io-playground/sample-gcr-registry-secrets
+
+If the --registry-secrets option is not specified, and a secrets.yml or
+secrets.json file exists in the balena directory (usually $HOME/.balena),
+this file will be used instead.
 
 Examples:
 
@@ -1598,8 +1602,12 @@ Sample registry-secrets YAML file:
 		username: '_json_key'
 		password: '{escaped contents of the GCR keyfile.json file}'
 
-If an option is not specified, and a secrets.yml or secrets.json file exists in
-the balena directory (usually $HOME/.balena), this file will be used instead.
+For a sample project using registry secrets with the Google Container Registry,
+check: https://github.com/balena-io-playground/sample-gcr-registry-secrets
+
+If the --registry-secrets option is not specified, and a secrets.yml or
+secrets.json file exists in the balena directory (usually $HOME/.balena),
+this file will be used instead.
 
 Examples:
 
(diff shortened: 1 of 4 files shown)


lib/actions/build.coffee
```diff
@@ -109,17 +109,15 @@ module.exports =
...
-			delete params.source
-
-			validateComposeOptions(sdk, options)
+		# `build` accepts `[source]` as a parameter, but compose expects it
+		# as an option. swap them here
+		options.source ?= params.source
+		delete params.source
 
+		Promise.resolve(validateComposeOptions(sdk, options))
+		.then ->
```
pdcastro: The key fix in this PR, here and in `deploy.coffee`, is to await (`.then`) the promise returned by `validateComposeOptions` (which is declared `async`). `validateComposeOptions` loads the registry secrets from a file and assigns the result to `options['registry-secrets']`. Previously, by chance, sometimes this operation would be completed "in time" before `options['registry-secrets']` was accessed -- especially likely if the `--application` option was provided, in which case a network request is then awaited. But if `--application` was not given, the chance would increase that the registry secrets would not be parsed in time, and authentication would fail as experienced by a customer.

lib/utils/messages.ts
```diff
@@ -45,5 +45,9 @@ Sample registry-secrets YAML file:
 		username: '_json_key'
 		password: '{escaped contents of the GCR keyfile.json file}'
 
-If an option is not specified, and a secrets.yml or secrets.json file exists in
-the balena directory (usually $HOME/.balena), this file will be used instead.`;
+For an example using the Google Container Registry, check this page:
+https://gist.github.com/pdcastro/d9739b9b81ad75ab5957bd5b8b405080
```
pdcastro: Support thread: https://www.flowdock.com/app/rulemotion/s-sales/threads/eMnbQeRuqCdCzNX0q1MzR-v5ZyA

lib/utils/messages.ts
```diff
@@ -45,5 +45,9 @@ Sample registry-secrets YAML file:
...
+
+If the --registry-secrets option is not specified, and a secrets.yml or
+secrets.json file exists in the balena directory (usually $HOME/.balena),
+this file will be used instead.`;
```
pdcastro: Replaced "an option" with "the --registry-secrets option" because, when the help is rendered with e.g. "balena help push", the paragraph is printed on its own, and I've added a new paragraph before this one. (One would still reason that "secrets.yml" most likely applies to "--registry-secrets", but spelling it reduces ambiguity.)

doc/cli.markdown
```diff
@@ -1443,8 +1443,12 @@ Sample registry-secrets YAML file:
 		username: '_json_key'
 		password: '{escaped contents of the GCR keyfile.json file}'
 
-If an option is not specified, and a secrets.yml or secrets.json file exists in
-the balena directory (usually $HOME/.balena), this file will be used instead.
+For an example using the Google Container Registry, check this page:
+https://gist.github.com/pdcastro/d9739b9b81ad75ab5957bd5b8b405080
```
CameronDiver: I really don't think we should be adding a gist to a production codebase - can you add this to the docs somewhere and then link there please?

pdcastro: Agreed. As discussed in the CLI check-in, I've created a sample `balena-io-playground` project and replaced the URL.

Issue #1389, opened 2019-08-08
[Error]   Build failed The command '/bin/sh -c apk update && apk add jq' returned a non-zero code: 1


lyly0906:
balena login is ok!
![image](https://user-images.githubusercontent.com/26811002/62696084-43b6bf00-ba0a-11e9-9ad0-447682a4e7aa.png)

balena devices is ok!
![image](https://user-images.githubusercontent.com/26811002/62696015-22ee6980-ba0a-11e9-9103-e3bb1eafdd19.png)

but ,deploy is wrong! how can i resovle this problem?
![image](https://user-images.githubusercontent.com/26811002/62696209-7e205c00-ba0a-11e9-9599-076717210cd2.png)

root@ubuntu:/opt/balena-sense-hat-python-starter# balena build --deviceType raspberrypi3 --arch armv7hf --emulated
[Info]    Creating default composition with source: /opt/balena-sense-hat-python-starter
[Info]    Building for armv7hf/raspberrypi3
[Info]    Emulation is enabled
[Build]   Built 1 service in 0:03
[Build]   main Preparing...
[Error]   Build failed
The command '/bin/sh -c apk update && apk add jq' returned a non-zero code: 1

Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, or "export DEBUG=1"
on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/

pdcastro:
@lyly0906, thanks for getting in touch. The key error message seems to be:
```
The command '/bin/sh -c apk update && apk add jq' returned a non-zero code: 1
```
It means that the following command failed to execute:
```
apk update && apk add jq
```
I assume / expect that the command above was executed (attempted) during the image build, perhaps as part of a RUN instruction of a Dockerfile, even though I don't see that command in the Dockerfile of the `balena-sense-hat-python-starter` project:
https://github.com/balena-io-playground/balena-sense-hat-python-starter/blob/master/Dockerfile

In any case, this sounds like a general support / troubleshooting issue with balena deployments or the `balena-sense-hat-python-starter` project, rather a `balena-cli` bug, so for further support I would suggest that a thread was created in the balenaCloud forums:
https://forums.balena.io/c/balena-cloud

[...]

Issue #1355, opened 2019-07-18
balena build: SecretRemovalError: The command '/bin/sh -c apk update && apk add jq' returned a non-zero code: 1


xginn8:
Introduced in v11.7.0 as part of multibuild, somehow [one](https://github.com/balena-io-modules/resin-multibuild/blob/fb3339396a4495888fd060ea5fbcbb08555a007c/lib/build-secrets/Dockerfile.remove) of these [two](https://github.com/balena-io-modules/resin-multibuild/blob/fb3339396a4495888fd060ea5fbcbb08555a007c/lib/build-secrets/Dockerfile) is injected improperly or incompletely into an `armv7hf` Debian build which fails with the following stacktrace:

```
SecretRemovalError: The command '/bin/sh -c apk update && apk add jq' returned a non-zero code: 1
    at Object.<anonymous> (/usr/lib/node_modules/balena-cli/node_modules/resin-multibuild/build/index.js:122:19)
    at Generator.throw (<anonymous>)
    at rejected (/usr/lib/node_modules/balena-cli/node_modules/resin-multibuild/build/index.js:5:65)
    at <anonymous>
```

For a simple reproduction:

Dockerfile:
```
FROM balenalib/raspberrypi3-debian:jessie-build
```
cmd:
```
balena build --deviceType raspberrypi3 --arch armv7hf
```

Downgrading to 11.6.1 works around the issue for now (cc @CameronDiver & @pdcastro).

pdcastro:
This issue was fixed in CLI [v11.9.0](https://github.com/balena-io/balena-cli/releases).

misk:
I am using 11.9.1 and the issue is still there

**`> balena push 192.168.1.217 -s .`**
`SecretRemovalError: The command '/bin/sh -c apk update && apk add bash jq' returned a non-zero code: 139
    at Object.<anonymous> (/usr/local/lib/balena-cli/node_modules/resin-multibuild/build/index.js:124:19)
    at Generator.throw (<anonymous>)
    at rejected (/usr/local/lib/balena-cli/node_modules/resin-multibuild/build/index.js:5:65)`

**`> balena -v`**
`[debug] original argv0="/usr/local/lib/balena-cli/bin/node" argv=[/usr/local/lib/balena-cli/bin/node,/usr/local/lib/balena-cli/bin/run,-v] length=3
[...]

PR #1403, opened 2019-08-21
 Fix 'SecretRemovalError' and enable certain emulated builds on a remote device

pdcastro:
In addition to fixing the bug originally reported in #1355 (SecretRemovalError), this PR includes a second commit that enables emulated builds on remote devices (remote docker daemon) where the remote device runs a different OS than the OS where the CLI is running. For example, the "balena build -e -h <IP> -p 2375" command with the CLI running on a Mac laptop (macOS), using balenaEngine on an Intel NUC device (Linux/balenaOS), building an image targeting the ARM/RPi architecture. Previously, the QEMU setup code in the CLI assumed that docker ran on the same OS as the CLI (Node `os.platform()`), and this PR introduces querying the docker daemon instead (`dockerode.info()`). This is relevant because, for example, "Docker for Mac" has built-in "binfmt_misc" support and does not require additional emulation setup, whereas balenaEngine on Linux/balenaOS requires explicit QEMU setup by the CLI.

Resolves: #1355 
Resolves: #1389
Depends-on: https://github.com/balena-io-modules/resin-multibuild/issues/56
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -95,7 +95,7 @@ export async function checkBuildSecretsRequirements(
 	sourceDir: string,
 ) {
 	const [metaObj, metaFilename] = await loadBuildMetatada(sourceDir);
-	if (!_.isEmpty(metaObj['build-secrets'])) {
+	if (metaObj && !_.isEmpty(metaObj['build-secrets'])) {
 		const dockerUtils = await import('./docker');
 		const isBalenaEngine = await dockerUtils.isBalenaEngine(docker);
 		if (!isBalenaEngine) {
--- a/lib/utils/qemu-ts.ts
+++ b/lib/utils/qemu-ts.ts
@@ -1,0 +1,61 @@
+/**
+ * @license
+ * Copyright 2019 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Dockerode = require('dockerode');
+
+import Logger = require('./logger');
+import { getQemuPath, installQemu } from './qemu';
+
+export { QEMU_BIN_NAME } from './qemu';
+export * from './qemu';
+
+export async function installQemuIfNeeded(
+	emulated: boolean,
+	logger: Logger,
+	arch: string,
+	docker: Dockerode,
+): Promise<boolean> {
+	if (!emulated || !(await platformNeedsQemu(docker))) {
+		return false;
+	}
+	const fs = await import('mz/fs');
+	if (!(await fs.exists(await getQemuPath(arch)))) {
+		logger.logInfo(`Installing qemu for ${arch} emulation...`);
+		await installQemu(arch);
+	}
+	return true;
+}
+
+/**
+ * Check whether the Docker daemon (including balenaEngine) requires explicit
+ * QEMU emulation setup. Note that Docker for Mac (macOS), and reportedly also
+ * Docker for Windows, have built-in support for binfmt_misc, so do not require
+ * explicity QEMU setup. References:
+ * - https://en.wikipedia.org/wiki/Binfmt_misc
+ * - https://docs.docker.com/docker-for-mac/multi-arch/
+ * - https://www.ecliptik.com/Cross-Building-and-Running-Multi-Arch-Docker-Images/
+ * - https://stackoverflow.com/questions/55388725/run-linux-arm-container-via-qemu-binfmt-misc-on-docker-lcow
+ *
+ * @param docker Dockerode instance
+ */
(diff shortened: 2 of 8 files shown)


lib/utils/compose_ts.ts
```diff
@@ -95,7 +95,7 @@ export async function checkBuildSecretsRequirements(
 	sourceDir: string,
 ) {
 	const [metaObj, metaFilename] = await loadBuildMetatada(sourceDir);
-	if (!_.isEmpty(metaObj['build-secrets'])) {
+	if (metaObj && !_.isEmpty(metaObj['build-secrets'])) {
```
pdcastro: This fixes an error thrown when the `balena.yml` file contains only comments.

lib/utils/qemu-ts.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export { QEMU_BIN_NAME } from './qemu';
+export * from './qemu';
+
+export async function installQemuIfNeeded(
+	emulated: boolean,
+	logger: Logger,
+	arch: string,
+	docker: Dockerode,
+): Promise<boolean> {
```
pdcastro: Converted from CoffeeScript, adding the `docker` argument to allow querying the Docker daemon (balenaEngine) for its operating system.

lib/utils/qemu-ts.ts
```diff
@@ -0,0 +1,71 @@
...
+ * - https://stackoverflow.com/questions/55388725/run-linux-arm-container-via-qemu-binfmt-misc-on-docker-lcow
+ *
+ * @param docker Dockerode instance
+ */
+async function platformNeedsQemu(docker: Dockerode): Promise<boolean> {
+	const dockerInfo = await docker.info();
+	// Docker for Mac reports dockerInfo.OSType as 'linux'
+	// https://stackoverflow.com/questions/38223965/how-can-i-detect-if-docker-for-mac-is-installed
+	const isDockerForMac = /Docker for Mac/i.test(dockerInfo.OperatingSystem);
+	return dockerInfo.OSType === 'linux' && !isDockerForMac;
```
pdcastro: `platformNeedsQemu` was converted from CoffeeScript and the logic was modified to query the docker daemon for the operating system (Linux, Mac) instead of using `os.platform()`. "Docker for Mac" has built-in support for [binfmt_misc](https://en.wikipedia.org/wiki/Binfmt_misc) and does not require explicit QEMU setup by the CLI, whereas Docker for Linux or balenaEngine/balenaOS do. Before this change, running the CLI on macOS and specifying a remote balenaEngine socket on Intel NUC with ARM emulation (e.g. `balena build ... -h 192.168.0.20 -p 2375 --emulated`) would fail because the CLI would incorrectly assume that balenaEngine was Docker for Mac.

With regard to Docker for Windows, I am not sure if it ever worked before... and apparently there are different cases to handle: [Linux Containers and Windows Containers](https://stackoverflow.com/questions/55388725/run-linux-arm-container-via-qemu-binfmt-misc-on-docker-lcow).

If you know or think that the `balena build` command used to work with Docker for Windows, let me know and I'll try to find a way of testing it.

lib/utils/qemu.coffee
```diff
@@ -1,18 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
 Promise = require('bluebird')
 
 exports.QEMU_VERSION = QEMU_VERSION = 'v4.0.0-balena'
 exports.QEMU_BIN_NAME = QEMU_BIN_NAME = 'qemu-execve'
 
-exports.installQemuIfNeeded = Promise.method (emulated, logger, arch) ->
```
pdcastro: `installQemuIfNeeded`, `hasQemu` and `platformNeedsQemu` were converted to TypeScript in `qemu-ts.ts`.

tests/utils/qemu.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+import { expect } from 'chai';
+
+describe('resin-multibuild consistency', function() {
+	it('should use the same values for selected constants', async () => {
+		const { QEMU_BIN_NAME: MQEMU_BIN_NAME } = await import('resin-multibuild');
+		const { QEMU_BIN_NAME } = await import('../../build/utils/qemu-ts');
+		expect(QEMU_BIN_NAME).to.equal(MQEMU_BIN_NAME);
```
pdcastro: "Peace of mind" test as both `balena-cli` and `resin-multibuild` have `QEMU_BIN_NAME` constants that should be equal. I thought of defining `QEMU_BIN_NAME` in `resin-multibuild` only and then importing it in `balena-cli`, however this would require either importing at module level / top of file (which hurts CLI start-up time), or importing `resin-multibuild` in the body of several functions, which felt... inelegant.

lib/utils/qemu-ts.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+async function hasQemu(arch: string): Promise<boolean> {
+	const fs = await import('mz/fs');
+	const qemuPath = await getQemuPath(arch);
+	try {
+		await fs.stat(qemuPath);
+	} catch (error) {
```
CameronDiver: Does this not trip up the linter? I think you need a `_` before `error`

pdcastro: Thanks! The linter was silent about it, but I've used the opportunity to tighten the logic and explicitly test for `err.code === 'ENOENT'`.

pdcastro: Actually... I've replaced the whole `hasQemu` function with simply:
`await fs.exists(await getQemuPath(arch))`

PR #1405, opened 2019-08-23
Deploy scripts: edit GitHub release descriptions given semver ranges

pdcastro:
Change-type: patch


automation/deploy-bin.ts
```diff
@@ -57,3 +58,201 @@ export async function release() {
...
+	let newDescription = `
+		Please note: the "login" command in this release is affected by a
+		security issue fixed in versions
+		[7.10.10](https://github.com/balena-io/balena-cli/releases/tag/v7.10.10),
+		[8.1.1](https://github.com/balena-io/balena-cli/releases/tag/v8.1.1),
+		[9.15.7](https://github.com/balena-io/balena-cli/releases/tag/v9.15.7),
+		[10.17.6](https://github.com/balena-io/balena-cli/releases/tag/v10.17.6),
+		[11.7.3](https://github.com/balena-io/balena-cli/releases/tag/v11.7.3)
+		and later. If you need to use this version, avoid passing your password,
+		keys or tokens as command-line arguments. ${editID}`;
```
pdcastro: The wording of this message was discussed and "approved" in today's CLI check-in, and also mentioned in the product flow: https://www.flowdock.com/app/rulemotion/r-product/threads/qBCIi79h1qqt3KyBSPBF1BV2fpD

automation/deploy-bin.ts
```diff
@@ -57,3 +58,201 @@ export async function release() {
...
+		}
+	}
+}
+
+/**
+ * Add a warning description to CLI releases affected by a mixpanel tracking
+ * security issue (#1359). This function can be executed "manually" with the
+ * following command line:
+ *
+ * npx ts-node --type-check -P automation/tsconfig.json automation/run.ts fix1359
```
pdcastro: Indeed I have executed the command line above on a Travis VM (because the GITHUB_TOKEN environment variable is available there), and the warning message has already been added to the affected CLI releases. I thought it would be nonetheless useful to commit this code in case we needed to run something similar again in the future.

automation/deploy-bin.ts
```diff
@@ -57,3 +58,201 @@ export async function release() {
...
+	// Run only on Linux/Node10, instead of all platform/Node combinations.
+	// (It could have been any other platform, as long as it only runs once.)
+	if (process.platform !== 'linux' || semver.major(process.version) !== 10) {
+		return;
+	}
+	const owner = 'balena-io';
+	const repo = 'balena-cli';
+	const affectedVersions =
+		'2.6.1 - 7.10.9 || 8.0.0 - 8.1.0 || 9.0.0 - 9.15.6 || 10.0.0 - 10.17.5 || 11.0.0 - 11.7.2';
+	const editID = '[AA100]';
```
pdcastro: This "edit ID" [AA100] is appended to the automatically added warning messages as an easy (additional) way of programmatically determining whether a given CLI release message / description has already been updated to include the relevant warning. The thinking is that, in the future, some CLI release descriptions may be further modified (manually or automatically) and, if this function was re-executed, the edit ID provides a way to prevent the overwriting of those messages (an alternative to having an "edit ID" would be to perform (sub)string equality comparison or `.includes()`, but this could fail in case of modified formatting -- e.g. someone manually editing the message via the GitHub web interface.

pdcastro:
@balena-ci retest

automation/deploy-bin.ts
```diff
@@ -57,3 +58,201 @@ export async function release() {
...
+) {
+	const perPage = 30;
+	const octokit = getOctokit();
+	const options = await octokit.repos.listReleases.endpoint.merge({
+		owner,
+		repo,
+		per_page: perPage,
+	});
+	let errCount = 0;
+	for await (const response of octokit.paginate.iterator(options)) {
```
dfunckt: Beautiful!

automation/run.ts
```diff
@@ -24,7 +24,16 @@ import {
...
 } from './build-bin';
-import { release } from './deploy-bin';
+import {
+	release,
+	updateDescriptionOfReleasesAffectedByIssue1359,
+} from './deploy-bin';
+
+function exitWithError(error: Error | string): never {
+	console.error(`Error: ${error}`);
+	process.exit(1);
```
dfunckt: Your comment about satisfying the compiler made me lookup the [docs for `process.exit()`](https://nodejs.org/dist/latest-v10.x/docs/api/process.html#process_process_exit_code) and I bumped on something interesting (that at least I didn't know):

> Calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.

Based on that, you could theoretically get in a situation where you asked Node to log the error before exiting but it will never be printed to output. In fact, it calls out this exact code as a counter-example. The manual suggests you set `process.exitCode` instead and let it die naturally when the event loop has no more work to do, or just throw an error.

> Rather than calling process.exit() directly, the code should set the process.exitCode and allow the process to exit naturally by avoiding scheduling any additional work for the event loop. If it is necessary to terminate the Node.js process due to an error condition, throwing an uncaught error and allowing the process to terminate accordingly is safer than calling process.exit().

I've never seen this happening in practice, and it doesn't *really* matter for this script, but I wonder if we're using this pattern in the CLI which is expected to run in a variety of platforms, in which case I think it makes sense to err on the safe side and replace any `process.exit()`s with exceptions that are caught by an unhandled exception/rejection handler to cleanly print an error (instead of printing stacktraces as it happens by default) and let the process exit naturally.

pdcastro: Indeed. One reason why the CLI doesn't terminate the process with an uncaught error is that uncaught errors are actually caught and reported via Sentry/Raven. Errors that we don't want to be reported via Sentry/Raven are usually "handled" by calling `exitWithExpectedError`, which calls `process.exit()`. One actual side effect of calling `exitWithExpectedError` that I am aware of, related to pending async operations, is that the mixpanel reporting is indeed interrupted before normal/graceful completion. But it's one of those things that "have been working well enough for years," changing the code would not fix any reported bug and might actually introduce new bugs and hurt performance a little bit...

PR #1413, opened 2019-08-28
Refactor oclif-specific code and convert "env rename" and "envs" to oclif

pdcastro:
Change-type: patch
Signed-off-by: Paulo Castro paulo@balena.io


pdcastro:
Replaced with PR #1414

Issue #1391, opened 2019-08-09
Homogenise oclif action loading


LucianBuzzo:
see https://github.com/balena-io/balena-cli/pull/1383#discussion_r312458518

Currently to add a new oclif command to the CLI you need modify lines in:
- `help_ts.ts`
- `app.ts`
- `oclif-utils.ts`
- `capitanodoc.ts`

This situation could be improved so that is simpler and cleaner to add new commands in oclif

pdcastro:
FYI, I'm actually doing some of this work at the moment, as part of the implementation of some other features. PR to follow soon.

Issue #1153, opened 2019-04-03
Add support for setting per-service variables (multi-container / multi-service)


pdcastro:
Probably using a syntax like:
`balena env add varName varValue --application myApp --service myService`

but the exact syntax / command may be decided later.

pdcastro:
Vaguely related issues:
* [Cannot SSH to container to MC application](https://github.com/balena-io/balena-cli/issues/856)
* [Support multi-container local push](https://github.com/balena-io/balena-cli/issues/836)
* [Update docs to reflect MC support](https://github.com/balena-io/balena-cli/issues/806)
* [document that balena sync doesn't work with multicontainer](https://github.com/balena-io/balena-cli/issues/1139)

magiconair:
Please focus on getting feature parity with the UI first. You don't have to solve all the multi-container issues at once. It is possible to set a service env via the UI so just expose that feature via the CLI through an extra cmd line arg.

magiconair:
We're working around this by prefixing config variables for different services, e.g. `FOO_CONFIG`. However, this has the effect that when we update the configuration for a single service, all services are restarted.

magiconair:
@pdcastro Any update here?

pdcastro:
@magiconair, thanks for checking on this issue's progress. This issue has indeed been discussed in a backlog priority meeting just a few days ago, and work on it should start shortly (hopefully next week, if not then soon after).

pdcastro:
Status update: some work has been done on "strategic pre-requisites" for this issue (PR #1413). I then came across a balena-ci (our continuous integration system) issue that is blocking further progress, but I expect to resolve that CI issue one way or another this week.

magiconair:
@pdcastro Any update? You must have an API that the browser uses to perform this function. Can't you just hook this up to the CLI until your API refactor is ready? It's been more than 6 months to get some feature parity between the UI and the CLI. How are we supposed to script and automate with Balena if we cannot use integrate the functions the platform offers?

pdcastro:
[...]

PR #1414, opened 2019-09-02
Migrate "env rename" and "envs" to oclif and refactor the preparser

pdcastro:
In addition to addressing #1391, this PR is a strategic pre-requisite for #1153.

Resolves: #1391 
Change-type: patch


pdcastro:
@balena-ci retest

nazrhom:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

lib/actions-oclif/env/rm.ts
```diff
@@ -89,23 +89,23 @@ export default class EnvRmCmd extends Command {
...
-						resource: 'device_environment_variable',
-						id: params.id,
-					});
-				} else {
-					return balena.pine.delete({
-						resource: 'application_environment_variable',
-						id: params.id,
-					});
-				}
-			});
```
pdcastro: Block converted from promise `.then` to `async/await`, plus some code deduplication and improved error handling.

lib/actions-oclif/env/rm.ts
```diff
@@ -89,23 +89,23 @@ export default class EnvRmCmd extends Command {
...
+			);
+		} catch (err) {
+			if (err.message === 'Aborted') {
+				return patterns.exitWithExpectedError(err);
```
pdcastro: If the user answers "No" to the question, call `exitWithExpectedError` to avoid Raven/Sentry error reporting (it's not a reportable error to answer "No" to a question...).

lib/actions/environment-variables.ts
```diff
@@ -1,154 +0,0 @@
-/*
```
pdcastro: This file implemented `balena envs` and `balena env rename` with Capitano. This PR migrates those commands to oclif.

lib/actions/help.coffee
```diff
@@ -90,15 +90,20 @@ general = (params, options, done) ->
 
 	if options.verbose
 		console.log('\nAdditional commands:\n')
-		print parse(groupedCommands.secondary).concat(getOclifHelpLinePairs()).sort()
+		secondaryCommandPromise = getOclifHelpLinePairs()
+		.then (oclifHelpLinePairs) ->
+			print parse(groupedCommands.secondary).concat(oclifHelpLinePairs).sort()
```
pdcastro: `getOclifHelpLinePairs()` was changed from sync to async, hence this refactoring.

lib/actions/help_ts.ts
```diff
@@ -16,20 +16,29 @@
...
+	for (const convertedCmd of convertedCommands) {
+		const [topic, cmd] = convertedCmd.split(':');
+		const pathComponents = ['..', 'actions-oclif', topic];
+		if (cmd) {
+			pathComponents.push(cmd);
+		}
+		// note that `import(path)` returns a promise
+		cmdClasses.push(import(path.join(...pathComponents)));
+	}
+	return Bluebird.map(cmdClasses, getCmdUsageDescriptionLinePair);
```
pdcastro: Previously, this function hardcoded the list of commands converted from oclif to Capitano. Now the list is substantially centralised in the `preparser` module, and imported here.

lib/app.ts
```diff
@@ -14,156 +14,18 @@
...
-export interface AppOptions {
-	// Prevent the default behaviour of flushing stdout after running a command
-	noFlush?: boolean;
-}
-
-/**
- * Simple command-line pre-parsing to choose between oclif or Capitano.
- * @param argv process.argv
- */
-function routeCliFramework(argv: string[], options: AppOptions): void {
```
pdcastro: The pre-parser code in this module was moved to `preparser.ts`.

lib/utils/oclif-utils.ts
```diff
@@ -17,11 +17,6 @@
 
 import * as Config from '@oclif/config';
 
-export const convertedCommands = {
-	'env:add': 'env add',
-	'env:rm': 'env rm',
```
pdcastro: This object was not used / not referenced in the code, hence deleted (it had been left behind from a previous refactoring).

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "balena-cli",
   "version": "11.9.3",
   "description": "The official balena CLI tool",
-  "main": "./build/actions/index.js",
+  "main": "./build/app.js",
```
pdcastro: I don't think this change has a practical effect, but it better reflects the entrypoint.

package.json
```diff
@@ -29,6 +29,7 @@
       "node_modules/raven/lib/instrumentation/*.js"
     ],
     "assets": [
+      "build/**/*.js",
```
pdcastro: Some oclif-specific code (common-flags.ts) was not being packaged by `pkg`.

tests/commands/help.spec.ts
```diff
@@ -59,9 +59,9 @@ Additional commands:
     device shutdown <uuid>                shutdown a device
     devices supported                     list all supported devices
     env add name [value]                  add an environment or config variable to an application or device
-    env rename <id> <value>               rename an environment variable
+    env rename id value                   change an environment variable for an application or device
```
pdcastro: "rename" is a bit misleading, because the command changes the variable's **value** rather than its **name.** Perhaps the verb should have been "set".

tests/helpers.ts
```diff
@@ -1,55 +1,60 @@
...
 
-	// @ts-ignore
-	process.stdout.write = (log: string) => {
+	const stdoutHook = (log: string | Buffer) => {
 		// Skip over debug messages
-		if (!log.startsWith('[debug]')) {
+		if (typeof log === 'string' && !log.startsWith('[debug]')) {
 			out.push(log);
 		}
-		oldStdOut(log);
```
pdcastro: The intention of the original author was to keep printing stdout and stderr to the screen in addition to capturing it, but it was not working (the output was not being printed to the screen). I've used a 4-year-old npm package (`intercept-stdout`) instead of rewriting it myself.

lib/actions-oclif/env/add.ts
```diff
@@ -71,69 +73,55 @@ export default class EnvAddCmd extends Command {
...
-		const _ = await import('lodash');
+		const cmd = this;
 		const balena = (await import('balena-sdk')).fromSharedOptions();
 		const { exitWithExpectedError } = await import('../../utils/patterns');
 
-		const cmd = this;
+		if (params.value == null) {
+			params.value = process.env[params.name];
 
-		await Bluebird.try(async function() {
```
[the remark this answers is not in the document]
LucianBuzzo: @pdcastro For ease of review, this kind of refactor should be in a separate commit, or even better, a separate PR

tests/commands/help.spec.ts
LucianBuzzo: 🤦‍♂ This is the kind of mistake that will haunt us for a long time - is there a list of breaking changes we want to make anywhere? We should definitely make a not of this

tests/helpers.ts
```diff
@@ -1,55 +1,60 @@
...
 			!log.startsWith('[debug]') &&
 			// TODO stop this warning message from appearing when running
 			// sdk.setSharedOptions multiple times in the same process
 			!log.startsWith('Shared SDK options')
 		) {
 			err.push(log);
 		}
-		oldStdErr(log);
 	};
+	const unhookIntercept = require('intercept-stdout')(stdoutHook, stderrHook);
```
LucianBuzzo: Can you import this dependency at the top of the file using `import` instead of an inline `require`?

lib/actions-oclif/env/add.ts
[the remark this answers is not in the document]
pdcastro: Done - I've split the changes to `add.ts` and `rm.ts` to a separate commit. I don't like multiple PRs for such simple refactoring because:
* Each PR creates a new CLI version and release, and CLI users and product/commercial folks (Shaun, Alex...) will be looking for the "value" of each new release.
* Users will be advised of the new CLI version available for download.
* Each PR/release results in hundreds of megabytes of assets created, uploaded and stored to the GitHub releases page.
* The CLI's automated testing still has significant room for improvement (as you know all too well, #1399), :-) and when bugs are reported, users and developers find themselves manually testing different versions to isolate the problem. The fewer "superfluous" releases, the better.

tests/commands/help.spec.ts
pdcastro: I've now created a GDoc to keep track of such suggested changes: https://docs.google.com/document/d/1uHE-_LLu5gJUQZICQJtlo76EQCm90z0CMAwpkfw6s8w/edit?usp=sharing
I've also further improved the command description to include _"change the value of..."_ so that at least the description is clear about what the command does.

tests/helpers.ts
pdcastro: Check my change: is it good enough to have a `require()` at the top of the function? Placing an import at the top of the file:
```
import * as intercept from 'intercept-stdout';
```
Results in:
```
TSError: ⨯ Unable to compile TypeScript:
tests/helpers.ts:18:28 - error TS7016: Could not find a declaration file for module 'intercept-stdout'. 'balena-io/balena-cli/node_modules/intercept-stdout/intercept-stdout.js' implicitly has an 'any' type.
```

LucianBuzzo: @pdcastro We should do one of the following:
1. Stub the typings for the `intercept-stdout` module in the CLI
2. PR typings to the `intercept-stdout` repo
3. PR typings for the `intercept-stdout` module to DefinitelyTyped

We could in fact do all three of these things, start with 1), if 2) fails, do 3).

pdcastro: As requested: added file `typings/intercept-stdout/index.d.ts` to `balena-cli`, and also created a PR at `sfarthin/intercept-stdout`: https://github.com/sfarthin/intercept-stdout/pull/14

pdcastro:
force push = rebase on master

lib/actions-oclif/env/rename.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+interface ArgsDef {
+	id: number;
+	value: string;
+}
+
+export default class EnvRenameCmd extends Command {
+	public static description = stripIndent`
+		Change the value of an environment variable for an app or device.
```
dfunckt: I wonder how we ended up with a command called `rename` that changes the value instead of the name :/ Do you plan a major CLI release soon @pdcastro? Definitely worth to rename the command to `set` or something similar when we do. In fact, you could do this here properly and have a shim for the `rename` command with a deprecation warning. Does oclif make this kind of thing difficult?

lib/actions-oclif/envs.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+import { Command, flags } from '@oclif/command';
+import { stripIndent } from 'common-tags';
+import * as _ from 'lodash';
+
+import * as cf from '../utils/common-flags';
+import { CommandHelp } from '../utils/oclif-utils';
+
+type ApplicationVariable = import('balena-sdk').ApplicationVariable;
+type DeviceVariable = import('balena-sdk').DeviceVariable;
```
dfunckt: Why do you need to alias these types? Doesn't the following work?

```
import { ApplicationVariable, DeviceVariable } from 'balena-sdk';
```

lib/actions-oclif/env/rename.ts
pdcastro: Thanks - no major release currently planned, but I created a document (as suggested by Lucian) to keep track of suggested breaking changes for a future major release: https://docs.google.com/document/d/1uHE-_LLu5gJUQZICQJtlo76EQCm90z0CMAwpkfw6s8w/edit?usp=sharing

However I think that you and Lucian have only made points about the `rename` command because I noticed and highlighted its oddity in the first place. We have a substantial backlog of issues/bugs that have been reported by internal and external users (things that are "truly bothering" them), so I would give priority to addressing those issues.

lib/actions-oclif/envs.ts
pdcastro: I hadn't thought about it in terms of "aliasing" the types, and yes what you've suggested works, but... I meant to exercise a TypeScript 2.9 feature of importing types only, without generating a `require` statement in transpiled JS, while avoiding the "need" for two `import 'balena-sdk'` statements: one at the top of the file for types only, and one inside the `run()` function for the runtime importing of the module.

I understand that the import statement you suggested is also "likely" to avoid generating a `require` statement in transpiled JS (depending on the `tsc` being sufficiently clever), but it leads to CLI code as follows:
```
import { ApplicationVariable, DeviceVariable } from 'balena-sdk';

export default class EnvsCmd extends Command {
	public async run() {
		const balena = (await import('balena-sdk')).fromSharedOptions();
```

One would look at the code above and think (like I myself did in the past) _"oh why have that inner `await import('balena-sdk')` if there's already an `import` statement at the top of the file? Let me "fix" this "code duplication" by moving everything to the top of the file:"_
```
import { ApplicationVariable, DeviceVariable, fromSharedOptions } from 'balena-sdk';

export default class EnvsCmd extends Command {
	public async run() {
		const balena = fromSharedOptions();
```
But the latter would actually cause a `require` statement to be generated at the top of the file in transpiled code. (And we generally want to avoid that in order to reduce CLI start time, even though oclif's on-demand importing of command files makes it less of an issue than with Capitano.)

Anyway, in summary, the reason for using `type x = import('y').x` is to make it clear (?) and explicit that it's a type import only, without producing a `require` in transpiled code. Of course, I wasn't that successful if you had to ask the question... :-)

tests/helpers.ts
pdcastro: As there wasn't activity on [the PR for the intercept-stdout package](https://github.com/sfarthin/intercept-stdout/pull/14) after a week, I've progressed to step 3 and created a PR against DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38373

pdcastro:
force push = rebase on master (bringing the PR up to date)
@LucianBuzzo, note that even though Akis has approved the PR, merge remains blocked by your request for changes. Please check if I've addressed the changes you had requested. Thanks!

lib/actions-oclif/envs.ts
LucianBuzzo: @pdcastro if you need to convey meaning, you can do so with a comment. The format @dfunckt suggested makes a lot more sense and is the pattern we use throughout every other project in the company.

pdcastro: Code changed as suggested, thanks.

Issue #1419, opened 2019-09-09
xxhash was compiled against a different Node.js version


tangb:
- **balena CLI version:** e.g. 11.2.1 (output of the `"balena version"` command)
balena-cli v11.9.3

- **Operating system version:** e.g. Windows 10, Ubuntu 18.04, macOS 10.14.5
windows10

- **32/64 bit OS and processor:** e.g. 32-bit Windows on 64-bit Intel processor
64bits

- **Install method:** npm or zip or executable installer
zip archive available https://github.com/balena-io/balena-cli/releases/tag/v11.9.3

- **If npm install, Node.js and npm version:** e.g. Node v8.16.0 and npm v6.4.1
npm 6.11.2
node v10.16.3
---

Hello

I'm trying to run balena-cli command line under windows 10  
`balena.exe local flash "C:\temp\raspbian.zip" --drive "\\.\PHYSICALDRIVE1" --yes`

and I got the following error:
```
The module '\\?\c:\Users\tanguy.bonneau\AppData\Roaming\CleepDesktop\balena-cli\xxhash\build\Release\hash.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 57. This version of Node.js requires
NODE_MODULE_VERSION 64. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).

Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, or "export DEBUG=1"
on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```

I understand the error, but I don't figure out how to find a workaround while the binary is delivered in the release archive (no way to rebuild it).

Maybe you can help me ?
I saw a new release of xxhash has been released few days ago after 3 years of inactivity.

Thank you very much for your answer ;-)

pdcastro:
@tangb, thank you for reporting this issue! I have reproduced it and I believe it is a bug in our CI (continuous integration) build scripts, which I will investigate further and fix. Meanwhile, I have tested the CLI's "executable installer" for Windows and it seems that it is not affected by this issue, so you could use it instead of the standalone zip package:
[...]

PR #1423, opened 2019-09-10
Fix Node.js version errors in standalone package on Windows

pdcastro:
Resolves: #1419
Change-type: patch

### DIFF
--- a/automation/build-bin.ts
+++ b/automation/build-bin.ts
@@ -19,11 +19,13 @@ import { run as oclifRun } from '@oclif/dev-cli';
 import * as archiver from 'archiver';
 import * as Bluebird from 'bluebird';
 import { execFile, spawn } from 'child_process';
+import { stripIndent } from 'common-tags';
 import * as filehound from 'filehound';
 import * as fs from 'fs-extra';
 import * as path from 'path';
 import { exec as execPkg } from 'pkg';
 import * as rimraf from 'rimraf';
+import * as semver from 'semver';
 import * as shellEscape from 'shell-escape';
 import * as util from 'util';
 
@@ -106,7 +108,7 @@ export async function runUnderMsys(argv?: string[]) {
 async function buildPkg() {
 	const args = [
 		'--target',
-		'node10',
+		'host',
 		'--output',
 		'build-bin/balena',
 		'package.json',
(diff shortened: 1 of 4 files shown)


automation/build-bin.ts
```diff
@@ -106,7 +108,7 @@ export async function runUnderMsys(argv?: string[]) {
 async function buildPkg() {
 	const args = [
 		'--target',
-		'node10',
+		'host',
```
pdcastro: This is the key fix, and it's my bad. In a previous PR (#1328, CLI version 11.4.0), I had specifically changed this option from 'host' to 'node10', aiming at improving consistency among the standalone zip packages (pkg) between Linux / Mac / Windows and Travis / Balena CI, having them all use the same Node 10 version rather than whatever Node was installed, and thus arguably helping with testing (fewer differences to account for). But I misunderstood how it works: `pkg` goes as far as downloading its own patched Node.js interpreter that gets bundled in the final CLI executable, but that interpreter was not used to build the native modules, which were built in the `npm install` step using whatever `node` is available in the path. As it happens, Node 10 was used in the balena CI setup for Linux and Mac, so that change didn't matter on those platforms (and reason why the problem was not detected any sooner, including in my manual testing). But on Windows, the CI setup still uses Node 8 so this led to native modules being compiled with Node 8 but bundled with Node 10 in the standalone package. (Ironically and perversely, one of the main benefits of standalone zip packages is to ensure compatibility between native modules and Node.js, and this bug ensured the very opposite... on Windows, between versions 11.4.0 and 11.9.3.)

To prevent a future regression, I have added a simple `testPkg()` function (below in this same file) that executes `balena version -j` (the built standalone executable) and ensures that the reported Node.js major version matches the major version of `process.version`.

automation/build-bin.ts
```diff
@@ -151,6 +153,69 @@ async function buildPkg() {
 	);
 }
 
+/**
+ * Run some basic tests on the built pkg executable.
+ */
+async function testPkg() {
```
pdcastro: So far we didn't have any automated testing for standalone zip packages. This function tests only `balena version -j` (comparing the output against `process.version`, to ensure that the native modules were compiled with a compatible version of Node.js). In the future, we should probably look at somehow integrating some of the the CLI source code / command tests (mocha, chai) so that they also test the built pkg executable.

package.json
```diff
@@ -172,7 +172,7 @@
     "dockerode": "^2.5.8",
     "dockerode-options": "^0.2.1",
     "ejs": "^2.5.7",
-    "etcher-sdk": "^2.0.5",
+    "etcher-sdk": "^2.0.14",
```
pdcastro: This change is in a separate commit. It's not completely random that this commit is bundled in this PR, because the the `balena-cli` issue (#1419) resolved by this PR regarded the `balena local flash` command, with an error that pointed to the `xxhash` native module, which is a dependency of [etcher-sdk](https://github.com/balena-io-modules/etcher-sdk). As it happens, `etcher-sdk` v2.0.14 bumped `xxhash` in relation to Node 12 support.

Note that, as indicated in the `npm-shrinkwrap.json` file, the actual `etcher-sdk` version bump was from `2.0.13` to `2.0.14`, rather than from `2.0.5` to `2.0.14` as `package.json` might lead you to believe. The `etcher-sdk` [changelog](https://github.com/balena-io-modules/etcher-sdk/blob/master/CHANGELOG.md) for `2.0.14` points only to the `xxhash` bump, hence this commit / PR also being of type `patch` rather than `minor`.

Issue #1304, opened 2019-06-11
Developer needs to enable device local mode with CLI


roman-mazur:
- **Cli version:** 11.1.0
- **Standalone install:** no, native installer
- **Node version:** 12.4.0
- **Npm version:** 6.9.0
- **Operating system and architecture:** MacOS, x86_64

It currently does not seem to be possible to enable local mode from the command line.

tobyhijzen:
I would like to have this too

thgreasi:
Let me open an SDK issue for this.
See: https://github.com/balena-io/balena-sdk/issues/732

thgreasi:
The related local mode methods are now available in the balena-sdk v9.10.0 .

hedss:
@thgreasi Awesome, great work! I'm writing masterclass tutorials atm. We really, really do need this in and I'd like to prioritise it at the next CLI checkin. cc @shaunmulligan

pdcastro:
We've discussed this issue in today's CLI check-in and I just wanted to leave a comment on what we agreed the command could look like:
```
balena device config local-mode enable/disable <UUID>
```
So we would introduce a `device config` command that sets configuration options such as `local-mode` or `public-url` to values such as `enable` or `disable`. At the same time, we agreed we should migrate the `device public-url` command to the config form:
```
balena device config public-url enable/disable/status <UUID>
```
... and add a deprecation message to the the old `device public-url` form.

srlowe:
@pdcastro Regarding the suggestion above (`balena device config local-mode enable/disable <UUID>`)...

It seem oclif doesn't support this 'depth' of command, ie. `balena <topic> <suptopic> <command>` (I tested). I'm not sure how much work it would be to support that, but given that we have already hacked the framework to support spaces instead of colons, it's probably not a good idea to hack it further.

`balena device-config local-mode enable/disable <UUID>` is an option, but doesn't seem very nice.

Another option would be for `balena device config` to be a kind of supercommand (that accepts a command param, e.g. `local-mode`)

pdcastro:
> Another option would be for balena device config to be a kind of supercommand (that accepts a command param, e.g. local-mode)

@srlowe, I think this supercommand option is reasonable, not by patching oclif but by coding our own trivial parser of the first two positional arguments (`if (params.cmd === 'local-mode') ...`). So `device config` is the oclif command, while `local-mode` and `enable` are positional arguments to `device config`.

srlowe:
[...]

PR #1425, opened 2019-09-11
Update balena-sdk to v12 and mitigate MaxListenersExceededWarning

pdcastro:
Connects-to: #1153 
Connects-to: #1304 
Change-type: patch


lib/utils/logger.ts
```diff
@@ -64,6 +64,15 @@ class Logger {
...
 	}
 
+	private static logger: Logger;
+
+	public static getLogger() {
+		if (!this.logger) {
+			this.logger = new Logger();
+		}
+		return this.logger;
+	}
```
pdcastro: This new `getLogger()` method and its usage throughout CLI are in their own commit. These changes are an answer to the following warnings:
```
$ balena tunnel test-rpi -p 22222:43213
(node:43572) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 drain listeners added. Use emitter.setMaxListeners() to increase limit
(node:43572) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit
(node:43572) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added. Use emitter.setMaxListeners() to increase limit
```
I ran the CLI with warning stack traces enabled and the traces pointed to instances of Logger. It would have been the combination of Logger instances with other event listener usage elsewhere that would have combined to produce the warnings, so it's still that possible that a memory leak (dangling listeners) exist somewhere. But the static Logger change was sufficient to prevent the warning, hence at least mitigating the issue for now.

package.json
```diff
@@ -135,7 +135,7 @@
     "shell-escape": "^0.2.0",
     "sinon": "^7.4.1",
     "ts-node": "^8.1.0",
-    "typescript": "3.4.3"
+    "typescript": "^3.4.5"
```
pdcastro: Updated to match `balena-sdk`.

package.json
```diff
@@ -151,7 +151,7 @@
     "balena-device-status": "^3.1.2",
     "balena-image-manager": "^6.0.0",
     "balena-preload": "^8.2.1",
-    "balena-sdk": "^11.18.0",
+    "balena-sdk": "^12.10.0",
```
pdcastro: Mainly a pre-req for #1304, but also useful for #1153 and other issues.

lib/utils/cloud.ts
```diff
@@ -1,9 +1,26 @@
...
+ * limitations under the License.
+ */
+
+import { BalenaSDK, Service } from 'balena-sdk';
 import memoize = require('lodash/memoize');
 
 export const serviceIdToName = memoize(
 	async (sdk: BalenaSDK, serviceId: number): Promise<string | undefined> => {
-		const serviceName = await sdk.pine.get({
+		const serviceName = await sdk.pine.get<Service>({
```
pdcastro: Required as a result of the new type information added to `balena-sdk` v12.

lib/utils/helpers.ts
```diff
@@ -150,7 +152,7 @@ export function getApplication(applicationName: string) {
 	// that off to a special handler (before importing any modules)
 	const match = applicationName.split('/');
 
-	const extraOptions = {
+	const extraOptions: BalenaSdk.PineOptionsFor<BalenaSdk.Application> = {
```
pdcastro: Required as a result of the new type information added to `balena-sdk` v12.

pdcastro:
In addition to the existing automated tests, I've done some manual testing (on macOS) with commands such as:
```
balena login
balena apps
balena deploy --build
balena push
balena devices
balena device rename
balena envs
balena env rename
```
... and I found no problems related to the `balena-sdk` v12 major version bump.

lib/utils/logger.ts
```diff
@@ -19,6 +19,13 @@ import _ = require('lodash');
...
 import { StreamLogger } from 'resin-stream-logger';
 
+/**
+ * General purpose logger class with support for log streams and colours.
+ * Call `Logger.getLogger()` to retrieve a global shared instance of this
+ * class. The `new Logger()` pattern is not recommended because it may lead
+ * to Node printing "MaxListenersExceededWarning" warning messages to the
+ * console.
+ */
 class Logger {
```
pdcastro: > _I wonder how could we preserve some of those "why"-style comments _in_ the code_

Thanks for the review! I've added some JSDoc comments to the Logger class and to the `getLogger()` method. Also, I've added the `protected` modified to the Logger `constructor`, which forces the use of `getLogger()` (preventing the `new Logger()` usage, at least in Typescript code). I hope this is not a step too far...

Issue #1433, opened 2019-09-16
`balena move` should allow having mixed device type fleets


thgreasi:
Currently it only shows apps of the same device type as the device's.
It should allow list apps of a compatible architecture.

PR #1434, opened 2019-09-16
Support moving devices to applications of a compatible architecture

thgreasi:
Connects-to: #1433
Change-type: minor

Resolves: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist


lib/actions/device.coffee
```diff
@@ -366,11 +366,22 @@ exports.move =
...
+				compatibleDeviceTypes = deviceTypes.filter (dt) ->
+						balena.models.os.isArchitectureCompatibleWith(deviceDeviceType.arch, dt.arch) &&
+						!!dt.isDependent == !!deviceDeviceType.isDependent &&
+						dt.state != 'DISCONTINUED'
+
+				return patterns.selectApplication (application) ->
+					return _.every [
+						_.some(compatibleDeviceTypes, (dt) -> dt.slug == application.device_type)
+						device.belongs_to__application[0].app_name isnt application.app_name
+					]
```
pdcastro: If I was to be picky... I would say that this logic block could be a separate function (`selectCompatibleArchApps()`), written in Typescript in some other file, imported and called here. But I am approving the PR in any case.

Issue #1379, opened 2019-08-06
Detached mode should use stdout for parseable output only


zrzka:
**From the support:**

hey folks, we're updating our build tooling to use --detached mode in balena build, and it's great, but with one small issue...
Because detached mode reports its status as a JSON object on stdout, the fact that your DEBUG output also goes to stdout is now a problem. Imagine the following pipeline:
RELID=$(balena push APPNAME --detached | jq -r .releaseId)
If DEBUG=true when that runs, jq will attempt to parse the debug output and choke horribly.
Ideally, when running in detached mode, only the parseable output would ever go to stdout, and debug logs etc would be sent to stderr.

PR #1438, opened 2019-09-18
Send all debug output to stderr

CameronDiver:
Closes: https://github.com/balena-io/balena-cli/issues/1379
Change-type: patch

---
##### Contributor checklist

### DIFF
--- a/automation/build-bin.ts
+++ b/automation/build-bin.ts
@@ -22,6 +22,7 @@ import { execFile, spawn } from 'child_process';
 import { stripIndent } from 'common-tags';
 import * as filehound from 'filehound';
 import * as fs from 'fs-extra';
+import * as _ from 'lodash';
 import * as path from 'path';
 import { exec as execPkg } from 'pkg';
 import * as rimraf from 'rimraf';
@@ -348,7 +349,18 @@ async function getSubprocessStdout(
 		child.stderr.on('data', (data: Buffer) => {
 			try {
 				const stderr = data.toString();
-				reject(new Error(`"${execPath}": non-empty stderr "${stderr}"`));
+
+				// ignore any debug lines, but ensure that we parse
+				// every line provided to the stderr stream
+				const lines = _.filter(
+					stderr.trim().split(/\r?\n/),
+					line => !line.startsWith('[debug]'),
+				);
+				if (lines.length > 0) {
+					reject(
+						new Error(`"${execPath}": non-empty stderr "${lines.join('\n')}"`),
+					);
+				}
 			} catch (err) {
 				reject(err);
 			}
--- a/lib/utils/logger.ts
+++ b/lib/utils/logger.ts
@@ -63,8 +63,10 @@ class Logger {
 		};
 
 		_.forEach(this.streams, function(stream, key) {
-			if (key !== 'debug' || process.env.DEBUG) {
+			if (key !== 'debug') {
 				stream.pipe(process.stdout);
+			} else if (process.env.DEBUG) {
+				stream.pipe(process.stderr);
 			}
 		});
 
--- a/lib/actions/push.ts
+++ b/lib/actions/push.ts
@@ -268,7 +268,7 @@ export const push: CommandDefinition<
 
 		const source = options.source || '.';
 		if (process.env.DEBUG) {
-			console.log(`[debug] Using ${source} as build source`);
+			console.error(`[debug] Using ${source} as build source`);
 		}
 
 		const dockerfilePath = validateSpecifiedDockerfile(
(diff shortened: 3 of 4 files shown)


lib/utils/logger.ts
```diff
@@ -63,8 +63,10 @@ class Logger {
 		};
 
 		_.forEach(this.streams, function(stream, key) {
-			if (key !== 'debug' || process.env.DEBUG) {
+			if (key !== 'debug') {
 				stream.pipe(process.stdout);
+			} else if (process.env.DEBUG) {
+				stream.pipe(process.stderr);
```
pdcastro: Good! Arguably I think all logging output (whatever the log level: info, warn, debug, error...) should always have been printed to stderr, and only "application output" (like the output of `balena version` or `balena envs`) should be printed to stdout. But such change would break compatibility and would require a major version bump. I've added this note for future consideration in the GDoc: [CLI - breaking changes for next major release](https://docs.google.com/document/d/1uHE-_LLu5gJUQZICQJtlo76EQCm90z0CMAwpkfw6s8w/edit?usp=sharing)

pdcastro:
@CameronDiver, the build is failing in a function that tests the built standalone zip package:
```
Testing standalone package "C:\Users\Administrator\concourse\work\volumes\live\f44e29b8-9516-4359-6e1d-b661dcb9cced\volume\build-bin\balena.exe"...
Error creating or testing standalone zip package:
 Error: "C:\Users\Administrator\concourse\work\volumes\live\f44e29b8-9516-4359-6e1d-b661dcb9cced\volume\build-bin\balena.exe": non-empty stderr "[debug] original argv0="C:\Users\Administrator\concourse\work\volumes\live\f44e29b8-9516-4359-6e1d-b661dcb9cced\volume\build-bin\balena.exe" argv=[C:\Users\Administrator\concourse\work\volumes\live\f44e29b8-9516-4359-6e1d-b661dcb9cced\volume\build-bin\balena.exe,C:\snapshot\volume\bin\balena,version,-j] length=4
"
```
The test code runs `balena version -j` and expects the output in stdout, and if anything is printed to stderr, the test fails:
https://github.com/balena-io/balena-cli/blob/v11.11.0/automation/build-bin.ts#L169-L171
```
// Run `balena version -j`, parse its stdout as JSON, and check that the
// reported Node.js major version matches semver.major(process.version)
const stdout = await getSubprocessStdout(pkgBalenaPath, ['version', '-j']);
```
Could it be that something in the PR caused `balena version -j` to print to `stderr` instead of `stdout`?

pdcastro:
Btw, FYI, when investigating the test failure, a rebase will be needed as well as I've just merged another PR.

pdcastro:
> Could it be that something in the PR caused balena version -j to print to stderr instead of stdout?

Duh, the error output includes the answer to my own question:
```
non-empty stderr "[debug] original argv0="/tmp/build/8e514326/versioned-source/build-bin/balena" argv=[/tmp/build/8e514326/versioned-source/build-bin/balena,/snapshot/versioned-source/bin/balena,version,-j] length=4"
```
So it's a debug message (above) that gets printed to stderr, causing the test to fail. Perhaps we could do something similar to what some other test code does, and ignore lines starting with `[debug]`:
https://github.com/balena-io/balena-cli/blob/v11.11.1/tests/helpers.ts#L40
```
if (!log.startsWith('[debug]'))
```

automation/build-bin.ts
```diff
@@ -348,7 +348,9 @@ async function getSubprocessStdout(
 		child.stderr.on('data', (data: Buffer) => {
 			try {
 				const stderr = data.toString();
-				reject(new Error(`"${execPath}": non-empty stderr "${stderr}"`));
+				if (!stderr.trim().startsWith('[debug]')) {
+					reject(new Error(`"${execPath}": non-empty stderr "${stderr}"`));
+				}
```
pdcastro: Hmm I'm not sure if the 'data' event is necessarily triggered one text line at a time... If it was then this would be fine, but if it could create a 'data' event with, say, 2 lines as follows, then a real error would be ignored:
```
[debug] First line can be ignore, it's a debug message
Error: unable to create standalone zip package because this PR broke it
```
If `data.toString()` consisted of the 2 lines above, the error would be ignored.

pdcastro:
I've pre-(re)-approved this PR to avoid delays, but it would be good to look into the question of a single `'data'` event being generated for multiple lines of text at once, in which case the `getSubprocessStdout()` function may need a bit of extra logic to prevent ignoring standalone zip package errors when the multi-line string starts with `'[debug]'`.

CameronDiver:
@pdcastro that's fair enough, I think it's line buffered anyway, but I can happily add in some other code to make sure.

PR #1452, opened 2019-09-26
Update TROUBLESHOOTING doc re 'balena ssh' line wrapping / cursor behavior

pdcastro:
Change-type: patch

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -83,3 +83,28 @@ Try resetting the ownership by running:
 ```sh
 $ sudo chown -R <user> $HOME/.balena
 ```
+
+## Broken line wrapping / cursor behavior with `balena ssh`
+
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -83,3 +83,30 @@ Try resetting the ownership by running:
 ```sh
 $ sudo chown -R <user> $HOME/.balena
 ```
+
+## Broken line wrapping / cursor behavior with `balena ssh`
+
+Users sometimes come across broken line wrapping / cursor behavior in text terminals, for example when long command lines are typed in a `balena ssh` session, or a misbehaving cursor when using text editors like `vim` or `nano`. As far as we know, this is not something specific to balena CLI, being also commonly reported issues with standard remote terminal tools like `ssh` or `telnet`. It is often a remote shell configuration issue (files like `/etc/profile`, `~/.bash_profile`, `~/.bash_login`, `~/.profile`, etc., or the lack of any such file at all), including UTF-8 misconfiguration, the use of ASCII control characters in shell prompt formatting (e.g. `$PS1` env var) or the output of tools or log files that use colored text. The issue can sometimes be fixed by resizing the client terminal window, or by running one or more of the following commands on the shell:
```
hedss: Probably worth removing the 'As far as we know' line, as we then describe why it usually isn't.

pdcastro: Done

pdcastro:
> Didn't love the google search urls, but I'm fine with them.

I've removed the 2 google search URLs. They were relevant, but perhaps too explicitly LMGTFY, right? :-)

Issue #1428, opened 2019-09-13
`balena os initialize` no longer initialises an image


hedss:
- **balena CLI version:** 11.9.5
- **Operating system version:** macOS 10.14.5
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** npm
- **If npm install, Node.js and npm version:** Node v10.16.0 and npm v6.9.0

Running the command results in balenaCLI returning the help text (as in an invalid command has been run):
```
$ DEBUG=1 balena os initialize balena-cloud-freee-raspberrypi3-2.38.0+rev1-dev-v9.15.7.img --type raspberrypi3 --drive /dev/disk3 -y
[debug] original argv0="node" argv=[/Users/heds/.nvm/versions/node/v10.16.0/bin/node,/Users/heds/.nvm/versions/node/v10.16.0/bin/balena,os,initialize,balena-cloud-freee-raspberrypi3-2.38.0+rev1-dev-v9.15.7.img,--type,raspberrypi3,--drive,/dev/disk3,-y] length=10
Initializing device

Note: Initializing the device may ask for administrative permissions
because we need to access the raw devices directly.
Going to erase /dev/disk3.
If asked please type your computer password to continue
[debug] original argv0="/Users/heds/.nvm/versions/node/v10.16.0/bin/node" argv=[/Users/heds/.nvm/versions/node/v10.16.0/bin/node,/Users/heds/.nvm/versions/node/v10.16.0/bin/balena,internal,sudo,internal osinit balena-cloud-freee-raspberrypi3-2.38.0+rev1-dev-v9.15.7.img raspberrypi3 {"drive":"/dev/disk3"}] length=5
Usage: balena [COMMAND] [OPTIONS]

If you need help, or just want to say hi, don't hesitate in reaching out
through our discussion and support forums at https://forums.balena.io

For bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/

Primary commands:

    help [command...]                       show help
    login                                   login to balena
    push <applicationOrDevice>              Start a remote build on the balena cloud build servers or a local mode device
    logs <uuidOrDevice>                     show device logs
    ssh <applicationOrDevice> [serviceName] SSH into the host or application container of a device
    apps                                    list all applications
    app <name>                              list a single application
    devices                                 list all devices
    device <uuid>                           list a single device
    tunnel <deviceOrApplication>            Tunnel local ports to your balenaOS device
[...]

PR #1455, opened 2019-10-01
Fix 'os initialize' (auto sudo execution)

pdcastro:
Resolves: #1428 
Change-type: patch

### DIFF
--- a/lib/actions/internal.coffee
+++ b/lib/actions/internal.coffee
@@ -54,35 +54,3 @@ exports.scanDevices =
 			.then (hostnameOrIp) ->
 				console.error("==> Selected device: #{hostnameOrIp}")
 		.nodeify(done)
-
-exports.sudo =
-	signature: 'internal sudo <command>'
-	description: 'execute arbitrary commands in a privileged subprocess'
-	help: '''
-		Don't use this command directly!
-
-		<command> must be passed as a single argument. That means, you need to make sure
-		you enclose <command> in quotes (eg. balena internal sudo 'ls -alF') if for
-		whatever reason you invoke the command directly or, typically, pass <command>
-		as a single argument to spawn (eg. `spawn('balena', [ 'internal', 'sudo', 'ls -alF' ])`).
-
-		Furthermore, this command will naively split <command> on whitespace and directly
-		forward the parts as arguments to `sudo`, so be careful.
-	'''
-	hidden: true
-	action: (params, options, done) ->
-		os = require('os')
-		Promise = require('bluebird')
-
-		return Promise.try ->
-			if os.platform() is 'win32'
-				windosu = require('windosu')
-				windosu.exec(params.command, {})
-			else
-				{ spawn } = require('child_process')
-				{ wait } = require('rindle')
-				cmd = params.command.split(' ')
-				ps = spawn('sudo', cmd, stdio: 'inherit', env: process.env)
-				wait(ps)
-
-		.nodeify(done)
--- a/lib/utils/helpers.ts
+++ b/lib/utils/helpers.ts
@@ -20,11 +20,10 @@ import chalk from 'chalk';
 import _ = require('lodash');
 import os = require('os');
 import visuals = require('resin-cli-visuals');
-import rindle = require('rindle');
+import * as ShellEscape from 'shell-escape';
 
 import { InitializeEmitter, OperationState } from 'balena-device-init';
 
-const waitStreamAsync = Bluebird.promisify(rindle.wait);
 const balena = BalenaSdk.fromSharedOptions();
 
 export function getGroupDefaults(group: {
@@ -59,19 +58,44 @@ export function stateToString(state: OperationState) {
 	}
 }
 
-export function sudo(
+/**
+ * Execute a child process with admin / superuser privileges, prompting the user for
+ * elevation as needed, and taking care of shell-escaping arguments in a suitable way
+ * for Windows and Linux/Mac.
+ *
+ * @param command Unescaped array of command and args to be executed. If isCLIcmd is
+ * true, the command should not include the 'node' or 'balena' components, for example:
+ * ['internal', 'osinit', ...]. This function will add argv[0] and argv[1] as needed
+ * (taking process.pkg into account -- CLI standalone zip package), and will also
+ * shell-escape the arguments as needed, taking into account the differences between
+ * bash/sh and the Windows cmd.exe in relation to escape characters.
+ * @param msg Optional message for the user, before the password prompt
+ * @param stderr Optional stream to which stderr should be piped
+ * @param isCLIcmd (default: true) Whether the command array is a balena CLI command
+ * (e.g. ['internal', 'osinit', ...]), in which case process.argv[0] and argv[1] are
+ * added as necessary, depending on whether the CLI is running as a standalone zip
+ * package (with Node built in).
+ */
+export async function sudo(
 	command: string[],
-	{ stderr, msg }: { stderr?: NodeJS.WritableStream; msg?: string } = {},
+	{
+		stderr,
+		msg,
+		isCLIcmd,
+	}: { stderr?: NodeJS.WritableStream; msg?: string; isCLIcmd?: boolean } = {},
 ) {
-	const { executeWithPrivileges } = require('./sudo');
+	const { executeWithPrivileges } = await import('./sudo');
 
 	if (os.platform() !== 'win32') {
 		console.log(
-			msg || 'If asked please type your computer password to continue',
+			msg ||
+				'Admin privileges required: you may be asked for your computer password to continue.',
 		);
 	}
-
-	return executeWithPrivileges(command, stderr);
+	if (isCLIcmd == null) {
+		isCLIcmd = true;
+	}
+	await executeWithPrivileges(command, stderr, isCLIcmd);
 }
 
 export function runCommand(command: string): Bluebird<void> {
@@ -106,7 +130,7 @@ export async function getOsVersion(
 	return init.getImageOsVersion(image, manifest);
 }
 
-export function osProgressHandler(step: InitializeEmitter) {
+export async function osProgressHandler(step: InitializeEmitter) {
 	step.on('stdout', process.stdout.write.bind(process.stdout));
 	step.on('stderr', process.stderr.write.bind(process.stderr));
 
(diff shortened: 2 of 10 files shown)


lib/actions/internal.coffee
```diff
@@ -59,7 +59,8 @@ exports.sudo =
 	signature: 'internal sudo <command>'
 	description: 'execute arbitrary commands in a privileged subprocess'
 	help: '''
-		Don't use this command directly!
+		WARNING: THIS INTERNAL COMMAND IS DEPRECATED and will be removed in the next
+		major version release of the CLI.
```
pdcastro: * The only use I had found of `"internal sudo"` was removed in this PR, so I even thought of deleting the command straight away... But technically it would be a breaking change, hence opting for the warning instead. No hurry anyway.
* I've failed to see the advantage `"internal sudo"` over simply calling a function like `helpers.sudo(cmd)` with the desired command (CLI command or otherwise) as argument. Especially given that the implementation of internal sudo simply does `spawn('sudo')` or `windosu.exe()`, which is what `helpers.sudo()` does. The privilege elevation happens on the child process only.

lib/utils/sudo.ts
```diff
@@ -14,34 +14,78 @@
...
-	};
+): Promise<void> {
+	// whether the CLI is already running with admin / super user privileges
+	const isElevated = await (await import('is-elevated'))();
+	const { shellEscape } = await import('./helpers');
+	const escapedArgs = shellEscape(command);
 
-	const args = process.argv
-		.slice(0, 2)
-		.concat(['internal', 'sudo', command.join(' ')]);
```
pdcastro: The previous code would call "`internal sudo`" here, which in turn would call `spawn` or `windosu`. It seemed to me like an unnecessary additional child process (executing the CLI itself as a child, and the CLI is unfortunately slow-ish), so I moved the implementation of "`internal sudo`" here (with modifications) and called it directly.

lib/utils/sudo.ts
```diff
@@ -14,34 +14,78 @@
...
+			stdio: ['inherit', 'inherit', stderr ? 'pipe' : 'inherit'],
+		};
+		let ps: ChildProcess;
+		let errMsgCmd: string;
 
-	const ps = spawn(args[0], args.slice(1), opts);
+		if (isElevated) {
+			// already running with privileges: simply spawn the command
+			errMsgCmd = escapedArgs.join(' ');
+			ps = spawn(escapedArgs[0], escapedArgs.slice(1), opts);
```
pdcastro: `if (isElevated)` addresses the following cases:
* A small performance improvement by avoiding calling sudo or windosu when the CLI is already being executed a root / admin. In this case, the Windows / Linux / Mac code paths all run this same `spawn` line (as opposed to `windosu` vs `sudo`).
* Some Docker containers in CI environments may not even have `sudo` installed -- so it's one fewer dependency to worry about. It's actually often how I test the CLI on Linux on a balenaOS NUC computer, using a minimalistic Debian container that doesn't come with sudo unless I `apt-get` it.

lib/utils/sudo.ts
```diff
@@ -14,34 +14,78 @@
...
+	escapedArgs: string[],
+	stderr?: NodeJS.WritableStream,
+): Promise<void> {
 	if (stderr) {
-		ps.stderr.pipe(stderr);
+		const msg = stripIndent`
+			Error: unable to elevate privileges. Please run the command prompt as an Administrator:
+			https://www.howtogeek.com/194041/how-to-open-the-command-prompt-as-administrator-in-windows-8.1/
+		`;
+		throw new Error(msg);
```
pdcastro: * `windosu` can't capture stderr / stdout, hence the error for CLI commands that require stderr capture, like the interactive forms of `balena join`, `balena leave`. The user can get around it by re-executing the shell as with admin privileges.
* I pinged Juan, Lorenzon and Alexis about it: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/4i6tUS12v6y-AHAMu9qT9VXSop9
* Alexis recommended `sudo-prompt` as an alternative to `windosu`. I tested it, and found that while `windosu` doesn't capture enough, `sudo-prompt` captures too much: captures and buffers (`Buffer` / `string` rather than `stream`) stdout and stderr. This is a problem because the elevated privilege child process is expected to interact with the user via stdin/stdout (interactive text menus) before producing a capturable output in stderr. (The whole design is questionable, but it's how it is.)
* I actually had a go at patching both/either `windosu` and `sudo-prompt`, perhaps against best advice (I blame Page, :-) because of his defence of user-friendliness of the on-demand privilege elevation on Windows, versus the inconvenient of re-starting the shell as admin...) Sadly it proved trickier than I had hoped. The `sudo-prompt` package is more promising in terms of patching (for future attempts), but both `windosu` and `sudo-prompt` spawn child processes that would have to align / coordinate with the CLI's objective of capturing stderr only, leaving stdin and stdout alone. It sounds like a simple objective, but it proved a hair-pulling sort of debugging and I eventually gave up. As discussed in the previous CLI check-in, this had never worked before on Windows (stderr capture), so even the advice of re-executing the shell as admin is an improvement, and we have higher-priority work to focus on.

patches/resin-cli-visuals+1.4.3.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/node_modules/resin-cli-visuals/build/widgets/drive/index.js b/node_modules/resin-cli-visuals/build/widgets/drive/index.js
+index 475c0bc..338fc3c 100644
+--- a/node_modules/resin-cli-visuals/build/widgets/drive/index.js
++++ b/node_modules/resin-cli-visuals/build/widgets/drive/index.js
+@@ -21,7 +21,7 @@ chalk = require('chalk');
+ 
+ Promise = require('bluebird');
+ 
+-drivelist = Promise.promisifyAll(require('drivelist'));
++drivelist = require('drivelist');
```
pdcastro: So I found (the hard way) that `resin-cli-visuals` was broken (following an update to the npm `drivelist` package), preventing `os intialize` from working. I intend to submit a PR (simple enough) in the `resin-cli-visuals` repo, but for interim testing I found it convenient to have this patch.

package.json
```diff
@@ -33,7 +33,9 @@
       "build/actions-oclif",
       "build/auth/pages/*.ejs",
       "build/hooks",
-      "node_modules/resin-discoverable-services/services/**/*"
+      "node_modules/resin-discoverable-services/services/**/*",
+      "node_modules/windosu/*.bat",
+      "node_modules/windosu/*.cmd"
```
pdcastro: This is one of the fixes required to get `windosu` to work with standalone zip packages. Even with the changes in this PR, I found that `os intialize` still doesn't work with standalone zip package on Windows specifically. It works with a npm install on Windows, it works with standazone zip package on Linux / Mac, but it doesn't work with standalone zip package on Windows. Presently clueless as to why. I'm inclined to create a CLI issue instead of spending more time on this.

pdcastro:
@balena-ci retest

lib/actions/internal.coffee
CameronDiver: I think we could probably delete it, a major version change talks about the public facing interface, this was never designed to be public facing (it's also super weird that it existed in the first place...) I'd say remove.

lib/utils/helpers.ts
```diff
@@ -59,19 +60,37 @@ export function stateToString(state: OperationState) {
...
+	// contains a "/snapshot" path that should be ignored.
+	const fullCommand = (process as any).pkg
+		? [process.argv[0], ...command]
+		: [process.argv[0], process.argv[1], ...command];
 
 	if (os.platform() !== 'win32') {
 		console.log(
-			msg || 'If asked please type your computer password to continue',
+			msg ||
+				'Admin privileges required: you may be asked for your computer password to continue.',
```
CameronDiver:
```suggestion
				'Admin privileges required: you may be asked for your password to continue.',
```

package.json
CameronDiver: As long as we're moving forwards, I'd be ok with that.

patches/resin-cli-visuals+1.4.3.patch
CameronDiver: Yeah I'd definitely prefer we upstream this before merging of this PR.

lib/actions/internal.coffee
pdcastro: All right then... I've added a commit that deletes `internal sudo`.

patches/resin-cli-visuals+1.4.3.patch
pdcastro: Upstream merged, PR updated to use `resin-cli-visuals` v1.4.4.

package.json
pdcastro: I've managed to identify why `os initialize` was not working in the combination of standalone zip package and Windows, and fixed it. It was because `windosu` would spawn a child process using `process.execPath` as the path of the Node.js interpreter. However, in a standalone zip package, Node.js is built-in in the `balena.exe` binary and `process.execPath` is actually `balena.exe`. When a standalone zip package is being used, the OS may not have a separate Node.js installation at all. To solve this, I came up with a new internal command, `pkgExec`, that `require`s the module given as argument, and I patched `windosu` to use it when a standalone zip package is detected. This was done in a separate commit.

lib/utils/helpers.ts
pdcastro: Hmm "computer password" was already there before. I actually considered it and I think it's helpful, specially for beginners, because `sudo` is being executed behind the scenes (users may not know it) and one might think that "your password" is the balenaCloud password, or some other password. I also thought of "login password", but login to what service? "Computer password" is actually somewhat clearer, I think.

pdcastro:
@balena-ci retest

pdcastro:
I've tested the following combinations:

| | Install | root/admin | regular user |
| -- | -- | -- | -- |
| Debian 9 | pkg | ✔ | ✔ |
| Debian 9 | npm | ✔ | ✔ |
| Mac | pkg | ✔ | ✔ |
| Mac | native | ✔ | ✔ |
| Windows cmd.exe | pkg | ✔ | ✔ |
| Windows cmd.exe | native | ✔ | ✔ |

`balena os initialize` does not work when the Linux variant of balena CLI is used on a WSL (Windows Subsystem for Linux) Unbuntu shell. The error is `lsblk: failed to access sysfs directory: /sys/dev/block: No such file or directory`. But it appears to be a more fundamental issue of raw Linux device access on WSL.

Issue #1459, opened 2019-10-02
Should provide a way to delete config env vars


thgreasi:
- **balena CLI version:** v11.13.1

---

Currently `balena envs add` allows users to add config vars, by checking the prefix of envvar name to distinguish whether it's an env var or a config var. On the other hand the `balena envs rm` doesn't provide a way to delete a config var.

PR #1460, opened 2019-10-02
Allow deleting config vars with 'env rm'

thgreasi:
Resolves: #1459
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -607,13 +607,14 @@ produce verbose output
 
 ## env rm ID
 
-Remove an environment variable from an application or device, as selected
-by command-line options.
+Remove a configuration or environment variable from an application or device,
+as selected by command-line options.
 
 Note that this command asks for confirmation interactively.
 You can avoid this by passing the `--yes` boolean option.
 
 The --device option selects a device instead of an application.
+The --config option selects a config var instead of an env var.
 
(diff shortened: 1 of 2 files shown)


doc/cli.markdown
```diff
@@ -636,6 +638,10 @@ environment variable numeric database ID
 
 Selects a device environment variable instead of an application environment variable
 
+#### -c, --config
+
+Selects a configuration variable instead of an application/device environment variable
```
pdcastro: Suggestion:
* Here it could say simply `Selects a configuration variable instead of an environment variable`, which I think avoids implying that `application/device` is a distinction that applies to `environment` variables but not to `configuration` variables. (Alternatively but more verbose: `Selects an application/device configuration variable instead of an application/device environment variable`)
* A few paragraphs above, where the overall command description currently reads:
```
Remove an environment variable from an application or device, as selected
by command-line options.
...
The --device option selects a device instead of an application.
```
Perhaps it should instead read:
```
Remove a configuration or environment variable from an application or device,
as selected by command-line options.
...
The --device option selects a device instead of an application.
The --config option selects a config var instead of an env var.
```

thgreasi:
Thanks for the suggestions @pdcastro . I amended your suggestions and pushing in a minute.

Issue #1132, opened 2019-03-11
Commands that interact with Docker need explicit `--docker` flag on Windows 10


dfunckt:
For some reason, despite docker-modem itself [seemingly handling the case properly](https://github.com/apocas/docker-modem/blob/master/lib/modem.js#L27) users need to explicitly pass `--docker=//./pipe/docker_engine` on Windows 10. Otherwise the error is:

```
$ balena deploy testApp2
Docker seems to be unavailable. Is it installed and running?
```

adipascu:
Can reproduce on Ubuntu 19.04 with docker rootless.

The `DOCKER_HOST` env variable is populated but not used, have to run the command as `balena build --deviceType raspberrypi3 --arch armv7hf --emulated --docker=/run/user/1000/docker.sock` to work.

pdcastro:
Fixed in CLI v11.14.2 ([releases](https://github.com/balena-io/balena-cli/releases)). `unix://` and `npipe://` URLs in the DOCKER_HOST env var are now also supported.

PR #1464, opened 2019-10-07
Fix 'balena build --emulated' on Windows (including default docker socket)

pdcastro:
Resolves: #1132 
Depends-on: https://github.com/balena-io-modules/resin-multibuild/pull/65
Depends-on: https://github.com/balena-io-modules/docker-qemu-transpose/pull/23
Change-type: patch


lib/utils/docker-coffee.coffee
```diff
@@ -114,12 +113,12 @@ generateConnectOpts = (opts) ->
...
-			# If no explicit options are provided, use the env
-			connectOpts = buildDockerodeOpts(process.env.DOCKER_HOST)
 		else
-			# No options anywhere, assume default docker local socket
-			connectOpts.socketPath = '/var/run/docker.sock'
+			# Use docker-modem defaults which take the DOCKER_HOST env var into account
+			# https://github.com/apocas/docker-modem/blob/v2.0.2/lib/modem.js#L16-L65
+			Modem = require('docker-modem')
+			defaultOpts = new Modem()
+			connectOpts[opt] = defaultOpts[opt] for opt in ['host', 'port', 'socketPath']
```
pdcastro: These changes fix two related issues:
* #1132, where the default socketPath value on Windows should be `//./pipe/docker_engine` instead of `/var/run/docker.sock`.
* https://github.com/balena-io/balena-cli/issues/1132#issuecomment-532422385 When the DOCKER_HOST environment variable is set to a value such as `unix:///run/user/1000/docker.sock`, the previous implementation using `require('dockerode-options')` was unable to extract the correct path. I found that the `docker-modem` v2 package [supports `unix://` and `npipe://` URLs](https://github.com/apocas/docker-modem/blob/v2.0.2/lib/modem.js#L16-L65) in the DOCKER_HOST env var, so I've changed the implementation to use it instead.

pdcastro:
force push = updating `docker-qemu-transpose` to v1.0.1 because of the issue described in https://github.com/balena-io-modules/docker-qemu-transpose/pull/24

PR #1467, opened 2019-10-09
Selected chore commits and docs updates

pdcastro:
Change-type: patch


.resinci.yml
```diff
@@ -2,34 +2,32 @@
 npm:
   platforms:
     - name: linux
-      os: alpine
+      os: ubuntu
```
pdcastro: Replacing alpine with Ubuntu/Debian because the CLI does not even run on Alpine... Though the tests were clearly passing. The tests also take a long time to run, so I "distributed" the versions 8 / 10 / 12 among the different architectures instead of testing all 3 versions in every architecture.

CameronDiver: This is a little worrying, I'd much rather we ran every version on every environment, how long do the tests usually take?

INSTALL.md
```diff
@@ -13,8 +13,16 @@ There are 3 options to choose from to install balena's CLI:
 Some specific CLI commands have a few extra installation steps: see section [Additional
 Dependencies](#additional-dependencies).
 
-> **Windows users:** We now have a [YouTube video tutorial](https://www.youtube.com/watch?v=2LApclXFqsg)
-for installing and getting started with the balena CLI on Windows!
+> **Windows users:**
+> * We now have a [YouTube video tutorial](https://www.youtube.com/watch?v=2LApclXFqsg) for
```
CameronDiver:
```suggestion
> * We have a [YouTube video tutorial](https://www.youtube.com/watch?v=2LApclXFqsg) for
```

I worry about how the text will "age"

INSTALL.md
```diff
@@ -69,17 +77,31 @@ If you are a Node.js developer, you may wish to install the balena CLI via [npm]
...
+* Install the Windows Driver Kit (WDK) which is needed to compile some native Node modules:
+  * [WDK for Windows 10](https://docs.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk)
+  * [WDK for earlier versions of Windows](https://docs.microsoft.com/en-us/windows-hardware/drivers/other-wdk-downloads)
+* Install Node from the [Nodejs website](https://www.howtogeek.com/194041/how-to-open-the-command-prompt-as-administrator-in-windows-8.1/)
+* Install the `windows-build-tools` npm package (which provides Python 2.7 and more), running the following command in an [administrator
+  console](https://www.howtogeek.com/194041/how-to-open-the-command-prompt-as-administrator-in-windows-8.1/):  
   `npm install -g --production windows-build-tools`
 
-With those in place, the CLI installation command is:
+With those dependencies in place, the balena CLI installation command is:
```
CameronDiver:
```suggestion
With these dependencies in place, the balena CLI installation command is:
```

Not 100% sure on this

.resinci.yml
pdcastro: OK. Actually I think I was misunderstanding that `.resinci.yml` file, because it is configuring the `npm` job while the `repo.yml` file specifies `type: node-cli`. I am testing some changes and I have asked @nazrhom on Flowdock as well: https://www.flowdock.com/app/rulemotion/d-build-pipelines/threads/VkarI_cmcz1_gmQsyYyOaP15wzC

INSTALL.md
pdcastro: Indeed. What about "the"? _"With the dependencies in place, ..."_

INSTALL.md
pdcastro: Done

INSTALL.md
```diff
@@ -69,17 +77,31 @@ If you are a Node.js developer, you may wish to install the balena CLI via [npm]
...
+* The [MSYS2 shell](https://www.msys2.org/) may be used to provide `git`, `ssh`, `rsync`, `make`
+  and `g++`:
+  * `pacman -S git openssh rsync gcc make`
+  * [Set a Windows environment variable](https://www.onmsft.com/how-to/how-to-set-an-environment-variable-in-windows-10): `MSYS2_PATH_TYPE=inherit`
+  * Note that a bug in the MSYS2 launch script (`msys2_shell.cmd`) makes text-based
+    interactive CLI menus to misbehave. [Check this Github issue for a
+    workaround](https://github.com/msys2/MINGW-packages/issues/1633#issuecomment-240583890).
+* Install the Windows Driver Kit (WDK) which is needed to compile some native Node modules:
+  * [WDK for Windows 10](https://docs.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk)
+  * [WDK for earlier versions of Windows](https://docs.microsoft.com/en-us/windows-hardware/drivers/other-wdk-downloads)
```
pdcastro: The WDK installation step addresses an error reported in the forums, which I had also recently observed myself in a fresh Windows 10 installation:
https://forums.balena.io/t/balena-cli-install-err/5606
```
embedder : error : Could not open file 'C:\Program Files (x86)\Windows Kits\10\redist\wdf\x64\WdfCoInstaller01011.dll'. [C:\Users\confusetador\AppData\Roaming\npm\node_modules\balena-cli\node_modules\winusb-driver-generator\build\libwdi.vcxproj]
    EMBED  C:\Program Files (x86)\Windows Kits\10\redist\wdf\x64\WdfCoInstaller01011.dll
gyp ERR! build error
```

.resinci.yml
pdcastro: Following tests, it seems that the platforms specified in `.resinci.yml` for the `npm` job type aren't actually triggering any tests. The `repo.yml` file specifies `type: node-cli`, so it is the `node-cli` job (rather than the `npm` job) that matters. If I wanted to add Node 12 to the tests, I would have to submit a PR to `resin-concourse` and update the Windows and Mac on-premises computers to install Node 12. Therefore, for this PR, I've reverted the previous `.resinci.yml` file, and I will leave any changes to it to another future PR.

pdcastro:
* I've renamed this PR to "Selected chore commits and docs updates", and added another 3 small commits. The commits are relatively small and independent from each other.
* The PR no longer adds Node 12 tests (I've reverted `.resinci.yml` as per comment), but one of the commits gets tests to pass on Node 12.
* The documentation commit has not changed.

INSTALL.md
CameronDiver: "These" or "the" I think are both fine, probably would prefer "these" but also not too bothered.

pdcastro: Thanks - replaced with "these".

Issue #1064, opened 2019-01-03
Should support managing tags using release commit hashes


thgreasi:
As discussed in Prod call 2019-01-03
Before adding support for release uuids to our API & SDK tagging methods, we should extend the tagging action to also accept release commit hashes 
`balena tag <tagKey> [tagValue] --release <releaseIdOrCommit>`

Since commit hashes are not unique
* neither across a single app (we can have a successful & a failed release with the same commit hash on the same app)
* not across different apps (we can have successful releases with the same commit hash across different apps)

we should check whether the parameter is a string or a number, and if it's indeed a string, then fetch all releases filtering on the commit hash and
* complete the operation if there is a single release is returned
* error with an ambiguous error balena error (`Release is ambiguous: ${commit}`) if more than one releases are returned, similarly with what the CLI currently does for ambiguous errors on actions accepting the application name.
  * ideally have a specific error for this added to balena-errors
    See: https://github.com/balena-io-modules/balena-errors/blob/master/lib/errors.ts#L214-L230
  * alternatively offer a list with all the matching releases and let the user choose one of them, similarly to how we offer a list of applications
    See: https://github.com/balena-io/balena-cli/blob/01eb4b473de40dec0aa2ee77d599239c222c5b1a/lib/utils/promote.ts#L176-L188

See: https://www.flowdock.com/app/rulemotion/r-product/threads/fIQp3RaIKRo8Cpn4__Ofmr6az2w

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_glgz)

thgreasi:
Depends-on: https://github.com/balena-io/balena-sdk/issues/464

PR #1470, opened 2019-10-15
Support managing tags using release commit hashes

thgreasi:
The sdk version in the shrinkwrap already supports setting tags by commit hashes and as a result this already works in the cli as of v11.9.6. This PR just adds some docs and some extra handling when the commit param prefix is all numeric.

Resolves: #1064
Change-type: minor

---
##### Contributor checklist

### DIFF
--- a/lib/actions/tags.ts
+++ b/lib/actions/tags.ts
@@ -17,7 +17,10 @@ limitations under the License.
 import { ApplicationTag, DeviceTag, ReleaseTag } from 'balena-sdk';
 import { CommandDefinition } from 'capitano';
 import { stripIndent } from 'common-tags';
-import { normalizeUuidProp } from '../utils/normalization';
+import {
+	disambiguateReleaseParam,
+	normalizeUuidProp,
+} from '../utils/normalization';
 import * as commandOptions from './command-options';
 
 export const list: CommandDefinition<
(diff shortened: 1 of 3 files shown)


lib/utils/normalization.ts
```diff
@@ -25,3 +26,26 @@ export function normalizeUuidProp(
 			params[propName + '_raw'] || _.toString(params[propName]);
 	}
 }
+
+export const disambiguateReleaseParam = async (
```
pdcastro: `export async function disambiguateReleaseParam(`

lib/actions/tags.ts
```diff
@@ -267,10 +290,13 @@ export const remove: CommandDefinition<
 				return balena.models.device.tags.remove(options.device, params.tagKey);
 			}
 			if (options.release) {
-				return balena.models.release.tags.remove(
+				const releaseParam = await disambiguateReleaseParam(
+					balena,
 					options.release,
-					params.tagKey,
+					options.release_raw!,
```
pdcastro: Instead of the exclamation mark here, it may be possible to use a question mark (`paramRaw?`) in the declaration of `disambiguateReleaseParam` -
```js
export async function disambiguateReleaseParam(
	balena: BalenaSDK,
	param: string | number,
	paramRaw?: string,
) {
```
Or not - it was just a thought.

thgreasi:
To clarify, I saw that on support yesterday and just took the extra time to do it properly, since I couldn't sleep before PRing this :)

lib/actions/tags.ts
thgreasi: I was thinking about this as well, but I think we should better go for `paramRaw: string | undefined` instead of `?` since the arg isn't meant to be optional.

Issue #1332, opened 2019-06-30
Error: Cannot find module 'web-streams-polyfill'


pdcastro:
* CLI v11.4.1 on Windows, executable (native) installer only, prints "null" as reported in #1331 and #1333, and the underlying cause appears to be a missing npm module called `web-streams-polyfill`.
```
C:\projects\resin-cli> balena login
internal/modules/cjs/loader.js:584
    throw err;
    ^
Error: Cannot find module 'web-streams-polyfill'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:582:15)
    at Function.resolveFilenameOptimized [as _resolveFilename] (C:\Program Files\balena-cli\client\node_modules\fast-boot2\index.js:57:33)
```
* It appears that the `npm-shrinkwrap.json` file generated from scratch in CLI release v11.4.1 by the `npm` tool version 6.4.1 (which ships with, for example, node 8.16.0) is "unstable", meaning that subsequently running:
```
rm -rf node_modules
npm install
```
... results in modifications being made to the `npm-shrinkwrap.json` file with several dependencies incorrectly deleted, including the `web-streams-polyfill` module. And then, running `npm list` prints several "invalid" and "missing" errors:
```
$ npm list > /dev/null
npm ERR! invalid: node-web-streams@0.2.2 /Users/paulo/balena/balena-io/balena-cli/node_modules/balena-request/node_modules/node-web-streams
npm ERR! missing: web-streams-polyfill@^1.3.2, required by node-web-streams@0.2.2
npm ERR! invalid: resin-discoverable-services@1.2.2 /Users/paulo/balena/balena-io/balena-cli/node_modules/balena-sync/node_modules/resin-discoverable-services
npm ERR! missing: bonjour@git+https://github.com/resin-io/bonjour.git#fixed-mdns, required by resin-discoverable-services@1.2.2
npm ERR! missing: ip@^1.1.4, required by resin-discoverable-services@1.2.2
npm ERR! invalid: usb@1.3.10 /Users/paulo/balena/balena-io/balena-cli/node_modules/node-raspberrypi-usbboot/node_modules/usb
```
* It appears to be a bug in `npm` (older versions only?), and a fix seems to be to recreate the `npm-shrinkwrap.json` file, again from scratch, using the `npm` tool version 6.9.0 (which ships with, for example, node 10.16.0):
```
$ npm --version
6.9.0
$ rm -rf npm-shrinkwrap.json node_modules
[...]

PR #1475, opened 2019-10-17
Add npm preinstall check for npm version 6.9.0 or later

pdcastro:
Older npm versions cause the npm-shrinkwrap.json file to be incorrectly
updated. This should avoid regression bugs related to issue #1332.

Change-type: patch


automation/check-npm-version.js
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env node
+'use strict';
+
+/** Check that semver v1 is greater than or equal to semver v2 */
+function semverGte(v1, v2) {
```
CameronDiver: Can we get a comment here detailing why we have our own version here (to avoid an `npm install`) so it's not mistakenly removed by anybody in the future please?

pdcastro: Yes, reinventing the wheel... The problem with `require('semver')` is that we wanted to run this script in the npm 'preinstall' hook, at which point no dependencies have been installed. As is, this script depends on Node.js only.

package.json
```diff
@@ -54,7 +55,7 @@
     "test:fast": "npm run build:fast && npm run test",
     "ci": "npm run test && catch-uncommitted",
     "watch": "gulp watch",
-    "prettify": "prettier --write \"{lib,tests,automation,typings}/**/*.ts\" --config ./node_modules/resin-lint/config/.prettierrc",
+    "prettify": "prettier --write \"{lib,tests,automation,typings}/**/*.[tj]s\" --config ./node_modules/resin-lint/config/.prettierrc",
```
CameronDiver: These expansions can fail when ran on certain shells, I've been bitten by this myself

pdcastro: In this case, it is `prettier` itself that interprets the glob, not a shell: https://prettier.io/docs/en/cli.html

By the way, I've also tested it on Windows, with `cmd.exe`, which is fairly different from sh or bash...

automation/check-npm-version.js
pdcastro: Done!

patches/@oclif+dev-cli+1.22.0.patch
```diff
@@ -75,11 +75,17 @@ index 3e613e0..129d041 100644
 +        const sources = [
 +            'bin', 'build', 'patches', 'typings', 'CHANGELOG.md', 'INSTALL.md',
 +            'LICENSE', 'package.json', 'npm-shrinkwrap.json', 'README.md',
-+            'TROUBLESHOOTING.md',
++            'TROUBLESHOOTING.md', 'automation/check-npm-version.js',
```
pdcastro: I've had to update this `oclif` patch because `oclif` copies the CLI source to a tmp folder to generate the native installers and runs `npm install` on that folder (as part of `npm run build:installer`), and now we have the `preinstall` hook that calls `check-npm-version.js`.

PR #1477, opened 2019-10-17
Move npm version check from npm 'preinstall' to git 'pre-commit' hook

pdcastro:
This should allow end users to use npm v6.4.1 that ships with Node 8, while still "probabilistically" requiring CLI developers to use npm v6.9.0 or later.

Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
   },
   "preferGlobal": true,
   "files": [
-    "automation/check-npm-version.js",
     "bin/",
     "build/",
     "doc/",
@@ -40,7 +39,6 @@
     ]
   },
   "scripts": {
-    "preinstall": "node automation/check-npm-version.js",
     "postinstall": "patch-package",
     "prebuild": "rimraf build/ build-bin/",
     "build": "npm run build:src",
@@ -73,6 +71,11 @@
   "license": "Apache-2.0",
   "engines": {
     "node": ">=8.0"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "node automation/check-npm-version.js"
+    }
   },
   "oclif": {
     "bin": "balena",
@@ -130,6 +133,7 @@
     "gulp-coffee": "^2.2.0",
     "gulp-inline-source": "^2.1.0",
     "gulp-shell": "^0.5.2",
+    "husky": "^3.0.9",
     "intercept-stdout": "^0.1.2",
     "mocha": "^6.2.0",
     "nock": "^10.0.6",
--- a/automation/check-npm-version.js
+++ b/automation/check-npm-version.js
@@ -52,9 +52,22 @@ function checkNpmVersion() {
 		.trim();
 	const requiredVersion = '6.9.0';
 	if (!semverGte(npmVersion, requiredVersion)) {
+		// In case you take issue with the error message below:
+		//   "At this point, however, your 'npm-shrinkwrap.json' file has
+		//    already been damaged"
+		// ... and think: "why not add the check to the 'preinstall' hook?",
+		// the reason is that it would unnecessarily prevent end users from
+		// using npm v6.4.1 that ships with Node 8.  (It is OK for the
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -40,15 +40,15 @@
     ]
   },
   "scripts": {
-    "preinstall": "node automation/check-npm-version.js",
     "postinstall": "patch-package",
-    "prebuild": "rimraf build/ build-bin/",
+    "prebuild": "npm run npmcheck && rimraf build/ build-bin/",
```
pdcastro: I am of two minds about this PR and I don't know if it should be merged:
* In favour: it allows CLI end users (as opposed to CLI developers) to install the CLI with npm v6.4.1 that ships with Node 8. It will damage their `npm-shrinkwrap.json` file but it's OK, because the damaged file won't be reused.
* Against: it allows a CLI contributor to submit a PR with a broken `npm-shrinkwrap.json` file that could end up being merged undetected, then breaking the CLI for all users. This is possible because merely running `npm install` damages the file, and the CLI contributor does not strictly have to run `npm run build` or `npm test` in order to submit a PR. **For example, a PR that merely updates the markdown documentation files (README, INSTALL...).** The CI will run `npm install` and `npm test` with a good npm version, but this neither detects the issue nor undamages the `npm-shrinkwrap.json` file.

thgreasi:
@pdcastro on the other hand we might be able to just use a pre-commit hook for the npm-shriwnkwrap.json and run this check in there.
We do have a similar hook in the UI for deduplicating our yarn.lock. What do you think?
See: https://github.com/balena-io/resin-ui/blob/aa6e6a8c56d85281766b4f4715d30b7efedd154b/.lintstagedrc#L12

package.json
```diff
@@ -74,6 +73,11 @@
   "engines": {
     "node": ">=8.0"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run npmcheck"
+    }
```
pdcastro: > on the other hand we might be able to just use a pre-commit hook for the npm-shriwnkwrap.json and run this check in there. We do have a similar hook in the UI for deduplicating our yarn.lock. What do you think?
See: https://github.com/balena-io/resin-ui/blob/aa6e6a8c56d85281766b4f4715d30b7efedd154b/.lintstagedrc#L12

@thgreasi, I think it's a good idea. I understand that `resin-ui` and `resin-api` use `husky` for this. I've added husky above (using `package.json` rather than `.huskyrc` for its configuration). I think the check could be added in addition to `prebuild`, because git hooks are also not guaranteed to run on the client side. Belts and braces..

pdcastro: Ha. There's a complication with `prebuild` as well. The balena CI build was failing on Windows because it uses Node 8 and runs `npx npm@6.9.0 install` instead of `npm install` -- it was the CI's workaround to avoid the npm bug. This means that a `preinstall` check for v6.9.0 passes, but a `prebuild` check fails as the default npm is v6.4.1. It's a to-do item to update balena CI on Windows to use Node 10 and/or Node 12...
So we are left with fewer choices for this PR, and the husky pre-commit hook may be the best one, at least for now.

pdcastro:
@balena-ci retest

package.json
pdcastro: I've decided to merge this PR sooner than later, to avoid users' CIs out there breaking unnecessarily with Node 8. We can still revisit this decision and readopt preinstall. I am also considering that until yesterday, we had no checks at all, and having a husky pre-commit check is an improvement over that.

I've tested all installation methods in the draft release for v11.16.4.

package.json
thgreasi: @pdcastro I generally prefer `.huskyrc` since it keeps the package.json cleaner and we can easily omit unwanted files from the published ones (eg in the SDK). Plus sounds nearer to the single responsibility principle. That's of course just imo and just a NIT, but it's great anyway that you added that hook as part of this PR :+1:

Issue #1491, opened 2019-11-01
Warnings (masked by Sentry) when calling CLI without arguments.


srlowe:
- **balena CLI version:** 11.17.1
- **Operating system version:** Ubuntu 19.10
- **32/64 bit OS and processor:** e.g. 64
- **Install method:** npm
- **If npm install, Node.js and npm version:** e.g. Node v12.13.0 and npm v6.12.0

---

When calling the CLI without any arguments it should behave as when called with `-h` or `--help`.  However the no-arguments route triggers a call to sentry, which turns out to be masking the warnings `UnhandledPromiseRejectionWarning` and `DeprecationWarning` (details below).

This appears to be caused by a call to `capitano.execute` which is not promisified.

We should probably review how sentry is involved in non-production environments, as this would not currently be caught be tests.

```
(node:17763) UnhandledPromiseRejectionWarning: TypeError: done is not a function
    at /home/scott/Dev/balena-cli/lib/actions/help.coffee:96:14
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:17763) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:17763) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

PR #1493, opened 2019-11-01
Fix warnings when executing CLI without arguments.

srlowe:
An edge case call to `capitano.execute` was not promisified, which was generating warnings in Node 12.

Note that the test will not catch the problem until we stop sentry masking errors/warnings.  Has been tested with sentry disabled though.

Change-type: patch
Resolves: #1491

---
##### Contributor checklist

### DIFF
--- a/lib/app-capitano.coffee
+++ b/lib/app-capitano.coffee
@@ -25,8 +25,8 @@ capitano.permission 'user', (done) ->
 
 capitano.command
 	signature: '*'
-	action: ->
-		capitano.execute(command: 'help')
+	action: (params, options, done) ->
+		capitano.execute(command: 'help', done)
 
 capitano.globalOption
 	signature: 'help'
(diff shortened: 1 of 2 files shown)


lib/app-capitano.coffee
```diff
@@ -26,7 +26,8 @@ capitano.permission 'user', (done) ->
 capitano.command
 	signature: '*'
 	action: ->
-		capitano.execute(command: 'help')
+		capitanoExecuteAsync = Promise.promisify(capitano.execute)
+		capitanoExecuteAsync(command: 'help')
```
pdcastro: Nice 👍   What might be, I think, more "idiomatic" of Capitano (i.e. matching the [documentation](https://www.npmjs.com/package/capitano#capitanoexecutecli-callback) and how most CLI commands are currently implemented, e.g. [help.coffee](https://github.com/balena-io/balena-cli/blob/v11.17.1/lib/actions/help.coffee#L146-L150)) is to use a callback:
```js
		capitano.execute(command: 'help', done)
```
I've tested it with Node 12, it seems to fix the issue just as well.

srlowe: Sure, will modify +1

Issue #1485, opened 2019-10-22
CLI still lists discontinued device types


alisondavis17:
CLI Version: 11.16.6

with command `$ balena app create` the list of device types presented still lists discontinued device types.

alisondavis17:
And also with command `$ balena devices supported`

PR #1501, opened 2019-11-07
Hide discontinued device types in `balena devices supported`, `balena app create`.

srlowe:
Hide discontinued device types in `balena devices supported`, `balena app create`.
Additionally:
 - Add tests for actions.
 - Convert devices supported action to TypeScript.

Resolves: #1485
Change-type: patch

---
##### Contributor checklist

### DIFF
--- a/lib/actions/device_ts.ts
+++ b/lib/actions/device_ts.ts
@@ -19,6 +19,31 @@ import { stripIndent } from 'common-tags';
 import { normalizeUuidProp } from '../utils/normalization';
 import * as commandOptions from './command-options';
 
+export const supported: CommandDefinition<{}, {}> = {
+	signature: 'devices supported',
+	description: 'list all supported devices',
+	help: stripIndent`
+		Use this command to get the list of all supported devices.
+
+		Examples:
+
+			$ balena devices supported
+	`,
+	async action(_params, _options) {
+		const sdk = (await import('balena-sdk')).fromSharedOptions();
+		const visuals = await import('resin-cli-visuals');
+		const _ = await import('lodash');
+
+		let deviceTypes = await sdk.models.config.getDeviceTypes();
+		const fields = ['slug', 'name'];
+		deviceTypes = _.sortBy(deviceTypes, fields).filter(
+			dt => dt.state !== 'DISCONTINUED',
+		);
+		const output = await visuals.table.horizontal(deviceTypes, fields);
+		console.log(output);
+	},
+};
+
 // tslint:disable-next-line:no-namespace
 namespace OsUpdate {
 	export interface Args {
--- a/tests/commands/device/supported.spec.ts
+++ b/tests/commands/device/supported.spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
 import * as _ from 'lodash';
-import { runCommand } from '../../helpers';
+import { cleanOutput, runCommand } from '../../helpers';
 
 const HELP = `
 Usage: devices supported
@@ -11,18 +11,6 @@ Examples:
 
 \t$ balena devices supported
 `;
-
-const cleanOutput = (output: string[] | string) => {
-	return _(_.castArray(output))
-		.map(log => {
-			return log.split('\n').map(line => {
-				return line.trim();
-			});
-		})
-		.flatten()
-		.compact()
-		.value();
-};
 
 describe('balena devices supported', function() {
 	it('should list currently supported devices', async () => {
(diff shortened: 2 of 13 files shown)


srlowe:
I had planned to include more in depth tests for `app create`, but have shelved these for now due to ongoing discussion about testing approach, in order for this issue to be closed.

lib/actions/device_ts.ts
```diff
@@ -19,6 +19,35 @@ import { stripIndent } from 'common-tags';
...
+			.getDeviceTypes()
+			.then(deviceTypes => {
+				const fields = ['slug', 'name'];
+				deviceTypes = _.sortBy(deviceTypes, fields).filter(
+					dt => dt.state !== 'DISCONTINUED',
+				);
+				return visuals.table.horizontal(deviceTypes, fields) as string;
+			})
+			.then(console.log)
+			.nodeify(done);
```
pdcastro: Something I noticed, which I think was already the case in the CoffeeScript implementation: the code is both returning a promise **and** explicitly calling the `done()` callback, so maybe `done()` will be called twice, once here explicitly and once by the Capitano framework, when it detects that a promise was returned. Perhaps completely harmless... or perhaps there's some side effect (?). You could leave it as it is, or you could choose to call `done()` here and return `undefined` (or any non-promise value) (in which case, the action could not be declared `async`, because an `async` function always returns a promise), or you could return a promise and not call `done()` here (i.e. remove the `.nodeify(done)` line and rename the action argument to `_done`).

Also: if you chose to declare the action `async` (thus necessarily returning a promise), then I thought of mentioning a "current team preference" (not just CLI, several other modules / projects too) of using async/await instead of promise chains in the implementation of an `async` function. For example, it might have been possible to write the code above (I haven't tested...) with something like:
```
async action(_params, _options, _done) {
    ...
    let deviceTypes = await sdk.models.config.getDeviceTypes();
    const fields = ['slug', 'name'];
    deviceTypes = _.sortBy(deviceTypes, fields).filter(dt => dt.state !== 'DISCONTINUED');
    const horizontalTable = await visuals.table.horizontal(deviceTypes, fields);
    console.log(horizontalTable);
```

tests/commands/device/supported.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+Usage: devices supported
+
+Use this command to get the list of all supported devices.
+
+Examples:
+
+\t$ balena devices supported
+`;
+
+const cleanOutput = (output: string[] | string) => {
```
pdcastro: Is this a copy-and-paste of the same function in `tests/helpers.ts`? Maybe it can be imported instead?
Btw, I think it's just a personal preference of mine, but I'd declare the function with the `function` keyword instead of `const`:
```
function cleanOutput(output: string[] | string) {
```

lib/actions/device_ts.ts
pdcastro: What I wrote above about `done()` being called twice is probably wrong (or conditionally correct...). When a promise is returned by the action, it looks like Capitano will only "automatically" call `done()` if `done` is not declared as an argument of the action function:

https://github.com/balena-io/capitano/blob/v1.9.0/lib/command.coffee#L84
`(_value) => callback() if @action.length < 3`

I.e., it should work if the action is declared with two arguments:
`async action(_params, _options)`
though I'm not sure if the TypeScript type checking would be happy with that.

tests/commands/device/supported.spec.ts
srlowe: Ah, actually looks like I'd added the function to `tests/helpers.ts` myself in this PR, but then forgot to import it here, doh :)  Will update.

Regarding the style, yeah I share your preference!  I think I had originally just copied this from `tests/commands/help.spec.ts` and so had not modified the style.

I can also change the other tests (e.g. help.spec.ts) to import this function while I'm here.

tests/helpers.ts
```diff
@@ -78,3 +79,15 @@ export const balenaAPIMock = () => {
 			configVarSchema: [],
 		});
 };
+
+export const cleanOutput = (output: string[] | string) => {
```
srlowe: @pdcastro BTW, what's your preferred style for function declaration when it's directly exported, such as here?

lib/actions/device_ts.ts
srlowe: Ah yes, thanks.   I had straight translated from the coffeescript, and hadn't really analysed the logic well enough. I'll look into that (It seems this `.nodify(done)` pattern is used extensively in the coffeescript actions).

Thanks for the info on team preference re promises & async/await - very valuable!  I usually assume that the existing code is doing it the preferred was, and so follow that.  Please keep these kind of comments coming if you notice anything :)

Right, while we are translating we might as well change to the preferred way of doing things too, I'll change this to an `async` function.

tests/helpers.ts
pdcastro: I would use the `function` keyword here as well:
```
export function cleanOutput(output: string[] | string) {
```

Issue #1504, opened 2019-11-07
`balena os configure --config` is not injecting the given config.json file


pdcastro:
- **balena CLI version:** e.g. 11.17.3 (but also reproducible with much older releases)
- **Operating system version:** macOS 10.14.6 (but it does not appear to be OS related)
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** zip
---

```
balena os configure balena.img -a test-rpi --config my-config.json
```

**Expected behaviour:** the given `my-config.json` file is inserted in the given `balena.img` image file.
**Actual behaviour:** the given `my-config.json` file appears to be ignored

Of interest: if the `--config.json` option is not used, the CLI asks questions interactively such as WiFi SSID and password (which appear to be correctly inserted in the `/system-connections/resin-wifi` file inside `balena.img`), and a working `config.json` file is also generated and inserted.

pdcastro:
Related to: #1395 ['balena config' should add network settings in the right place for balenaOS 2.x](https://github.com/balena-io/balena-cli/issues/1395)

PR #1507, opened 2019-11-09
Fix 'os configure --config', and migrate it to oclif + TypeScript

pdcastro:
The first commit introduces the `ExpectedError` type: see comments I added to the PR files. The second commit fixes the `--config` issue (#1504) and migrates `os configure` to oclif + Typescript. I suggest reviewing the commits separately.

Resolves: #1504
Change-type: minor

### DIFF
--- a/lib/actions-oclif/os/configure.ts
+++ b/lib/actions-oclif/os/configure.ts
@@ -1,0 +1,431 @@
+/**
+ * @license
+ * Copyright 2019 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Command, flags } from '@oclif/command';
+import BalenaSdk = require('balena-sdk');
+import { stripIndent } from 'common-tags';
+import * as _ from 'lodash';
+
+import { ExpectedError } from '../../errors';
+import * as cf from '../../utils/common-flags';
+import { CommandHelp } from '../../utils/oclif-utils';
+
+interface FlagsDef {
+	advanced?: boolean;
+	app?: string;
+	application?: string;
+	config?: string;
+	'config-app-update-poll-interval'?: number;
+	'config-network'?: string;
+	'config-wifi-key'?: string;
+	'config-wifi-ssid'?: string;
+	device?: string; // device UUID
+	'device-api-key'?: string;
+	'device-type'?: string;
+	help?: void;
+	version?: string;
+}
+
+interface ArgsDef {
+	image: string;
+}
+
+interface DeferredDevice extends BalenaSdk.Device {
+	belongs_to__application: BalenaSdk.PineDeferred;
+}
+
+interface Answers {
+	appUpdatePollInterval: number; // in minutes
+	deviceType: string; // e.g. "raspberrypi3"
+	network: 'ethernet' | 'wifi';
+	version: string; // e.g. "2.32.0+rev1"
+	wifiSsid?: string;
+	wifiKey?: string;
+}
+
+const deviceApiKeyDeprecationMsg = stripIndent`
+	The --device-api-key option is deprecated and will be removed in a future release.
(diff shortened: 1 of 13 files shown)


lib/errors.ts
```diff
@@ -20,7 +20,9 @@ import * as _ from 'lodash';
 import * as os from 'os';
 import * as Raven from 'raven';
 
-import * as patterns from './utils/patterns';
+export class ExpectedError extends Error {}
```
pdcastro: This commit (changes to `errors.ts` and `patterns.ts`) introduces the `ExpectedError` type. An "expected" error is an error we explicitly test for (e.g. some data validation), and which does not require Sentry.io reporting. We currently have a "situation" in the CLI where:
* Several commands are arguably incorrectly handling "expected" errors with `throw Error("some reason")`.
* Most commands follows the arguably correct pattern of calling the `exitWithExpectedError()` function (which calls `process.exit(1)` before anything is reported to Sentry.io).

The former is deemed incorrect, and the latter correct, simply because the top-level `handleError()` function in `errors.ts` would report any error that reached it to Sentry.io. However, [ordinarily calling process.exit() is not a recommended practice](https://stackoverflow.com/a/37592669) as it may prematurely abort async operations (like perhaps updating config files - theoretically it could cause data corruption). When possible, it is preferable to set `process.exitCode` and allow the process to exit gracefully. This commit modifies the `handleError()` function so that it does not report instances of `ExpectedError` to Sentry.io, and set `process.exitCode` instead of calling `process.exit()` for expected errors. This allows `throw ExpectedError("some reason")` as an alternative correct way of handling errors in CLI commands.

lib/utils/patterns.ts
```diff
@@ -77,16 +78,28 @@ export function authenticate(options: {}): Bluebird<void> {
 		});
 }
 
-export async function exitIfNotLoggedIn(): Promise<void> {
+export async function checkLoggedIn(): Promise<void> {
 	const balena = getBalenaSdk();
 	if (!(await balena.auth.isLoggedIn())) {
-		exitWithExpectedError(stripIndent`
+		throw new NotLoggedInError(stripIndent`
```
pdcastro: `NotLoggedInError` is an instance of `ExpectedError`, so it uses the new error handling capability that avoids reporting the error to Sentry.io while also avoiding the premature call to `process.exit()`. I am calling `checkLoggedIn()` (instead of `exitIfNotLoggedIn()`) in another PR I'm working on.

lib/actions-oclif/os/configure.ts
```diff
@@ -0,0 +1,430 @@
...
+			Unsupported platform error: the 'balena os configure' command currently requires
+			the Windows Subsystem for Linux in order to run on Windows. It was tested with
+			the Ubuntu 18.04 distribution from the Microsoft Store. With WSL, a balena CLI
+			release for Linux (rather than Windows) should be installed: for example, the
+			standalone zip package for Linux. (It is possible to have both a Windows CLI
+			release and a Linux CLI release installed simulatenously). For more information
+			on WSL and the balena CLI installation options, please check:
+			- https://docs.microsoft.com/en-us/windows/wsl/about
+			- https://github.com/balena-io/balena-cli/blob/master/INSTALL.md
+		`);
```
pdcastro: I've created a separate CLI issue for this:
#1509 ["balena os configure" hangs on Windows (not WSL)](https://github.com/balena-io/balena-cli/issues/1509)

lib/actions-oclif/os/configure.ts
```diff
@@ -0,0 +1,430 @@
...
+		);
+	}
+	if (!options.application && options['device-type']) {
+		throw new ExpectedError(
+			"The '--device-type' option can only be used in conjunction with the '--application' option",
+		);
+	}
+	if (options['device-api-key']) {
+		console.error(stripIndent`
+			-------------------------------------------------------------------------------------------
```
hedss: Intended NL?

lib/errors.ts
```diff
@@ -105,21 +103,35 @@ const messages: {
 };
 
 export async function handleError(error: any) {
-	let message = interpret(error);
-	if (message == null) {
+	const { printErrorMessage } = await import('./utils/patterns');
+
+	process.exitCode =
```
hedss: Nice!

lib/actions-oclif/os/configure.ts
pdcastro: Not sure I understand the question. At run time, what gets printed by the CLI is:
```
$ ./bin/balena-dev os configure --device-api-key foo -a test-rpi os.img
-------------------------------------------------------------------------------------------
Warning: The --device-api-key option is deprecated and will be removed in a future release.
A suitable key is automatically generated or fetched if this option is omitted.
-------------------------------------------------------------------------------------------
? Network Connection (Use arrow keys)
❯ ethernet
  wifi
...
```

PR #1529, opened 2019-12-07
Add multicontainer (microservices) support for 'balena env(s)'

pdcastro:
Introduces multicontainer / microservices support for the balena envs / env add / env rm / env rename commands. The `--service` option is added to all these commands, and the `--all` and `--json` options are added to the `balena envs` command. 

```
$ ./bin/balena help envs
List the environment or config variables of an application, device or service.

USAGE
  $ balena envs

OPTIONS
  -a, --application=application  application name
  -c, --config                   show configuration variables only
  -d, --device=device            device UUID
  -h, --help                     show CLI help
  -j, --json                     produce JSON output instead of tabular output
  -s, --service=service          service name
  -v, --verbose                  produce verbose output

  --all                          include app-wide, device-wide variables that apply to the selected device or service.
                                 Variables are still filtered out by type with the --config option.

DESCRIPTION
  List the environment or configuration variables of an application, device or
  service, as selected by the respective command-line options. (A service is
  an application container in a "microservices" application.)

  The --config option is used to list "configuration variables" that control
  balena platform features, as opposed to custom environment variables defined
  by the user. The --config and the --service options are mutually exclusive
  because configuration variables cannot be set for specific services.

  The --all option is used to include application-wide (fleet), device-wide
  (multiple services on a device) and service-specific variables that apply to
  the selected application, device or service. It can be thought of as including
  "inherited" variables: for example, a service inherits device-wide variables,
  and a device inherits application-wide variables. Variables are still filtered
  out by type with the --config option, such that configuration and non-
  configuration variables are never listed together.

  When the --all option is used, the printed output may include DEVICE and/or
  SERVICE columns to distinguish between application-wide, device-specific and
  service-specific variables. As asterisk in these columns indicates that the
  variable applies to "all devices" or "all services".

  If you are parsing the output in a script, please select the JSON format with
  the '-j' option. This avoids future compatibility issues if columns are added,
  renamed or reordered. Also, when the JSON format is selected, an empty JSON
  array ([]) is printed instead of an error message when no variables exist for
  the given query. When querying variables for a device, note that the application
  name may be null in JSON output (or 'N/A' in tabular output) if the application
  linked to the device is no longer accessible by the current user (for example,
  in case the current user has been removed from the application by its owner).

EXAMPLES
  $ balena envs --application MyApp
  $ balena envs --application MyApp --all --json
  $ balena envs --application MyApp --service MyService
  $ balena envs --application MyApp --all --service MyService
  $ balena envs --application MyApp --config
  $ balena envs --device 7cf02a6
  $ balena envs --device 7cf02a6 --all --json
  $ balena envs --device 7cf02a6 --config --all --json
  $ balena envs --device 7cf02a6 --all --service MyService
```

The code was substantially rewritten, and a number of test cases were added. Backwards compatibility is preserved in the tabular output format for the `envs` command, when the new command-line options are ***not*** used.

Connects-to: #1153 
Change-type: minor

### DIFF
--- a/lib/actions-oclif/envs.ts
+++ b/lib/actions-oclif/envs.ts
@@ -19,6 +19,7 @@ import { ApplicationVariable, DeviceVariable } from 'balena-sdk';
 import { stripIndent } from 'common-tags';
 import * as _ from 'lodash';
 
+import { ExpectedError } from '../errors';
 import * as cf from '../utils/common-flags';
 import { CommandHelp } from '../utils/oclif-utils';
 
@@ -68,10 +69,12 @@ export default class EnvsCmd extends Command {
 		const { flags: options } = this.parse<FlagsDef, {}>(EnvsCmd);
 		const balena = (await import('balena-sdk')).fromSharedOptions();
 		const visuals = await import('resin-cli-visuals');
-		const { exitWithExpectedError } = await import('../utils/patterns');
+		const { checkLoggedIn } = await import('../utils/patterns');
 		const cmd = this;
-
 		let environmentVariables: ApplicationVariable[] | DeviceVariable[];
+
+		await checkLoggedIn();
+
 		if (options.application) {
 			environmentVariables = await balena.models.application[
 				options.config ? 'configVar' : 'envVar'
@@ -81,11 +84,11 @@ export default class EnvsCmd extends Command {
 				options.config ? 'configVar' : 'envVar'
 			].getAllByDevice(options.device);
 		} else {
-			return exitWithExpectedError('You must specify an application or device');
+			throw new ExpectedError('You must specify an application or device');
 		}
 
 		if (_.isEmpty(environmentVariables)) {
-			return exitWithExpectedError('No environment variables found');
+			throw new ExpectedError('No environment variables found');
 		}
 
 		cmd.log(
--- a/lib/actions-oclif/envs.ts
+++ b/lib/actions-oclif/envs.ts
@@ -15,11 +15,7 @@
  * limitations under the License.
  */
 import { Command, flags } from '@oclif/command';
-import {
-	ApplicationVariable,
-	DeviceVariable,
-	EnvironmentVariableBase,
-} from 'balena-sdk';
+import * as SDK from 'balena-sdk';
 import { stripIndent } from 'common-tags';
 import * as _ from 'lodash';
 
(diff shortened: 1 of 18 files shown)


tests/balena-api-mock.ts
```diff
@@ -65,19 +177,20 @@ class BalenaAPIMock {
...
-	public expectOptionalWhoAmI() {
-		this.scope
-			.get('/user/v1/whoami')
-			.optionally()
-			.reply(200, {
-				id: 99999,
-				username: 'testuser',
-				email: 'testuser@test.com',
-			});
+	public expectWhoAmI(persist = false, optional = true) {
```
srlowe: Right, this is nicer! Perhaps in future we can add these options generically for all of these expectations in some way.

tests/helpers.ts
```diff
@@ -21,9 +21,10 @@ import * as nock from 'nock';
 import * as path from 'path';
 
 import * as balenaCLI from '../build/app';
-import { configureBluebird } from '../build/app-common';
+import { configureBluebird, setMaxListeners } from '../build/app-common';
 
 configureBluebird();
+setMaxListeners(25); // it appears that 'nock' adds a bunch of listeners - bug?
```
srlowe: Hmm, could be.    Perhaps related to how we are instantiating in each test ('Resetting' nock back to original state is not straighforward, and perhaps something is not cleaned up, and gets added cumulatively).  I wonder if the problem goes away if a single test is run (e.g. with `it.only(...`)

pdcastro:
@srlowe, @hedss, thanks for the review. I've just added a commit ("Add 'Application' column to 'balena envs --all' output") that in addition to the extra Application column, performs service name validation (previously, an inexistent service name would be silently ignored, now an error is raised). The extra Application column can be useful when the variables are queried by device UUID, and asterisks are printed for the DEVICE or SERVICE columns -- then asterisk then means that the variable applies to all devices or services in the application listed in the APPLICATION column. Also, in case the application is inaccessible (#1530), the APPLICATION is listed as 'N/A' in tabular output or `null` in JSON output.
I could actually squash that commit before merging the PR, if you're happy with it.

lib/actions-oclif/envs.ts
```diff
@@ -241,20 +282,24 @@ async function getAppVars(
...
  */
 async function getDeviceVars(
-	sdk: import('balena-sdk').BalenaSDK,
+	sdk: SDK.BalenaSDK,
+	fullUUID: string,
+	appName: string | undefined,
 	options: FlagsDef,
 ): Promise<EnvironmentVariableInfo[]> {
-	const deviceUUID = options.device!;
+	const printedUUID = options.json ? fullUUID : options.device!;
```
pdcastro: I've found that not all SDK methods accept short 7-char UUIDs (instead requiring the full 32-char UUID). So I thought that, when the user chooses JSON output (which is meant for programmatic use), the full UUID should always be printed instead of the short UUID. However in tabular format (which is meant for human consumption, and "visual neatness" is a little bit more of a consideration), the same UUID given by the user in the command line (`options.device`) is used -- if the user provided a short UUID, a short UUID is printed; if the user provided a full UUID, a full UUID is printed.

pdcastro:
@srlowe, @hedss, I've added three commits for `env add`, `env rm` and `env rename` respectively, and squashed the commit I had mentioned earlier (that added the Application column to the output of `balena envs`). Sorry for the changes after your earlier review. If you now review the last three commits, the PR should be good to go:
```
Add multicontainer (microservices) support for 'balena env add'
Add multicontainer (microservices) support for 'balena env rm'
Add multicontainer (microservices) support for 'balena env rename'
```
(I was wavering between 4 semver-minor PRs or just this one, and finally opted for just this one as they are all about multicontainer support for balena envs / env add / env rm / env rename.)
**Thank you!**

pdcastro:
force push = corrected a couple of typos and added command-line examples to the 'help' output for `balena rm` and `balena rename`.

Issue #1530, opened 2019-12-11
Some actions throwing errors when device assigned to inaccesible application.


srlowe:
In some normal and expected situations, a users device can end up associated with an application which they no longer have access to (e.g. when a user is a collaborator on an application, and later has access revoked).  This causes the associated applications array to be empty when the API is queried for that device by the user.

Some actions in the CLI assume a device will always have an application associated, and will throw an error if it is missing:
 - `balena devices`
 - `balena device <uuid>`
 - `balena device move <uuid>`

srlowe:
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/h98mNEGp0rFQ5umSF68uw-144a3

PR #1533, opened 2019-12-12
Fix issues with devices associated with inaccessible applications.

srlowe:
Fixes three CLI actions which assumed a device would always have an application associated, and threw an error if missing.  (Associated application can become inaccessible when e.g. a user is no longer a collaborator on an application)

```
balena devices
balena device <uuid>
balena device move <uuid>
```
Change-type: patch
Resolves: #1530

---
##### Contributor checklist


lib/actions/device.coffee
```diff
@@ -53,7 +53,8 @@ exports.list =
 		.tap (devices) ->
 			devices = _.map devices, (device) ->
 				device.dashboard_url = balena.models.device.getDashboardUrl(device.uuid)
-				device.application_name = device.belongs_to__application[0].app_name
+				device.application_name =
+					if device.belongs_to__application[0] then device.belongs_to__application[0].app_name else 'N/a'
```
pdcastro: Not 100% sure, but I would think that `belongs_to__application` would be an empty array or even undefined in case the application was not accessible, so that `belongs_to__application[0]` might raise a TypeError:
```
$ node
Welcome to Node.js v12.13.0.
Type ".help" for more information.
> let a;
undefined
> a[0]
Thrown:
TypeError: Cannot read property '0' of undefined
```

pdcastro: The CoffeeeScript [Existential Operator](https://coffeescript.org/#existential-operator) might come in handy in this case. Maybe (needs testing) the syntax would be:
```
if device.belongs_to__application?[0] then ...
```

srlowe: @pdcastro I believe in these cases `belongs_to__application` is sent as empty array from API.  E.g.:

```
{
  "d": [
	{
	  "belongs_to__application": [
	  ],
```

The original error that Shaun reported would suggest this:
```
TypeError: Cannot read property 'app_name' of undefined
```
Otherwise we'd expect the error you mentioned.

Still, probably good idea to check array too! (Will update)

srlowe: Updated with existential operator and tested for missing `belongs_to__application`.

tests/helpers.ts
```diff
@@ -23,6 +23,10 @@ import * as path from 'path';
 import * as balenaCLI from '../build/app';
 import { configureBluebird } from '../build/app-common';
 
+// REMOVE When setMaxListeners merged
```
pdcastro: This meant PR #1529, right? I've finally merged it, which also means this PR now needs rebasing - sorry! :-)

PR #1534, opened 2019-12-13
[Experimental] Add ability to install with "npm install git://github.com/balena-io/balena-cli"

pdcastro:
Flowdock discussion:
https://www.flowdock.com/app/rulemotion/resin-tech/threads/EwR2oqZ-rYR4NaVkw3z06tZucyI

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,11 @@
     "bin/",
     "build/",
     "doc/",
+    "gulpfile.coffee",
+    "tsconfig.json",
     "lib/",
     "patches/",
+    "typings/",
     "*.md",
     "npm-shrinkwrap.json"
   ],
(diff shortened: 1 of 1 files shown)


pdcastro:
Closing the PR because:
* The `--production` flag does not work (`npm install -g --production`), because the added `npm run build:fast` step requires a number of 'devDependencies' packages declared in `package.json`.
* It broke the Windows build. The usual "npm install -g --production balena-cli" would probably fail too.
* The changes to the `postinstall` script means that running a plain "npm install" on the repo would cause "npm run build:fast" to run too, which is an extra step that takes some time and reduces people's workflow flexibility. (Despite the name "build:fast", it's more accurately described as "less slow".) :-) We would then also want to amend the CI scripts to avoid running "build:fast" multiple times, as currently it is a separate step.
* https://www.flowdock.com/app/rulemotion/resin-tech/threads/EwR2oqZ-rYR4NaVkw3z06tZucyI

Issue #467, opened 2017-03-29
Expose the list of supported device types and their aliases


emirotin:
With aliases.

And also mention that device types can have alternative names.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_1da1)

pimterry:
Basic support for this is already done, see `resin devices supported`. Closing.

lekkas:
@pimterry just a heads up that the support conversation where this ticket originated from (https://app.frontapp.com/open/cnv_4c4n9h) suggested that we would look into presenting the users with any existing device type aliases as well (e.g. `raspberrypi` alias for `raspberry-pi`). 

I'm not sure where we stand about aliases now, do we ignore them completely? If not, are they documented somewhere?

PR #1553, opened 2019-12-26
Add --verbose and --json options to the 'devices supported' command

pdcastro:
The command was also converted to oclif.

(The reason why I chose to work on this otherwise low-priority PR is that I actually needed the information of available device aliases while writing the spec for the CLI (balenaOS) Configuration project, and it was quick and easy enough to put together.)

Resolves: #467
Change-type: minor

### DIFF
--- a/lib/actions-oclif/devices/supported.ts
+++ b/lib/actions-oclif/devices/supported.ts
@@ -1,0 +1,118 @@
+/**
+ * @license
+ * Copyright 2016-2019 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Command, flags } from '@oclif/command';
+import * as SDK from 'balena-sdk';
+import { stripIndent } from 'common-tags';
+import * as _ from 'lodash';
+
+import * as cf from '../../utils/common-flags';
+import { CommandHelp } from '../../utils/oclif-utils';
+
+interface FlagsDef {
+	discontinued: boolean;
+	help: void;
+	json?: boolean;
+	verbose?: boolean;
+}
+
+interface DeviceTypeWithAliases extends SDK.DeviceType {
+	aliases?: string[];
+}
+
+export default class DevicesSupportedCmd extends Command {
+	public static description = stripIndent`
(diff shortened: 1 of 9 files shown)


lib/actions-oclif/devices/supported.ts
```diff
@@ -0,0 +1,119 @@
...
+
+interface DeviceTypeWithAliases extends SDK.DeviceType {
+	aliases?: string[];
+}
+
+export default class DevicesSupportedCmd extends Command {
+	public static description = stripIndent`
+		List the supported device types (like 'raspberrypi3' or 'intel-nuc').
+
+		List the supported device types (like 'raspberrypi3' or 'intel-nuc').
```
srlowe: Is this a duplicate line?

lib/actions-oclif/devices/supported.ts
```diff
@@ -0,0 +1,119 @@
...
+		'$ balena devices supported',
+		'$ balena devices supported --verbose',
+		'$ balena devices supported -vj',
+	];
+
+	public static usage = (
+		'devices supported ' +
+		new CommandHelp({ args: DevicesSupportedCmd.args }).defaultUsage()
+	).trim();
+	// public static usage = 'devices supported';
```
srlowe: Should this stay?

tests/commands/device/supported.spec.ts
```diff
@@ -30,7 +20,7 @@ describe('balena devices supported', function() {
 
 		const { out, err } = await runCommand('devices supported -h');
 
-		expect(cleanOutput(out)).to.deep.equal(cleanOutput([HELP_RESPONSE]));
+		expect(cleanOutput(out)).to.contain('$ balena devices supported');
```
srlowe: Yeah, I think the help text testing is maybe a bit over the top at the moment.  This is a more reasonable basic check that it's working.

lib/actions-oclif/devices/supported.ts
pdcastro: It is an intentional duplicate because of they way the oclif actions process / parse this string for help output. The first line is used in at the very top of help output, and the duplicate line starts the DESCRIPTION block:
```
$ ./bin/balena help devices supported
List the supported device types (like 'raspberrypi3' or 'intel-nuc').

USAGE
  $ balena devices supported

OPTIONS
  ...

DESCRIPTION
  List the supported device types (like 'raspberrypi3' or 'intel-nuc').

  The --verbose option adds extra columns/fields to the output, including the
  "STATE" column whose values are one of 'beta', 'release' or 'discontinued'.
  ...
```
The first line is also the one used in `balena help -v`.

lib/actions-oclif/devices/supported.ts
pdcastro: Well spotted! Deleted.

tests/commands/device/supported.spec.ts
pdcastro: Yes, I kept having to amend the test for any small changes to the description, so I thought of making this change. :-)

lib/actions-oclif/devices/supported.ts
srlowe: Ah, good to know, thanks.

Issue #1445, opened 2019-09-23
Balena push/build fails to understand code as a multicontainer project


imrehg:
- **balena CLI version:** 11.11.2
- **Operating system version:** ArchLinux
- **32/64 bit OS and processor:** 64-bit
- **Install method:** npm
- **If npm install, Node.js and npm version:*

```
~> npm version  
{
  npm: '6.11.3',
  ares: '1.15.0',
  brotli: '1.0.7',
  cldr: '35.1',
  http_parser: '2.8.0',
  icu: '64.2',
  llhttp: '1.1.4',
  modules: '72',
  napi: '4',
  nghttp2: '1.39.2',
  node: '12.10.0',
  openssl: '1.1.1c',
  tz: '2019a',
  unicode: '12.1',
  uv: '1.31.0',
  v8: '7.6.303.29-node.16',
  zlib: '1.2.11'
}
```

When deploying this repo to one of our applications: https://github.com/monicahq/monica, with balena push (just changing the version in the docker-compose.yml to 2.1, the build fails with)

```
master /tmp/monica> balena push monica
[Info]     Starting build for monica, user imrehg
[Info]     Dashboard link: https://dashboard.balena-cloud.com/apps/1513257/devices
[Error]    Could not detect project type: Service main: package.json: engines.node must be specified
[Error]    Not deploying release.
Remote build failed

Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, or "export DEBUG=1"
on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```
Similar thing for `balena build`
```
master /tmp/monica> balena build -a monica
[Info]    Compose file detected
[Info]    Building for armv7hf/raspberrypi3
[Build]   Built 2 services in 0:02
[Build]   monicahq Preparing...
[Build]   mysql    Preparing...
[Error]   Build failed
package.json: engines.node must be specified
```

Discussed in [Flowdock](https://www.flowdock.com/app/rulemotion/resin-tech/threads/y0Bh0eD-duERaY5DcDqi5vlG9Qu)

Dko1905:
I have the same issue. 
* balena CLI version: 11.19.1
* Operating system version: Manjaro/ArchLinux
* 32/64 bit OS and processor: 64-bit
[...]

PR #1558, opened 2020-01-08
Prevent file ignorer from ignoring Dockerfile, docker-compose.yml

srlowe:
Prevent file ignorer from ignoring `Dockerfile`, `docker-compose.yml`, even if listed in `.dockerignore`

Previously: 
- If `Dockerfile` was included in `.dockerignore`, then then push/build would fail with the message `Error: package.json: engines.node must be specified`.  This was due to `Dockerfile` being missing from the bundle, and `resin-bundle-resolve` matching on the 'node' resolver, rather than docker.
- If `docker-compose.yml` was included in `.dockerignore`, then the push/build would fail due to the default `docker-compose` file being used rather than the real one.

Note, including `Dockerfile` in `.dockerignore` is valid according to docker docs (not sure about `docker-compose.yml` but we should protect against it anyway). 
 https://docs.docker.com/engine/reference/builder/#dockerignore-file

Change-type: patch
Resolves: #1445

### DIFF
--- a/lib/utils/ignore.ts
+++ b/lib/utils/ignore.ts
@@ -115,6 +115,11 @@ export class FileIgnorer {
 			return true;
 		}
 
+		// Don't ignore Dockerfile (with or without extension) or docker-compose.yml
+		if (/^Dockerfile$|^Dockerfile\.\S+/.test(path.basename(relFile)) || path.basename(relFile) === 'docker-compose.yml') {
+			return true;
+		}
+
 		const dockerIgnoreHandle = dockerIgnore();
 		const gitIgnoreHandle = ignore();
 


lib/utils/ignore.ts
```diff
@@ -115,6 +115,11 @@ export class FileIgnorer {
 			return true;
 		}
 
+		// Don't ignore Dockerfile or docker-compose.yml
+		if (path.basename(relFile) === 'Dockerfile' || path.basename(relFile) === 'docker-compose.yml') {
+			return true;
```
pdcastro: I find it uncomfortable that we hardcode these exceptions (it feels like it is the user's mistake if they list 'Dockerfile' in their `.dockerignore` and/or `.gitignore` files), but there is a precedent (`.balena` just above...), it's probably for the best in practice (given support requests) and it is what previous CLI check-in meetings had decided. So, I agree. :-)

In this case, we should probably also hardcode exceptions for `Dockerfile.template`, and the likes of `Dockerfile.armv7hf` and `Dockerfile.raspberrypi3` (see [documentation](https://www.balena.io/docs/learn/develop/dockerfile/#multiple-dockerfiles)). It think it's enough to add a regex along the lines of `(Dockerfile|Dockerfile\.\S+)`.

srlowe: Yeah, I can't really think of an alternative approach.  I think it stems from the fact that we are using `.dockerignore` in a subtly different way to docker.  (e.g. Adding `Dockerfile` to `.dockerignore` is even suggested in the docker documentation).

srlowe: One small downside is that the monica project included `Dockerfile.dev`, which would now be included in the build.

srlowe:
@balena-ci retest

pdcastro:
`retest` won't help until they solve the unavailable balena CI machines or networking nodes in the Seattle office:
* https://www.flowdock.com/app/rulemotion/o-seattle/threads/a45SZFJvu54w2FuavfCidYXc22l
* https://ci.balena-dev.com/teams/main/pipelines/github-events-resin/jobs/node-cli/builds/331
```
no workers satisfying: platform 'darwin', tag 'onprem'
no workers satisfying: platform 'windows', tag 'onprem'
```

srlowe:
@balena-ci retest

Issue #1557, opened 2020-01-07
balena join/leave hangs on Windows


pdcastro:
- **balena CLI version:** e.g. 11.21.0 (older versions as well)
- **Operating system version:** e.g. Windows 10
- **32/64 bit OS and processor:** 64-bit Windows on 64-bit Intel processor
- **Install method:** executable installer

---
The following hangs on Windows 10, when the CLI uses the built-in `ssh` client that ships with Windows 10 (or that gets installed with Windows Update).

```
balena join <ip-addr> -a appName
balena leave <ip-add>
```

Adding some debugging code, I figured that the ssh client was printing the following message to stderr:

```
GetConsoleMode on STD_INPUT_HANDLE failed with 6
```

And it is cause by a bug in the ssh client:

https://github.com/PowerShell/Win32-OpenSSH/issues/856

Updating the ssh client fixes it, but I will also submit a PR that works around the issue in CLI code.

pdcastro:
CLI release v11.21.4 works around the Windows ssh client bug by attaching the CLI's stdin to the ssh subprocess.

PR #1559, opened 2020-01-09
Fix join and leave commands on Windows (hanging on stdin and argument escaping)

pdcastro:
Resolves: #1557 
Change-type: patch

### DIFF
--- a/lib/utils/helpers.ts
+++ b/lib/utils/helpers.ts
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import { InitializeEmitter, OperationState } from 'balena-device-init';
 import BalenaSdk = require('balena-sdk');
 import Bluebird = require('bluebird');
 import chalk from 'chalk';
@@ -22,7 +23,7 @@ import os = require('os');
 import visuals = require('resin-cli-visuals');
 import * as ShellEscape from 'shell-escape';
 
-import { InitializeEmitter, OperationState } from 'balena-device-init';
+import { ExpectedError } from '../errors';
 
 const balena = BalenaSdk.fromSharedOptions();
 
@@ -190,10 +191,21 @@ export function getApplication(applicationName: string) {
 	return balena.models.application.get(applicationName, extraOptions);
 }
 
-// A function to reliably execute a command
-// in all supported operating systems, including
-// different Windows environments like `cmd.exe`
-// and `Cygwin`.
+/**
+ * Choose between 'cmd.exe' and '/bin/sh' for running the given command string,
+ * depending on the value of `os.platform()`.
+ * When writing new code, consider whether it would be possible to avoid using a
+ * shell at all, using the which() function in this module to obtain a program's
+ * full path, executing the program directly and passing the arguments as an
+ * array instead of a long string. Avoiding a shell has several benefits:
+ *   - Avoids the need to shell-escape arguments, especially nested commands.
+ *   - Bypasses the incompatibilities between cmd.exe and /bin/sh.
+ *   - Reduces the security risks of lax input validation.
+ * Code example avoiding a shell:
+ *    const program = await which('ssh');
+ *    const args = ['root@192.168.1.1', 'cat /etc/os-release'];
+ *    const child = spawn(program, args);
+ */
 export function getSubShellCommand(command: string) {
 	if (os.platform() === 'win32') {
 		return {
@@ -293,6 +305,21 @@ export function getManualSortCompareFunction<T, U = T>(
 }
 
 /**
+ * Decide whether the current shell (that executed the CLI process) is a Windows
+ * 'cmd.exe' shell, including PowerShell, by checking a few environment
+ * variables.
+ */
+export function isWindowsComExeShell() {
+	return (
+		// neither bash nor sh (e.g. not MSYS, MSYS2, Cygwin, WSL)
+		process.env.SHELL == null &&
+		// Windows cmd.exe or PowerShell
+		process.env.ComSpec != null &&
+		process.env.ComSpec.endsWith('cmd.exe')
+	);
+}
+
+/**
  * Shell argument escaping compatible with sh, bash and Windows cmd.exe.
  * @param arg Arguments to be escaped
  * @param detectShell Whether to use the SHELL and ComSpec environment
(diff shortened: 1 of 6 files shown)


lib/utils/promote.ts
```diff
@@ -29,14 +29,6 @@ export async function join(
 	deviceHostnameOrIp?: string,
 	appName?: string,
 ): Promise<void> {
-	logger.logDebug('Checking login...');
-	const isLoggedIn = await sdk.auth.isLoggedIn();
-	if (!isLoggedIn) {
-		logger.logInfo("Looks like you're not logged in yet!");
-		logger.logInfo("Let's go through a quick wizard to get you started.\n");
-		await runCommand('login');
```
pdcastro: Running 'login' in this way was not working for some reason: on macOS, it would print the CLI 'help' and then run login, and on Windows, it was failing altogether with "BalenaRequestError: Request error: Invalid binding". I spent a bit of time trying to figure it out but gave up (not high enough of a priority) and decided it was better to just add `permission: 'user'` in `join.ts`, which is how every other Capitano command is currently implemented in relation to the requirement of logging in, including the `balena leave` command.

lib/utils/helpers.ts
```diff
@@ -292,6 +304,21 @@ export function getManualSortCompareFunction<T, U = T>(
...
+ */
+export function isWindowsComExeShell() {
+	return (
+		// neither bash nor sh (e.g. not MSYS, MSYS2, Cygwin, WSL)
+		process.env.SHELL == null &&
+		// Windows cmd.exe or PowerShell
+		process.env.ComSpec != null &&
+		process.env.ComSpec.endsWith('cmd.exe')
+	);
+}
```
pdcastro: This is a refactoring of the `shellEscape` function. It was done during debugging. It turned out not to be relevant for the bug fixes, but it makes the the `shellEscape` function a bit easier to read.

lib/utils/helpers.ts
```diff
@@ -356,3 +375,28 @@ export async function workaroundWindowsDnsIssue(ipOrHostname: string) {
...
+	} catch (err) {
+		if (err.code === 'ENOENT') {
+			throw new ExpectedError(
+				`'${program}' program not found. Is it installed?`,
+			);
+		}
+		throw err;
+	}
+	return programPath;
+}
```
pdcastro: The `which` function was added as it allows us to avoid using a shell to invoke 'ssh'. Avoiding a shell avoids having to deal with the incompatibilities between the Windows `cmd.exe` on and the Unix `/bin/sh`.

lib/utils/promote.ts
```diff
@@ -117,26 +109,29 @@ async function configure(deviceIp: string, config: any): Promise<void> {
 	const json = JSON.stringify(config);
 	const b64 = Buffer.from(json).toString('base64');
 	const str = `"$(base64 -d <<< ${b64})"`;
-	await execCommand(deviceIp, `os-config join '${str}'`, 'Configuring...');
+	await execCommand(deviceIp, `os-config join ${str}`, 'Configuring...');
```
pdcastro: The single quotes around `'${str}'` are one of the reasons why `balena join` was failing to run on Windows. On macOS and Linux, `/bin/sh` was removing the single quotes when running 'ssh' as intended by the original developer. But on Windows, `cmd.exe` does not remove the single quotes.

lib/utils/promote.ts
```diff
@@ -117,26 +109,29 @@ async function configure(deviceIp: string, config: any): Promise<void> {
...
+		throw new ExpectedError(stripIndent`
 			Device "${deviceIp}" is incompatible and cannot join or leave an application.
-			Please select or provision device with balenaOS newer than ${MIN_BALENAOS_VERSION}.`);
+			Please select or provision a device with balenaOS newer than ${MIN_BALENAOS_VERSION}.`);
 	}
 }
 
 async function getDeviceType(deviceIp: string): Promise<string> {
-	const output = await execBuffered(deviceIp, 'cat /etc/os-release');
+	const output = await getDeviceOsRelease(deviceIp);
```
pdcastro: During debugging, I observed that `balena join` was invoking `ssh` twice to run `cat /etc/os-release`. So I created `getDeviceOsRelease` using lodash's memoize to cache the result.

lib/utils/ssh.ts
```diff
@@ -36,17 +35,41 @@ export async function exec(
...
-	const command = `ssh \
-		-t \
-		-p 22222 \
-		-o LogLevel=ERROR \
-		-o StrictHostKeyChecking=no \
-		-o UserKnownHostsFile=/dev/null \
-		root@${deviceIp} \
-		${cmd}`;
+	const { which } = await import('./helpers');
+	const program = await which('ssh');
```
pdcastro: This call to `which()` replaces the previous call to `getSubShellCommand()`. It allows us to avoid using a shell (`cmd.exe` on Windows or `/bin/sh` on Unix) to invoke `ssh`, which removes the need to shell-escape arguments, which was the cause of one of the bugs on Windows.

lib/utils/ssh.ts
```diff
@@ -36,17 +35,41 @@ export async function exec(
...
+	// https://github.com/PowerShell/Win32-OpenSSH/issues/856
+	// but users whould have to manually download and install a new client.
+	// Note that "ssh -n" does not solve the problem, but should in theory
+	// prevent the ssh client from using the CLI process stdin, even if it
+	// is connected with 'inherit'.
+	const stdio: StdioOptions = [
+		'inherit',
+		stdout ? 'pipe' : 'inherit',
+		'inherit',
+	];
```
pdcastro: Replacing 'ignore' with 'inherit' was the fix for the CLI hanging on Windows. Apparently it only happens because of a bug in the Microsoft ssh client that ships with Windows. To avoid ssh actually using the CLI process stdin, I've also added the `-n` flag when invoking `ssh`, although currently I don't think it matters.

I have also changed stderr from `ignore` to `inherit`. This was not necessary to avoid the hanging bug, but it means that error messages are not swept under the carpet. An error message regarding the `stdin` problem on Windows was actually being printed to stderr, but hidden. This made the investigation harder, as initially we had no idea that the problem was with the ssh client.

lib/utils/helpers.ts
srlowe: Very descriptive comments!

pdcastro: Too much? :-) Occasionally I write more lines of comment than lines of code. :-D

srlowe: Nono, not too much!  Great IMO.

pdcastro:
force push = rebase on master

PR #1572, opened 2020-01-16
Add test cases for the push, build and deploy commands

pdcastro:
This PR adds test code for the push, build and deploy commands, using [nock](https://www.npmjs.com/package/nock) to mock HTTP responses from the balenaCloud API, builders and Docker daemon. In this PR, tests are limited to verifying that the produced tar stream (intercepted in the body of HTTP POST requests) matches the expected files in a very simple test project. The tests only cover a fraction of the functionality provided by those commands, but it is a starting point that we can build on in future pull requests.

The PR adds and modifies files under the `tests` folder only, and doesn't touch production code.


pdcastro:
@balena-ci retest

tests/balena-api-mock.ts
```diff
@@ -16,70 +16,122 @@
...
+
+const jHeader = { 'Content-Type': 'application/json' };
 
+export class BalenaAPIMock extends NockMock {
 	constructor() {
-		nock.cleanAll();
+		super('https://api.balena-cloud.com');
+	}
+
+	public expectGetApplication(opts: ScopeOpts = {}) {
```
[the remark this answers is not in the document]
srlowe: Good idea re standardisation.   I think we might start having some issues once we continue to add more of these methods though to this file though.  E.g. when I first added `expectWhoAmI` and `expectMixpanel`, I put them in `balena-api-mock.ts` for convenience, because they are used in every/most CLI actions.  I think this file will get huge and hard to manage if we keep putting them here though (and will obscure the general/framework parts).  For the ones that are only used in one or a few actions, maybe we should brainstorm about where would be the most appropriate place for these to live.  Maybe in or nearer the particular test specs themselves? E.g. perhaps something like a `call-helpers.ts` file in each test spec folder?

tests/balena-api-mock.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2019 Balena Ltd.
+ * Copyright 2019-2020 Balena Ltd.
```
srlowe: Do we have a company-wide approach for what the copyright message should be?

tests/nock-mock.ts
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Base class for tests using nock to intercept HTTP requests.
+ * Subclasses include BalenaAPIMock, DockerMock and BuilderMock.
+ */
+export class NockMock {
+	public readonly scope: nock.Scope;
+	// Expose `scope` as `expect` to allow for better semantics in tests
+	public readonly expect = this.scope;
+	protected static instanceCount = 0;
```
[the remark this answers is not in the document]
srlowe: Ok, I'm not totally sure about this, but if it works.... :)  I don't like nock's global behaviors - we shouldn't have to do stuff like this IMO, but hey.

package.json
```diff
@@ -140,9 +140,9 @@
     "nock": "^11.0.7",
     "parse-link-header": "~1.0.1",
     "pkg": "^4.4.0",
-    "prettier": "1.17.0",
+    "prettier": "^1.19.1",
```
pdcastro: This matches the version of `prettier` that the newer `resin-lint` version (3.1.1) depends on. Reminder: the CLI has a `npm-shrinkwrap.json` file, therefore the specification `"^1.19.1"` does not mean that `prettier` will auto update to newer versions whenever `npm install` is run.

automation/build-bin.ts
```diff
@@ -69,34 +71,6 @@ export const finalReleaseAssets: { [platform: string]: string[] } = {
...
- * Run the MSYS2 bash.exe shell in a child process (child_process.spawn()).
- * The given argv arguments are escaped using the 'shell-escape' package,
- * so that backslashes in Windows paths, and other bash-special characters,
- * are preserved. If argv is not provided, defaults to process.argv, to the
- * effect that this current (parent) process is re-executed under MSYS2 bash.
- * This is useful to change the default shell from cmd.exe to MSYS2 bash on
- * Windows.
- * @param argv Arguments to be shell-escaped and given to MSYS2 bash.exe.
- */
-export async function runUnderMsys(argv?: string[]) {
```
pdcastro: A bunch of functions, like this one, were refactored from `automation/build-bin.ts` to a new file, `automation/utils.ts`.

automation/build-bin.ts
```diff
@@ -313,62 +287,24 @@ export async function buildOclifInstaller() {
...
-		});
-		child.on('exit', (code: number) => {
-			if (code) {
-				reject(new Error(`"${execPath}": non-zero exit code "${code}"`));
-			} else {
-				resolve(stdout);
-			}
-		});
-	});
+export async function catchUncommitted(): Promise<void> {
```
pdcastro: This function implements the main functionality of this commit: conditionally calling `catch-uncommitted` (via `npx`).

automation/run.ts
```diff
@@ -20,14 +20,13 @@ import * as _ from 'lodash';
...
-	fixPathForMsys,
-	ROOT,
-	runUnderMsys,
+	catchUncommitted,
 } from './build-bin';
 import {
 	release,
 	updateDescriptionOfReleasesAffectedByIssue1359,
 } from './deploy-bin';
+import { fixPathForMsys, ROOT, runUnderMsys } from './utils';
```
pdcastro: These functions were refactored from `automation/build-bin.ts` to the new `automation/utils.ts`.

automation/utils.ts
```diff
@@ -0,0 +1,174 @@
+/**
```
pdcastro: Note that most of the code in this file was simply copied-and-pasted (refactored) from `automation/build-bin.ts`.

package.json
```diff
@@ -41,7 +41,7 @@
   "scripts": {
     "postinstall": "patch-package",
     "prebuild": "rimraf build/ build-bin/",
-    "build": "npm run build:src",
+    "build": "npm run build:src && npm run catch-uncommitted",
```
pdcastro: Core change in this commit.

package.json
```diff
@@ -52,6 +52,7 @@
     "pretest": "npm run build",
     "test": "mocha --timeout 6000 -r ts-node/register \"tests/**/*.spec.ts\"",
     "test:fast": "npm run build:fast && npm run test",
+    "catch-uncommitted": "ts-node --type-check -P automation/tsconfig.json automation/run.ts catch-uncommitted",
     "ci": "npm run test && catch-uncommitted",
```
pdcastro: Regarding the pre-existing `ci` script, it is still referred to in the `.travis.yml` file. I'm leaving it alone (just like `.travis.yml`) just in case we need/want to (temporarily) revert to Travis and AppVeyor. Just a couple of weeks ago, balena CI was unavailable for 10 days... Luckily we didn't have any urgent CLI PR during that period, but if we had, we might have reverted to Travis and AppVeyor.

package.json
```diff
@@ -124,7 +125,7 @@
     "@types/tar-stream": "1.6.0",
     "@types/through2": "2.0.33",
     "@types/which": "^1.3.2",
-    "catch-uncommitted": "^1.3.0",
+    "catch-uncommitted": "^1.4.0",
```
pdcastro: This was actually necessary to avoid failing on uncommitted changes to `.versionbot/CHANGELOG.yml`.

automation/utils.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/**
+ * Call which(programName) and spawn() with the given arguments. Throw an error
+ * if the process exit code is not zero.
+ */
+export async function whichSpawn(
+	programName: string,
+	args: string[],
+): Promise<void> {
```
pdcastro: This is the only new function / new code in this file.

lib/utils/ignore.ts
```diff
@@ -116,7 +116,10 @@ export class FileIgnorer {
 		}
 
 		// Don't ignore Dockerfile (with or without extension) or docker-compose.yml
-		if (/^Dockerfile$|^Dockerfile\.\S+/.test(path.basename(relFile)) || path.basename(relFile) === 'docker-compose.yml') {
+		if (
+			/^Dockerfile$|^Dockerfile\.\S+/.test(path.basename(relFile)) ||
+			path.basename(relFile) === 'docker-compose.yml'
```
pdcastro: This change was necessary even before upgrading `prettier` and `resin-lint` -- it was the motivation for the `catch-uncommitted` commit. But with the upgrade to `prettier` and `resin-lint`, it became just another prettification change in this commit...

Issue #1498, opened 2019-11-06
Allow custom network connections to be generated


hedss:
Currently we allow the configuration of a Wifi connection via `balena os config`, which will create an appropriate wifi connection in the `system-connections` directory of the boot partition.

We should allow the generation of custom network configurations (eg. modem configs) as well, and allow them to be injected.

I *thought* I'd raised this under another issue, but can't find it.

cc @majorz

pdcastro:
Related to issue #1395 ['balena config' should add network settings in the right place for balenaOS 2.x](https://github.com/balena-io/balena-cli/issues/1395)

richbayliss:
@hedss do you think this issue is closable now, due to the merging of the above PR? You talk about file *generation* which is a different, so I wasn't sure.

hedss:
@richbayliss Yes, and shall do so now! cc @pdcastro

Issue #1162, opened 2019-04-11
User wants to use CLI to apply custom network modifications on the master image


gelbal:
User wants to modify the balenaOS image file with their custom cellular modem connection config. Ideally the user wants the master image to already include these files/modifications, and then burn that to the SD card from etcher.

Relevant support thread:
https://jel.ly.fish/#/72e0dbf8-b146-417c-9de3-6a7abb6d68b5

--- 

I guess similar to how `balena config generate` works to create a custom config.json file.

richbayliss:
Feature added in #1576

Issue #957, opened 2018-08-27
The CLI should allow configuring NetworkManager in images


CameronDiver:
Currently the only way to do this is to mount the flashed SD card, and alter the `resin-boot` partition, or alter the image pre-flash.

It would be better if we could provide resin-cli with an image and a NetworkManager configuration file, and it would be ready to flash, in the same way that preload works.

- **resin-cli version:**
N/a
- **Operating system and architecture:**
N/a

pdcastro:
Related to (duplicated by?) #1498

PR #1576, opened 2020-01-21
configure: Allow passing system-connection files to `os configure` command

richbayliss:
Allow passing files to `os configure` via `--system-connection` to allow
pre-configuration of network configuration for cellular networks.

Change-type: minor

### DIFF
--- a/lib/actions-oclif/os/configure.ts
+++ b/lib/actions-oclif/os/configure.ts
@@ -17,12 +17,17 @@
 
 import { Command, flags } from '@oclif/command';
 import BalenaSdk = require('balena-sdk');
+import Bluebird = require('bluebird');
 import { stripIndent } from 'common-tags';
 import * as _ from 'lodash';
+import * as path from 'path';
 
 import { ExpectedError } from '../../errors';
 import * as cf from '../../utils/common-flags';
 import { CommandHelp } from '../../utils/oclif-utils';
+
+const BOOT_PARTITION = 1;
+const CONNECTIONS_FOLDER = '/system-connections';
 
 interface FlagsDef {
 	advanced?: boolean;
@@ -38,6 +43,7 @@ interface FlagsDef {
 	'device-type'?: string;
 	help?: void;
 	version?: string;
+	'system-connection': string[];
 }
 
 interface ArgsDef {
@@ -67,7 +73,7 @@ export default class OsConfigureCmd extends Command {
 
 		Configure a previously downloaded balenaOS image for a specific device type or
 		balena application.
-		
+
 		Configuration settings such as WiFi authentication will be taken from the
 		following sources, in precedence order:
 		1. Command-line options like \`--config-wifi-ssid\`
@@ -76,6 +82,13 @@ export default class OsConfigureCmd extends Command {
 
 		The --device-type option may be used to override the application's default
 		device type, in case of an application with mixed device types.
+
+		The --system-connection (-c) option can be used to inject NetworkManager connection
+		profiles for additional network interfaces, such as cellular/GSM or additional
+		WiFi or ethernet connections. This option may be passed multiple times in case there
+		are multiple files to inject. See connection profile examples and reference at:
+		https://www.balena.io/docs/reference/OS/network/2.x/
+		https://developer.gnome.org/NetworkManager/stable/nm-settings.html
 
 		${deviceApiKeyDeprecationMsg.split('\n').join('\n\t\t')}
 	`;
@@ -144,6 +157,13 @@ export default class OsConfigureCmd extends Command {
 		version: flags.string({
 			description: 'balenaOS version, for example "2.32.0" or "2.44.0+rev1"',
 		}),
+		'system-connection': flags.string({
+			multiple: true,
+			char: 'c',
+			required: false,
+			description:
+				"paths to local files to place into the 'system-connections' directory",
+		}),
 	};
 
 	public async run() {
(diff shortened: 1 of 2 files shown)


lib/actions-oclif/os/configure.ts
```diff
@@ -144,6 +150,11 @@ export default class OsConfigureCmd extends Command {
 		version: flags.string({
 			description: 'balenaOS version, for example "2.32.0" or "2.44.0+rev1"',
 		}),
+		'system-connection': flags.string({
+			multiple: true,
+			required: false,
+			description: 'paths to local files to place into the \'system-connections\' directory',
```
pdcastro: * Let's add something like `char: 'c'` here (equivalent to Capitano's 'alias') so one can write:
```
... -c a.con -c b.con -c c.con
```
instead of
```
... --system-connection a.con --system-connection b.con --system-connection c.con
```

* It would be nice to add a few words to the output of `balena help os configure`, explaining that additional NetworkManager connection profiles can be added using the `--system-connection` option. It could also mention the filename that gets generated by default for WiFi settings taken from `--config-wifi-key` and `--config-wifi-ssid`. (What happens if the user provides that same filename as a `--system-connection` argument? Looks like `--system-connection` overwrites the 'default' filename, which is OK.)

lib/actions-oclif/os/configure.ts
```diff
@@ -213,14 +225,36 @@ export default class OsConfigureCmd extends Command {
...
+				image,
 				deviceTypeManifest,
 				configJson || {},
 				answers,
 			),
 		);
+
+		if (options["system-connection"]) {
+			const files = await Bluebird.map(options["system-connection"], async filePath => {
+				const content = await fs.promises.readFile(filePath, 'utf8');
```
pdcastro: What the difference between `fs.promises.readFile` and `fs.readFile`? (given `const fs = await import('mz/fs');`)

lib/actions-oclif/os/configure.ts
```diff
@@ -213,14 +225,36 @@ export default class OsConfigureCmd extends Command {
...
+				const name = path.basename(filePath);
+
+				return {
+					name,
+					content
+				};
+			});
+
+			await Bluebird.each(files, async ({ name, content }) => {
```
pdcastro: `Bluebird.each` writes files one a time, rather than in parallel. I guess it's intentional.

lib/actions-oclif/os/configure.ts
richbayliss: @pdcastro I missed the `mz/fs` import, I'll use that instead to keep the usage consistent

lib/actions-oclif/os/configure.ts
richbayliss: I didn't know if parallel might have consequences with it being a disk image file, so I played it safe in the knowledge that the files *should* be small text files.

lib/actions-oclif/os/configure.ts
richbayliss: so the `os configure` command puts the wifi settings into the `config.json` and doesn't create the `system-connection` file; this would be populated to a file out on boot.

I thought that some help was already added? If you're referring to the main bulk of the text, then sure 👍

The filenames are copied over, so yes, it would overwrite anything that existed. I think this is OK since it's a pre-configuration step.

pdcastro: The main bulk of the text, yes. It seems like a big feature that should be explained; this is (partially?) addressing 3 CLI issues often requested by users and support agents:
* https://github.com/balena-io/balena-cli/issues/957
* https://github.com/balena-io/balena-cli/issues/1162
* https://github.com/balena-io/balena-cli/issues/1498
* Others?

So it would be good to explain that if users want to configure additional interfaces to the default WiFi of interactive prompts, they can use the `--system-connection` option to point to NetworkManager profiles that the user wrote. And point to documentation that describes the file format and such, like https://www.balena.io/docs/reference/OS/network/2.x/#wifi-setup

lib/actions-oclif/os/configure.ts
```diff
@@ -77,6 +83,11 @@ export default class OsConfigureCmd extends Command {
 		The --device-type option may be used to override the application's default
 		device type, in case of an application with mixed device types.
 
+		The --system-connection option can be used to inject a
+		NetworkManager system-connection config file, for providing cellular/network
+		settings. It may be passed multiple times in case there are multiple files
+		to inject.
```
pdcastro: Suggestion:
```
		The --system-connection (-c) option can be used to inject NetworkManager connection
		profiles for additional network interfaces, such as cellular/GSM or additional
		WiFi or cable connections. This option may be passed multiple times in case there
		are multiple files to inject. See connection profile examples and reference at:
		https://www.balena.io/docs/reference/OS/network/2.x/
		https://developer.gnome.org/NetworkManager/stable/nm-settings.html
```

pdcastro:
I think `Change-type: minor` rather than patch? Because it's adding a user-visible new feature. Also a few connects to:

Change-type: minor
Connects-to: #957
Connects-to: #1162
Connects-to: #1498

lib/actions-oclif/os/configure.ts
richbayliss: @pdcastro nice wording, I'm happy with that :+1:

richbayliss:
Wow! I didn't consider how many open issues this would resolve :+1: :tada:

PR #1581, opened 2020-01-24
Update dependencies

Page-:
I checked through all the version bumps and they all look safe, I also brought in a couple of majors I know are safe

Change-type: minor


automation/deploy-bin.ts
```diff
@@ -59,17 +62,12 @@ export async function release() {
...
 
-let cachedOctokit: any;
-
 /** Return a cached Octokit instance, creating a new one as needed. */
-function getOctokit(): any {
-	if (cachedOctokit) {
-		return cachedOctokit;
-	}
-	const Octokit = require('@octokit/rest').plugin(
+const getOctokit = _.once(function() {
```
Page-: This change isn't strictly necessary, but it makes the code simpler and avoided me having to figure out the correct type for `cachedOctokit` (it's now inferred automatically via `require('@octokit/rest') as typeof _ocktokit`)

lib/actions-oclif/devices/supported.ts
```diff
@@ -29,10 +29,6 @@ interface FlagsDef {
 	verbose?: boolean;
 }
 
-interface DeviceTypeWithAliases extends SDK.DeviceType {
-	aliases?: string[];
```
Page-: This was added to the SDK types themselves so it became unnecessary to extend them

package.json
```diff
@@ -129,86 +129,88 @@
...
-    "mixpanel": "^0.10.1",
+    "mixpanel": "^0.10.3",
     "mkdirp": "^0.5.1",
     "moment": "^2.24.0",
-    "moment-duration-format": "^2.2.2",
+    "moment-duration-format": "^2.3.2",
     "mz": "^2.7.0",
     "node-cleanup": "^2.1.2",
-    "oclif": "^1.13.1",
+    "node-unzip-2": "^0.2.8",
```
Page-: This is identical to `"unzip2": "balena-io-library/node-unzip-2#v0.2.8"` in terms of contents and is just pointing it to the npm version to avoid the need for the git dep

package.json
```diff
@@ -129,86 +129,88 @@
...
-    "balena-device-status": "^3.1.2",
-    "balena-image-manager": "^6.1.0",
-    "balena-preload": "^8.2.1",
-    "balena-sdk": "^12.16.0",
-    "balena-settings-client": "^4.0.0",
-    "balena-sync": "^10.0.3",
+    "balena-config-json": "^2.1.1",
+    "balena-device-init": "^5.0.1",
+    "balena-device-status": "^3.2.0",
+    "balena-errors": "^4.2.1",
```
Page-: I added this explicit dependency because the cli does directly depend on it and so if dependencies change their balena-errors version it can break the cli (either because it's no longer available in node_modules, or the wrong version is available)

tests/commands/device/supported.spec.ts
```diff
@@ -60,7 +60,7 @@ describe('balena devices supported', function() {
 
 		// Experimental devices should be listed as beta
 		expect(lines.some(l => l.includes('EXPERIMENTAL'))).to.be.false;
-		expect(lines.some(l => l.includes('BETA'))).to.be.true;
+		expect(lines.some(l => l.includes('NEW'))).to.be.true;
```
Page-: https://github.com/balena-io/balena-sdk/issues/798

Page-:
Features wise I think it's just the following:
* `resin-doodles` fixes the holiday doodles date check, so it adds them as a feature
* `balena-device-status` - "Add support for the Ordered, Preparing, Shipped device statuses"
* `ignore` supports the latest .gitignore rules so will match git's behaviour better

Issue #1579, opened 2020-01-22
Proxy fails with Node.js >= 10.16.0 (Token authentication failed)


pdcastro:
- **balena CLI version:** many, including the current latest, 11.22.0
- **Operating system version:** all
- **32/64 bit OS and processor:** any (tested on 64-bit)
- **Install method:** any, if it uses / bundles Node.js versions >= 10.16.0
- **If npm install, Node.js and npm version:** see table in a comment

---

The CLI fails to use a proxy if the Node.js version **(not the CLI version!)** is 10.16.0 or later. This applies to any "modern" (say, past year or two) CLI release, depending only on the version of Node.js bundled or used. See installer / version matrix in a comment below.

```
$ export BALENARC_PROXY='http://127.0.0.1:8888'
$ balena login --token <long-token-string-copied-from-dashboard>
 _            _
| |__   __ _ | |  ____  _ __    __ _
| '_ \ / _` || | / __ \| '_ \  / _` |
| |_) | (_) || ||  ___/| | | || (_) |
|_.__/ \__,_||_| \____/|_| |_| \__,_|

Logging in to balena-cloud.com
Token authentication failed

Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, or "export DEBUG=1"
on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For CLI bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```

* The failure happens around [these lines in `auth.coffee`](https://github.com/balena-io/balena-cli/blob/v11.22.0/lib/actions/auth.coffee#L90-L95)
* `balena-sdk` hides the underlying error (https://github.com/balena-io/balena-sdk/issues/805), but with some digging it was: `Cannot read property 'proxy' of undefined`.
* Applies to Node.js versions >=10.16.0 only (see data points in a comment below).
* It's probably linked to `global-tunnel-ng` [not supporting newer versions of Node.js](https://github.com/np-maintain/global-tunnel#global-tunnel), thus also closely related to issue #1335.

pdcastro:
Some data points, all using CLI v11.22.0, consistent with the belief that the issue applies to node versions >= 10.16.0:

OS | Installation type | Node.js version | Proxy status
-- | -- | -- | --
Windows | standalone zip | 8.16.0 | OK
[...]

Issue #1335, opened 2019-07-01
Consider switching global-tunnel-ng to global-agent


gajus:
`global-tunnel-ng` used in this package [does not work in Node v11 and above](https://github.com/np-maintain/global-tunnel#global-tunnel).

If you require a global proxy, consider switching to https://github.com/gajus/global-agent.

The required setup is:

```js
import {
  createGlobalProxyAgent
} from 'global-agent';

createGlobalProxyAgent({
  environmentVariableNamespace: '',
  forceGlobalAgent: true
});

```

pdcastro:
Thanks for reporting this! We intend to support Node 12 (currently we recommend Node 8 or Node 10) and this issue sounds like a pre-requisite.

Issue #1580, opened 2020-01-22
Allow configurable proxy exceptions


pdcastro:
As reported in a [support forum thread](https://forums.balena.io/t/how-to-use-balena-deploy-to-build-on-device-running-development-image/58490), some users, being behind a corporate proxy, would like to use the `balena deploy --build` command to build images on local devices (keeping proprietary source code in premises) and upload the built images to balenaCloud. While the CLI allows the configuration of a proxy, the difficulty is that the corporate proxy server blocks the local traffic between the developer machine (laptop) and the local device (balenaEngine listening on TCP port 2375), probably because of the "unusual" port number.

Some proposed solutions:
* Boolean flag to exclude [all private IPv4 ranges](https://en.wikipedia.org/wiki/Private_network), and 'localhost'.
* Environment variable:
```
export no_proxy="localhost,127.0.0.1,192.168.168.*"
```
* Adding a list of addresses or host names to `~/.balenarc.yml`.
* Something else.

pdcastro:
**Known workaround:**
Instead of using `balena deploy --build` in a single step (that communicates with both balenaCloud and the local device), split the operation in 3 separate steps:

1. **Without** proxy configuration, use `balena build` **without specifying an app name,** so that the CLI does not need to contact balenaCloud, reaching the local device directly:
```sh
balena build -h <deviceIP> -p 2375 --deviceType raspberrypi3 --arch armv7hf
```
2. Still without proxy configuration, transfer the images from the device (balenaEngine) to your local machine (Docker):
```sh
ssh root@<deviceIP> -p 22222 balena-engine save <my_image> | docker load
```
3. Now, **with** proxy configuration, upload the image to balenaCloud:
```sh
BALENARC_PROXY=<proxy> balena deploy <my_app> <my_image>
```

pdcastro:
Another known workaround: run an intermediate proxy server on the development machine, and configure it to exclude private addresses from the proxy and "forward" other connections to the "upstream proxy". For example
* SquidMan on macOS: https://squidman.net/squidman/index.html
* Squid on Linux: http://www.squid-cache.org/,
  * Ubuntu: https://help.ubuntu.com/lts/serverguide/squid.html

pdcastro:
Apparently this feature would come "free of charge" if we adopted the [global-agent](https://www.npmjs.com/package/global-agent) npm package, which is one of the possible solutions for high-priority issue #1580.

pdcastro:
[...]

PR #1582, opened 2020-01-26
Fix proxy support and add proxy exclusion feature (Node.js >= 10.16.0)

pdcastro:
Resolves: #1579
Resolves: #1335 
Connects-to: #1580
Change-type: minor


README.md
```diff
@@ -64,19 +64,65 @@ $ balena login
...
+below. Eventually, all installation types will have support for it.
+
+OS | Installation type | BALENARC_NO_PROXY environment variable support
+-- | ----------------- | ----------------------------------------------
+Windows | standalone zip | Supported with CLI v11.24.0 and later
+Windows | native/GUI     | Not supported
+macOS   | standalone zip | Not supported
+macOS   | native/GUI     | Supported with CLI v11.24.0 and later
+Linux   | standalone zip | Not supported
+Any     | npm            | Supported with Node.js >= v10.16.0 and CLI >= v11.24.0
```
pdcastro: This "complex situation" comes from the fact that the proxy exclusion feature is implemented by the `global-agent` package, which only supports Node 10 and later (see more details in `app-common.ts`). Different installation types use / bundle different versions of Node.js, which a balena CI configuration (changes require pull requests to `resin-concourse`, which we will do in due course).

lib/actions/ssh.ts
```diff
@@ -227,9 +227,9 @@ export const ssh: CommandDefinition<
 		const applicationOrDevice =
 			params.applicationOrDevice_raw || params.applicationOrDevice;
 		const bash = await import('bash');
-		// TODO: Make this typed
-		const hasbin = require('hasbin');
-		const { getSubShellCommand } = await import('../utils/helpers');
+		const { getProxyConfig, getSubShellCommand, which } = await import(
+			'../utils/helpers'
+		);
```
pdcastro: The `hasbin` package was removed, and the implementation modified to use the `which` package that the CLI was already using elsewhere. The reasons were twofold:
* `hasbin` was simply not working in my Linux and macOS tests with Node 10 and Node 12, neither the master branch (before this PR) nor with/after this PR. I didn't really understand why, but I tested that `which` worked all right on Linux, macOS and Windows.
* I realised that this was the only occurrence of `hasbin` in the CLI and any dependencies, so `hasbin` could be removed from `package.json` and `npm-shrinkwrap.json`.

lib/app-common.ts
```diff
@@ -15,6 +15,36 @@
...
 
+class CliSettings {
+	public readonly settings: any;
+	constructor() {
+		// TODO figure out why the typescript compiler attempts to type-check
+		// the `balena-settings-client` module (and then fails with errors) if
+		// a straighforward `require('balena-settings-client')` statement is
+		// used here. It may even be a compiler bug, because `tsconfig.json`
+		// has a `"skipLibCheck": true` setting.
+		this.settings = require(['balena', 'settings', 'client'].join('-'));
```
pdcastro: This was baffling. See comment above. I only created the `CliSettings` class to isolate this oddity and avoid spreading these `require/join` lines in multiple places in the file.
* `require('balena-settings-client')` fails on type checking the `balena-settings-client` module, which is a CLI dependency and should not be type checked.
* `require(['balena', 'settings', 'client'].join('-'))` works, presumably because the typescript compiler does not attempt to execute the join at compilation time in order to obtain the module name.

lib/app-common.ts
```diff
@@ -53,37 +83,135 @@ function checkNodeVersion() {
...
+ * details are optional): http://username:password@domain.com:1234
+ *
+ * A proxy exclusion list in the NO_PROXY variable is only supported when
+ * `global-agent` is used, i.e. with Node.js v10.16.0 or later. The format
+ * is specified at: https://www.npmjs.com/package/global-agent#exclude-urls
+ * Patterns are matched with matcher: https://www.npmjs.com/package/matcher
+ * 'localhost' and '127.0.0.1' are always excluded. If NO_PROXY is not defined,
+ * this function adds exclusion patterns for all private IPv4 address ranges.
+ */
+async function setupGlobalHttpProxy(settings: CliSettings) {
```
pdcastro: See comment above. `global-agent` fixes the Node.js support for Node 10 and Node 12 (issue #1579), **and** also provides the proxy exclusion feature requested in issue #1580.  This is why, by the way, I didn't provide separate commits for fixing the issue and adding the feature -- it's hard to separate them.

lib/utils/helpers.ts
```diff
@@ -379,20 +379,87 @@ export async function workaroundWindowsDnsIssue(ipOrHostname: string) {
...
+	username?: string;
+	password?: string;
+	proxyAuth?: string;
+}
+
+/**
+ * Check whether a proxy has been configured (whether global-tunnel-ng or
+ * global-agent) and if so, return a ProxyConfig object.
+ */
+export function getProxyConfig(): ProxyConfig | undefined {
```
pdcastro: I've written some test code for this function.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,11 +1,24 @@
...
-See: <url> <!-- Refer to any external resource, like a PR, document or discussion -->
-Depends-on: <url> <!-- This change depends on a PR to get merged/deployed first -->
-Change-type: major|minor|patch <!-- The change type of this PR -->
+Resolves: # <!-- Refer an issue of this repository that this PR fixes -->  
+Change-type: major|minor|patch <!-- See https://semver.org/ -->  
+Depends-on: <url> <!-- This change depends on a PR to get merged/deployed first -->  
+See: <url> <!-- Refer to any external resource, like a PR, document or discussion -->  
 
 ---
-##### Contributor checklist
```
pdcastro: This Contributor Checklist was put together years ago. I found it was a bit odd, especially because a "checklist" is usually a list of things that someone should do (before submitting the PR), usually starting with imperative verbs:
```
[ ] Read something
[ ] Write test code
[ ] Run something
```
But _"Introduces security considerations"_ is not something someone should _do_. So I replaced the checklist with some "guidance paragraphs". Suggestions are welcome.

This change is incidental and piggybacked on this PR, but it's in a separate commit. If it becomes contentious or something like that, I'll move it to a separate PR.

INSTALL.md
```diff
@@ -141,11 +141,13 @@ especially if you're using a user-managed node install such as [nvm](https://git
...
-    for Ubuntu, for example).
-    Check the [README](https://github.com/balena-io/balena-cli/blob/master/README.md) file
-    for proxy configuration instructions.
+  * The [`proxytunnel`](http://proxytunnel.sourceforge.net/) package (command-line tool) is needed
+    for the `balena ssh` command to work behind a proxy. It is available for Linux distributions
+    like Ubuntu/Debian (`apt install proxytunnel`), and for macOS through
+    [Homebrew](https://brew.sh/). Windows support is limited to the Windows Subsystem for Linux
+    (e.g., by installing Ubuntu through the Microsoft App Store). Check the
+    [README](https://github.com/balena-io/balena-cli/blob/master/README.md) file for proxy
+    configuration instructions.
```
pdcastro: @richbayliss or @wrboyce, do you know if `proxytunnel` is only meant to work with openBalena? As documented by Rich on a forum post: https://forums.balena.io/t/howto-ssh-into-host-device/5312

I was testing `balena ssh` in the CLI's ***master branch*** (before this PR) with balenaCloud (not openBalena) and a proxy server (squid), but never managed to get it working when `proxytunnel` is used. Tested both Linux (Debian) and macOS, same result:
```
$ DEBUG=1 ./bin/balena ssh b1878c618483983448f6b863d30df408 -v
[debug] original argv0="node" argv=[/Users/paulo/.nvm/versions/node/v10.15.3/bin/node,/Users/paulo/balena/balena-io/balena-cli/bin/balena-dev,ssh,b1878c618483983448f6b863d30df408,-v] length=5
[Debug]   Fetching application by name b1878c618483983448f6b863d30df408 (string)
[Debug]   Application not found
[Debug]   Fetching device by UUID b1878c618483983448f6b863d30df408 (string)
vpncmd: ssh -vvv -t -o LogLevel=ERROR -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ProxyCommand=proxytunnel\ --proxy\=127\\\.0\\\.0\\\.1\\\:8888\ --dest\=\\\%h\\\:\\\%p  gh_paulo_castro@ssh.balena-devices.com host b1878c618483983448f6b863d30df408
OpenSSH_7.9p1, LibreSSL 2.7.3
debug1: Reading configuration data /Users/paulo/.ssh/config
debug1: /Users/paulo/.ssh/config line 107: Applying options for *
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 48: Applying options for *
debug1: Executing proxy command: exec proxytunnel --proxy=127\\.0\\.0\\.1\\:8888 --dest=\\ssh.balena-devices.com\\:\\22
debug1: identity file /Users/paulo/.ssh/id_rsa type 0
debug1: identity file /Users/paulo/.ssh/id_rsa-cert type -1
debug1: identity file /Users/paulo/.ssh/id_dsa type -1
debug1: identity file /Users/paulo/.ssh/id_dsa-cert type -1
debug1: identity file /Users/paulo/.ssh/id_ecdsa type -1
debug1: identity file /Users/paulo/.ssh/id_ecdsa-cert type -1
debug1: identity file /Users/paulo/.ssh/id_ed25519 type -1
debug1: identity file /Users/paulo/.ssh/id_ed25519-cert type -1
debug1: identity file /Users/paulo/.ssh/id_xmss type -1
debug1: identity file /Users/paulo/.ssh/id_xmss-cert type -1
debug1: Local version string SSH-2.0-OpenSSH_7.9
Via 127.0.0.1:8888 -> ssh.balena-devices.com:22
HTTP return code: 403 Forbidden
ssh_exchange_identification: Connection closed by remote host
```

.github/PULL_REQUEST_TEMPLATE.md
srlowe: Yeah I agree.  I wonder if that was introduced in response to something bad that happened, or just for good measure.

pdcastro:
> I wonder if we should auto-exclude private ip6 addresses too? I.e. private ( fc00::) and link local (fe80::)

Yeah, I had thought of IPv6 too, but I looked at that the documentation of both `global-tunnel-ng` and `global-agent` and neither had any comments or examples regarding IPv6. The thing is that the NO_PROXY exclusion expects a colon as `hostname:port` separator, and an an IPv6 address like `fe80::1` looks like a hostname of `fe80:` and a port number of `1`. Maybe `global-agent` understands square bracket notation like `[::1]:8080` -- I would need to test, and I wish they'd have documented it!

pdcastro:
> I wonder if we should auto-exclude private ip6 addresses too? I.e. private ( fc00::) and link local (fe80::)

> I would need to test

I have now tested it -- on macOS only, not an extensive test, but what I found is that `global-agent` does not appear to even try to proxy IPv6 requests, even without any IPv6 exclusion pattern. In addition, I have tested the push / build / deploy commands with IPv6 (without proxy) and found that none work with IPv6, so it's not like proxy is the main problem there. :-) I have created CLI issue #1585 for IPv6 support to be added. For these reasons, plus the fact that there is no clarity on whether the square bracket notation should or should not be used to differentiate IPv6 addresses from host names in the proxy exclusion list (when/if they add proxy support for IPv6), I decided not to add IPv6 exclusion patterns at this time.

Issue #1502, opened 2019-11-07
Debug mode still active with env var `DEBUG=0` 


srlowe:
`DEBUG=1` for debugging implies that it can be disabled with `DEBUG=0`.  However debugging is still active, and you need to `unset DEBUG` to deactivate.

- **balena CLI version:** e.g. 11.17.3
- **Operating system version:** Ubuntu 19.10
- **32/64 bit OS and processor:** 64
- **Install method:** npm
- **If npm install, Node.js and npm version:** Node v12.13.0, npm v6.12.0

pdcastro:
Perhaps the values 0, "no" and "false" (case insensitive) could be treated as disabling debug (in addition to unsetting the variable), and any other non-empty value treated as enabling debug. (If an env var can hold the empty string (can it?), it should probably be considered as disabling debug as well.)

srlowe:
Linux env vars can be empty, but not windows.

codewithcheese:
Hi @srlowe @pdcastro. I'm interested to resolve this issue as my first contribution to `balena-cli`.

What do you think about this approach?

Running a `find` shows that `DEBUG` truthy evaluation is done in a bunch of files. I would create a new function called `isDebugEnabled` in `lib/utils/helpers.ts` that would return `false` for `undefined` `"0"`, `"no"`, `"false"`, or `""` and `true` for all other values.

I would then replace all usages of `DEBUG` truthy evaluation with the function call.

Since the changes touches many file, I'm concerned how I would test. Any tips?

![Selection_004](https://user-images.githubusercontent.com/989237/73549693-6a6c2480-4486-11ea-8e15-3bffa77a6cba.png)

codewithcheese:
Now, thinking supporting other "falsey" values only serves to add confusion. For example do we also support capitalization? `export DEBUG=NO` `export DEBUG=YES`?

It seems reasonable that someone would want to switch from DEBUG=1 to DEBUG=0. I doubt it's reasonable to support people who wish to freestyle with truthiness :)

PR #1586, opened 2020-01-31
Debug mode can now be disabled with DEBUG=0 env var

codewithcheese:
Added `isDebugModeEnabled` to evaluate if debug mode is enabled, debug mode can now be disabled with DEBUG=0
Updated all conditional usages of `process.env.DEBUG` to use `isDebugModeEnabled`

Resolves: #1502
Change-type: patch
Signed-off-by: Thomas Manning <thomasm@balena.io>

### DIFF
--- a/automation/run.ts
+++ b/automation/run.ts
@@ -28,6 +28,13 @@ import {
 } from './deploy-bin';
 import { fixPathForMsys, ROOT, runUnderMsys } from './utils';
 
+// DEBUG set to falsy for negative values else is truthy
+process.env.DEBUG = ['0', 'no', 'false', '', undefined].includes(
+	process.env.DEBUG?.toLowerCase(),
+)
+	? ''
+	: '1';
+
 function exitWithError(error: Error | string): never {
 	console.error(`Error: ${error}`);
 	process.exit(1);
--- a/lib/app.ts
+++ b/lib/app.ts
@@ -23,6 +23,13 @@ export async function run(
 	cliArgs = process.argv,
 	options: import('./preparser').AppOptions = {},
 ) {
+	// DEBUG set to falsy for negative values else is truthy
+	process.env.DEBUG = ['0', 'no', 'false', '', undefined].includes(
+		process.env.DEBUG?.toLowerCase(),
+	)
+		? ''
+		: '1';
+
 	// The 'pkgExec' special/internal command provides a Node.js interpreter
 	// for use of the standalone zip package. See pkgExec function.
 	if (cliArgs.length > 3 && cliArgs[2] === 'pkgExec') {


lib/utils/debug-mode.ts
```diff
@@ -0,0 +1,7 @@
+export function isDebugModeEnabled() {
+	return (
+		process.env.DEBUG !== '0' &&
+		process.env.DEBUG !== '' &&
+		process.env.DEBUG !== undefined
```
pdcastro: As for the logic itself, my two cents would be:
```
	return (
		!!process.env.DEBUG &&
		!['0', 'no', 'false'].includes(process.env.DEBUG.toLowerCase())
	);
```
I take note of your good reasoning in the linked issue:
> Now, i'm thinking supporting other "falsey" values only serves to add confusion. For example do we also support capitalization? export DEBUG=NO export DEBUG=YES?

I get it, but I think it doesn't align with "reducing friction" for users. We would not document / recommend anything other than DEBUG=1 (it would add confusion), but if someone out there in a hurry was to execute `export DEBUG=no` in a bash shell, your version of `isDebugModeEnabled()` would return `true`. Should it really? 

Re capitalization (good point!), `toLowerCase()` covers it.

codewithcheese: Yes, I see you point. One more alternative would be to stick to the letter of the docs and ONLY `DEBUG=1` enables debug mode. Presumably developers learn about the `DEBUG` option from reading the docs at which point they learn that the acceptable value is `1` not `true` or `yes`. Then any other value besides `1` would NOT enable debug mode.

lib/utils/debug-mode.ts
```diff
@@ -0,0 +1,7 @@
+export function isDebugModeEnabled() {
```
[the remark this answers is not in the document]
codewithcheese: @pdcastro thanks for the thorough feedback. Yes, I tried to place the function in `utils/helpers` but that broke the build somehow, so I moved it to its own file. I don't understand the build process well enough to debug exactly why. I had to presume it was because of side effects as you mentioned. 

How about `isDebugMode()`?  It's not obvious from the naming of `evalDebug` and `DEBUG` that the value is a `boolean`. `isDebug` is indeed shorter but does saving the 4 characters of `mode` justify the loss of context?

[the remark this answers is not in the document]
srlowe: @pdcastro @codewithcheese Bare in mind that it was not a user that opened this issue (only me being pedantic).  So I think we should solve it as conveniently as possible.  If we can stick with the global `if (process.env.DEBUG)` then I think that's preferable, since we don't have to change any code, and no need for extra imports into modules.  With that in mind, personally I prefer Paulo's approach of the env var "cleaning" function `evalDebug`, since it only needs to be called once, and doesn't change anything else.  I don't think the name of this function is very important, since it's not used anywhere else.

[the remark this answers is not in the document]
codewithcheese: I can see you both share a concern about adding yet another module and changing the usage in so many places. I have updated the PR with an inline assignment of `process.env.DEBUG` at the entrypoints.

lib/utils/debug-mode.ts
codewithcheese: I have updated the PR to assign DEBUG to falsy for negative values.

automation/build-bin.ts
```diff
@@ -41,6 +41,13 @@ export const packageJSON = loadPackageJson();
 export const version = 'v' + packageJSON.version;
 const arch = process.arch;
 
+// DEBUG set to falsy for negative values else is truthy
+process.env.DEBUG = ['0', 'no', 'false', '', undefined].includes(
+	process.env.DEBUG?.toLowerCase(),
+)
+	? ''
+	: '1';
```
pdcastro: `automation/run.ts` executes before `automation/build-bin.ts`, so `run.ts` is closer to being an entrypoint. It does created subprocesses, but if I recall correctly, the environment variables are inherited by the child processes.
Another separate entrypoint (executed in a `package.json` script) is `automation/capitanodoc/index.ts`, although I haven't seen occurrences of `process.env.DEBUG` there.

pdcastro:
One more suggestion: given that the CHANGELOG file will be updated with the commit's first line only, it might look clearer to users if it read:
* Something like: `Debug mode can now be disabled with DEBUG=0 env var`
* Instead of: `Debug mode can now be disabled with negative strings`

Only because, out of context, the meaning of "negative strings" may not be clear to many people.

automation/build-bin.ts
codewithcheese: Thanks moved to `automation/run.ts`

PR #1589, opened 2020-02-03
Automatically lint files on commit

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,11 @@
+{
+	"*.coffee": [
+		"resin-lint"
+	],
+	"*.ts": [
+		"resin-lint --typescript --fix"
+	],
+	"tests/**/*.ts": [
+		"resin-lint --typescript --tests"
+	]
+}
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "node automation/check-npm-version.js && node automation/check-doc.js"
+      "pre-commit": "node automation/check-npm-version.js && lint-staged && node automation/check-doc.js"
     }
   },
   "oclif": {
@@ -144,6 +144,7 @@
     "gulp-shell": "^0.5.2",
     "husky": "^4.2.3",
     "intercept-stdout": "^0.1.2",
+    "lint-staged": "^10.0.8",
     "mocha": "^6.2.2",
     "mock-require": "^3.0.3",
     "nock": "^11.9.1",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -648,6 +648,15 @@
       "resolved": "https://registry.npmjs.org/@resin.io/valid-email/-/valid-email-0.1.0.tgz",
       "integrity": "sha1-DnUwmoQ8AUqAqhSC+WmQYvL6UV0="
     },
+    "@samverschueren/stream-to-observable": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.0.tgz",
+      "integrity": "sha512-MI4Xx6LHs4Webyvi6EbspgyAb4D2Q2VtnCQ1blOJcoLS6mVa8lNN2rkIy1CVxfTUpoyIbCTkXES1rLXztFD1lg==",
+      "dev": true,
+      "requires": {
+        "any-observable": "^0.3.0"
+      }
+    },
     "@sentry/apm": {
       "version": "5.13.2",
       "resolved": "https://registry.npmjs.org/@sentry/apm/-/apm-5.13.2.tgz",
@@ -1447,6 +1456,12 @@
       "version": "0.3.2",
       "resolved": "https://registry.npmjs.org/ansicolors/-/ansicolors-0.3.2.tgz",
       "integrity": "sha1-ZlWX3oap/+Oqm/vmyuXG6kJrSXk="
+    },
+    "any-observable": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/any-observable/-/any-observable-0.3.0.tgz",
+      "integrity": "sha512-/FQM1EDkTsf63Ub2C6O7GuYFDsSXUwsaZDurV0np41ocwq0jthUAYCmhBX9f+KwlaCgIuWyr/4WlUQUBfKfZog==",
+      "dev": true
     },
     "any-promise": {
       "version": "1.3.0",
@@ -3966,6 +3981,12 @@
         "mimic-response": "^1.0.0"
       }
     },
+    "dedent": {
+      "version": "0.7.0",
+      "resolved": "https://registry.npmjs.org/dedent/-/dedent-0.7.0.tgz",
+      "integrity": "sha1-JJXduvbrh0q7Dhvp3yLS5aVEMmw=",
+      "dev": true
+    },
     "deep-eql": {
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/deep-eql/-/deep-eql-3.0.1.tgz",
@@ -4788,6 +4809,12 @@
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/ejs/-/ejs-3.0.1.tgz",
       "integrity": "sha512-cuIMtJwxvzumSAkqaaoGY/L6Fc/t6YvoP9/VIaK0V/CyqKLEQ8sqODmYfy/cjXEdZ9+OOL8TecbJu+1RsofGDw=="
+    },
+    "elegant-spinner": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/elegant-spinner/-/elegant-spinner-1.0.1.tgz",
+      "integrity": "sha1-2wQ1IcldfjA/2PNFvtwzSc+wcp4=",
+      "dev": true
     },
     "emoji-regex": {
       "version": "7.0.3",
@@ -6817,6 +6844,12 @@
       "integrity": "sha1-6td0q+5y4gQJQzoGY2YCPdaIekE=",
       "dev": true
     },
+    "get-own-enumerable-property-symbols": {
+      "version": "3.0.2",
+      "resolved": "https://registry.npmjs.org/get-own-enumerable-property-symbols/-/get-own-enumerable-property-symbols-3.0.2.tgz",
+      "integrity": "sha512-I0UBV/XOz1XkIJHEUDMZAbzCThU/H8DxmSfmdGcKPnVhu2VfFqr34jr9777IyaTYvxjedWhqVIilEDsCdP5G6g==",
+      "dev": true
+    },
     "get-port": {
       "version": "3.2.0",
       "resolved": "https://registry.npmjs.org/get-port/-/get-port-3.2.0.tgz",
@@ -7679,6 +7712,12 @@
         }
       }
     },
+    "human-signals": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/human-signals/-/human-signals-1.1.1.tgz",
+      "integrity": "sha512-SEQu7vl8KjNL2eoGBLF3+wAjpsNfA9XMlXAYj/3EdaNfAlxKthD1xjEQfGOUhllCGGJVNY34bRr6lPINhNjyZw==",
+      "dev": true
+    },
     "humanize": {
       "version": "0.0.9",
       "resolved": "https://registry.npmjs.org/humanize/-/humanize-0.0.9.tgz",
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -55,7 +55,7 @@
     "catch-uncommitted": "ts-node --type-check -P automation/tsconfig.json automation/run.ts catch-uncommitted",
     "ci": "npm run test && catch-uncommitted",
     "watch": "gulp watch",
-    "prettify": "prettier --write \"{lib,tests,automation,typings}/**/*.[tj]s\" --config ./node_modules/resin-lint/config/.prettierrc",
+    "prettify": "resin-lint --typescript --fix automation/ lib/ typings/ tests/ && prettier --write \"{lib,tests,automation,typings}/**/*.js\" --config ./node_modules/resin-lint/config/.prettierrc",
     "lint": "resin-lint lib/ tests/ && resin-lint --typescript automation/ lib/ typings/ tests/",
```
pdcastro: There were two pre-existing npm scripts here: `prettify` and `lint`, which used to run `prettier` and `resin-lint` respectively. The pre-existing `lint` script was left unchanged (notably without the new `--fix` option), and the `prettify` script was modified to run both `resin-lint` and `prettier`. 

Note also the `build:src` npm script, which calls `npm run prettify && npm run lint`.

Wouldn't it be better (clearer?) to leave the `prettify` script unchanged, and modify the `lint` script only? And then, as `resin-lint` itself calls prettier (?), perhaps `build:src` does not need to call `npm run prettify` anymore (enough to call `npm run lint`).

.lintstagedrc
```diff
@@ -0,0 +1,9 @@
+{
+	"*.ts": [
+		"resin-lint --typescript --fix",
+		"resin-lint --typescript"
```
pdcastro: I guess `resin-lint` is called twice because the first call attempts a fix, and the second call verifies that the result was successful. Sounds right?

Page-: Yes, and the `--fix` just fixes what it can, it's possible there are things that can not be automatically fixed (eg shadowed variables)

Page-: I've just checked and `--fix` will also report any it cannot auto-fix so I'll remove the redundant call

package.json
Page-: The prettify script is to auto-format the code so it can be useful to call it explicitly (eg on an update to automatically apply everywhere), I can remove it as a separate call entirely and just embed it in the lint script instead if you want

As an aside I also never put linting as part of a normal test run (only a CI/prepublish test run) because it adds a lot of time to the iteration cycle unnecessarily (I see you actually have a `test:fast` to handle that, but the default is the slow version which is what any new contributors will likely hit due to `npm test`)

pdcastro: I ran some tests with `resin-lint` and `lint-staged` to understand them better. I was also revisiting the npm scripts, and I have some thoughts "about the whole thing" :-) (not just the specific changes proposed in this PR so far).

I understand the concept of leaving "comprehensive checks" for the CI, and using `lint-staged` for quickly doing what is most likely the right thing on a subset of the repo files (the staged ones), thus saving dev time. Under this concept, `npm test` might not necessarily involve running a full slow rebuild of all source files (currently it does, as a result of a `pretest` npm script: `"pretest": "npm run build"`). But one "serious enough issue" I came across is that `coffeelint` may print important warnings that do not cause the build to fail (so the CI tests would pass and the PR would be allowed to be merged), and _if the developer had seen the warnings,_ they could have done something about it. For example, if I add a reference to an inexistent variable `fooz` and manually run `npm run lint` (which is one of the steps called by `npm run build`), I get on my laptop terminal, with high-contrast yellow colour:

```
$ npm run lint
...
> resin-lint lib/ tests/ && resin-lint --typescript automation/ lib/ typings/ tests/

  ⚡ lib/actions/app.coffee
     ⚡ #59: Undefined identifier "fooz".
  ✓ lib/actions/auth.coffee
```

The `coffeelint` process returns a success exit code because this is "just a warning," but it was actually a bug I just introduced: the variable `fooz` indeed does not exist. Yet, under current arrangements, the build will pass, and unless the developer manually runs `npm run build` (or `npm run lint`), they won't see the warning. On this point:
* I was searching for a way (configuration, option) to cause `coffeelint` to exit with a failure error code for _any warnings,_ but there isn't such an option. It is possible to configure a rule's "warn level", e.g. make the "undefined identifier" rule become an error instead of a warning. The configuration currently comes from the `resin-lint` repo, and if we changed it globally, it would affect other repos. We could override the config in `balena-cli`, but:
* What about all other non-error rules? (Other than "undefined identifier".) Would we set all/most rules to error level instead of warn level? Maybe, but until recently, the balena-cli repo had some coffeelint warnings that were "OK", not bugs -- we got rid of some warnings by code refactoring "only to please coffeelint," and in other cases we added linter annotations (e.g. `# coffeelint: disable-next-line=check_scope ("Variable is assigned to but never read")`).  This is a possible solution, but I have another one in mind (keep reading). :-)

By the way, I tested adding `"*.coffee"` to the `.lintstagedrc` file:
```
	"*.coffee": [
		"resin-lint",
	],
```

But the problem there is that:
* `lint-staged` currently hides the output of the commands it runs (for as long as the command returns a success code).
* `lint-staged` only executes the lint commands for the staged files (its _raison d'être..._), but the changes the developer made _to the staged files_ might cause "serious warnings" _in other files that weren't touched_ (dependencies between files), and because it's warnings rather than errors, the CI will not fail and will allow the PR to be merged.

Thus, while it would be good if `lint-staged` showed rather hid the output of commands, it would still not be the end of the story. (For the record, as far as personal preference goes, I dislike that `lint-staged` hides the output of prettier, tslint, coffeelint, etc. If a pre-commit hook is automatically processing and modifying files that I am about to commit, at least I'd like to know which files were modified; I call it basic courtesy.) :-)

**I agree that it is a CI / build problem if a PR is allowed to be merged without bringing "serious warnings" to the attention of developers; this can't be simply blamed on `lint-staged`.** But the fact that `lint-staged` hides linter output doesn't help.

Finally... I do have a suggestion / "solution" in mind. A compromise. I propose that:
* We modify `npm test` to skip prettifying and linting (i.e. replace `"pretest": "npm run build"` with `"pretest": "npm run build:fast"`). We could also modify `npm test:fast` to skip building 'lib' source files altogether, in case the developer is only modifying the test files themselves.
* We require (in the "normal" development flow) that the developer executes `npm run build` on their laptop ***once*** before committing. The way I have mind to detect this is by checking some file timestamps:
  * We add a command like "touch last-build" to the `npm run build` steps, where "touch" is the [Unix touch command](http://man7.org/linux/man-pages/man1/touch.1.html) or, for Windows compatibility, the npm [touch package](https://www.npmjs.com/package/touch). (Maybe we can use the timestamp of the `build` folder; this needs testing.)
  * We write a small script to be executed by `lint-staged` (added to `.lintstagedrc`) with the list of staged files as arguments. The script compares the timestamp of the `last-build` file with the timestamps of the staged files. If the stages files are newer, then the script exits with a failure code and error message, which `lint-staged` prints: "`please run 'npm run build' before committing`". When the developer does that, usually only once before committing, they will see the full output of coffeelint, tslint and prettier, being aware of any warnings and knowing which files have been modified before committing (if nothing else, `git status` or `git diff` would show exactly what changed _before_ committing).

Thoughts? :-)   (cc: @srlowe)

pdcastro: As things stand, another reason to require developers to run `npm run build` (only once, at least once) on their laptop, before committing (pre-commit hook), is that it updates / auto generates the `cli.markdown` file that contains the CLI command reference documentation. When developers forget to run `npm run build`, the CI build fails on a check for "uncommitted changes" (the CI itself runs `npm run build`, which causes `cli.markdown` to be modified on the build worker, and in turn results in `git status` reporting uncommitted changes). Given how slow the CI builds are, ultimately it saves time to require developers to run `npm run build` ***once*** before committing.

.lintstagedrc
```diff
@@ -0,0 +1,8 @@
+{
+	"*.ts": [
```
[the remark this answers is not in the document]
Page-: Yes there's overlap because all the autoformatting is handled by the first check and only test specific changes by the test specific matching, ie all .ts files should lint correctly and then we run an extra check specifically for tests.  We could split it up so we duplicate the standard linting checks/options in both places but it would make this more complex/less understandable imo and more prone to copy/paste errors where only one is update, for the sake of saving a small amount of time on a commit.

For the time saved I checked and with resin-lint 3.2.0 it would be ~1s per file saved which is actually fairly significant so I optimized resin-lint somewhat and with 3.2.2 it'll be down to ~.5s saved (and I'll probably find some more savings before this)

package.json
Page-: > coffeelint may print important warnings that do not cause the build to fail (so the CI tests would pass and the PR would be allowed to be merged), and *if the developer had seen the warnings*

This is an issue right now where it's perfectly possible to not even run the lint checks locally let alone notice the issue so this PR doesn't fundamentally change that at all. The best bet for having them actually be seen is to add a prepush hook that runs the full lint because that will maximize the odds that the developer will see it, since it will have appeared at least once in their workflow (it does come at the cost that a push is slower, so depending on how much you check then that can add up)

> we got rid of some warnings by code refactoring "only to please coffeelint,"

Yep, I've come across non-warnings in coffeescript because the tooling is just not that good, realistically I'd say the best approach is just to convert the files to typescript to gain all the benefits and stop messing about with deprecated code (I know it's easier said than done but I've found that with every file converted future ones get easier)

> the changes the developer made to the staged files might cause "serious warnings" in other files that weren't touched (dependencies between files), and because it's warnings rather than errors

the cross file issues that might be introduced would only be relevant for typescript files (coffeescript files are all treated independently) at which point it gets handled by the typescript checker and they'd all be errors so that's not an issue in practice

> We require (in the "normal" development flow) that the developer executes npm run build on their laptop once before committing. 

I would be against this because it adds yet more friction to contributing which is what I'm trying to reduce, also if a person never bothers to `npm install` in the first place the script to check won't even be run. Also I fully expect all sorts of issues where it complains despite you having actually run it due to timestamp checking weirdness, and realistically it's all in an effort to support coffeescript which we should be aiming to get rid of

pdcastro: >> coffeelint may print important warnings that do not cause the build to fail

> This is an issue right now where it's perfectly possible to not even run the lint checks locally let alone notice the issue so this PR doesn't fundamentally change that at all.

True, but I'm thinking there's a better solution, which is checking that `npm run build` was executed once, before committing, instead of having `lint-staged` lint the staged files only. This is because `npm run build` will:
* Update the CLI documentation in the `cli.markdown` file (without which the build will fail with an "uncommitted changes" error, if the PR modified any CLI command flags. The fix for the build failure is to run `npm run build` on the laptop -- currently there isn't a way around it).
* It would show coffeelint warnings. For all its shortcomings, it is still a "software quality" tool we have for the substantial amount of coffeescript code remaining in the CLI. (After we convert all coffeescript code to typescript, this argument may go away, but we are not there yet.)
* It would show, rather than hide, prettier and tslint output.

>> We require (in the "normal" development flow) that the developer executes `npm run build` on their laptop once before committing.

> I would be against this because it adds yet more friction to contributing which is what I'm trying to reduce, also if a person never bothers to npm install in the first place the script to check won't even be run.

In practice, as things stand, running `npm run build` is already a requirement for any PR that modifies CLI command options -- but instead of the failure being timely pointed out pre-commit, it's pointed out 15 minutes later, when the build fails (`cli.markdown`, "uncommitted changes").

Also, for PRs more complex updating documentation, I would expect contributors to run the CLI at least on their own machine, to test their changes, and this requires running `npm run build` anyway, at least once. So I am not convinced that an automated check that `npm run buid` was executed is adding friction. In many cases it would actually reduce friction, by helping errors and warnings being caught sooner rather than later.

srlowe: I don't have very strong opinions on a lot of this, but if I understand the discussion correctly, then I think the things I'm not too keen on with the proposed approach are:
 - Potential for hiding warnings.
 - Linting a subset of the code outside of the context of the rest of the codebase.
 - Committing automatically made changes which the developer has not seen.

Page-: > Potential for hiding warnings.

Definitely a real issue that should be fixed, the only thing is I would say the same potential is there right now so it doesn't materially change it (best way to fix that imo is to finish the typescript conversion)

> Linting a subset of the code outside of the context of the rest of the codebase.

This is in addition to the existing lint checks to save developer time/reduce the barrier to entry, the final say on linting should always lie in the CI checks

> Committing automatically made changes which the developer has not seen.

I was concerned about this when we first started adding it around the place as well so all I can really say is that in practice it hasn't been an issue and has saved a great deal of frustration

Page-:
> btw, I learned to like prettier!

Yep, I was not a fan either but over time it really grows on you, and in fact I had the same experience with this auto-linting (actually the auto-linting probably helped prettier grow on me even quicker)

Page-:
I've split out the non-contentious parts to https://github.com/balena-io/balena-cli/pull/1595 so they can be merged whilst we deal with the on-commit parts

PR #1594, opened 2020-02-05
Convert lib/actions/app to typescript

Page-:
Change-type: patch


lib/actions/app.ts
```diff
@@ -0,0 +1,227 @@
...
+			})
+			.then(
+				(
+					applications: Array<
+						Application & { device_count?: number; online_devices?: number }
+					>,
+				) => {
+					applications.forEach(application => {
+						application.device_count = _.size(application.owns__device);
+						application.online_devices = _.sumBy(application.owns__device, d =>
```
Page-: I switched this from a `.filter(...).length` to this `.sumBy(...)` to avoid creating an intermediary array unnecessarily, it probably doesn't matter for the CLI specifically but I have seen real performance hits from it so as a best practice I converted it

lib/actions/app.ts
```diff
@@ -0,0 +1,227 @@
...
+
+		return balena.models.application
+			.getAll({
+				$select: ['id', 'app_name', 'device_type'],
+				$expand: { owns__device: { $select: 'is_online' } },
+			})
+			.then(
+				(
+					applications: Array<
+						Application & { device_count?: number; online_devices?: number }
```
Page-: This is to allow us to add the additional properties

lib/actions/app.ts
```diff
@@ -0,0 +1,227 @@
...
+`,
+	options: [commandOptions.yes],
+	permission: 'user',
+	async action(params, options, done) {
+		const balena = (await import('balena-sdk')).fromSharedOptions();
+		const patterns = await import('../utils/patterns');
+
+		return patterns
+			.confirm(
+				options.yes ?? false,
```
pdcastro: ooh [Nullish Coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)! :-) I think it's the first usage in the CLI repo. In this case, though, I think `!!options.yes` would have worked as well.

Page-:
tested

PR #1599, opened 2020-02-05
Switch from opn to its new name of open

Page-:
Info: https://github.com/sindresorhus/open/blob/v7.0.2/readme.md
>Note: The original open package was previously deprecated in favor of this package, and we got the name, so this package is now named open instead of opn

Changes: https://github.com/sindresorhus/open/releases
Notable stuff looks to be a bunch of windows fixes, WSL2 support, and typescript typings


Page-:
Tested on WSL1 and powershell, needs a retest on osx though

pdcastro:
> needs a retest on osx though

Successfully retested on macOS ✔

PR #1601, opened 2020-02-05
Convert lib/actions/auth to typescript

Page-:
Depends on #1600


lib/actions/auth.ts
```diff
@@ -0,0 +1,209 @@
...
+									'Token authentication failed',
+								);
+							}
+						}),
+					);
+			} else if (loginOptions.credentials) {
+				return patterns.authenticate(loginOptions);
+			} else if (loginOptions.web) {
+				console.info('Connecting to the web dashboard');
+				return auth.login().return();
```
pdcastro: The original code was `return auth.login()` -- I assume that `.return()` was needed to satisfy type checking. Nothing wrong, but given that this line relies on `login()` returning a Bluebird promise, I suggest either:
* Explicitly add a Bluebird return type to the `login()` function in `lib/auth/index.ts`, so at least it is clearer that it returns a Bluebird promise (instead of `async/await`), or
* Alternatively, replacing this line with: (I think)
```
    await auth.login();
    return;
```

lib/actions/auth.ts
```diff
@@ -0,0 +1,209 @@
...
+			}
+
+			return patterns.askLoginType().then(loginType => {
+				if (loginType === 'register') {
+					const signupUrl = 'https://dashboard.balena-cloud.com/signup';
+					require('opn')(signupUrl, { wait: false });
```
pdcastro: hmm maybe it's OK, but the original code here was:
```
require('open')(signupUrl, { url: true, wait: false })
```
note 'opn' _vs. 'open', and the `'url'` attribute.

Page-: Yeah I haven't rebased this on my other changes yet - this will definitely need updating (and can also become `import()` with the added types from `open`)

lib/actions/auth.ts
Page-: I'll look into also making use of async/await to simplify this

Page-: done

lib/actions/auth.ts
Page-: done

Page-:
Local testing looked good

Issue #1596, opened 2020-02-05
balena push in Raspberry 3 tutorial fails: "could not load any valid magic files"


saintaardvark:
- **balena CLI version:** e.g. 11.25.2
- **Operating system version:** Fedora 31 (also tested with Debian 9.1)
- **32/64 bit OS and processor:** 64-bit OS and processor
- **Install method:** zip
- **If npm install, Node.js and npm version:** N/A

---
Hi there -- as reported in [this forum post](https://forums.balena.io/t/fail-to-push-application-to-balena-cloud/61043/2), I'm having problems pushing an application with `balena push`; using version 11.25.2, it fails with the error "could not load any valid magic files".  This does *not* happen with version 11.21.5 (an older version of the CLI I picked at random).  Details follow.

Following [the Raspberry Pi 3 tutorial](https://www.balena.io/docs/learn/getting-started/raspberrypi3/nodejs/), I created an app named `Test`; I downloaded [the balena-node-hello-world-master repo](https://github.com/balenalabs/simple-server-node/archive/master.zip) and unzipped it; and ran `balena push Test` from within the new directory.  Here's what I saw (running again with `DEBUG=1` exported):

```
12:08 $ balena version
[debug] original argv0="balena" argv=[/home/aardvark/balena-cli-11.25.2/balena,/snapshot/versioned-source/bin/balena,version] length=3
[debug] new argv=[/home/aardvark/balena-cli-11.25.2/balena,/snapshot/versioned-source/bin/balena,version] length=3
11.25.2
✔ ~/dev/balena-node-hello-world-master 
12:12 $ balena apps
[debug] original argv0="balena" argv=[/home/aardvark/balena-cli-11.25.2/balena,/snapshot/versioned-source/bin/balena,apps] length=3
ID      APP NAME DEVICE TYPE  ONLINE DEVICES DEVICE COUNT
1578870 Test     raspberrypi3 1              1
✔ ~/dev/balena-node-hello-world-master 
12:12 $ balena push Test
[debug] original argv0="balena" argv=[/home/aardvark/balena-cli-11.25.2/balena,/snapshot/versioned-source/bin/balena,push,Test] length=4
[debug] Using . as build source
could not load any valid magic files!

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For CLI bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```
[...]

PR #1603, opened 2020-02-05
Fix "could not load any valid magic files"

pdcastro:
Resolves: #1596 
Change-type: patch

### DIFF
--- a/lib/utils/eol-conversion.ts
+++ b/lib/utils/eol-conversion.ts
@@ -14,11 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import mmmagic = require('mmmagic');
-import fs = require('mz/fs');
-import Logger = require('./logger');
-
-const globalLogger = Logger.getLogger();
 
 // Define file size threshold (bytes) over which analysis/conversion is not performed.
 const LARGE_FILE_THRESHOLD = 10 * 1000 * 1000;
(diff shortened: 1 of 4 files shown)


pdcastro:
@balena-ci retest

lib/utils/eol-conversion.ts
```diff
@@ -14,11 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import mmmagic = require('mmmagic');
```
pdcastro: I've moved these imports inside the relevant functions, mainly so that the `mmmagic` module doesn't even need to be loaded if `process.platform !== 'win32'`.

srlowe: @pdcastro Interesting, I guess these distributions have no `file` command.  Thanks for the quick fix!  I suppose we could also consider testing for `file` on non-win OSs, in case there is any value in having the functionality on linux (e.g. if you've copied windows format files for some reason).

srlowe:
@balena-ci retest

lib/utils/eol-conversion.ts
srlowe: Just leaving note here that user had problem on Fedora 31 and Debian 9.11

srlowe:
@balena-ci retest

srlowe:
Mac build step is currently failing on CI. Some signing issue.  Waiting for the right people to come online.

srlowe:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

PR #1604, opened 2020-02-06
Add .gitattributes to check out with the correct line-ending on windows

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


.gitattributes
```diff
@@ -0,0 +1,6 @@
+# Set all files to use line feed endings (since we can't match only ones without an extension)
+* eol=lf
+# And then reset all the files with extensions back to default
+*.* -eol
+
+*.sh text eol=lf
```
pdcastro: Would any of this affect test file `tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh` when running `git push`?
https://github.com/balena-io/balena-cli/blob/master/tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh

Page-: No it still gets checked out with crlf because that's what's in the git repo itself. These settings are about core.autocrlf behaviour

PR #1606, opened 2020-02-06
Speed up tests

Page-:
Resolves: #   
Depends-on: <url>   
See: <url>   

---


package.json
```diff
@@ -42,21 +42,22 @@
...
-    "build:standalone": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:standalone",
-    "build:installer": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:installer",
+    "build:automation": "tsc -P automation/tsconfig.json --noEmit",
+    "build:doc": "mkdirp doc/ && ts-node --transpile-only -P automation/tsconfig.json automation/capitanodoc/index.ts > doc/cli.markdown",
+    "build:standalone": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts build:standalone",
+    "build:installer": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts build:installer",
     "package": "npm run build:fast && npm run build:standalone && npm run build:installer",
-    "release": "ts-node --type-check -P automation/tsconfig.json automation/run.ts release",
+    "release": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts release",
     "pretest": "npm run build",
```
pdcastro: If we had a pre-commit hook that checked that `npm run build` was executed (aborting the commit otherwise) e.g. by comparing file timestamps of staged files (hint!) ;-) ,  then we could further speed up tests with:
```
"pretest": "npm run build:fast",
"test:fast": "mocha --timeout 6000 -r ts-node/register/transpile-only \"tests/**/*.spec.ts\"",
```
I think that `"test:fast"` is currently "broken" in that it actually triggers `pretest` (because it calls `npm run test`), which makes it redundant and slower than `npm test`. My suggestion for `test:fast` above is a copy-and-paste of `test`, except that it doesn't trigger `pretest` and doesn't run `build` at all. (It would (only) be useful when the developer makes changes to test code only, not feature code.)

.gitattributes
```diff
@@ -0,0 +1,6 @@
+# Set all files to use line feed endings (since we can't match only ones without an extension)
+* eol=lf
+# And then reset all the files with extensions back to default
+*.* -eol
+
+*.sh text eol=lf
```
pdcastro: This `.gitattributes` file "belongs" to another open PR, doesn't it? Better remove from this one?

package.json
```diff
@@ -42,21 +42,22 @@
     "postinstall": "patch-package",
     "prebuild": "rimraf build/ build-bin/",
     "build": "npm run build:src && npm run catch-uncommitted",
-    "build:src": "npm run prettify && npm run lint && npm run build:fast && npm run build:doc",
+    "build:src": "npm run prettify && npm run lint && npm run build:automation && npm run build:fast && npm run build:doc",
     "build:fast": "gulp build && tsc",
-    "build:doc": "mkdirp doc/ && ts-node --type-check -P automation/tsconfig.json automation/capitanodoc/index.ts > doc/cli.markdown",
-    "build:standalone": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:standalone",
-    "build:installer": "ts-node --type-check -P automation/tsconfig.json automation/run.ts build:installer",
+    "build:automation": "tsc -P automation/tsconfig.json --noEmit",
```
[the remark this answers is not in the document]
Page-: Yep, also this type checks everything at once whereas the ts-node type-checking only type checks files as they're loaded, and so can potentially miss files that are unused

package.json
Page-: I was actually focused on speeding up the CI tests specifically as they run multiple times and having them be less demanding should hopefully help reduce problems 🤞
Speeding up/simplifying the local dev flow specifically is something I'm more focused on in #1589 (particularly around automating it) and is more invasive so I'd rather not bundle it here

.gitattributes
Page-: Yeah, I built on top of the .gitattributes PR so I could actually run the tests to time the improvement, and I expected that one to be an easy merge :|

pdcastro: Meaning that... you'd wait for #1604 to be merged first?

Page-: Well that was my plan (based on thinking it would "just work") but if you're ok with these changes specifically I can drop that commit and merge this by itself whilst I figure out how to make #1604 work properly on mac

Page-: I don't want to rebase unnecessarily as it'll trigger unnecessary node-cli jobs

Page-: I dropped the commit since that one has a bunch more to do it turns out

Issue #1611, opened 2020-02-10
Pushing is only working the secound time


AlexProgrammerDE:
- **balena CLI version:** 11.25.10
- **Operating system version:** Windows 10
- **32/64 bit OS and processor:** 64-bit Windows on 64-bit AMD processor
- **Install method:** executable installer and git bash

---

Hi everyone! I updated today to a newer version of the cli and had last updated after the fix for my issue #1554 This is my log:

**Edit:** See also: https://github.com/balena-io/balena-cli/issues/1611#issuecomment-584201285

```
alex-@My-PC MINGW64 ~/Downloads/balena-sound-master (master)
$ balena push rpi-4-1

alex-@My-PC MINGW64 ~/Downloads/balena-sound-master (master)
$ balena push rpi-4-1
[Info]       Starting build for rpi-4-1, user my_account
[Info]       Dashboard link: https://dashboard.balena-cloud.com/apps/1566937/devices
[Info]       Building on arm01
[Info]       Pulling previous images for caching purposes...
[Success]    Successfully pulled cache images
[airplay]    Step 1/5 : FROM balenalib/raspberrypi4-64:buster
[spotify]    Step 1/4 : FROM balenalib/raspberrypi3:buster
[bluetooth]  Step 1/16 : FROM balenalib/raspberrypi3:stretch
[airplay]     ---> 0b799f14cdae
[airplay]    Step 2/5 : ENV DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host/run/dbus/system_bus_socket
[airplay]    Using cache
[airplay]     ---> e8bcc9dd088d
[airplay]    Step 3/5 : RUN install_packages shairport-sync
[airplay]    Using cache
[airplay]     ---> 8c190c35b181
[airplay]    Step 4/5 : COPY start.sh /usr/src/
[spotify]     ---> f505656294ab
[spotify]    Step 2/4 : RUN curl -sSL https://dtcooper.github.io/raspotify/key.asc | apt-key add -v - 2> /dev/null   && echo 'deb https://dtcooper.github.io/raspotify jessie main' | tee /etc/apt/sources.list.d/raspotify.list   && install_packages raspotify
[airplay]    Using cache
[airplay]     ---> fd052211cac2
[airplay]    Step 5/5 : CMD [ "/bin/bash", "/usr/src/start.sh" ]
[airplay]    Using cache
[airplay]     ---> 38409254060e
[airplay]    Successfully built 38409254060e
[...]

PR #1615, opened 2020-02-12
Fix balena push "Segmentation fault" on Windows (replace 'mmmagic' with 'isBinaryFile')

pdcastro:
Connects-to: #1611
Change-type: patch


lib/utils/eol-conversion.ts
```diff
@@ -136,3 +117,123 @@ export async function readFileWithEolConversion(
 		return fileBuffer;
 	}
 }
+
+/**
+ * Attempt to detect the encoding of a data buffer.
+ * Code copied and modified from the npm package 'isbinaryfile' (MIT licence)
+ *   https://github.com/gjtorikian/isBinaryFile/blob/master/src/index.ts
```
pdcastro: To emphasize it, this function was copied from the `isbinaryfile` npm module (MIT licence). The original code is at https://github.com/gjtorikian/isBinaryFile/blob/4.0.4/src/index.ts#L62

The modifications I made were:
* The original function would return true/false (true=binary), and the new function returns a string naming the detected encoding (with `'binary'` representing a generic binary encoding, and 'utf-8' encompassing 'ascii').
* The original function would probabilistically conclude that a file is binary if more than 10% of characters were in the ASCII ranges 0 to 7 and 14 to 32. Because of our purpose of "editing" text files to replace CRLF with LF, I changed the detection to conservatively conclude that a file is binary if even a single character is found in those ranges, **except for** character ESC (escape, number 27 = 0x1B) which is used in ANSI escape sequences for text color.
* I've added some test code, but I didn't worry too much about the UTF-8 algorithm details because I didn't change it, and the original code had already been tested by more than 2 million downloads a week...

tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh
```diff
@@ -1,2 +1,2 @@
-#!/bin/sh
-echo 'this file was saved with Windows CRLF line endings'
+#!/bin/sh
+echo 'this file was saved with Windows CRLF line endings'
```
pdcastro: @Page-, I've had to re-add this file (unmodified) in order to avoid `git status` reporting modifications (and preventing branch switching, pushing...) despite not touching the file:
* Merely cloning the CLI master branch (fresh, on a new folder) and running `git status` would report that the file was modified.
* I tried with different values of `autocrlf` in my `~/.giconfig` file. No difference.
* Both before and after this commit:
  * `ls -l` reported the file size unchanged (70 bytes).
  * `hexdump -C` showed it contained CRLF as expected.
  * `md5` showed the same hash (1fffe84f62b4609d481dc0c00ca0887b).
* We already had test cases checking the file size before and after the CLI performs EOL conversion -- the tests verify that the file size is 70 before conversion, and 68 after conversion (2 lines). The tests passed before and after this commit.

The git-scm documentation for the `.gitattributes` file (which we recently added to the repo) seems to warn about this potential condition, with the fix being to re-add the file:
> https://git-scm.com/docs/gitattributes#_code_eol_code
> **eol**
> This attribute sets a specific line-ending style to be used in the working directory. It enables end-of-line conversion without any content checks, effectively setting the text attribute. **Note that setting this attribute on paths which are in the index with CRLF line endings may make the paths to be considered dirty. Adding the path to the index again will normalize the line endings in the index.**

_"Adding the path to the index again will normalize the line endings in the index"_ -- this is thus what I am trying to achieve with this commit.

pdcastro: Also: `git checkout `, `git rm --cached -r .`, `git reset --hard` ... make not difference. Well, re-cloning the CLI on a new folder makes no difference...
```
$ git checkout -- tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh

$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh

no changes added to commit (use "git add" and/or "git commit -a")

$ git reset --hard master
HEAD is now at d48672f v11.25.17

$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh

no changes added to commit (use "git add" and/or "git commit -a")

$ git checkout 1611-replace-mmmagic
error: Your local changes to the following files would be overwritten by checkout:
	tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh
Please commit your changes or stash them before you switch branches.
Aborting
```

pdcastro: Steps to reproduce:
```
$ docker run -it ubuntu:18.04
$ apt-get update && apt-get install -y git
$ git clone https://github.com/balena-io/balena-cli.git
$ cd balena-cli
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh

no changes added to commit (use "git add" and/or "git commit -a")
```
Steps to fix, following on from the above:
```
# set username and email, so that git allows creating a commit
$ git config --global user.name "test"
$ git config --global user.email "test"
$ git add tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh
$ git commit -am "fixed"
$ git status
On branch master
Your branch is ahead of 'origin/master' by 1 commit.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
```

lib/utils/eol-conversion.ts
pdcastro: Btw, my first attempt was to add `isbinaryfile` to `package.json` and then patch the function with `patch-package`, but the patch file turned out to be larger than the original function and harder to read. So I thought I'd just copy-and-paste the function, and add some notes about it.

lib/utils/eol-conversion.ts
```diff
@@ -111,7 +92,7 @@ export async function readFileWithEolConversion(
 	}
 
 	// Skip further processing of files that don't contain CRLF
-	if (!fileBuffer.includes('\r\n', 0, encoding)) {
+	if (!fileBuffer.includes('\r\n')) {
```
pdcastro: The `encoding` argument to `includes()` refers to the encoding of the `'\r\n'` string, not the encoding of the string in the `fileBuffer`.

srlowe: Ah, nice catch.  (Though I think a side-effect-less mistake)

pdcastro:
> I think it makes far more sense to move this to an external module that can be reused elsewhere, there's nothing in this that I can see which is specific to the cli

That's not a bad idea! :-) As it is in this PR (without mmmagic dependency, which is a binary module that includes a 5MB database), `eol-conversion.ts` turns out to be a very small and handy pure-JS EOL conversion library... We might want to make it a bit more generic than a `readFile` replacement: perhaps just a function that takes in a Buffer, detects the encoding and returns a converted Buffer as needed. Perhaps not for this relatively urgent PR, though.

lib/utils/eol-conversion.ts
pdcastro: In the end I didn't use `patch-package` -- actually this PR deletes a patch file that was being used to patch `mmmagic`. As for forking the module, I like better your other suggestion of creating a new small npm package dedicated to EOL conversion -- also because the "patch" to `isBinaryFile` isn't a bug fix, it actually slightly re-purposes the function/package to detecting encodings that are suitable for EOL conversion.  Maybe for this PR we can live with the function copy-and-paste, as we need to release a fix for Windows CLI users out there.

lib/utils/eol-conversion.ts
```diff
@@ -136,3 +117,123 @@ export async function readFileWithEolConversion(
...
+ *
+ * Note: pure ASCII data is identified as 'utf-8' (ASCII is indeed a subset
+ * of UTF-8).
+ *
+ * @param fileBuffer File contents whose encoding should be detected
+ * @param bytesRead Optional "file size" if smaller than the buffer size
+ */
+export async function detectEncoding(
+	fileBuffer: Buffer,
+	bytesRead?: number,
```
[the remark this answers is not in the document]
pdcastro: Nice - done!

pdcastro:
@balena-ci retest

Issue #1616, opened 2020-02-12
Add '--cachefrom' option to balena build / deploy commands


pdcastro:
User reports that cached images are not used by balena build / deploy commands. Shaun found / believes that the solution would be to use the `cachefrom` option of image build Docker API.

* Forum thread: https://forums.balena.io/t/balena-cli-is-not-using-the-cache-on-fresh-machine/62266
* Flowdock: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/uD3tiekE56YA1zpVEuoEq8z1SGX
* API: https://docs.docker.com/engine/api/v1.40/#operation/ImageBuild

samothx:
User reports this fix has worked for him..

pdcastro:
Feature released in CLI version 11.26.0.

PR #1617, opened 2020-02-12
Add '--cachefrom' option to balena build/deploy commands

pdcastro:
Resolves: #1616
Change-type: minor

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -1968,6 +1968,10 @@ The alias to the generated image
 
 Set a build-time variable (eg. "-B 'ARG=value'"). Can be specified multiple times.
 
+#### --cache-from &#60;image-list&#62;
+
+Comma-separated list (no spaces) of image names for build cache resolution. Implements the same feature as the "docker build --cache-from" option.
+
 #### --nocache
 
 Don't use docker layer caching when building
@@ -2096,6 +2100,10 @@ The alias to the generated image
 #### --buildArg, -B &#60;arg&#62;
 
 Set a build-time variable (eg. "-B 'ARG=value'"). Can be specified multiple times.
+
+#### --cache-from &#60;image-list&#62;
+
+Comma-separated list (no spaces) of image names for build cache resolution. Implements the same feature as the "docker build --cache-from" option.
 
 #### --nocache
 
--- a/lib/utils/docker-coffee.coffee
+++ b/lib/utils/docker-coffee.coffee
@@ -84,6 +84,13 @@ exports.appendOptions = (opts) ->
 			alias: 'B'
 		},
 		{
+			signature: 'cache-from'
+			parameter: 'image-list'
+			description: '
+				Comma-separated list (no spaces) of image names for build cache resolution.
+				Implements the same feature as the "docker build --cache-from" option.'
+		},
+		{
 			signature: 'nocache'
 			description: "Don't use docker layer caching when building"
 			boolean: true
@@ -160,6 +167,8 @@ exports.generateBuildOpts = (options) ->
 		opts.t = options.tag
 	if options.nocache?
 		opts.nocache = true
+	if options['cache-from']?.trim()
+		opts.cachefrom = options['cache-from'].split(',').filter((i) -> !!i.trim())
 	if options.squash?
 		opts.squash = true
 	if options.buildArg?


Langhalsdino:
There is one typo in the docker-coffee.coffee file. It is suposed to be:
```
if options.cachefrom?
		opts.cachefrom = options.cachefrom.split(',').filter((i) -> !!i.trim())
```

sadly it does not work. Since the cachefrom option is not used by the docker build process and i can not figure out why 🙈

pdcastro:
> sadly it does not work. Since the cachefrom option is not used by the docker build process and i can not figure out why 🙈

Thanks @Langhalsdino, I'll update it. There was a [comment from Shaun in the forum thread](https://forums.balena.io/t/balena-cli-is-not-using-the-cache-on-fresh-machine/62266/9) where he pointed out special metadata requirements for an image to be used as cache. Quoting the docker documentation:
> https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources
> To use an image as a cache source, cache metadata needs to be written into the image on creation. This can be done by setting --build-arg BUILDKIT_INLINE_CACHE=1 when building the image. After that, the built image can be used as a cache source for subsequent builds.

Did you try that? I'm not sure if the CLI could / should produce some metadata as well.

Langhalsdino:
> > https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources
> > To use an image as a cache source, cache metadata needs to be written into the image on creation. This can be done by setting --build-arg BUILDKIT_INLINE_CACHE=1 when building the image. After that, the built image can be used as a cache source for subsequent builds.
> 
> Did you try that? I'm not sure if the CLI could / should produce some metadata as well.

Yes i tried it and it does not seem to work. My current guess is, that the Dockerfile will get treaded similar to how docker compose gets treated, therefore we might need to implement something like [this](https://docs.docker.com/compose/compose-file/compose-file-v2/#cache_from).

From my understanding of the code, compose.coffee will be used even-though no docker-compose  file is setup. Furthermore i added some logging in in order to see, what the task during the build_project looks like and i found the following:
```
{ external: false,
  serviceName: 'main',
  buildStream:
   Pack {
     _readableState:
      ReadableState {
        objectMode: false,
        highWaterMark: 16384,
        buffer: BufferList { length: 1758 },
        length: 76864512,
        pipes: null,
        pipesCount: 0,
        flowing: null,
        ended: true,
        endEmitted: false,
        reading: false,
        sync: false,
        needReadable: false,
        emittedReadable: true,
        readableListening: false,
        resumeScheduled: false,
        destroyed: false,
        defaultEncoding: 'utf8',
        awaitDrain: 0,
        readingMore: false,
        decoder: null,
        encoding: null },
     readable: true,
     domain: null,
     _events: [Object: null prototype] {},
     _eventsCount: 0,
     _maxListeners: undefined,
     _drain: [Function: noop],
     _finalized: true,
     _finalizing: false,
     _destroyed: false,
     _stream: null },
  resolved: true,
  buildMetadata:
   BuildMetadata {
     metadataDirectories: [ '.balena/', '.resin/' ],
     metadataFiles: {} },
  dockerfilePath: 'Dockerfile',
  context: '.',
  dockerfile:
   'MY_DOCKER_FILE_LIFES_HERE ',
  architecture: 'aarch64',
  projectType: 'Standard Dockerfile',
  tag: 'app_main',
  dockerOpts:
   { t: 'app_main',
     cachefrom:
      [ 'registry.gitlab.com/...' ],
     buildargs: { BUILDKIT_INLINE_CACHE: '1' } },
 }
```

Langhalsdino:
I do not know why, but it started working now with docker-compose files and regular dockerfiles with no change at all :D 

Thanks for the PR :)

pdcastro:
Many thanks for testing it @Langhalsdino. I've tested it too, and indeed it appears to work as intended! 👍  (Usually it's more difficult, this PR was relatively straightforward!) :-)
Note that I've just pushed an update, renaming the option from `cachefrom` to `cache-from`, because:
* It aligns with the `docker build --cache-from` option.
* We've decided to standardise the balena CLI options across all commands (which are currently mixed) as dash-separated instead of camelCase. (This will happen over time, in major releases.)
* Most Unix command-line tools adopt dash-separated options.

And the whole PR is still subject to further review from the balena team, before it gets merged.

pdcastro:
@balena-ci retest

Langhalsdino:
Just out of curiosity why are you force pushing the branch all the time?

pdcastro:
> Just out of curiosity why are you force pushing the branch all the time?

3.1 reasons:
* To update the code without extra commits. It's nicer to have somewhat meaningful commits in the repo history, and the commit messages are automatically copied to the CHANGELOG file when the PR is merged (it's no use to CLI users to have a CHANGELOG line that reads _"renamed `--cachefrom` to `--cache-from`"_ if the former was never released in the first place...).
* To bring my branch up to date (rebased on master) when the master branch gets updated by other unrelated PRs getting merged. I could however have waited for reviewer approval before doing that, so I'd only do it once.
* Sometimes the `store-npm` CI step (publishing to the npm registry) fails with an error along the lines of _"you can't publish over a previously published version",_ and what I have empirically / anecdotally observed is that changing the commit hash gets around the error. It's in my to-do list to understand what's really going on there, and maybe do something about it...
  * Reason 3.1 would be to force balena CI to re-run the tests on Windows / macOS / Linux. Recently -- like right now -- the macOS tests have been failing for unrelated reasons, apparently networking problems with the physical machine's local network (our CI machines are not on the cloud, as this allows us to plug special hardware/devices to run some tests against them). However if the purpose is simply to re-run the tests, it's not necessary to force push: a better / easier way is to add a "balena-ci retest" comment to the PR.

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

PR #1622, opened 2020-02-17
Add a script to automate nested changelogs

thgreasi:
This is similar to the script that we have in the API and the UI.
See: https://github.com/balena-io/balena-api/blob/master/bin/update-module.ts
See: https://github.com/balena-io/balenacloud-dashboard/blob/master/scripts/update-module.ts

Change-type: patch

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,27 @@ reformats the code (based on configuration in the `node_modules/resin-lint/confi
 file). Beyond that, we have a preference for Javascript promises over callbacks, and for
 `async/await` over `.then()`.
 
+## Updating upstream dependencies
+
+In order to get proper nested changelogs, when updating upstream modules that are in the repo.yml
+(like the balena-sdk), the commit body has to contain a line with the following format:
+```
+Update balena-sdk from 12.0.0 to 12.1.0
+```
+
+Since this is error prone, it's suggested to use the following npm script:
+```
+npm run update balena-sdk ^12.1.0
+```
+This will create a new branch (only if you are currently on master), run `npm update` with the
+version you provided as a target and commit the package.json & npm-shrinkwrap.json. The script by
+default will set the `Change-type` to `patch` or `minor`, depending on the semver change of the
+updated dependency, but if you need to use a different one (eg `major`) you can specify it as an
+extra argument:
+```
+npm run update balena-sdk ^12.14.0 patch
+npm run update balena-sdk ^13.0.0 major
+
 ## Common gotchas
 
 One thing that most CLI bugs have in common is the absence of test cases exercising the broken
--- a/automation/update-module.ts
+++ b/automation/update-module.ts
@@ -1,0 +1,140 @@
+import { exec } from 'child_process';
+import * as semver from 'semver';
+
+const changeTypes = ['major', 'minor', 'patch'] as const;
+
+const validateChangeType = (maybeChangeType: string = 'minor') => {
+	maybeChangeType = maybeChangeType.toLowerCase();
+	switch (maybeChangeType) {
+		case 'patch':
+		case 'minor':
+		case 'major':
+			return maybeChangeType;
+		default:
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "watch": "gulp watch",
     "prettify": "prettier --write \"{lib,tests,automation,typings}/**/*.js\" --config ./node_modules/resin-lint/config/.prettierrc",
     "lint": "resin-lint lib/ tests/ && resin-lint --typescript --fix automation/ lib/ typings/ tests/",
+    "update": "npx ts-node@^8.0.0 --transpile-only ./automation/update-module.ts",
```
pdcastro: Is `npx ts-node@^8.0.0` necessary? The `devDependencies` section already requires `ts-node@^8.6.2`, so maybe it's enough to have:
```
"update": "ts-node --transpile-only ./automation/update-module.ts",
```

CONTRIBUTING.md
```diff
@@ -81,6 +81,27 @@ reformats the code (based on configuration in the `node_modules/resin-lint/confi
...
+```
+Update balena-sdk from 12.0.0 to 12.1.0
+```
+
+Since this is error prone, it's suggested to use the following npm script:
+```
+npm run update balena-sdk ^12.1.0
+```
+This will create a new branch (only if you are currently on master), run `npm update` with the
+version you provided as a target and commit the package.json & npm-shrinkwrap.lock. The script by
```
pdcastro: `npm-shrinkwrap.json` ? (lock vs json)

automation/update-module.ts
```diff
@@ -0,0 +1,134 @@
...
+		`
+Usage: npm run update ${upstreamName} $version [$changeType=minor]
+
+Upstream names: ${upstreams.map(({ repo }) => repo).join(', ')}
+`,
+	);
+	return process.exit(1);
+};
+
+(async () => {
```
pdcastro: Hey, give the function a name and add a short jsdoc description of its purpose... It's quite a large one! :-)

package.json
pdcastro: Also: other `ts-node` entries above include `-P automation/tsconfig.json`

nazrhom:
@balena-ci retest

thgreasi:
Sorry for not having a separate "Review changes" commit, but I marked this as a draft b/c I was doing changes and just wanted to test what balenaCI would do with my commits.
Will drop the sdk update commit, just want to double check that the nested changelog is indeed generated correctly.

CONTRIBUTING.md
hedss: `npm-shrinkwrap.json` gets published to NPM, so the deps will be set on package download, `package-lock.json` never gets published. It's something that's bugged me for a while (there are reasons for it, npm-shrinkwrap was the legacy way of doing this and also respects all deps in the lock file, not just top-level; it also means that you get exactly what you specify when your NPM package is downloaded as a module, instead of just what's compatible with `package.json`).

automation/update-module.ts
```diff
@@ -0,0 +1,141 @@
...
+const compareSemverChangeType = (oldVersion: string, newVersion: string) => {
+	const oldSemver = semver.parse(oldVersion)!;
+	const newSemver = semver.parse(newVersion)!;
+
+	for (const changeType of changeTypes) {
+		if (oldSemver[changeType] !== newSemver[changeType]) {
+			return changeType;
+		}
+	}
+	return;
```
hedss: I'm missing something here, why the `return`?

CONTRIBUTING.md
pdcastro: hmm I meant the typo: `npm-shrinkwrap.lock` _vs._ `npm-shrinkwrap.json`

automation/update-module.ts
thgreasi: Hmm :thinking: I don't really know. Removing it.

PR #1623, opened 2020-02-18
Add pre-commit check for cli.markdown updates and coffeelint execution

pdcastro:
This PR introduces a pre-commit file timestamp comparison between `doc/cli.markdown` (CLI markdown documentation) and any staged files under the `lib` and `typings` folders. If `cli.markdown` is older, an error is printed and the commit aborted:
```
$ git commit -a
husky > pre-commit (node v10.18.1)
--------------------------------------------------------------------------------
ERROR: at least one staged file: "lib/utils/promote.ts"
has a more recent modification timestamp than the documentation file:
"doc/cli.markdown"

This probably means that `npm run build` or `npm test` have not been executed,
and this error can be fixed by doing so. Running `npm run build` or `npm test`
before commiting is currently a requirement (documented in the CONTRIBUTING.md
file) for three reasons:
1. Because it updates the CLI markdown documentation (in case any command-line
   options were updated, added or removed).
2. Because it reveals coffeelint warnings that the CI would currently ignore
   (in case any Coffeescript files were modified).
3. Given that balena-cli CI builds/tests are currently fairly slow, running
   tsc or resin-lint on the laptop catches Typescript type check errors much
   sooner and typically results in reduced overall waiting time.

If you need/wish to bypass this check without running `npm run build`, run
`npx touch -am "doc/cli.markdown"` and try again.
--------------------------------------------------------------------------------
husky > pre-commit hook failed (add --no-verify to bypass)
```

Change-type: patch

### DIFF
--- a/automation/check-doc.js
+++ b/automation/check-doc.js
@@ -1,0 +1,93 @@
+/**
+ * @license
+ * Copyright 2020 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const { stripIndent } = require('common-tags');
+const _ = require('lodash');
+const { fs } = require('mz');
+const path = require('path');
+const simplegit = require('simple-git/promise');
+
+const ROOT = path.normalize(path.join(__dirname, '..'));
+
+/**
+ * Compare the timestamp of cli.markdown with the timestamp of staged files,
+ * issuing an error if cli.markdown is older. Besides the purpose of ensuring
+ * that cli.markdown is updated, it effectively also ensures that coffeelint
+ * is executed (via `npm run build` or `npm test`) on the developers laptop,
+ * so that there is at least a chance that the developer will spot any linter
+ * warnings (that could reveal bugs) sooner than later.  (The CI does not
+ * currently fail in case of coffeelint warnings.)
+ * If cli.markdown does not require updating and the developer cannot run
+ * `npm run build` on their laptop, the error message suggests a workaround
+ * using `touch`.
+ */
+async function checkBuildTimestamps() {
+	const git = simplegit(ROOT);
+	const docFile = path.join(ROOT, 'doc', 'cli.markdown');
+	const [docStat, gitStatus] = await Promise.all([
+		fs.stat(docFile),
+		git.status(),
+	]);
+	const stagedFiles = _.uniq([
+		...gitStatus.created,
+		...gitStatus.staged,
+		...gitStatus.renamed.map(o => o.to),
+	])
+		// select only staged files that start with lib/ or typings/
+		.filter(f => f.match(/^(lib|typings)[/\\]/))
+		.map(f => path.join(ROOT, f));
+
+	const fStats = await Promise.all(stagedFiles.map(f => fs.stat(f)));
+	fStats.forEach((fStat, index) => {
+		if (fStat.mtimeMs > docStat.mtimeMs) {
+			const fPath = stagedFiles[index];
+			throw new Error(stripIndent`
+				--------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


automation/check-doc.js
```diff
@@ -0,0 +1,90 @@
...
+				   options were updated, added or removed).
+				2. Because it reveals coffeelint warnings that the CI would currently ignore
+				   (in case any Coffeescript files were modified).
+				3. Given that balena-cli CI builds/tests are currently fairly slow, running
+				   tsc or resin-lint on the laptop catches Typescript type check errors much
+				   sooner and typically results in reduced overall waiting time.
+
+				If you need/wish to bypass this check without running \`npm run build\`, run
+				\`npx touch -am "${docFile}"\` and try again.
+				--------------------------------------------------------------------------------
```
pdcastro: The reason for advising `npx touch` instead of simply `touch` is Windows compatibility (`touch` is a Unix utility).

package.json
```diff
@@ -148,6 +148,7 @@
     "publish-release": "^1.6.1",
     "resin-lint": "^3.2.4",
     "rewire": "^3.0.2",
+    "simple-git": "^1.131.0",
```
pdcastro: [simple-git](https://www.npmjs.com/package/simple-git) is a lightweight wrapper around executing `git`, conveniently parsing the output as objects and arrays. The `git` executable must still be available.

automation/check-doc.js
```diff
@@ -0,0 +1,90 @@
...
+ * If cli.markdown does not require updating and the developer cannot run
+ * `npm run build` on their laptop, the error message suggests a workaround
+ * using `touch`.
+ */
+async function checkBuildTimestamps() {
+	const git = simplegit();
+	const docFile = path.join('doc', 'cli.markdown');
+	const [docStat, gitStatus] = await Promise.all([
+		fs.stat(docFile),
+		git.status(),
```
pdcastro: `git.status()` provides similar information as running `git status` on a command line, with the result returned in the form of objects and arrays. Here it is used to identify staged files (including newly created or renamed files).

I had originally intended to use `lint-staged` (as I had suggested in some comments in PR #1589) instead of anything like `git status`, because `lint-staged`'s purpose is exactly to run scripts with staged files as arguments. Alas, I found that `lint-staged` modifies the timestamps of staged files, which completely foiled my timestamp comparison plan! No matter: `git.status()` to the rescue. It should be OK for as long as this script is executed **before** `lint-staged`.

package.json
```diff
@@ -52,7 +52,7 @@
     "release": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts release",
     "pretest": "npm run build",
     "test": "mocha --timeout 6000 -r ts-node/register/transpile-only \"tests/**/*.spec.ts\"",
-    "test:fast": "npm run build:fast && npm run test",
+    "test:fast": "npm run build:fast && mocha --timeout 6000 -r ts-node/register/transpile-only \"tests/**/*.spec.ts\"",
```
pdcastro: This modification is in its own commit because it is an independent fix. What happens is that `npm test` triggers the `pretest` script that executes `npm run build`. Before this change, `npm run test:fast` would run `npm run build:fast` but then also run `npm run build` (because of `prestest`), therefore building twice and being slower than `npm test`.

pdcastro:
@balena-ci retest

srlowe:
@balena-ci retest

automation/check-doc.js
```diff
@@ -0,0 +1,90 @@
...
+		}
+	});
+}
+
+async function run() {
+	try {
+		await checkBuildTimestamps();
+	} catch (err) {
+		console.error(err.message);
+		process.exit(1);
```
hedss:
```suggestion
		process.exitCode = 1;
```
Whilst it doesn't usually happen, `libuv` can still be holding onto file handles after it's accessed files, and ensuring that it cleans itself up before exiting at the next event loop just makes things a bit better. `process.exitCode` ensures that the process still exists non-zero, but allows it to run the rest of the codepath in the current event loop and clean up first (and looking at the calling convention here, this is fine).

pdcastro:
Force push:
* Replaced `process.exit(1)` with `process.exitCode = 1` ([Hedley's comment](https://github.com/balena-io/balena-cli/pull/1623#discussion_r380852211)).
* Added the project root folder as an argument to simple-git (now `simplegit(ROOT)`, previously `simplegit()`) so that the script can be executed from any current working dir, not just the project root.

pdcastro:
Force pushed to update the PR branch (rebase on master). I had not merged the PR before because I wanted to run some additional tests on Windows. Tested now, looks good, will merge, thanks for the reviews! 👍

PR #1626, opened 2020-02-21
Update type deps

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


package.json
```diff
@@ -227,7 +227,7 @@
     "resin-compose-parse": "^2.1.0",
     "resin-doodles": "^0.1.1",
     "resin-image-fs": "^5.0.8",
-    "resin-multibuild": "^4.3.4",
+    "resin-multibuild": "^4.4.1-update-deps-9aede6efa97b9da634e42cc6f11c46b81ff5c077",
```
Page-: Depends on https://github.com/balena-io-modules/resin-multibuild/pull/72 for unpinning dockerode

lib/actions-oclif/envs.ts
```diff
@@ -377,7 +377,7 @@ function stringifyVarArray<T = Dictionary<any>>(
 					result[key] = value;
 				}
 			},
-			{},
+			{} as Dictionary<any>,
```
Page-: This is due to updated lodash types, which means the type of this is preserved, and the default `{}` would fail on the `result[key] = value` assignment

automation/capitanodoc/index.ts
```diff
@@ -57,11 +57,11 @@ function importCapitanoCommands(jsFilename: string): CapitanoCommand[] {
 	const commands: CapitanoCommand[] = [];
 
 	if (actions.signature) {
-		commands.push(_.omit(actions, 'action'));
+		commands.push(_.omit(actions, 'action') as any);
 	} else {
 		for (const actionName of Object.keys(actions)) {
 			const actionCommand = actions[actionName];
-			commands.push(_.omit(actionCommand, 'action'));
+			commands.push(_.omit(actionCommand, 'action') as any);
```
Page-: This is due to updated lodash typings, previously because of `actionCommand` being `any` the result of the `_.omit` was automatically `any` but now it's a bit tighter and clashes with the `commands` typing

Issue #1624, opened 2020-02-18
Build emulation does not work on non standard Dockerfile names


mwohlert:
- **balena CLI version:** 11.25.18
- **Operating system version:** Arch Linux
- **32/64 bit OS and processor:** 64bit
- **Install method:** npm
- **If npm install, Node.js and npm version:** Node v12.6.0 and npm v6.9.0
---

When using a Dockerfile with a diffrent name qemu detection does not work and emulator will not be used.

Example:
Dockerfile
```
FROM balenalib/armv7hf-debian:latest
RUN echo 'hello world'
```

Command:
`balena build --application testApp --source . --logs --emulated --dockerfile Dockerfile.debug`

This fails on x86 with:
```
[Info]    Building for testApp/q-svr-001
[Build]   Building services...
[Build]   main Preparing...
[Info]    Emulation is enabled
[Build]   main Step 1/2 : FROM balenalib/armv7hf-debian:latest
[Build]   main  ---> 98236338c3e5
[Build]   main Step 2/2 : RUN echo 'hello world'
[Build]   main  ---> Running in d798ee31cabe
[Build]   main standard_init_linux.go:211: exec user process caused "exec format error"
[Build]   Built 1 service in 0:20
[Error]   Build failed
The command '/bin/sh -c echo 'hello world'' returned a non-zero code: 1

Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, "$env:DEBUG = 1" on
powershell, or "export DEBUG=1" on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For CLI bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```

However if the Dockerfile is actually named Dockerfile the output is the following:
```
[Info]    Building for testApp/q-svr-001
[Build]   Building services...
[Build]   main Preparing...
[Info]    Emulation is enabled
[Build]   main Step 1/3 : FROM balenalib/armv7hf-debian:latest
[Build]   main  ---> 98236338c3e5
[Build]   main Step 2/3 : COPY [".balena/qemu-execve","/tmp/qemu-execve"]
[Build]   main  ---> 1f5a579b2cd1
[Build]   main Step 3/3 : RUN ["/tmp/qemu-execve","-execve","/bin/sh","-c","echo 'hello world'"]
[...]

PR #1628, opened 2020-02-25
Fix build/deploy commands with QEMU emulation and alternative Dockerfile

pdcastro:
Resolves: #1624
Change-type: patch
Depends-on: https://github.com/balena-io-modules/docker-qemu-transpose/pull/28

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -384,20 +384,6 @@ async function validateSpecifiedDockerfile(
 		`);
 	}
 
-	const { dir, ext, name } = path.parse(fullDockerfilePath);
-	if (ext) {
-		const nativePathMinusExt = path.join(dir, name);
-		if (await fs.exists(nativePathMinusExt)) {
-			throw new ExpectedError(stripIndent`
-				Error: "${name}" exists on the same folder as "${nativeDockerfilePath}".
-				When an alternative Dockerfile name is specified, a file with the same base name
-				(minus the file extension) must not exist in the same folder. This is because
-				the base name file will be auto generated and added to the tar stream that is
-				sent to balenaEngine or the Docker daemon, resulting in duplicate Dockerfiles
-				and undefined behavior.
-			`);
-		}
-	}
 	return toPosixPath(nativeDockerfilePath);
 }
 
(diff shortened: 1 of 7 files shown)


lib/utils/compose_ts.ts
```diff
@@ -384,20 +384,6 @@ async function validateSpecifiedDockerfile(
...
-			throw new ExpectedError(stripIndent`
-				Error: "${name}" exists on the same folder as "${nativeDockerfilePath}".
-				When an alternative Dockerfile name is specified, a file with the same base name
-				(minus the file extension) must not exist in the same folder. This is because
-				the base name file will be auto generated and added to the tar stream that is
-				sent to balenaEngine or the Docker daemon, resulting in duplicate Dockerfiles
-				and undefined behavior.
-			`);
-		}
-	}
```
pdcastro: This project validation block was deleted because it regarded the fact that a specified dockerfile like `Dockerfile.debug` could become `Dockerfile` in the tar stream that gets sent to the Docker daemon. `docker-qemu-transpose` version 1.0.3 (depends-on PR balena-io-modules/docker-qemu-transpose/pull/28) changes that. (There's the related but different case of `Dockerfile.template` becoming `Dockerfile`, but this is not covered by this validation check which is specific to specified dockerfiles.)

package.json
```diff
@@ -186,7 +186,7 @@
     "denymount": "^2.3.0",
     "docker-modem": "^2.0.4",
     "docker-progress": "^4.0.0",
-    "docker-qemu-transpose": "^1.0.2",
+    "docker-qemu-transpose": "^1.0.3-27-specified-dockerfile-bde92fe1c1252f71aebb1e1df91dc0395ecf3cf1",
```
pdcastro: This will be changed to `^1.0.3` after depends-on PR balena-io-modules/docker-qemu-transpose/pull/28 is merged and released.

repo.yml
```diff
@@ -4,3 +4,5 @@ publishMetadata: true
 upstream:
   - repo: 'balena-sdk'
     url: 'https://github.com/balena-io/balena-sdk'
+  - repo: 'docker-qemu-transpose'
+    url: 'https://github.com/balena-io-modules/docker-qemu-transpose'
```
pdcastro: @thgreasi, just to confirm, any upstream repo can be added to this file, right? Like I've done for `docker-qemu-transpose`.

pdcastro:
Force push: updated version of `docker-qemu-transpose` to `^1.0.3` in `package.json`.

PR #1636, opened 2020-02-27
Simplify lazy-loading of balena-sdk

Page-:
This also avoids instantiating multiple balena-sdk

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/utils/lazy.ts
+++ b/lib/utils/lazy.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2016-2017 Balena
+Copyright 2020 Balena
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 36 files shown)


Page-:
@balena-ci retest

Page-:
@balena-ci retest

lib/utils/lazy.ts
```diff
@@ -0,0 +1,32 @@
...
+	return (): T => {
+		if (!cached) {
+			cached = fn();
+		}
+		return cached;
+	};
+};
+
+export const getBalenaSdk = once(() =>
+	(require('balena-sdk') as typeof BalenaSdk).fromSharedOptions(),
```
pdcastro: This line could be written as:
```js
(require('balena-sdk') as typeof import('balena-sdk')).fromSharedOptions(),
```
... in which case the module-level import line could be deleted:

~~`import * as BalenaSdk from 'balena-sdk';`~~

I imagine that it was your intention for `lazy.ts` to be generic/reusable for other functions besides `getBalenaSdk` (which may not depend on `balena-sdk`), so it would be desirable to avoid the module-level import. I gather that the current module-level import is for Typescript type information only, and gets deleted from built javascript. And I've also seen your draft PR that replaces `import` with `import type` for clarity and certainty (nice!), though it seems to be blocked by required changes to `prettier` (?). Nonetheless, at least while we don't have `import type`, I think the intention and behaviour are made clearer by moving the import inside `getBalenaSdk` as suggested. Built javascript appears to be unchanged.

Page-: I'd rather wait for the `import type` as I think the `import()` is a bit weird, and yeah the reason we're waiting on prettier is that it fails to parse the `import type` into AST so it can't prettify it and fails but I wouldn't expect it to take long to be resolved

PR #1638, opened 2020-02-28
Simplify lazy-loading of resin-cli-visuals with a shared function

Page-:
This also avoids current unnecessary requiring of resin-cli-visuals for every command - on my machine this shaves off around 220ms on any command that does not actually use resin-cli-visuals and helps make them feel much snappier

Change-type: patch

### DIFF
--- a/lib/utils/lazy.ts
+++ b/lib/utils/lazy.ts
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import * as BalenaSdk from 'balena-sdk';
+import * as visuals from 'resin-cli-visuals';
 
 // Equivalent of _.once but avoiding the need to import lodash for lazy deps
 const once = <T>(fn: () => T) => {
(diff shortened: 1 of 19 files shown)


lib/utils/lazy.ts
```diff
@@ -30,3 +31,7 @@ const once = <T>(fn: () => T) => {
 export const getBalenaSdk = once(() =>
 	(require('balena-sdk') as typeof BalenaSdk).fromSharedOptions(),
 );
+
+export const getVisuals = once(
+	() => require('resin-cli-visuals') as typeof visuals,
```
pdcastro: NIT: for clarity (code reader's first impression of `lazy.ts`), I would have preferred replacing the module-level import (line 18) with a local import here:
```
	() => require('resin-cli-visuals') as typeof import('resin-cli-visuals'),
```

Page-: I understand the sentiment but I do believe that `import type` is the right solution for it and I expect it to be usable pretty soon (so it's worth waiting a little bit, but if it was further out I'd definitely agree with the `import()`)

PR #1639, opened 2020-02-28
Make use of capitano's promise support to simplify the code

Page-:
There is still usage of the callback method for capitano in the codebase but only for code that actually uses a callback interface somewhere, anything purely promise based now relies on the promise handling (and as you can see I got to delete a load of code :D )

### DIFF
--- a/lib/actions/api-key.ts
+++ b/lib/actions/api-key.ts
@@ -18,8 +18,8 @@ export const generate: CommandDefinition<{
 
 			$ balena api-key generate "Jenkins Key"
 	`,
-	async action(params, _options, done) {
-		getBalenaSdk()
+	async action(params) {
+		await getBalenaSdk()
 			.models.apiKey.create(params.name)
 			.then(key => {
 				console.log(stripIndent`
@@ -29,7 +29,6 @@ export const generate: CommandDefinition<{
 
 					This key will not be shown again, so please save it now.
 				`);
-			})
-			.finally(done);
+			});
 	},
 };
--- a/lib/actions/app.ts
+++ b/lib/actions/app.ts
@@ -54,7 +54,7 @@ Examples:
 		},
 	],
 	permission: 'user',
-	async action(params, options, done) {
+	async action(params, options) {
 		const balena = getBalenaSdk();
 
 		const patterns = await import('../utils/patterns');
@@ -82,8 +82,7 @@ Examples:
 				console.info(
 					`Application created: ${application.app_name} (${application.device_type}, id ${application.id})`,
 				),
-			)
-			.nodeify(done);
+			);
 	},
 };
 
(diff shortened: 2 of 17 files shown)


lib/actions/api-key.ts
```diff
@@ -18,8 +18,8 @@ export const generate: CommandDefinition<{
 
 			$ balena api-key generate "Jenkins Key"
 	`,
-	async action(params, _options, done) {
-		getBalenaSdk()
+	async action(params) {
```
pdcastro: I haven't tested this to confirm so I don't know for sure that it doesn't work, but... I see that Capitano detects promises by testing whether the "promise object" has a `.then` property:
```
					if callback?
						if actionPromise?.then?
							# Asynchronous action function
							actionPromise.then(
```
And while this test works with Bluebird promises, it doesn't seem to work with native async promises / functions. So perhaps Capitano would need to be patched for some of the changes in this PR to work. Needs confirmation.

pdcastro: Testing with the interactive interpreter, note how **`f.then`** is `undefined`, whereas **`g.then`** is `Function`:
```js
$ node
> async function f() {
... await new Promise(resolve => setTimeout(resolve, 1000));
... }
undefined
> f.then
undefined
> const Bluebird = require('bluebird');
undefined
> const g = Bluebird.resolve(new Promise(resolve => setTimeout(resolve, 1000)));
Promise [Object] {
  _bitField: 0,
  _fulfillmentHandler0: undefined,
  _rejectionHandler0: undefined,
  _promise0: undefined,
  _receiver0: undefined }
> g.then
[Function]
```

pdcastro: Perhaps the Capitano fix would be as simple as replacing:
```coffeescript
if actionPromise?.then?
```
with:
```coffeescript
if actionPromise?.then? or actionPromise[Symbol.toStringTag] == 'AsyncFunction'
```
on this line: https://github.com/balena-io/capitano/blob/v1.9.1/lib/command.coffee#L78

srlowe: Yeah I think checking for `.then()` is the most common way to detect a promise object cross-implementation.

srlowe: I thought testing for presence of `.then()` was the most common implementation-agnostic way of detecting promise object.

srlowe: @pdcastro In your above example you are testing `f.then`, but shouldn't you be testing `f().then`?

e.g. this line in capitano: https://github.com/balena-io/capitano/blob/8547387f6b62c8a5777da07f9e29236568ebe57b/lib/command.coffee#L73

Page-: @pdcastro as @srlowe points out your test is flawed, you're testing for `Function.then` not `Promise.then`. The code definitely works, eg check out https://github.com/balena-io/balena-cli/blob/v11.28.5/lib/actions/auth.ts#L85 which is already using async with capitano's promise support

pdcastro: @srlowe, yes, you're right, I should have tested `f().then`, in which case the test works! Thanks! Now approving the PR. :-)

PR #1640, opened 2020-02-28
Lazy-load chalk

Page-:
This one only shaves off ~40ms but it still helps as it was previously loaded at the module level in one place so we paid the cost on every command (and 40ms here and there really adds up to how snappy the cli feels)

### DIFF
--- a/lib/utils/lazy.ts
+++ b/lib/utils/lazy.ts
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import * as BalenaSdk from 'balena-sdk';
+import chalk from 'chalk';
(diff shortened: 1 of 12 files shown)


lib/utils/lazy.ts
```diff
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import * as BalenaSdk from 'balena-sdk';
+import chalk from 'chalk';
```
pdcastro: Being the 3rd PR, you know what my comment here would be... :-)

Page-: I'll fix it the moment I can :P

PR #1642, opened 2020-02-29
Switch to object spreading in favor of _.assign

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/utils/promote.ts
+++ b/lib/utils/promote.ts
@@ -196,7 +196,7 @@ async function selectAppFromList(applications: BalenaSdk.Application[]) {
 	return selectFromList(
 		'Select application',
 		_.map(applications, app => {
-			return _.merge({ name: app.slug }, app);
+			return { name: app.slug, ...app };
 		}),
 	);
 }
--- a/tests/nock-mock.ts
+++ b/tests/nock-mock.ts
@@ -19,7 +19,6 @@ import { configureBluebird } from '../build/app-common';
 
 configureBluebird();
 
-import * as _ from 'lodash';
 import * as nock from 'nock';
 
 export interface ScopeOpts {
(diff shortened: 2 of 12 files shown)


lib/utils/promote.ts
```diff
@@ -196,7 +196,7 @@ async function selectAppFromList(applications: BalenaSdk.Application[]) {
 	return selectFromList(
 		'Select application',
 		_.map(applications, app => {
-			return _.merge({ name: app.slug }, app);
+			return { name: app.slug, ...app };
```
pdcastro: This may or may not be relevant, but:
* I think that inverting the assignment order (as follows) would be closer to the original behaviour in the corner case where the `app` object contained a `name` property of `undefined` value (impossible?) -
```js
return { ...app, name: app.slug };
```
This is because the [documentation of _.merge](https://lodash.com/docs/4.17.15#merge) says that _"Source properties that resolve to undefined are skipped if a destination value exists."_

* Another difference is that, with `_.merge`, _"Array and plain object properties are merged recursively"_ -- i.e., properties that are arrays and objects are recursively _cloned_, whereas the spreading syntax does not clone properties. But from my reading of the code, I think it is not a problem in this case.

tests/nock-mock.ts
```diff
@@ -54,38 +53,34 @@ export class NockMock {
 
 	public optGet(
 		uri: string | RegExp | ((uri: string) => boolean),
-		opts: ScopeOpts,
+		{ optional = false, persist = false }: ScopeOpts,
```
pdcastro: Nice trick! :-)

lib/utils/promote.ts
Page-: I'll have to test to make sure but I think the current order is correct, the docs are just pointing out undefined as special case (when the property exists as undefined as opposed to not existing)

pdcastro: You're right, sorry for the noise. (I hadn't realised that `app: BalenaSdk.Application` doesn't even have a `name` property.)

Page-: So I've done some testing on this and got:
```ts
_.merge({name: 'a'}, {name: ['b','c']})
// {name: ['b','c']}
_.merge({name: 'a'}, {name: 'b'})
// {name: 'b'}
```
so merge semantics don't apply when the target is a string (which I expected but double checked)
```ts
_.merge({name: 'a'}, {name: undefined})
// {name: 'a'}
{name: 'a', ...{name: undefined}}
// {name: undefined}
```
but there is the difference where the source is set as undefined then the target also gets set ^ Thankfully that won't apply here as `applications` comes over the network via JSON which cannot represent a value being set as undefined (only not set)

PR #1644, opened 2020-03-02
Don't try to setup a proxy agent when there's no proxy configured

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/app-common.ts
+++ b/lib/app-common.ts
@@ -172,7 +172,6 @@ async function setupGlobalAgentProxy(
 	env.GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE = '';
 
 	if (proxy) {
-		const _ = await import('lodash');
 		const proxyUrl: string =
 			typeof proxy === 'string' ? proxy : makeUrlFromTunnelNgConfig(proxy);
 
@@ -180,7 +179,7 @@ async function setupGlobalAgentProxy(
 
 		env.NO_PROXY = [
 			...requiredNoProxy,
-			...(noProxy ? _.filter((noProxy || '').split(',')) : privateNoProxy),
+			...(noProxy ? noProxy.split(',').filter(v => v) : privateNoProxy),
 		].join(',');
 	}
 
--- a/lib/app-common.ts
+++ b/lib/app-common.ts
@@ -87,6 +87,8 @@ export interface GlobalTunnelNgConfig {
 	sockets?: number;
 }
 
+type ProxyConfig = string | GlobalTunnelNgConfig;
+
 /**
  * Global proxy setup. Originally, `global-tunnel-ng` was used, but it only
  * supports Node.js versions older than 10.16.0. For v10.16.0 and later,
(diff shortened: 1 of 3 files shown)


lib/app-common.ts
```diff
@@ -115,35 +115,6 @@ export interface GlobalTunnelNgConfig {
  * default exclusion patterns are added for all private IPv4 address ranges.
  */
 async function setupGlobalHttpProxy(settings: CliSettings) {
-	const semver = await import('semver');
-	if (semver.lt(process.version, '10.16.0')) {
-		setupGlobalTunnelNgProxy(settings);
-	} else {
-		// use global-agent instead of global-tunnel-ng
-		await setupGlobalAgentProxy(settings);
-	}
```
pdcastro: I liked better having the separate `setupGlobalTunnelNgProxy` and `setupGlobalAgentProxy` functions, mainly because it makes it clearer / obvious that some proxy functionality only applies to each of them; for example, `BALENARC_NO_PROXY ` only applies to `setupGlobalAgentProxy`. Also, the suggested modification in this PR still runs a bit of unnecessary code if the version of Node is older than 10.16.0. What do you think of keeping the separate functions unmodified, and simply adding the following `if` clause at the top of the original `setupGlobalHttpProxy` function?

```
async function setupGlobalHttpProxy(settings: CliSettings) {
	if (
		!process.env.HTTP_PROXY &&
		!process.env.HTTPS_PROXY &&
		!process.env.http_proxy &&
		!process.env.https_proxy &&
		!settings.getCatch<string | GlobalTunnelNgConfig>('proxy')
	) {
		return;
	}
	... // the rest of the function is kept unmodified
```

This `if` clause matches the documented CLI proxy capabilities:
https://www.balena.io/docs/reference/balena-cli/#proxy-support

lib/app-common.ts
```diff
@@ -115,21 +117,36 @@ export interface GlobalTunnelNgConfig {
...
-	const semver = await import('semver');
-	if (semver.lt(process.version, '10.16.0')) {
-		setupGlobalTunnelNgProxy(settings);
-	} else {
-		// use global-agent instead of global-tunnel-ng
-		await setupGlobalAgentProxy(settings);
+	// `global-tunnel-ng` accepts lowercase variables with higher precedence
+	// than uppercase variables, but `global-agent` does not accept lowercase.
+	// Set uppercase versions for backwards compatibility.
+	const { env } = process
```
pdcastro: Build failing because prettier was not run (missing semicolon and other errors).

Page-: Yeah.. I'm used to every other repo running it automatically :P

lib/app-common.ts
```diff
@@ -115,22 +112,37 @@ export interface GlobalTunnelNgConfig {
...
+	if (env.http_proxy) {
```
Page-: I moved this code to shared so that we can rely on just the single form of `HTTP_PROXY` for both proxy setups and also have to check less vars in the `if` below

lib/app-common.ts
Page-: Fixed for each commit

lib/app-common.ts
```diff
@@ -115,21 +117,36 @@ export interface GlobalTunnelNgConfig {
...
+		env.HTTP_PROXY = env.http_proxy;
+	}
+	if (env.https_proxy) {
+		env.HTTPS_PROXY = env.https_proxy;
+	}
+	delete env.http_proxy;
+	delete env.https_proxy;
+
+	const proxy = settings.getCatch<ProxyConfig>('proxy');
+	if (proxy || env.HTTPS_PROXY != null || env.HTTP_PROXY != null) {
```
pdcastro: If I recall correctly, in some platform (Windows?) or shell (PowerShell?) it is possible to set an environment variable to the empty string (i.e. setting it in the shell, in a way that the CLI process sees it). However there's no need to load the proxy modules if an environment variable is set to the empty string, so I'd replace the `!= null` test with:
```
if (proxy || env.HTTPS_PROXY || env.HTTP_PROXY)
```
This would also be more consistent with the other tests a few lines above:
```
if (env.http_proxy) ...
if (env.https_proxy) ...
```

lib/app-common.ts
```diff
@@ -156,33 +171,19 @@ async function setupGlobalAgentProxy(settings: CliSettings) {
 
 	const env = process.env;
 	env.GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE = '';
+	env.NO_PROXY = [
+		...requiredNoProxy,
+		...(noProxy ? noProxy.split(',').filter(v => v) : privateNoProxy),
+	].join(',');
```
pdcastro: I've taken the liberty to add my own commit that moves the 4 lines above from the `if (proxy)` block, so that BALENARC_NO_PROXY is also considered when BALENARC_PROXY is not set (but HTTP(S)_PROXY are set).

lib/app-common.ts
Page-: No idea if you can do it in cmd/powershell but you can definitely do it in bash

Page-: Changed

PR #1608, opened 2020-02-07
Node 13 compatibility: update ext2fs module

srlowe:
Upgraded ext2fs version in `npm-shrinkwrap.json`

Resolves: # 1591, 
Change-type: patch

PR #1645, opened 2020-03-03
Update dependencies

Page-:
Depends on #1656


package.json
```diff
@@ -238,6 +238,7 @@
     "semver": "^7.1.1",
     "shell-escape": "^0.2.0",
     "split": "^1.0.1",
+    "stream-to-promise": "^2.2.0",
```
Page-: The cli actually uses this directly but was relying on a version pulled in via a dependency, I've installed it explicitly to avoid potential future surprises

package.json
```diff
@@ -218,17 +218,17 @@
     "node-unzip-2": "^0.2.8",
     "oclif": "^1.15.2",
     "open": "^7.0.2",
-    "patch-package": "6.1.2",
```
pdcastro: Nope! You had already updated this about a month ago, it introduced build errors and I pinned it back:
* https://github.com/balena-io/balena-cli/pull/1608/commits/05d478b759a79e7e3f0ae9c1d4e5e9adce246409
* https://github.com/ds300/patch-package/pull/202
* https://github.com/ds300/patch-package/issues/201

I created a PR with a fix upstream (link above), the maintainer thought the solution should be different so the PR was closed; they said they'd do it themselves in a week, then they didn't. I created an issue in their repo (also above).

Page-: I've pinned it but I think it's probably worth trying to remove those patches if at all possible since modifying into other packages can break expectations and is generally fragile (albeit I can also see the usefulness for short periods)

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
 	"compilerOptions": {
-		"declaration": true,
+		"declaration": false,
```
Page-: This is due to the `chalk` update meaning that lazy-loading fails because it can't re-export private types from chalk for declaration files. Since the cli is not intended to be used as a module the declaration types aren't useful and disabling them avoids the issue

tests/utils/eol-conversion.spec.ts
```diff
@@ -64,7 +64,6 @@ describe('detectEncoding() function', function() {
 		const sampleBinary = [
 			'ext2fs/build/Release/bindings.node',
 			'drivelist/build/Release/drivelist.node',
-			'resin-cli-visuals/node_modules/drivelist/build/Release/drivelist.node',
```
Page-: This file no longer exists because it's been deduped with the top-level drivelist

automation/deploy-bin.ts
```diff
@@ -64,7 +62,7 @@ export async function release() {
 
 /** Return a cached Octokit instance, creating a new one as needed. */
 const getOctokit = _.once(function() {
-	const Octokit = (require('@octokit/rest') as typeof _ocktokit).plugin(
+	const Octokit = (require('@octokit/rest') as typeof import('@octokit/rest')).Octokit.plugin(
```
Page-: This change was required due to types, but either form would work fine if I cast to `any`:
```
require('@octokit/rest')
[Function: DeprecatedOctokit] {
  Octokit: [Function: bound Octokit] { plugin: [Function: bound registerPlugin] },
  plugin: [Function: bound registerPlugin]
}
```

Page-: I'm assuming in 17.x they might remove the `.plugin` form in favor of `.Octokit.plugin` but I haven't checked

Page-:
@balena-ci retest

Page-:
@balena-ci retest

npm-shrinkwrap.json
```diff
@@ -24,6 +24,17 @@
         "js-tokens": "^4.0.0"
       },
       "dependencies": {
+        "chalk": {
```
pdcastro: @Page-, if I re-create the `npm-shrinkwrap.json` file through the following commands, I get a [different result](https://github.com/balena-io/balena-cli/pull/1650/commits/df8f7c212c42f333ea05e3ea2c4133b5a3d16f3e):
```
$ git checkout update-deps
$ rm -rf node_modules
$ npm install
```
```
$ node --version
v10.18.1
$ npm --version
6.13.4
```

My "different result" is in a draft PR (for the purpose of comparison / this discussion), based on the `update-deps` branch, which passes the build tests: https://github.com/balena-io/balena-cli/pull/1650
Skimming through the [differences](https://github.com/balena-io/balena-cli/pull/1650/commits/df8f7c212c42f333ea05e3ea2c4133b5a3d16f3e), I have the impression (could be wrong) that the shrinkwrap file in this PR updates/bumps more packages than the shrinkwrap file in the draft PR. Both satisfy the `package.json` requirements, but I think it could be said that the one in the draft PR is "more conservative", upgrading just as much as necessary to still satisfy the `package.json` file. Being conservative in this sense is probably a good thing: the "stability" gained from the shrinkwrap file.

Interestingly, `npm audit --production` appears to produce identical output in both cases.

What do you make of it? Out of interest, how did you create the `npm-shrinkwrap.json` file in this PR?

Page-: @pdcastro I think what's happened with yours is that based on the starting point that npm decided to keep a load of packages duplicated, eg your PR adds an additional 4 copies of the `chalk` vs this PR, an extra 3 copies of `ansi-escapes`, one extra `strip-ansi`, 3 extra `supports-color`, and so on. If you try running `npm dedupe; npm i` on yours I'd be interested if it ended up more similar (the `npm i` would be to add the extraneous net-keepalive shrinkwrapped packages back).

As for how I created it I did it via a bunch of `npm update $package` and occasional `npm rm $package; npm i $package` to trigger clean installs with fully updated deps of those packages rather than keeping old versions due to them currently being installed - that being important in order to make sure we reduce the number of copies of modules to shrink the install size, and reduce the time spent loading and compiling duplicate modules (and reduce the associated memory usage)

tsconfig.json
pdcastro: Just realised... This would mean that the `d.ts` files would no longer be created in the `build` folder, right? A consequence might be that the CLI test files written in Typescript would no longer be able to perform type checking, because they import directly from the build folder. For example:
https://github.com/balena-io/balena-cli/blob/v11.28.13/tests/helpers.ts#L26

What could we do to preserve type information in tests, one way or another?

pdcastro: Just realised... This would mean that the `d.ts` files would no longer be created in the `build` folder, right? A consequence might be that the CLI test files written in Typescript would no longer be able to perform type checking, because they import directly from the build folder. For example:
https://github.com/balena-io/balena-cli/blob/v11.28.13/tests/helpers.ts#L26

What could we do to preserve type information in tests, one way or another?

lib/actions/local/flash.ts
```diff
@@ -22,25 +22,7 @@ import { getChalk, getVisuals } from '../../utils/lazy';
...
-		}
-		drive = d;
-	} else {
-		const { DriveList } = await import('../../utils/visuals/drive-list');
-		const driveList = new DriveList(scanner);
-		drive = await driveList.run();
-	}
-	scanner.stop();
-	return drive;
+	return options.drive || getVisuals().drive('Select a drive');
```
Page-: This matches how we do it in other places of the cli so it's now consistent across commands. It also allows removing the direct dependency on inquirer-dynamic-list which is problematic

Page-:
@balena-ci retest

Issue #1649, opened 2020-03-05
Balena build with --emulated error, when balena push and build on cloud works perfectly


Razikus:
- **balena CLI version:** e.g. 11.2.1 (output of the `"balena version"` command)
11.28.11
- **Operating system version:** e.g. Windows 10, Ubuntu 18.04, macOS 10.14.5
Ubuntu 19.10
- **32/64 bit OS and processor:** e.g. 32-bit Windows on 64-bit Intel processor
64 bit Intel
- **Install method:** npm or zip or executable installer
np,
- **If npm install, Node.js and npm version:** e.g. Node v8.16.0 and npm v6.4.1
Node 10.19.0
np 6.13.4

---

Hello i have a problem with building on emulated environment
I'm using simple service wifi-connect based on balena example
docker-compose.yml:
```
version: "2.1"

services:
    wifi-connect:
        build: ./wifi-connect
        network_mode: "host"
        labels:
            io.balena.features.dbus: '1'
        cap_add:
            - NET_ADMIN
        environment:
            DBUS_SYSTEM_BUS_ADDRESS: "unix:path=/host/run/dbus/system_bus_socket"
```

wifi-connect fork here: https://github.com/Razikus/wifi-connect (just the same as original repo)

Everything is OK when i'm pushing code over balena push (it's builded && working correctly) , but that's not the case for me.

But when i'm using command
balena build --emulated --application Skryba --sources .

It's getting me an error "standard_init_linux.go:211: exec user process caused "exec format error"

```
debug] original argv0="node" argv=[/usr/bin/node,/usr/bin/balena,build,--emulated,--application,Skryba,--sources,.] length=8
[Debug]   Parsing input...
[Debug]   Loading project...
[Debug]   Resolving project...
[Debug]   docker-compose.yml file found at "/home/adam/dev/henk/balena-python-hello-world"
[Debug]   Creating project...
[Info]    Building for rpi/raspberry-pi
[Info]    Emulation is enabled
[Debug]   Found build tasks:.. -
[Debug]       wifi-connect: build [./wifi-connect]
[Debug]   Resolving services with [raspberry-pi|rpi]
[Debug]   Found project types: \
[Debug]       wifi-connect: Dockerfile.template
[Debug]   Prepared tasks; building...
[Build]   Built 1 service in 0:01
[Build]   wifi-connect [======>              ]  33% Step 2/6: standard_init_linux.go:211: exec user process caused "exec format error"
[Error]   Build failed
[...]

PR #1651, opened 2020-03-06
Fix `build --emulated` on Linux ("exec format error")

pdcastro:
Resolves: #1649
Change-type: patch

### DIFF
--- a/lib/utils/compose.coffee
+++ b/lib/utils/compose.coffee
@@ -230,7 +230,7 @@ exports.buildProject = (
 					containerQemuPath: "/tmp/#{qemu.QEMU_BIN_NAME}",
 					qemuFileMode: 0o555,
 				},
-				dockerfilePath
+				dockerfilePath or undefined,
 			).then (stream) ->
 				task.buildStream = stream
 			.return([ task, binPath ])


lib/utils/compose.coffee
```diff
@@ -230,7 +230,7 @@ exports.buildProject = (
 					containerQemuPath: "/tmp/#{qemu.QEMU_BIN_NAME}",
 					qemuFileMode: 0o555,
 				},
-				dockerfilePath
+				dockerfilePath or undefined,
```
pdcastro: A recent PR (#1628) introducing the feature of specified/alternative Dockerfile names in the context of QEMU emulation resulted in the `dockerfilePath` value on this line being the empty string (instead of undefined) when a name is **not** specified on the command line (i.e., when the `--dockerfile` option is **not** used).
But the [transposeTarStream function](https://github.com/balena-io-modules/docker-qemu-transpose/blob/v1.0.3/src/index.ts#L213) expects the value to be `undefined` (rather than the empty string) in order to use the default `'Dockerfile'` name, hence adding `or undefined` here.

I have started writing test cases to cover the `build --emulation` functionality, but it's taking some time to get right and this fix is relatively urgent. So I'd like to get this PR through without the automated test cases, and I'll add the tests cases in a subsequent PR.

PR #1653, opened 2020-03-06
Add test case for `build --emulated`

pdcastro:
This PR adds tests that would have caught the bug reported in #1649, introduced by #1628 and already fixed in #1651. There is also a commit that re-adds type checking for Typescript test files, and fixes some type issues that were already present in the master branch.

Connects-to: #1651
Connects-to: #1649 
Connects-to: #1628 
Change-type: patch


package.json
```diff
@@ -50,7 +50,7 @@
     "build:installer": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts build:installer",
     "package": "npm run build:fast && npm run build:standalone && npm run build:installer",
     "release": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts release",
-    "pretest": "npm run build",
+    "pretest": "npm run build && tsc -P tests/tsconfig.json --noEmit",
```
pdcastro: Type checking for Typescript test files had been disabled when the `test` script was modified to use `ts-node/register/transpile-only` instead of `ts-node/register`.  I've learned that `resin-lint --typescript` performs Typescript linting **without** type checking.

tests/commands/build.spec.ts
```diff
@@ -115,6 +116,85 @@ describe('balena build', function() {
...
 
+	it('should create the expected tar stream (--emulated)', async () => {
+		const projectPath = path.join(projectsPath, 'no-docker-compose', 'basic');
+		const transposedDockerfile =
+			stripIndent`
+			FROM busybox
+			COPY [".balena/qemu-execve","/tmp/qemu-execve"]
+			COPY ["./src","/usr/src/"]
+			RUN ["/tmp/qemu-execve","-execve","/bin/sh","-c","chmod a+x /usr/src/*.sh"]
+			CMD ["/usr/src/start.sh"]` + '\n';
```
pdcastro: This test case checks that the tar stream produced by `balena build --emulated` contains a duly transposed Dockerfile that prefixes relevant commands with `"/tmp/qemu-execve"`. (Like all other test cases in this file, the docker daemon is mocked and no image is actually built.)

.gitattributes
```diff
@@ -10,3 +10,6 @@ doc/cli.markdown text eol=lf
 # crlf for the eol conversion test files
 tests/test-data/projects/docker-compose/basic/service2/file2-crlf.sh eol=crlf
 tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh eol=crlf
+
+# Prevent auto merging of the npm-shrinkwrap.json file: see notes in CONTRIBUTING.md
+/npm-shrinkwrap.json merge=binary
```
pdcastro: @Page-, given your careful tidying up of dependencies in the `npm-shrinkwrap.json` file (#1645), I thought we should try and protect it from merge madness. :-)  I got inspiration from [this post](https://blog.scottlogic.com/2018/02/13/npm-shrikwrap-merges-and-conflicts.html), but I am not really super sure that this is the best approach. See also notes added to CONTRIBUTING.md. Thoughts?

pdcastro:
@balena-ci retest

pdcastro:
This PR is ready for review for 5 days... Anyone? :-)
Force push = rebase on master

.gitattributes
Page-: I think it makes sense to avoid auto-merges for the shrinkwrap since they could be fairly dangerous, especially as npm doesn't seem to care if the shrinkwrap makes sense or not and just installs it regardless (see the junk in `net-keepalive`s shrinkwrap that gets installed even though it's not in the package.json for the module)

pdcastro:
@balena-ci retest

pdcastro:
I have dropped the commit "Re-add type checking for Typescript test files" in favour of Page's PR #1656.

tests/commands/build.spec.ts
```diff
@@ -115,6 +116,85 @@ describe('balena build', function() {
...
+				'[Warn] Windows-format line endings were detected in some files. Consider using the `--convert-eol` option.',
+			);
+		}
+		const arch = 'rpi';
+		const deviceType = 'raspberry-pi';
+		const fsModPath = 'mz/fs';
+		const fsMod = await import(fsModPath);
+		const qemuModPath = '../../build/utils/qemu';
+		const qemuMod = require(qemuModPath);
+		const qemuBinPath: string = await qemuMod.getQemuPath(arch);
```
Page-: Is the `: string` actually necessary?
```suggestion
		const qemuBinPath = await qemuMod.getQemuPath(arch);
```

CONTRIBUTING.md
```diff
@@ -58,12 +83,55 @@ the standard Command Prompt or PowerShell can be used (though MSYS2 is still han
...
+```
+
 ## TypeScript vs CoffeeScript, and Capitano vs oclif
 
 The CLI was originally written in [CoffeeScript](https://coffeescript.org), but we decided to
 migrate to [TypeScript](https://www.typescriptlang.org/) in order to take advantage of static
 typing and formal programming interfaces. The migration is taking place gradually, as part of
-maintenance work or the implementation of new features.
+maintenance work or the implementation of new features. The recommended way of making the
+conversion is to first generate plain Javascript by running `npm run build` (look in the `build`
```
Page-: I actually use `npx decaffeinate --use-js-modules x.coffee` as it generates far less junk in the output (and I add the `--use-js-modules` as it makes the exports slightly more typescript-y but it's preference to use/not use it)

tests/commands/build.spec.ts
pdcastro: Done

CONTRIBUTING.md
pdcastro: Great hint, suggestion incorporated 👍

Issue #1523, opened 2019-11-21
Update Raven / Sentry library to the new Unified Node.js SDK


pdcastro:
The CLI is still using the "legacy" Raven Node.js SDK:
https://docs.sentry.io/clients/node/

It should be updated to use Sentry's new "Unified" Node.js SDK
https://docs.sentry.io/error-reporting/quickstart/?platform=node

The API is different, so it's not just a matter of updating the package version in `package.json`. Code changes will be needed.

This upgrade might resolve issue #1333. While doing the upgrade, revisit issue #703, in case any work is pending in that area.

Issue #703, opened 2017-10-25
Filter user data from Sentry error reports


lekkas:
- **resin-cli version:** Version independent
- **Operating system and architecture:** Architecture independent

For instance, directory paths can be omitted

pimterry:
I'm assuming you're on top of this @izavits as part of the analytics updates, but let me know if there's anything I should be doing about this myself :+1:

izavits:
@pimterry yes, I'll let you know when I need your help

pdcastro:
Related work was carried out in multiple PRs since this issue was created. I am not aware of anything specific that is pending, but it would be a good idea to double check, perhaps as part of the work in issue #1523.

pdcastro:
The following commit removed the sending of the full command line to [Sentry.io](https://sentry.io/) in the event of uncaught errors: https://github.com/balena-io/balena-cli/pull/1654/commits/36d3d1256ebd851e2edeab4380a97646126fb5ee

Issue #1333, opened 2019-07-01
CLI prints "null" and exits with error code 1


pdcastro:
This error is rare but can happen with any balena CLI command, any platform (Windows, Mac, Linux) and any installation type (npm dev, npm production, standalone zip, executable installer). The most recent example was issue #1331, and it is closely related to issue #1193.

It happens because of an issue / bug in the Sentry library, which is supposed to intercept errors and report them to Balena via the [sentry.io](https://sentry.io/) platform. It seems that the Sentry library (at least the old version being used) cannot cope with certain types of errors, and then itself fails and "masks" the underlying error, resulting in "null" being printed to the console:

```
$ balena login
null
```

That's a rather frustrating error, difficult to debug. A workaround (to reveal the underlying error) is to sidestep the Sentry library by commenting out (`//`) the `setupRaven()` line in the `build/app-common.js` file, so that it looks as follows:
```
    // setupRaven();
```

If the CLI was installed with the executable (native) installers, the file is located at:

OS | File
-- | --
Windows | `C:\Program Files\balena-cli\client\build\app-common.js`
macOS | `/usr/local/lib/balena-cli/build/app-common.js`

With the "npm installation", it depends on where Node.js was installed and whether `nvm` was used, but typical locations are:

OS | File
-- | --
Linux | /usr/local/lib/node_modules/balena-cli/build/app-common.js
nvm | ~/.nvm/versions/node/vX.Y.Z/lib/node_modules/balena-cli/build/app-common.js

This workaround is not possible in the standalone zip packages, as the source code is bundled in the executable binary.

The first step towards a proper fix is probably to update the Sentry library being used by the CLI (#1523). Unfortunately it's not just a matter of updating a dependency version, because Sentry did a re-writing of the public API.

pdcastro:
Draft pull request: https://github.com/balena-io/balena-cli/pull/1295

pdcastro:
"hello" support conversation: https://app.frontapp.com/open/cnv_1yv269v

pdcastro:
On macOS, in order to edit the `/usr/local/lib/balena-cli/build/app-common.js` file, you can use the `nano` command-line text editor. Open a terminal prompt and run:
```
$ sudo nano /usr/local/lib/balena-cli/build/app-common.js
Password:
```
[...]

Issue #1193, opened 2019-04-26
"Creation of this event was denied due to rate limiting" (Sentry errors mask real errors)


pdcastro:
We use Sentry.io to get CLI error reports. But they apply rate limiting and monthly caps, and when this happens, the error reported on the console to the user is the Sentry error instead of the real cause, making the situation worse and preventing debugging.
Any Sentry-related errors should be either omitted, or made less prominent in favour of the original error.

zvin:
error message (this happened while using `balena os download`:
```
Error: HTTP Error (429): Creation of this event was denied due to rate limiting
    at ClientRequest.<anonymous> (/usr/src/app/node_modules/balena-cli/node_modules/raven/lib/transports.js:56:15)
    at Object.onceWrapper (events.js:300:26)
    at ClientRequest.emit (events.js:210:5)
    at ClientRequest.EventEmitter.emit (domain.js:476:20)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:583:27)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:115:17)
    at TLSSocket.socketOnData (_http_client.js:456:22)
    at TLSSocket.emit (events.js:210:5)
    at TLSSocket.EventEmitter.emit (domain.js:476:20)
    at addChunk (_stream_readable.js:308:12)
    at readableAddChunk (_stream_readable.js:289:11)
    at TLSSocket.Readable.push (_stream_readable.js:223:10)
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:182:23) {
  response: IncomingMessage {
    _readableState: ReadableState {
      objectMode: false,
      highWaterMark: 16384,
      buffer: BufferList { head: null, tail: null, length: 0 },
      length: 0,
      pipes: null,
      pipesCount: 0,
      flowing: null,
      ended: false,
      endEmitted: false,
      reading: false,
      sync: true,
      needReadable: false,
      emittedReadable: false,
      readableListening: false,
      resumeScheduled: false,
      paused: true,
      emitClose: true,
      autoDestroy: false,
      destroyed: false,
      defaultEncoding: 'utf8',
      awaitDrain: 0,
      readingMore: true,
      decoder: [StringDecoder],
      encoding: 'utf8'
    },
    readable: true,
    _events: [Object: null prototype] { end: [Function: responseOnEnd] },
    _eventsCount: 1,
    _maxListeners: undefined,
    socket: TLSSocket {
[...]

PR #1654, opened 2020-03-06
Replace Raven SDK with "new" Sentry "Unified" SDK (fix "CLI prints 'null' and exits")

pdcastro:
Old, deprecated Raven SDK: https://docs.sentry.io/clients/node/
New Sentry SDK: https://docs.sentry.io/error-reporting/quickstart/?platform=node

Resolves: #1523 
Resolves: #703 
Connects-to: #1333 
Connects-to: #1193 

Change-type: patch

### DIFF
--- a/lib/events.ts
+++ b/lib/events.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2019 Balena Ltd.
+ * Copyright 2019-2020 Balena Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,12 +14,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import Promise = require('bluebird');
-import _ = require('lodash');
-import Mixpanel = require('mixpanel');
-import Raven = require('raven');
+import * as Sentry from '@sentry/node';
+import * as Bluebird from 'bluebird';
+import * as _ from 'lodash';
+import * as Mixpanel from 'mixpanel';
 
-import packageJSON = require('../package.json');
+import * as packageJSON from '../package.json';
 import { getBalenaSdk } from './utils/lazy';
 
 const getMixpanel = _.once<any>(() => {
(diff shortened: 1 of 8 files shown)


lib/hooks/prerun/track.ts
```diff
@@ -16,8 +16,12 @@
  */
 import { Hook } from '@oclif/config';
 
+let trackResolve: (result: Promise<any>) => void;
+
 // note: trackPromise is subject to a Bluebird.timeout, defined in events.ts
-export let trackPromise: PromiseLike<void>;
+export const trackPromise = new Promise((resolve, _reject) => {
```
Page-: There's no need to include unused params:
```suggestion
export const trackPromise = new Promise((resolve) => {
```

lib/hooks/prerun/track.ts
```diff
@@ -16,8 +16,12 @@
  */
 import { Hook } from '@oclif/config';
 
+let trackResolve: (result: Promise<any>) => void;
```
Page-: If you want to be fully accurate you can type this as
```suggestion
let trackResolve: ConstructorParameters<PromiseConstructor>[0];
```

lib/global.d.ts
```diff
@@ -20,6 +20,9 @@ interface Dictionary<T> {
 }
 
 declare module '*/package.json' {
```
Page-: If you add `"resolveJsonModule": true` then you can remove this altogether as typescript will automatically generate the types for JSON files

lib/events.ts
```diff
@@ -31,27 +31,36 @@ const getMixpanel = _.once<any>(() => {
...
+	return Bluebird.props({
 		balenaUrl: balena.settings.get('balenaUrl'),
 		username: balena.auth.whoami().catchReturn(undefined),
 		mixpanel: getMixpanel(),
 	})
 		.then(({ username, balenaUrl, mixpanel }) => {
-			return Promise.try(() => {
-				Raven.mergeContext({
-					user: {
+			return Bluebird.try(() => {
```
Page-: This `Bluebird.try` can be removed altogether - you're already in a promise chain

lib/hooks/prerun/track.ts
pdcastro: In this case it may be better not to be fully accurate, :-) because the suggested alternative type is somewhat obscure. It is the same type as that of the resolve argument to `Promise((resolve, reject) => ...)`, which my VS Code IDE tells to be `(value?: unknown) => void`. I replaced `unknown` with a less unknown type.

lib/global.d.ts
pdcastro: I tried it before, but it has side effects. Adding `"resolveJsonModule": true` to `tsconfig.json` somehow changes the build output location, which then breaks `npm run build`:
```
$ npm run build
...
{ Error: Cannot find module '/Users/paulo/balena/balena-io/balena-cli/build/actions/api-key.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at importCapitanoCommands (/Users/paulo/balena/balena-io/balena-cli/automation/capitanodoc/index.ts:56:18)
    at renderMarkdown (/Users/paulo/balena/balena-io/balena-cli/automation/capitanodoc/index.ts:46:8)
    at process._tickCallback (internal/process/next_tick.js:68:7) code: 'MODULE_NOT_FOUND' }
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! balena-cli@11.28.13 build:doc: `mkdirp doc/ && ts-node --transpile-only -P automation/tsconfig.json automation/capitanodoc/index.ts > doc/cli.markdown`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the balena-cli@11.28.13 build:doc script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

`resolveJsonModule` | Build output folder
--- | ---
false | `build/actions/api-key.js`
true | `build/lib/actions/api-key.js`

lib/events.ts
pdcastro: Indeed, done. 👍

lib/global.d.ts
Page-: That'll be because `package.json` is now part of the source files, which means the `tsc` build will include it in the output and hence bump everything down a level (ie `build/{package.json,src/}`)

pdcastro: Thanks! 👍  I was guessing the reason would be something like that, and it's good to have confirmation. I might make this change in some other PR. It would need some extra testing -- like how the standalone zip package behaves with it.

pdcastro:
@balena-ci retest

pdcastro:
Retesting because, curiously, the draft Github release is missing the standalone zip file for Linux, even though the build succeed and the output suggests the file was produced and uploaded to Github. Maybe some uncaught error somewhere. :-/

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

PR #1656, opened 2020-03-10
Add type checking for tests

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


tests/commands/push.spec.ts
```diff
@@ -323,10 +323,6 @@ describe('balena push: project validation', function() {
 		const expectedErrorLines = [
 			'The --nolive flag is only valid when pushing to a local mode device',
 		];
-		const expectedOutputLines = [
-			'[Warn] "docker-compose.y[a]ml" file found in parent directory: please check',
-			"[Warn] that the correct folder was specified. (Suppress with '--noparent-check'.)",
-		];
```
pdcastro: This change and a few others were also made in a PR of mine, #1653, several days older than this PR and still awaiting review... :-)

package.json
```diff
@@ -41,23 +41,23 @@
...
     "prebuild": "rimraf build/ build-bin/",
     "build": "npm run build:src && npm run catch-uncommitted",
-    "build:src": "npm run prettify && npm run lint && npm run build:automation && npm run build:fast && npm run build:doc",
+    "build:src": "npm run prettify && npm run lint && npm run build:fast && npm run build:test && npm run build:doc",
     "build:fast": "gulp build && tsc",
-    "build:automation": "tsc -P automation/tsconfig.json --noEmit",
-    "build:doc": "mkdirp doc/ && ts-node --transpile-only -P automation/tsconfig.json automation/capitanodoc/index.ts > doc/cli.markdown",
-    "build:standalone": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts build:standalone",
-    "build:installer": "ts-node --transpile-only -P automation/tsconfig.json automation/run.ts build:installer",
+    "build:test": "tsc -P ./tsconfig.dev.json --noEmit",
```
pdcastro: In PR #1653 (older and still awaiting review), this change takes the form:
```
"pretest": "npm run build && tsc -P tests/tsconfig.json --noEmit"
```
I don't mind it one way or another, but we can't have both. :-)

package.json
```diff
@@ -41,23 +41,23 @@
...
+    "build:doc": "mkdirp doc/ && ts-node --transpile-only automation/capitanodoc/index.ts > doc/cli.markdown",
```
pdcastro: Nice to get rid of the `-P` argument! 👍

Page-: Yeah I realized it doesn't actually do anything when using `--transpile-only` so it's nicer to have it cleaned up

package.json
Page-: This one covers type-checking for both automation and tests in one go so should be faster in theory. It does include the main code as well that could in theory be removed for speed, but it's nice to have one config to check everything imo, and it inherits all the compiler options from the production tsconfig so it will help to keep that standardized :)

Page-:
It addresses the part of it where the tests aren't type-checked, and it'll address the part about breaking the typings by blocking me from merging that PR until I fix them :P

tests/commands/push.spec.ts
Page-: I hadn't realized that addressed type-checking the tests as the description only talks about adding a test case, but I prefer this approach to type-checking as we merge automation/tests type-checking (which should be faster) and inherit from the main tsconfig which means we'll automatically inherit all of the compiler options/etc

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

package.json
pdcastro: I have now dropped the conflicting commit from PR #1653.

Issue #1657, opened 2020-03-10
`balena login` Error: spawn /snapshot/versioned-source/node_modules/open/xdg-open ENOENT


pdcastro:
- **balena CLI version:** 11.28.14 (ENOENT), 11.25.6 to 11.28.13 (null)
- **Operating system version:** Linux (Ubuntu 18.04)
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** standalone zip
- **If npm install, Node.js and npm version:** 

---

CLI version 11.28.14 :
```
$ ./balena-cli-11.28.14/balena login
 _            _
| |__   __ _ | |  ____  _ __    __ _
| '_ \ / _` || | / __ \| '_ \  / _` |
| |_) | (_) || ||  ___/| | | || (_) |
|_.__/ \__,_||_| \____/|_| |_| \__,_|

Logging in to balena-cloud.com
? How would you like to login? Web authorization (recommended)
Connecting to the web dashboard
Error: spawn /snapshot/versioned-source/node_modules/open/xdg-open ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:240:19)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)```
```

CLI versions 11.25.6 to 11.28.13 :
```
$ ./balena-cli-11.25.6/balena login
 _            _
| |__   __ _ | |  ____  _ __    __ _
| '_ \ / _` || | / __ \| '_ \  / _` |
| |_) | (_) || ||  ___/| | | || (_) |
|_.__/ \__,_||_| \____/|_| |_| \__,_|

Logging in to balena-cloud.com
? How would you like to login? Web authorization (recommended)
Connecting to the web dashboard
null
```

Traced down to PR #1599

PR #1659, opened 2020-03-11
Fix 'balena login' web auth on Linux standalone zip install (xdg-open ENOENT)

pdcastro:
Reported in #1657 and [flowdock](https://www.flowdock.com/app/rulemotion/public-s-community/threads/_4Mbt-BaW0WVsHYT_T140lMPKLe). Issue introduced by the `opn` → `open` upgrade in #1599, affecting CLI releases 11.25.6 to 11.28.14.

Resolves: #1657
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,8 @@
       "build/auth/pages/*.ejs",
       "build/hooks",
       "node_modules/resin-discoverable-services/services/**/*",
+      "node_modules/opn-cli/node_modules/opn/xdg-open",
+      "node_modules/open/xdg-open",
       "node_modules/windosu/*.bat",
       "node_modules/windosu/*.cmd"
     ]
--- a/patches/open+7.0.2.patch
+++ b/patches/open+7.0.2.patch
@@ -1,0 +1,15 @@
+diff --git a/node_modules/open/index.js b/node_modules/open/index.js
+index 5fe49ad..c9194ba 100644
+--- a/node_modules/open/index.js
++++ b/node_modules/open/index.js
+@@ -10,7 +10,9 @@ const pAccess = promisify(fs.access);
+ const pExecFile = promisify(childProcess.execFile);
+ 
+ // Path to included `xdg-open`.
+-const localXdgOpenPath = path.join(__dirname, 'xdg-open');
++const localXdgOpenPath = process.pkg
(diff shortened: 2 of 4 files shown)


pdcastro:
@jtonello, if you would like to double check that this PR fixes the issue with the standalone zip install on your laptop, you can download the zip file from the draft release at:
https://github.com/balena-io/balena-cli/releases/tag/untagged-71502efb5850c9dbd21e

package.json
```diff
@@ -33,6 +33,8 @@
       "build/auth/pages/*.ejs",
       "build/hooks",
       "node_modules/resin-discoverable-services/services/**/*",
+      "node_modules/opn-cli/node_modules/opn/xdg-open",
+      "node_modules/open/xdg-open",
```
pdcastro: I believe the 2 lines above are redundant (`pkg` would automatically include the `xdg-open` script in the standalone executable anyway), however I thought it would be desirable to make the dependency **explicit.** Also, in case dependencies are updated such that `node_modules/opn-cli/node_modules/opn/xdg-open` "disappears", we would quickly find out about it and this would prompt us to consider what else needs to be changed -- like the patch files in the `patches` folder.

patches/open+7.0.2.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/node_modules/open/index.js
++++ b/node_modules/open/index.js
+@@ -10,7 +10,9 @@ const pAccess = promisify(fs.access);
+ const pExecFile = promisify(childProcess.execFile);
+ 
+ // Path to included `xdg-open`.
+-const localXdgOpenPath = path.join(__dirname, 'xdg-open');
++const localXdgOpenPath = process.pkg
++	? path.join(path.dirname(process.execPath), 'xdg-open')
++	: path.join(__dirname, 'xdg-open');
```
pdcastro: This patch is the key fix to the issue.

pdcastro:
I am going to merge this PR sooner than later (now) without waiting for additional reviews, because it is fixing a relatively serious bug, and my testing indicates that the fix works. But post-merge reviews are still welcome, 👍  and changes can be made in another PR.

patches/open+7.0.2.patch
jtonello: Works perfectly, as tested on Ubuntu 18.04.

Issue #1658, opened 2020-03-10
`balena apps` fails with "_.trimEnd is not a function"


pdcastro:
- **balena CLI version:** 11.28.14
- **Operating system version:** Windows 10
- **32/64 bit OS and processor:** 64-bit Intel
- **Install method:** Windows native installer
- **If npm install, Node.js and npm version:** Node v8.15.1, installer built-in

---

```
C:\Users\paulo>set DEBUG=1
C:\Users\paulo>balena apps
[debug] original argv0="C:\Program Files\balena-cli\bin\\..\client\bin\node.exe" argv=[C:\Program Files\balena-cli\client\bin\node.exe,C:\Program Files\balena-cli\client\bin\run,apps] length=3
_.trimEnd is not a function
TypeError: _.trimEnd is not a function
    at C:\Program Files\balena-cli\client\node_modules\resin-cli-visuals\build\widgets\table.js:85:14
    at arrayMap (C:\Program Files\balena-cli\client\node_modules\resin-cli-visuals\node_modules\lodash\index.js:1406:25)
    at Function.map (C:\Program Files\balena-cli\client\node_modules\resin-cli-visuals\node_modules\lodash\index.js:6710:14)
    at trimRight (C:\Program Files\balena-cli\client\node_modules\resin-cli-visuals\build\widgets\table.js:84:18)
    at Object.exports.horizontal (C:\Program Files\balena-cli\client\node_modules\resin-cli-visuals\build\widgets\table.js:120:10)
    at balena.models.application.getAll.then (C:\Program Files\balena-cli\client\build\actions\app.js:78:51)
From previous event:
    at Command.action (C:\Program Files\balena-cli\client\build\actions\app.js:73:14)
    at <anonymous>
```

**However,** there is no error in a "fresh" installation of v11.28.14. It only happens when it is installed on top of an older, pre-installed version, for example v11.28.0. Thus:

**Steps to reproduce:**
1. Use Windows Explorer to delete the following folder and all its contents:
`C:\Program Files\balena-cli`

2. Install CLI v11.28.0: [balena-cli-v11.28.0-windows-x64-installer.exe](https://github.com/balena-io/balena-cli/releases/download/v11.28.0/balena-cli-v11.28.0-windows-x64-installer.exe)

[...]

PR #1662, opened 2020-03-12
Make windows installer remove old files before installation.

srlowe:
This is an npx patch to the [oclif/dev-cli build script](https://github.com/oclif/dev-cli/blob/master/src/commands/pack/win.ts) that creates the windows standalone installer.

This modifies the `nsis` installer script so that the installation section first deletes the `./client` folder, if it exists from a previous installation.

This has been tested extensively, including:
 - As a new install (to verify that a missing `./client` folder does not cause problems)
 - Installed over the top of v11.28.0 (which previously causes errors when doing `balena apps`)

(Background) Initially the idea was to run the uninstaller (if installed) as the first step of installation.  However I wasn't able to resolve one issue: the installer would not wait for the uninstaller to finish before continuing, which lead to the uninstaller and installer running in parallel.  This is caused by the uninstaller copying itself outside of the app directory, and then spawning a new process (so it can delete itself).   This should be suppressible according to the docs, but after multiple attempts with various changes, I could not get this to work (very slow testing/iteration cycle too).

Change-type: patch
Resolves: #1658

### DIFF
--- a/patches/@oclif+dev-cli+1.22.0.patch
+++ b/patches/@oclif+dev-cli+1.22.0.patch
@@ -11,7 +11,7 @@ index cd771cd..4a66939 100644
      }
  }
 diff --git a/node_modules/@oclif/dev-cli/lib/commands/pack/win.js b/node_modules/@oclif/dev-cli/lib/commands/pack/win.js
-index a9d4276..75c2f8b 100644
+index a9d4276..4ac508f 100644
 --- a/node_modules/@oclif/dev-cli/lib/commands/pack/win.js
 +++ b/node_modules/@oclif/dev-cli/lib/commands/pack/win.js
 @@ -3,11 +3,14 @@ Object.defineProperty(exports, "__esModule", { value: true });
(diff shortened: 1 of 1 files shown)


patches/@oclif+dev-cli+1.22.0.patch
```diff
@@ -47,6 +47,20 @@ index a9d4276..75c2f8b 100644
...
+@@ -89,6 +93,13 @@ VIAddVersionKey /LANG=\${LANG_ENGLISH} "ProductVersion" "\${VERSION}.0"
+ InstallDir "\$PROGRAMFILES${arch === 'x64' ? '64' : ''}\\${config.dirname}"
+ 
+ Section "${config.name} CLI \${VERSION}"
++  ; First remove any old client files.
++	; (Remnants of old versions were causing CLI errors)
++	; Initially tried running the Uninstall.exe, but was
++	; unable to make script wait for completion (despite using _?)
++	DetailPrint "Removing files from previous version."
++	RMDir /r "$INSTDIR\\client"
```
pdcastro: I like this solution, it might even run faster than running `Uninstall.exe`.

PR #1668, opened 2020-03-13
Switch to native type checks

Page-:
https://nodejs.org/docs/latest-v8.x/api/util.html#util_deprecated_apis


Page-:
@balena-ci retest

PR #1674, opened 2020-03-19
Add support for `root` property on oclif commands

srlowe:
This mirrors the `root` concept in capitano, allowing commands to specify whether elevated privileges are required to run a command.  When set to true, command will exit with an error if executed without root on Mac/Linux (or if executed by non-Administrator on Windows).

This uses the same underlying library as capitano to test for root/admin.

Change-type: patch

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -16,11 +16,36 @@
  */
 
 import Command from '@oclif/command';
+import { ExpectedError } from './errors';
 
-export default abstract class extends Command {
+export default abstract class BalenaCommand extends Command {
 	/**
 	 * When set to true, command will be listed in `help`,
 	 * otherwise listed in `help --verbose` with secondary commands.
 	 */
 	public static primary = false;
+
+	/**
+	 * Require elevated privileges to run.
+	 * When set to true, command will exit with an error
+	 * if executed without root on Mac/Linux
+	 * or if executed by non-Administrator on Windows.
+	 */
+	public static root = false;
+
+	protected async checkElevatedPrivileges() {
+		const root = (this.constructor as typeof BalenaCommand).root;
+		if (root) {
+			const isElevated = await (await import('is-elevated'))();
+			if (!isElevated) {
+				throw new ExpectedError(
+					'You need admin privileges to run this command',
+				);
+			}
+		}
+	}
+
+	protected async init() {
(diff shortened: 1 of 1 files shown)


lib/command.ts
```diff
@@ -16,11 +16,36 @@
...
+
+	/**
+	 * Require elevated privileges to run.
+	 * When set to true, command will exit with an error
+	 * if executed without root on Mac/Linux
+	 * or if executed by non-Administrator on Windows.
+	 */
+	public static root = false;
+
+	protected async _checkElevated() {
```
pdcastro: I remember someone telling me before that Typescript has special semantics for the underscore prefix (some type checks are disabled). I think your intention was to identify this method as non-public, right? Perhaps the `protected` label is sufficient, and the underscore prefix could be removed.

srlowe: @pdcastro Ah ok.  Yeah my intention was to make it clear that it should not be used in the subclass (which `protected` wont prevent).  Perhaps we could use `private`? (I remember there was some reason not to use `private` that you mentioned, but I can't remember what it was :-) )

srlowe: @pdcastro Maybe you are referring to function parameters that are prefixed by _,  in which case they are allowed to be unused without causing compiler errors.  BTW, oclif itself uses this convention (e.g. `protected _helpOverride()`)

pdcastro: `private` is a good choice if you need to prevent it being overridden in subclasses, but why prevent it?

"The reason not to use private" is simply that most developers seem to make a binary choice between public and private -- anything not intended for the public API is labeled private by default. But I think that `protected` is a better choice than `private` most of the time. It still indicates that the command is not part of the public API, yet does not unnecessarily prevent code reuse and specialisation. A subclass might want to run an additional step before or after calling `super.checkElevated()`, and why prevent it?

srlowe: @pdcastro It's not dangerous in this case, but I'd mainly want to prevent it to remove it (or make it clear it's not to be used) from things like autocomplete, where it causes noise.

pdcastro: Those were the only comments I had, and the PR is approved. :-)

srlowe: @pdcastro Ok, I removed that underscore, and renamed to something a little clearer.

PR #1676, opened 2020-03-24
Use balena-lint for javascript linting and add javascript type-checking

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/automation/check-npm-version.js
+++ b/automation/check-npm-version.js
@@ -7,16 +7,18 @@
  * We don't `require('semver')` to allow this script to be run as a npm
  * 'preinstall' hook, at which point no dependencies have been installed.
  */
-function semverGte(v1, v2) {
-	let v1Array, v2Array; // number[]
-	try {
-		const [, major1, minor1, patch1] = /v?(\d+)\.(\d+).(\d+)/.exec(v1);
-		const [, major2, minor2, patch2] = /v?(\d+)\.(\d+).(\d+)/.exec(v2);
-		v1Array = [parseInt(major1), parseInt(minor1), parseInt(patch1)];
-		v2Array = [parseInt(major2), parseInt(minor2), parseInt(patch2)];
-	} catch (err) {
-		throw new Error(`Invalid semver versions: '${v1}' or '${v2}'`);
+function parseSemver(version) {
+	const match = /v?(\d+)\.(\d+).(\d+)/.exec(version);
+	if (match == null) {
+		throw new Error(`Invalid semver version: ${version}`);
 	}
+	const [, major, minor, patch] = match;
+	return [parseInt(major, 10), parseInt(minor, 10), parseInt(patch, 10)];
+}
+
+function semverGte(v1, v2) {
+	let v1Array = parseSemver(v1);
+	let v2Array = parseSemver(v2);
 	for (let i = 0; i < 3; i++) {
 		if (v1Array[i] < v2Array[i]) {
 			return false;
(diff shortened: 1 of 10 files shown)


automation/check-npm-version.js
```diff
@@ -8,12 +8,21 @@
...
-	let v1Array, v2Array; // number[]
+	let v1Array;
+	let v2Array; // number[]
 	try {
 		const [, major1, minor1, patch1] = /v?(\d+)\.(\d+).(\d+)/.exec(v1);
 		const [, major2, minor2, patch2] = /v?(\d+)\.(\d+).(\d+)/.exec(v2);
-		v1Array = [parseInt(major1), parseInt(minor1), parseInt(patch1)];
-		v2Array = [parseInt(major2), parseInt(minor2), parseInt(patch2)];
+		v1Array = [
+			parseInt(major1, 10),
```
Page-: All these changes came from the linter

lib/utils/device/deploy.ts
```diff
@@ -536,7 +536,7 @@ export function generateTargetState(
 		// this composition entry (everything else in this
 		// function comes from the composition which doesn't
 		// change)
-		let contract = null;
+		let contract;
```
Page-: This was due to the `"noImplicitAny": false,` for the js type checking having an effect on how inference works, but this new form is an improvement imo anyway

tsconfig.js.json
```diff
@@ -0,0 +1,7 @@
+{
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"noImplicitAny": false,
```
Page-: This is necessary as otherwise the js files would have loads of errors due to implicit anys

package.json
```diff
@@ -93,6 +92,7 @@
     }
   },
   "devDependencies": {
+    "@balena/lint": "^4.1.1",
```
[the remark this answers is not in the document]
Page-: Yes it's 5.x that will drop node 8 support, that said it would only affect developing on the CLI with node 8 and not running it, so imo that would be fine to do without a major (but a major dropping node 8 would be really nice as it unlocks a bunch of stuff for the cli to use)

package.json
```diff
@@ -43,9 +43,9 @@
     "postinstall": "patch-package",
     "prebuild": "rimraf build/ build-bin/",
     "build": "npm run build:src && npm run catch-uncommitted",
-    "build:src": "npm run prettify && npm run lint && npm run build:fast && npm run build:test && npm run build:doc",
+    "build:src": "npm run lint && npm run build:fast && npm run build:test && npm run build:doc",
     "build:fast": "gulp build && tsc",
-    "build:test": "tsc -P ./tsconfig.dev.json --noEmit",
+    "build:test": "tsc -P ./tsconfig.dev.json --noEmit && npx tsc -P ./tsconfig.js.json --noEmit",
```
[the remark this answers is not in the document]
Page-: That's true, npm automatically runs these with a modified path like npx will

tsconfig.dev.json
```diff
@@ -4,6 +4,10 @@
 		"./automation/**/*.ts",
 		"./lib/**/*.ts",
 		"./tests/**/*.ts",
-		"./typings/**/*.ts"
+		"./typings/**/*.ts",
+		"./automation/**/*.js",
+		"./lib/**/*.js",
+		"./tests/**/*.js",
+		"./typings/**/*.js"
```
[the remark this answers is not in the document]
Page-: It works, that's pretty sweet - it also made me realize a bug where tsconfig.js.json was inheriting from tsconfig.json instead of tsconfig.dev.json :)

automation/check-npm-version.js
```diff
@@ -7,16 +7,18 @@
  * We don't `require('semver')` to allow this script to be run as a npm
  * 'preinstall' hook, at which point no dependencies have been installed.
  */
-function semverGte(v1, v2) {
-	let v1Array, v2Array; // number[]
-	try {
-		const [, major1, minor1, patch1] = /v?(\d+)\.(\d+).(\d+)/.exec(v1);
```
Page-: I rewrote this because typescript correctly complained that `/v?(\d+)\.(\d+).(\d+)/.exec(v1)` can return `null` which is not valid to destructure

automation/check-npm-version.js
```diff
@@ -27,24 +29,6 @@ function semverGte(v1, v2) {
 	return true;
 }
 
-function _testSemverGet() {
```
Page-: Typescript wasn't happy about this unused function, the `_` prefix only has special meaning in actual typescript code

PR #1677, opened 2020-03-25
Convert lib/actions/local/common.coffee to javascript

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


lib/actions/local/common.js
```diff
@@ -0,0 +1,99 @@
...
+	});
+
+	// List all containers, including those not running
+	return docker.listContainers({ all: true }).then(function(containers) {
+		containers = containers.filter(function(container) {
+			if (!filterSupervisor) {
+				return true;
+			}
+			return filterOutSupervisorContainer(container);
+		});
```
Page-: I switched this from a Bluebird.filter to an Array.filter as the native dockerode promises are not typed as Bluebird promises (they are in this case, but just having a generic promise type) so typescript complained, and we didn't actually need it to be a Bluebird.filter so it should be more performant/clearer with the new version

lib/actions/local/common.js
```diff
@@ -0,0 +1,99 @@
...
+	}
+	const form = require('resin-cli-form');
+	const docker = dockerUtils.createClient({
+		host: deviceIp,
+		port: dockerPort,
+		timeout: dockerTimeout,
+	});
+
+	// List all containers, including those not running
+	return docker.listContainers({ all: true }).then(function(containers) {
```
Page-: Switched to using the native dockerode promise functions because typescript was unaware of our manual promisification and it makes sense to use the native versions now they're available (they still return bluebird promises because docker-toolbelt passes in `{ Promise: Bluebird }`)

lib/utils/docker-coffee.d.ts
```diff
@@ -25,8 +25,13 @@ export interface BuildDockerOptions {
 	docker?: string; // dockerode DockerOptions.socketPath
 	dockerHost?: string; // dockerode DockerOptions.host
 	dockerPort?: number; // dockerode DockerOptions.port
+	host?: string;
+	port?: number;
+	timeout?: number;
```
Page-: These types needed to be added for the sake of lib/actions/local/common.js which uses them

lib/utils/docker-coffee.d.ts
```diff
@@ -25,8 +25,13 @@ export interface BuildDockerOptions {
...
 }
 
 export function getDocker(
 	options: BuildDockerOptions,
 ): Bluebird<DockerToolbelt>;
+
+export function createClient(options: BuildDockerOptions): DockerToolbelt;
```
Page-: this type needed to be added for the sake of lib/actions/local/common.js which uses it

lib/actions/local/common.js
```diff
@@ -0,0 +1,99 @@
+import * as Promise from 'bluebird';
```
Page-: This file will be transpiled like any typescript files, which allows us to use modern javascript functionality like async/await without having to worry about node 8 supporting it or not - I haven't made use of any of it outside of modules because I wanted this to be a minimalistic conversion for maximum safety.  
This file will also be type checked as much as possible and linted correctly (with actual lint errors unlike for coffeescript where it's only warnings)

Page-:
@balena-ci retest

Page-:
@balena-ci retest

PR #1556, opened 2020-01-03
Fix 'balena push' hanging on Windows

pdcastro:
Resolves: #1182
Resolves: #1342
Resolves: #1554
Change-type: patch
Signed-off-by: Paulo Castro <paulo@balena.io>

PR #1678, opened 2020-03-25
Convert commands `settings`, `note` to oclif

srlowe:
Convert commands `settings`, `note` to oclif. 
Add oclif support for piped input.

Change-type: patch


lib/actions-oclif/note.ts
```diff
@@ -0,0 +1,84 @@
+/**
```
srlowe: Capitano file was `notes.ts`, have renamed to `note.ts` to match command.

lib/actions-oclif/note.ts
```diff
@@ -0,0 +1,84 @@
...
+`;
+	public static examples = [
+		'$ balena note "My useful note" --device 7cf02a6',
+		'$ cat note.txt | balena note --device 7cf02a6',
+	];
+
+	public static args = [
+		{
+			name: 'note',
+			required: true,
```
pdcastro: I was going to say that the `note` args is not required (`required: false`) because when not provided, the command reads from stdin. However I see in the old implementation:
```
		if (_.isEmpty(params.note)) {	
			const { exitWithExpectedError } = await import('../utils/patterns');	
			exitWithExpectedError('Missing note content');	
		}
```

So the command never actually attempts reading from stdin, right? In which case, the argument is indeed required, and what we need to change is the help output, which should not mention reading from stdin...

lib/preparser.ts
```diff
@@ -133,6 +133,8 @@ export const convertedCommands = [
 	'env:rm',
 	'os:configure',
 	'version',
+	'settings',
+	'note',
```
pdcastro: The list was in alphabetical order until this PR. :-)  (The CLI has some 60 commands/subcommands, eventually alphabetical order will make it easier to find something in the list.)

lib/actions-oclif/note.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { flags } from '@oclif/command';
+import { stripIndent } from 'common-tags';
+import Command from '../command';
+import * as cf from '../utils/common-flags';
+import { getBalenaSdk } from '../utils/lazy';
+import { CommandHelp } from '../utils/oclif-utils';
+
+interface FlagsDef {
+	device: string; // device UUID
```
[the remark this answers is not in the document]
srlowe: Oh damn, I meant to do that, but forgot (finished this off after support shift).

lib/actions-oclif/note.ts
```diff
@@ -0,0 +1,84 @@
...
+	note: string;
+}
+
+export default class NoteCmd extends Command {
+	public static description = stripIndent`
+		Set a device note.
+
+		Use this command to set or update a device note.
+
+		If note command isn't passed, the tool attempts to read from \`stdin\`.
```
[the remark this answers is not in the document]
srlowe: Yes right, hadn't noticed that.

lib/actions-oclif/note.ts
srlowe: It's supposed to accept piped input. I though I had tested this, but in tests just now it seems it's not working.  Will look into it.

lib/actions-oclif/note.ts
```diff
@@ -0,0 +1,84 @@
...
+			description: 'note content',
+		},
+	];
+
+	public static usage = (
+		'note ' + new CommandHelp({ args: NoteCmd.args }).defaultUsage()
+	).trim();
+
+	public static flags: flags.Input<FlagsDef> = {
+		device: flags.string({
```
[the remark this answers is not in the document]
srlowe: Ah, thanks for the extension example.

lib/actions-oclif/note.ts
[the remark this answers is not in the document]
srlowe: @pdcastro How about using the `hidden` attribute with `--dev`?  I.e. we still support it, but it doesn't get mentioned in the help.  (This way we could reduce some of the multiple aliases without breaking user scripts)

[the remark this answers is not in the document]
pdcastro: The `hidden` attribute sounds good. 👍 (I don't think we've used it so far, probably good to test it...)

lib/actions-oclif/note.ts
srlowe: Yeah seems oclif doesn't support piped input, but there is mention of a workaround here (trying to figure out what he meant)

https://github.com/oclif/oclif/issues/245#issuecomment-576469479

pdcastro: Before this PR, it was not working with Capitano either:
```
$ balena note --device 123
Missing note content
```
Also as per code snippet I quoted above. Therefore, I'm happy with simply amending the help output to remove mention of stdin.

srlowe: @pdcastro these currently work with capitano though:
```
$ cat test.txt | balena note -d 123

$ bin/balena note -d 123 < test.txt
```

pdcastro: Oh okay. In this case we need to get it working with oclif as well, or leave the Capitano implementation for now. (We could also drop support from reading from stdin in a major CLI release like v12.) Someone has replied to your comment in the oclif issue, maybe we can implement something like that. We'd also need to test with cmd.exe, PowerShell and bash... to at least maintain what currently works.

srlowe: I've added piped input support, using the same library that capitano uses.  It should behave exactly the same whatever the platform, but testing now.

lib/actions-oclif/note.ts
```diff
@@ -0,0 +1,100 @@
...
+
+	public static args = [
+		{
+			name: 'note',
+			// required: true,
+			description: 'note content',
+		},
+	];
+
+	public static usage = 'note <|note>';
```
srlowe: I reverted to including the pipe (required string literal for now).

lib/actions-oclif/note.ts
srlowe: Now tested working on Windows cmd, powershell, msys

lib/actions-oclif/note.ts
```diff
@@ -0,0 +1,100 @@
...
+	public static authenticated = true;
+
+	public static readStdin = true;
+
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			NoteCmd,
+		);
+
+		params.note = params.note || this.stdin;
```
pdcastro: Interesting. If stdin "took a long time" to provide data and close, would the `init()` function block and prevent the execution flow from reaching this line? (Is that how it works?)

lib/command.ts
```diff
@@ -74,18 +83,29 @@ export default abstract class BalenaCommand extends Command {
...
 			await BalenaCommand.checkElevatedPrivileges();
 		}
 
-		if (requireAuthentication) {
+		if (ctr.authenticated) {
 			await BalenaCommand.checkLoggedIn();
 		}
+
+		if (ctr.readStdin) {
+			await this.getStdin();
```
pdcastro: hmm but stdin should only be read if the `notes` argument is not provided, no? Apparently the `key add` command conditionally reads from `stdin`:
https://github.com/balena-io/balena-cli/blob/v11.30.6/lib/actions/keys.ts#L99-L116

> I've added piped input support, using the same library that capitano uses

Does Capitano read from stdin regardless or whether the argument was provided? If there are no side effects then I suppose it is OK. Things that come mind are interactive menus, or a password prompt.

> Now tested working on Windows cmd, powershell, msys

And Linux? :-)  (msys uses a CLI release for Windows...)

srlowe: > hmm but stdin should only be read if the notes argument is not provided, no?

> Does Capitano read from stdin regardless or whether the argument was provided?

I'm not sure about capitano.  Can you elaborate on a situtation you are worried about?  I don't see reading the contents as harmful, whether or not we use them later.  If we wanted to do it conditionally we would need to have the base class responsible for triggering the parsing of commands (I plan to try to do this, but I wanted to keep things simple first, so that we can make some progress without getting bogged down).  (The fact that oclif relies on the concrete implemetation to call the parse method seems like a strange design choice and really limiting to me).

> And Linux? :-) (msys uses a CLI release for Windows...)

Yep, I always test on linux (but should have been clearer).

lib/actions-oclif/note.ts
srlowe: Yes, it's my understanding that init will block.  I suppose we could try to add a timeout here, but I think that might be over the top (and capitano has no such mechanism).

lib/command.ts
pdcastro: > Things that come mind are interactive menus, or a password prompt.
>> Can you elaborate on a situtation you are worried about?

Basically whether there'd be side effects that prevented something else from working. In #1556 we added `process.stdin.pause()` before the CLI exits, as otherwise the CLI was hanging trying to read from stding -- so that kind of situation (maybe it was Capitano's behaviour that required that line to be added, and I didn't even know about it...) But as you tested it and it works, it should be fine. Approving the PR. 👍

srlowe: Ok thanks, I'll keep an eye open for that! (I haven't managed to provoke anything strange in tests yet)

Issue #1679, opened 2020-03-26
Improve `balena device os-update` help to clarify that it only works for balenaCloud


srlowe:
The balena device os-update command works on balenaCloud only (not openBalena or "standalone" balenaOS).

This should be made clear in the command's help output.

PR #1680, opened 2020-03-26
Clarify `balena device os-update` help re balenaCloud

srlowe:
Change-type: patch
Resolves: #1679


lib/actions/device_ts.ts
```diff
@@ -39,6 +39,8 @@ export const osUpdate: CommandDefinition<OsUpdate.Args, OsUpdate.Options> = {
 		Notice this command will ask for confirmation interactively.
 		You can avoid this by passing the \`--yes\` boolean option.
 
+		Requires balenaCloud; will not work with standalone devices.
```
pdcastro:
```
Requires balenaCloud; will not work with openBalena or standalone balenaOS.
```

Issue #1681, opened 2020-03-27
balena ssh: sh: -c: line 0: unexpected EOF while looking for matching `"’


pdcastro:
https://forums.balena.io/t/ssh-to-service-in-local-mode-fails-when-using-local-address/86201/

```
$ balena ssh da3cefd.local main
[debug] original argv0=“D:\Program Files\balena-cli\client\bin…\bin\node.exe” argv=[D:\Program Files\balena-cli\client\bin\node.exe,D:\Program Files\balena-cli\client\bin\run,ssh,da3cefd.local,main] length=5
sh: -c: line 0: unexpected EOF while looking for matching `"’
sh: -c: line 1: syntax error: unexpected end of file
Connection to da3cefd.local closed.
```

I was able to reproduce it with a MSYS 2 shell on Windows 10, and I am working on a fix. Presumably it would also happen with MinGW / MSYS 1, and "git for Windows".

**Known workarounds**
* Run `balena ssh` on a native Windows “Command Prompt” (cmd.exe) or PowerShell, with a CLI release for Windows.
* Run `balena ssh` on a bash shell of the Windows Subsystem for Linux (WSL 1) (e.g. selecting Ubuntu in the Microsoft app store), and using a balena CLI release for **Linux** (standalone zip package).

PR #1685, opened 2020-03-27
Fix balena ssh to local device (MSYS/Windows and command pipe to service)

pdcastro:
Resolves: #1681
Change-type: patch


lib/actions/local/ssh.coffee
```diff
@@ -1,113 +0,0 @@
-###
```
pdcastro: This file implemented `balena local ssh` (as opposed to `balena ssh`), but since CLI v11.0.0 it it no longer available:
```
$ ./bin/balena local ssh
Note: the command "balena local ssh" was replaced in CLI version v11.0.0.
Please use "balena ssh" instead.
```
I don't remember why it wasn't deleted before: probably just forgotten.

lib/actions/ssh.ts
```diff
@@ -80,30 +79,29 @@ async function getContainerId(
...
 		const { deviceContainerEngineBinary } = await import('../utils/device/ssh');
 
-		const command = generateVpnSshCommand({
+		const sshBinary = await which('ssh');
+		const sshArgs = generateVpnSshCommand({
 			uuid,
 			verbose: false,
 			port: sshOpts.port,
-			command: `host ${uuid} '"${deviceContainerEngineBinary}" ps --format "{{.ID}} {{.Names}}"'`,
+			command: `host ${uuid} "${deviceContainerEngineBinary}" ps --format "{{.ID}} {{.Names}}"`,
```
pdcastro: The single quotes had to be removed because this PR reduces "one count of shell command parsing". Previously, this string was parsed by either `cmd.exe` or `bash` before being passed on to `ssh`, now it is given directly to `ssh`.

lib/actions/ssh.ts
```diff
@@ -80,30 +79,29 @@ async function getContainerId(
...
-			subShellCommand.args,
-			{
-				stdio: [null, 'pipe', null],
-			},
-		);
+		if (process.env.DEBUG) {
+			console.error(`[debug] [${sshBinary}, ${sshArgs.join(', ')}]`);
+		}
+		const subprocess = child_process.spawn(sshBinary, sshArgs, {
+			stdio: [null, 'pipe', null],
```
pdcastro: The documentation says that `null` defaults to `pipe`, so presumably `[null, 'pipe', null]` means the same as `pipe`. But I also seem to remember we had found corner cases on Windows (?) with certain versions of Node (?), so I thought I'd better leave it alone.

lib/actions/ssh.ts
```diff
@@ -80,30 +79,29 @@ async function getContainerId(
 		// container
 		const { child_process } = await import('mz');
 		const escapeRegex = await import('lodash/escapeRegExp');
-		const { getSubShellCommand } = await import('../utils/helpers');
+		const { which } = await import('../utils/helpers');
 		const { deviceContainerEngineBinary } = await import('../utils/device/ssh');
 
-		const command = generateVpnSshCommand({
+		const sshBinary = await which('ssh');
```
pdcastro: This translates `'ssh'` to something like:
* `/usr/bin/ssh` (macOS)
* `C:\WINDOWS\System32\OpenSSH\ssh.EXE` (Powershell, cmd.exe)
* `C:\msys64\usr\bin\ssh.EXE` (MSYS2)

... without relying on invoking a shell to do it, instead looking directly at the PATH env var (which is what the shells do), thus allowing us to avoid dealing with the quoting and escaping peculiarities of each shell / platform combination.

lib/actions/ssh.ts
```diff
@@ -80,30 +79,29 @@ async function getContainerId(
...
+			console.error(`[debug] [${sshBinary}, ${sshArgs.join(', ')}]`);
+		}
+		const subprocess = child_process.spawn(sshBinary, sshArgs, {
+			stdio: [null, 'pipe', null],
+		});
 		const containers = await new Promise<string>((resolve, reject) => {
-			let output = '';
-			subprocess.stdout.on('data', chunk => (output += chunk.toString()));
+			const output: string[] = [];
+			subprocess.stdout.on('data', chunk => output.push(chunk.toString()));
```
pdcastro: This is not any bug fix, just applying a best practice pattern (adding strings to an array and joining them later reduces the amount of memory copy, compared to repeated string concatenation).

lib/actions/ssh.ts
```diff
@@ -145,15 +143,17 @@ function generateVpnSshCommand(opts: {
...
+		...(opts.verbose ? ['-vvv'] : []),
+		'-t',
+		...['-o', 'LogLevel=ERROR'],
+		...['-o', 'StrictHostKeyChecking=no'],
+		...['-o', 'UserKnownHostsFile=/dev/null'],
+		...(opts.proxyCommand ? [opts.proxyCommand] : []),
+		...(opts.port ? ['-p', opts.port.toString()] : []),
+		`${opts.username}@ssh.${opts.proxyUrl}`,
+		opts.command,
+	];
```
pdcastro: This function now returns an array of arguments instead of a string, which avoids the need for escaping arguments in a shell-specific way (cmd.exe _vs._ bash).

lib/actions/ssh.ts
```diff
@@ -259,22 +257,16 @@ export const ssh: CommandDefinition<
 		}
 
 		// this will be a tunnelled SSH connection...
-		await exitIfNotLoggedIn();
+		await checkLoggedIn();
```
pdcastro: I've used the opportunity to replace `exitIfNotLoggedIn` (which calls `process.exit`) with `checkLoggedIn` (which throws a subclass of ExpectedError), which is our current best practice (it allows errors to bubble up to a centralised error processing function in `errors.ts`, and avoids the side effects of `process.exit` such as discarding data still in buffers -- corrupt files, partial console output).

lib/actions/ssh.ts
```diff
@@ -259,22 +257,16 @@ export const ssh: CommandDefinition<
...
-			const device = await sdk.models.device.get(uuid, {
-				$select: ['id', 'supervisor_version', 'is_online'],
-			});
-			id = device.id;
-			version = device.supervisor_version;
-		} catch (e) {
-			if (e instanceof BalenaDeviceNotFound) {
-				exitWithExpectedError(`Could not find device: ${uuid}`);
-			}
-		}
```
pdcastro: I found that this `try/catch` block was potentially hiding errors other than `BalenaDeviceNotFound`. And for `BalenaDeviceNotFound` itself, it is already handled in `errors.ts` and can be allowed to bubble up:
```
$ ./bin/balena ssh whatwasthat
BalenaDeviceNotFound: Device not found: whatwasthat
```

lib/utils/device/ssh.ts
```diff
@@ -93,36 +91,30 @@ export async function performLocalDeviceSSH(
...
+	return whichSpawn('ssh', [
+		...(opts.verbose ? ['-vvv'] : []),
+		'-t',
+		...['-p', opts.port ? opts.port.toString() : '22222'],
+		...['-o', 'LogLevel=ERROR'],
+		...['-o', 'StrictHostKeyChecking=no'],
+		...['-o', 'UserKnownHostsFile=/dev/null'],
+		`root@${opts.address}`,
+		...(command ? [command] : []),
+	]);
```
pdcastro: This is the key bug fix of issue #1681: replacing `getSubShellCommand` and the execution of either `cmd.exe` or `sh`, with a direct execution of the `ssh` binary without any shell as intermediary.

lib/utils/helpers.ts
```diff
@@ -187,35 +188,6 @@ export function getApplication(applicationName: string) {
...
- * array instead of a long string. Avoiding a shell has several benefits:
- *   - Avoids the need to shell-escape arguments, especially nested commands.
- *   - Bypasses the incompatibilities between cmd.exe and /bin/sh.
- *   - Reduces the security risks of lax input validation.
- * Code example avoiding a shell:
- *    const program = await which('ssh');
- *    const args = ['root@192.168.1.1', 'cat /etc/os-release'];
- *    const child = spawn(program, args);
- */
-export function getSubShellCommand(command: string) {
```
pdcastro: This function is being deleted because, of course, there was not any remaining usage of it, but also because I think it encourages a bad pattern of executing a shell in the first place. In the past I had already fixed bugs because of the usage of this function, and now again in this PR. All previous invocations of shells were limited to taking advantage of a shell's ability to select the full path of an executable given its "short name", like 'ssh' instead of `C:\WINDOWS\System32\OpenSSH\ssh.EXE`, at the cost of us having to deal with argument quoting and escaping in a shell-specific way. The `which()` function + spawn with argument in an array (instead of a plain string) is a better pattern.

lib/utils/helpers.ts
```diff
@@ -406,6 +378,40 @@ export async function which(
 	return programPath;
 }
 
+/**
+ * Call which(programName) and spawn() with the given arguments.
+ * Reject the promise if the process exit code is not zero.
+ */
+export async function whichSpawn(
```
pdcastro: This function is mainly to reduce boilerplate error handling associate with `spawn`, including checking the child process exit code.

README.md
```diff
@@ -88,21 +88,31 @@ HTTP(S) proxies can be configured through any of the following methods, in prece
...
+> server, it should be configured with the following rules in the `squid.conf` file:  
+> `acl SSL_ports port 22`  
+> `acl Safe_ports port 22`  
+
+#### Proxy exclusion
+
+The `BALENARC_NO_PROXY` variable may be used to exclude specified destinations from proxying.
+
+> * This feature requires balena CLI version 11.30.7 or later. In the case of the npm [installation
+>   option](https://github.com/balena-io/balena-cli/blob/master/INSTALL.md), it also requires
```
pdcastro: This is a correct but simplified statement. Earlier CLI versions between 11.24.0 and 11.30.7 added support for some combinations of installation type and platform (standalone zip / GUI installer / npm, **x** Windows / macOS / Linux). It all depends on the version of Node.js that gets bundled with the installer (BALENARC_NO_PROXY requires Node.js v10.16.0 or later). Two significant events took place since I had last updated these proxy instructions:
* [pkg](https://www.npmjs.com/package/pkg) updated "the Node 10 version" that they patch and bundle in the single-file binary from 10.15.3 to 10.17.0. I think this took place around early March, and I had not even realised.
* Last week, Page updated the host OS Node.js in the balena CI Windows worker from 8.15.1 to 10.19.0.

README.md
```diff
@@ -88,21 +88,31 @@ HTTP(S) proxies can be configured through any of the following methods, in prece
...
+>   Node.js version 10.16.0 or later.
```
pdcastro: This is a correct but simplified statement. Earlier CLI versions between 11.24.0 and 11.30.7 added support for some combinations of installation type and platform (standalone zip / GUI installer / npm, **x** Windows / macOS / Linux). It all depends on the version of Node.js that gets bundled with the installer (BALENARC_NO_PROXY requires Node.js v10.16.0 or later). Two significant events took place since I had last updated these proxy instructions:
* [pkg](https://www.npmjs.com/package/pkg) updated "the Node 10 version" that they patch and bundle in the single-file binary from 10.15.3 to 10.17.0. I think this took place around early March, and I had not even realised.
* Last week, Page updated the host OS Node.js in the balena CI Windows worker from 8.15.1 to 10.19.0.

lib/actions/ssh.ts
```diff
@@ -295,27 +290,32 @@ export const ssh: CommandDefinition<
...
-			const ProxyCommand = `proxytunnel ${bash.args(tunnelOptions, '--', '=')}`;
-			return `-o ${bash.args({ ProxyCommand }, '', '=')}`;
+			return [
+				'proxytunnel',
+				`--proxy=${p.host}:${p.port}`,
+				// ssh replaces these %h:%p variables in the ProxyCommand option
+				// https://linux.die.net/man/5/ssh_config
+				'--dest=%h:%p',
+				...(verbose ? ['--verbose'] : []),
+			];
```
pdcastro: Modified to return an array of unescaped strings instead of a bash-escaped string, which is platform centric and does not work with PowerShell / cmd.exe. By using the PROXYUSER and PROXYPASS env vars set above, it was possible to avoid shell escaping and quoting altogether. To be fair, `proxytunnel` is not available on PowerShell / cmd.exe anyway, but being shell-agnostic is a good thing. This also allowed removing the `bash` dependency from `package.json`.

lib/utils/device/ssh.ts
```diff
@@ -93,36 +91,30 @@ export async function performLocalDeviceSSH(
...
-			} ${shellCmd}`;
-		}
+		const containerId = containers[0]!.id;
+		const shellCmd = `/bin/sh -c "if [ -e /bin/bash ]; then exec /bin/bash; else exec /bin/sh; fi"`;
+		// stdin (fd=0) is not a tty when data is piped in, for example
+		// echo 'ls -la; exit;' | balena ssh 192.168.0.20 service1
+		// See https://www.balena.io/blog/balena-monthly-roundup-january-2020/#charliestipsntricks
+		//     https://assets.balena.io/newsletter/2020-01/pipe.png
+		const ttyFlag = isatty(0) ? '-t' : '';
+		command = `${deviceContainerEngineBinary} exec -i ${ttyFlag} ${containerId} ${shellCmd}`;
```
pdcastro: Using `exec -i` instead of `exec -ti` fixes a bug (error "the input device is not a TTY") when a command is piped to ssh, e.g.:
```
echo 'ls -la; exit;' | balena ssh 192.168.0.20 service1
```

package.json
```diff
@@ -177,7 +177,6 @@
     "balena-semver": "^2.2.0",
     "balena-settings-client": "^4.0.4",
     "balena-sync": "^10.2.0",
-    "bash": "0.0.1",
```
pdcastro: It's been version 0.0.1 for 9 years, :-) and just over 1K downloads a week. Anyway, no longer needed. The CLI also uses another dependency that overlaps in functionality: `shell-escape`.

PR #1693, opened 2020-04-02
Convert `internal scandevices`, `internal osinit` to typescript & oclif

srlowe:
Change-type: patch


lib/actions-oclif/internal/osinit.ts
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { stripIndent } from 'common-tags';
+import Command from '../../command';
+import { CommandHelp } from '../../utils/oclif-utils';
+
+// These are internal commands we want to be runnable from the outside.
+// One use-case for this is spawning the minimal operation with root privileges.
```
srlowe, line 23: Could anyone elaborate on these notes (taken from old version)? I'm not really sure what these are used for.

lib/actions-oclif/internal/osinit.ts
```diff
@@ -0,0 +1,76 @@
...
+
+interface ArgsDef {
+	image: string;
+	type: string;
+	config: string;
+}
+
+export default class OsinitCmd extends Command {
+	public static description = stripIndent`
+		Do actual init of the device with the preconfigured os image.
```
srlowe, line 33: Anyone know if  `actual` has any special meaning here?

lib/actions-oclif/internal/osinit.ts
```diff
@@ -0,0 +1,76 @@
...
+// These are internal commands we want to be runnable from the outside.
+// One use-case for this is spawning the minimal operation with root privileges.
+
+interface ArgsDef {
+	image: string;
+	type: string;
+	config: string;
+}
+
+export default class OsinitCmd extends Command {
```
srlowe, line 31: Could anyone provide example usage of this command?  (I haven't been able to fully test)

lib/actions-oclif/internal/osinit.ts
pdcastro: It sounds like the idea was to run "internal commands" with different privileges (either elevating to root, or dropping root), but my impression was always that it is unnecessary to have "internal commands" for that: it is possible to dynamically gain or drop privileges regardless of internal commands. For code reuse, functions / classes / modules are preferable / simpler than running "internal commands". See the following PR and comments where I deleted "internal sudo":
* https://github.com/balena-io/balena-cli/pull/1455/files
* https://github.com/balena-io/balena-cli/pull/1455#discussion_r334308357
* https://github.com/balena-io/balena-cli/pull/1455#discussion_r334308526

So my impression is that getting rid of internal commands would be a good thing.

lib/actions-oclif/internal/osinit.ts
pdcastro: It's an internal command, so do a text search of the CLI code for `osinit` and find out how it is being invoked...

Depending on how far you wanted to take the PR / refactoring, you could perhaps ***not*** create the `internal osinit` and `internal scandevices` commands, and instead just refactor the code as modules / functions (rather than commands) and have the callers import the modules and call the function.

lib/actions-oclif/internal/osinit.ts
pdcastro: iirc, this command flashes a SD card (does it really?). I guess what they meant by "actual" is that `balena os initialize` does some "preparations" ahead of the "actual (real) initialization" which is done by the internal `osinit` command.

typings/capitano/index.d.ts
```diff
@@ -41,6 +41,7 @@ declare module 'capitano' {
 		permission?: 'user';
 		root?: boolean;
 		primary?: boolean;
+		hidden?: booleanl;
```
pdcastro, line 44: `booleanl` - ending with `l` - is that a thing? :-)

srlowe: Huh.  That's not checked by the compiler? TIL.

lib/actions-oclif/internal/osinit.ts
srlowe: I see thanks.  I thought internal just meant "for internal use".

Regarding the extended refactor, I think I'll limit this PR to the ts/oclif conversion (not that it's done), and add a comment that this should be done at some point.

typings/capitano/index.d.ts
pdcastro: Something went wrong, because it is meant to be type checked:
https://github.com/balena-io/balena-cli/blob/v11.30.13/tsconfig.dev.json#L7

Also, if all manual and automated tests passed despite the typo, is the line even necessary?

lib/actions-oclif/internal/osinit.ts
pdcastro: > I haven't been able to fully test

The occurrence I found for `internal osinit` was in `lib/actions/os.coffee`, in the implementation of the `os initialize` command. For `scandevices`, I found an occurrence in `lib/utils/promote.ts`, used by the `balena join` and `balena leave` commands. Please test that these commands still work with the changes in this PR (I mean testing `os initialize`, `balena join`, `balena leave`).

pdcastro: https://www.flowdock.com/app/rulemotion/i-cli/threads/yveDZUzzT4f5g2INiLiD9s83UDJ

typings/capitano/index.d.ts
srlowe: @pdcastro It's not technically necessary anymore (and I guess that's why it wasn't checked, though I'm suprised even unused definitions don't undergo some kind of checking).  I added it, because I had initial converted the capitano command to ts, before the oclif conversion (I must have added the typo sometime after that I suppose, as it was working). 

I decided to leave it in, as technically the definition is not complete without it.

PR #1571, opened 2020-01-16
Add project directory validation for balena push / build / deploy

pdcastro:
Connects-to: #1421
Resolves: #1619 
Change-type: minor
Signed-off-by: Paulo Castro <paulo@balena.io>

PR #1694, opened 2020-04-02
Fix project directory validation for 'balena deploy' with pre-built image

pdcastro:
Resolves an issue reported in the following forum thread, regarding `balena deploy <image>` failing project directory validation when a pre-built image is given as argument:
https://forums.balena.io/t/balena-deploy-app-image-fails/89372

Change-type: patch

### DIFF
--- a/lib/actions/deploy.coffee
+++ b/lib/actions/deploy.coffee
@@ -198,7 +198,7 @@ module.exports =
 		require('events').defaultMaxListeners = 1000
 		sdk = getBalenaSdk()
 		{ ExpectedError } = require('../errors')
-		{ validateProjectDirectory } = require('../utils/compose_ts')
+		{ getRegistrySecrets, validateProjectDirectory } = require('../utils/compose_ts')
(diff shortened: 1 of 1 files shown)


lib/actions/deploy.coffee
```diff
@@ -225,16 +225,21 @@ module.exports =
...
+				validateProjectDirectory(sdk, {
+					dockerfilePath: options.dockerfile,
+					noParentCheck: options['noparent-check'] || false,
+					projectPath: options.source || '.',
+					registrySecretsPath: options['registry-secrets'],
+				})
+				.then ({ dockerfilePath, registrySecrets }) ->
+					options.dockerfile = dockerfilePath
+					options['registry-secrets'] = registrySecrets
+		.then ->
```
pdcastro, line 242: The explanation is that when I added project directory validation in #1571, I forgot that the `balena deploy <image>` command, for a pre-built image, does not use any project directory at all (nothing to build), so the current working directory should not be validated as if it was a project directory.

PR #1702, opened 2020-04-06
 Convert commands leave, join to oclif.

srlowe:
Convert commands leave, join to oclif.

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


lib/utils/oclif-utils.ts
```diff
@@ -26,9 +26,9 @@ export class CommandHelp {
...
 
 	protected arg(arg: Config.Command['args'][0]): string {
-		const name = arg.name.toUpperCase();
+		const name = arg.name;
 		if (arg.required) {
-			return `${name}`;
+			return `<${name}>`;
 		}
 		return `[${name}]`;
 	}
```
srlowe, line 34: These changes required to preserve camelCase in usages.

lib/utils/oclif-utils.ts
```diff
@@ -54,6 +54,5 @@ export function capitanoizeOclifUsage(
 ): string {
 	return (oclifUsage || '')
 		.toString()
-		.replace(/(?<=\s)[A-Z]+(?=(\s|$))/g, match => `<${match}>`)
-		.toLowerCase();
+		.replace(/(?<=\s)[A-Z]+(?=(\s|$))/g, match => `<${match}>`);
 }
```
srlowe, line 58: These changes required to preserve camelCase in usages.

tests/commands/help.spec.ts
```diff
@@ -25,8 +25,8 @@ Primary commands:
     preload <image>                         preload an app on a disk image (or Edison zip archive)
     build [source]                          Build a single image or a multicontainer project locally
     deploy <appName> [image]                Deploy a single image or a multicontainer project to a balena application
-    join [deviceIp]                         Promote a local device running balenaOS to join an application on a balena server
-    leave [deviceIp]                        Detach a local device from its balena application
+    join [deviceIpOrHostname]               promote a local device running balenaOS to join an application on a balena server
+    leave [deviceIpOrHostname]              detach a local device from its balena application
```
srlowe, line 29: Changed param name to better reflect meaning.

lib/utils/oclif-utils.ts
srlowe: (The regex replace on ln57 doesn't seem to have any effect as far as I can tell).

pdcastro: The regex was meant to add the angle brackets in Capitano style for the `balena help` output only -- so the angle brackets are not added to `cli.markdown`.

lib/utils/oclif-utils.ts
pdcastro: This would be a departure from oclif's framework default style:
https://github.com/oclif/plugin-help/blob/v2.2.3/src/command.ts#L115-L117

I understand that oclif intentionally uses all upper case, instead of angle brackets, as a visual indication of a value that should be replaced by the user in actual usage. It is a framework style where oclif differs from Capitano. This change is also affecting the oclif style already in use in documentation (`cli.markdown`) for previously converted CLI commands like `os configure`, `env add`, etc.

The `capitanoizeOclifUsage` function below, as the name suggests, attempts to re-format oclif's style to match Capitano, but it is used specifically for the `balena help` output that presents the command usage as one command per line, where the difference between oclif and Capitano would be too obvious / too hard on the eye.

doc/cli.markdown
```diff
@@ -806,7 +806,7 @@ select a service variable (may be used together with the --device option)
 
 do not prompt for confirmation before deleting the variable
 
-## env add NAME [VALUE]
+## env add &#60;name&#62; [value]
```
pdcastro, line 809: Also if we were to make this change to `env add` and other previously converted commands, we'd have to change other help sections as well, like ARGUMENTS and EXAMPLES below:

**Before this PR:** (usage line consistent with ARGUMENTS and EXAMPLES sections)
```
$ balena help env add
Add an environment or config variable to an application, device or service.

USAGE
  $ balena env add NAME [VALUE]

ARGUMENTS
  NAME   environment or config variable name
  VALUE  variable value; if omitted, use value from this process' environment

EXAMPLES
  $ balena env add TERM --application MyApp
  $ balena env add EDITOR vim --application MyApp
  $ balena env add EDITOR vim --application MyApp --service MyService
  $ balena env add EDITOR vim --device 7cf02a6
  $ balena env add EDITOR vim --device 7cf02a6 --service MyService
```
**After this PR:** (usage line inconsistent with ARGUMENTS and EXAMPLES sections)
```
$ ./bin/balena help env add
Add an environment or config variable to an application, device or service.

USAGE
  $ balena env add <name> [value]

ARGUMENTS
  NAME   environment or config variable name
  VALUE  variable value; if omitted, use value from this process' environment

EXAMPLES
  $ balena env add TERM --application MyApp
  $ balena env add EDITOR vim --application MyApp
  $ balena env add EDITOR vim --application MyApp --service MyService
  $ balena env add EDITOR vim --device 7cf02a6
  $ balena env add EDITOR vim --device 7cf02a6 --service MyService
```

lib/utils/oclif-utils.ts
srlowe: @pdcastro I guess we need make a decision on what the format should be across the board.  

Personally I prefer the capitano style.  The oclif style looks a bit ugly and jarring to my eyes (any muti-word param names become quite unreadable when all caps).

The capitanoizeOclifUsage function did not preserve camelCase names though, so made multi-word names hard to read whenlikethis, hence these changes.  I think we hadn't noticed this before, because there were no multi-word parameter names in any of oclif commands converted so far.

lib/actions-oclif/join.ts
```diff
@@ -0,0 +1,97 @@
...
+
+interface ArgsDef {
+	deviceIpOrHostname?: string;
+}
+
+export default class JoinCmd extends Command {
+	public static description = stripIndent`
+		Promote a local device running balenaOS to join an application on a balena server.
+
+		Move a local device to an application on another balena server.
```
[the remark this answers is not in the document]
srlowe: Yeah I agree there are too many verbs, I wasn't really sure which were most correct.   Suggestion looks ok to me.

I think we should think about the line duplication when deciding between the oclif/capitano styles.   There are a number of commands that don't have the duplicated lines.  Personally I'm not sure that writing the same content twice in the help output makes sense.

Eg. Why not: 
```
		Move a local device to an application on another balena server.

		Causes the device to "join" the new server. The device must be running balenaOS.
```

Which would render to (pls ignore the oclif argument style issues here):

```
Move a local device to an application on another balena server.

USAGE
  $ balena join [deviceIpOrHostname]

ARGUMENTS
  DEVICEIPORHOSTNAME  the IP or hostname of device

OPTIONS
  -a, --application=application  application name
  -h, --help                     show CLI help

DESCRIPTION
  Causes the device to "join" the new server. The device must be running balenaOS.
  ...

lib/utils/oclif-utils.ts
srlowe: We also have the inconsistency of how oclif shows it's options in help with `=`, e.g. `--application=application` (a format we don't use in any of our usage examples or blog post etc.)

doc/cli.markdown
srlowe: Yes, I think this needs to be part of the wider discussion.

lib/actions-oclif/join.ts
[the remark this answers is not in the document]
srlowe: Another approach (which I think makes sense) is to use the capitano style for 'balena help`, where usage is the first line (i.e. we don't have two descriptions in the help).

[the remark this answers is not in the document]
pdcastro: > I'm not sure that writing the same content twice in the help output makes sense

The duplicated line is printed in different places, and for some commands like `balena help push`, with lots of flags, the duplicated lines would not even be visible on the same "screen" (one needs to scroll the terminal window).

> Another approach (which I think makes sense) is to use the capitano style for 'balena help`, where usage is the first line

Sounds good to me. I don't assume that someone reading the DESCRIPTION section has actually read the first line that gets printed before USAGE anyway...

Issue #1725, opened 2020-04-15
`balena device os-update` progress bar resets to 0% at end of update.


srlowe:
The `balena device os-update` progress bar resets from 100% to 0% at end of update (though this does not affect the update).

PR #1726, opened 2020-04-15
Enable CLI upgrade of dev OS versions

srlowe:
Enable CLI upgrade of dev OS versions
Fix `device os-update` incorrectly showing 0% progress
Convert `device os-update` to use async/await

Change-type: minor
Resolves: #1725

### DIFF
--- a/lib/actions/device_ts.ts
+++ b/lib/actions/device_ts.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2016-2017 Balena
+Copyright 2016-2020 Balena Ltd.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@ limitations under the License.
 import { Device } from 'balena-sdk';
 import { CommandDefinition } from 'capitano';
 import { stripIndent } from 'common-tags';
+import { ExpectedError } from '../errors';
 import { getBalenaSdk } from '../utils/lazy';
 import { normalizeUuidProp } from '../utils/normalization';
 import * as commandOptions from './command-options';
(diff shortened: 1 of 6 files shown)


lib/actions/keys.ts
```diff
@@ -48,7 +48,7 @@ Examples:
 `,
 	permission: 'user',
 	async action(params) {
-		const key = await getBalenaSdk().models.key.get(params.id);
+		const key = await getBalenaSdk().models.key.get(parseInt(params.id, 10));
```
srlowe, line 51: Required by new SDK version type changes.

lib/utils/device/progress.ts
```diff
@@ -18,7 +18,7 @@ import { Device } from 'balena-sdk';
 
 export const getDeviceOsProgress = (device: Device) => {
 	if (!device.is_online) {
-		return 0;
+		return null;
```
srlowe, line 21: OS Update progress bar was resetting to 0% at end of process, as device went offline during reboot.  I think null is more appropriate in these cases.

lib/utils/patterns.ts
```diff
@@ -293,10 +293,7 @@ export function awaitDeviceOsUpdate(uuid: string, targetOsVersion: string) {
 				balena.models.device.getOsUpdateStatus(uuid),
 				balena.models.device.get(uuid).then(getDeviceOsProgress),
 			]).then(([osUpdateStatus, osUpdateProgress]) => {
-				if (
-					osUpdateStatus.status === 'update_done' ||
-					osUpdateStatus.status === 'done'
-				) {
+				if (osUpdateStatus.status === 'done') {
```
srlowe, line 296: Required due to SDK type changes.  API no longer sends `update_done` status.

lib/actions/keys.ts
```diff
@@ -86,7 +86,7 @@ Examples:
 			'Are you sure you want to delete the key?',
 		);
 
-		await getBalenaSdk().models.key.remove(params.id);
+		await getBalenaSdk().models.key.remove(parseInt(params.id, 10));
```
srlowe, line 89: Required by new SDK version type changes.

lib/utils/patterns.ts
srlowe: Yeah agreed.  Though this one had no effect other than breaking ts compilation.

lib/actions/device_ts.ts
```diff
@@ -55,62 +56,67 @@ export const osUpdate: CommandDefinition<OsUpdate.Args, OsUpdate.Options> = {
...
+		);
+		if (hupVersionInfo.versions.length === 0) {
+			throw new ExpectedError(
+				'There are no available Host OS update targets for this device',
+			);
+		}
+
+		// Get target OS version
+		let targetOsVersion = options.version;
+		if (targetOsVersion != null) {
```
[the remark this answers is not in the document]
srlowe: Hmm I not sure I'd agree that that would help with readability, I guess this is quite subjective.  If it were needed in more than one place then I'd agree that moving to a function is good from a modularity point of view, but as far as I know it's not.  Personally I think splitting simple code into functions can sometimes have the reverse effect.

srlowe:
> From: `Enable CLI upgrade of dev OS versions`
> To: `device os-update: allow host OS upgrade with development balenaOS images`

Ok will do.  Yeah this might explain some weird changelog lines that I saw in relation to previous commits I'd done with the similar description structure.  Will alter my style for future commits.

lib/actions/device_ts.ts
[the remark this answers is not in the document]
pdcastro: Subjective?! :-) Functions are great: they break large problems in smaller ones, define input/output contracts that help reason for correctness (when low coupling is possible), and facilitate unit testing (which in some companies is like religion). I am approving the PR anyway, but I see it a basic [software design](https://en.wikipedia.org/wiki/Software_design) principle/technique rather than my personal opinion.

srlowe:
@pdcastro > but I see it a basic software design principle/technique rather than my personal opinion.

I think functions are great for all the reasons you stated, but I don't think those reasons apply or ad value in this particular usage ;-)

lib/actions/keys.ts
thgreasi, line 51: Instead of doing these two `parseInt` calls, can't we just change the capitano arg type for this command?
From:
```
export const info: CommandDefinition<{ id: string }> = {
```
To:
```
export const info: CommandDefinition<{ id: number }> = {
```

srlowe: Yes, good point.

lib/actions/device_ts.ts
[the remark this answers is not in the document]
thgreasi: How about reformatting this to reduce nesting:
`if (targetOsVersion == null) {
  targetOsVersion = await form.ask({...});
}
// possibly make this an `else if` you prefer, but I do like the extra check anyway :)
if (!_.includes(hupVersionInfo.versions, targetOsVersion)) {
  throw...;
}

lib/actions/keys.ts
srlowe: Just tested that, and it seems to allow though strings(!?)  E.g., with this change:
```
$balena key 15561r
BalenaRequestError: Request error: Database error
```
Not really sure what's going on there with capitano, but I suggest we stick with the parseInt for now, since this will be converted to oclif shortly anyway.

thgreasi: Imo `$balena key 15561r` should error since that's not a correct key id.
What if you have a key named as `1234mysshkey` and also have a different ssh key with the db id `1234`. If the user runs `$balena key 1234mysshkey` the `parseInt` will parse the 1234 and delete the key that happens to have that db id.

srlowe: Ok let me add some better checks there in a separate PR.

Issue #1728, opened 2020-04-15
Improve input validation for `key`, `key rm`


srlowe:
`key`, `key rm` should throw an error when param is not value interger, rather than trying to interpret as integer.

PR #1729, opened 2020-04-15
improve input validation for `key`, `key rm`

srlowe:
Change-type: patch
Resolves: #1728

Relates to: https://github.com/balena-io/balena-cli/pull/1726#discussion_r408863313

### DIFF
--- a/lib/actions/keys.ts
+++ b/lib/actions/keys.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2016-2017 Balena
+Copyright 2016-2020 Balena Ltd.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -15,8 +15,16 @@ limitations under the License.
 */
 
 import { CommandDefinition } from 'capitano';
+import { ExpectedError } from '../errors';
 import { getBalenaSdk, getVisuals } from '../utils/lazy';
 import * as commandOptions from './command-options';
+
+function parseId(id: string): number {
+	if (/^[\d]+$/.exec(id) == null) {
+		throw new ExpectedError('The key id must be an integer');
+	}
+	return Number(id);
+}
 
 export const list: CommandDefinition = {
 	signature: 'keys',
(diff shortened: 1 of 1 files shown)


lib/actions/keys.ts
```diff
@@ -48,7 +56,7 @@ Examples:
 `,
 	permission: 'user',
 	async action(params) {
-		const key = await getBalenaSdk().models.key.get(parseInt(params.id, 10));
+		const key = await getBalenaSdk().models.key.get(parseId(params.id));
```
thgreasi, line 59: There is no need for all this, since capitano already handles this for you.
We only need to do a
```
if (typeof id !== `number'`) {
  // Hey user this should have been a number
  // exit
}
```
and then just change the typings to:
```
export const info: CommandDefinition<{ id: number }> = {
```
so that TS is happy.

pdcastro: Now that you mentioned it, iirc, Capitano uses a regex to check if the value "looks like a number", and if it does, converts it from string to number (which is why we have problems with UUIDs, when a leading zero is dropped in an _unwanted_ conversion from string to number). So if the user provides `'017'` Capitano will give us a number (`17` without the leading zero), but if the user provides `'17r'` Capitano will give us that string unchanged. In this case, perhaps the correct type annotation would be:
```
export const info: CommandDefinition<{ id: string | number }> = {
```
I.e. it should not reflect what "we would like" the type to be, but what Capitano actually provides, so that our code can handle all possible cases.

srlowe: @thgreasi , @pdcastro Thanks guys.  I had been confused by capitano's behaviour during tests, and there are no other examples of number arguments using capitano in the CLI.    However, since this is working correctly, and will be thrown away in the next couple of weeks anyway as we move to oclif I'll leave it as it for now.

srlowe: I noticed that it also silently drops multiple leading zeros.  So 0001 is processed as 1.  I think it's safer to throw an error in this case as it probably indicates a typo.  (e.g. 10023 mistyped as 0023).

pdcastro: Typo, or they typed a UUID, getting confused between database ID and device UUID. (People rarely type numbers with leading zeros if they mean for it to be an integer... Maybe the regex can be modified so that the first digit must not be zero. Which is probably what you already had in mind.) :-)

srlowe: @pdcastro Ah, I forgot that `Number()` also drops leading zeros :-(  Yeah so ideally needs altered regex.

Issue #1723, opened 2020-04-14
Error: ENOENT: no such file or directory, lstat '/usr/local/lib/balena-cli/lib' when running cli commands


szony:
- **balena CLI version:** 11.30.10
- **Operating system version:** MacOS 10.15.3
- **32/64 bit OS and processor:** 64-bit
- **Install method:** MacOS package
- **If npm install, Node.js and npm version:** e.g. Node v8.16.0 and npm v6.4.1

---

I'm using the cli in zsh and frequently I would come across this error when running different commands:

➜  repos balena devices -a tmp0
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
Error: ENOENT: no such file or directory, lstat '/usr/local/lib/balena-cli/lib'
    at Object.realpathSync (fs.js:1461:7)
    at Object.<anonymous> (/usr/local/lib/balena-cli/node_modules/is-installed-globally/index.js:8:29)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:690:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at lazy (/usr/local/lib/balena-cli/node_modules/import-lazy/index.js:2:51)
    at /usr/local/lib/balena-cli/node_modules/import-lazy/index.js:10:11
    at UpdateNotifier.notify (/usr/local/lib/balena-cli/node_modules/update-notifier/index.js:122:14)
    at Object.notify (/usr/local/lib/balena-cli/build/utils/update.js:22:18)
    at globalInit (/usr/local/lib/balena-cli/build/app-common.js:148:31)
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:832:11)

Typically, second iteration of the same command works just fine. I mainly use balena push, devices, ssh. The error comes up randomly with any of them. The /usr/local/lib/balena-cli/lib file doesn't exist on my system.

pdcastro:
[...]

PR #1730, opened 2020-04-15
Fix seemingly random ENOENT error (update 'is-installed-globally' dependency)

pdcastro:
`is-installed-globally` version 0.3.2 fixed the ENOENT issue:
* https://github.com/sindresorhus/is-installed-globally/issues/4
* https://github.com/sindresorhus/is-installed-globally/issues/6
* https://github.com/sindresorhus/is-installed-globally/pull/7

The CLI's `npm-shrinkwrap.json` file has two occurrences of affected `is-installed-globally` dependency:
```
    "is-installed-globally": {
      "version": "0.3.1",
...
        "is-installed-globally": {
          "version": "0.1.0",
```

This PR updates the first one, but leaves `is-installed-globally` v0.1.0 alone: there is a net of dependencies that I couldn't untangle. Even so, strictly this PR resolves issue #1723 because that issue qualified the problem as "when running cli commands", and the v0.1.0 dependency only applies when running `patch-package` in the `postinstall` npm script. `patch-package` v6.2.1 resolves the problem by removing the dependency altogether, but we cannot upgrade from `patch-package` 6.1 to 6.2 because of issue ds300/patch-package/issues/201. I created PR ds300/patch-package/pull/224 in their repo, which hopefully would allow us to upgrade `patch-package`. Meanwhile, this PR should alleviate the problem.

Change-type: patch
Resolves: #1723

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -8283,9 +8283,9 @@
       }
     },
     "is-installed-globally": {
-      "version": "0.3.1",
-      "resolved": "https://registry.npmjs.org/is-installed-globally/-/is-installed-globally-0.3.1.tgz",
-      "integrity": "sha512-oiEcGoQbGc+3/iijAijrK2qFpkNoNjsHOm/5V5iaeydyrS/hnwaRCEgH5cpW0P3T1lSjV5piB7S5b5lEugNLhg==",
+      "version": "0.3.2",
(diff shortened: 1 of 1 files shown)


leonardodino:
note: `is-installed-globally@0.1.0` doesn't causes `ENOENT` issues.

the possibility for this issue was introduced in `0.2.0`, and fixed in `0.3.1` 😄

PR #1734, opened 2020-04-17
convert commands `key`, `keys`, `key add`, `key rm` to oclif.

srlowe:
Also:
 - Display keys with `name` instead of `title`.
 - Check for empty key before calling API.

Change-type: patch


lib/actions-oclif/key/add.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+interface ArgsDef {
+	name: string;
+	path: string;
+}
+
+export default class KeyAddCmd extends Command {
+	public static description = stripIndent`
+		Add an SSH key to balena.
```
srlowe, line 36: I modified the help a little bit for consistency.

lib/actions-oclif/key/add.ts
```diff
@@ -0,0 +1,87 @@
...
+
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(KeyAddCmd);
+
+		let key: string;
+		if (params.path != null) {
+			const { promisify } = await import('util');
+			const readFileAsync = promisify((await import('fs')).readFile);
+			key = await readFileAsync(params.path, 'utf8');
+		} else if (this.stdin.length > 0) {
```
srlowe, line 79: Now checks if there's anything from stdin, rather than assuming there is if param.path is empty.

lib/actions-oclif/key/index.ts
```diff
@@ -0,0 +1,79 @@
...
+	};
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params } = this.parse<{}, ArgsDef>(KeyCmd);
+
+		const key = await getBalenaSdk().models.key.get(params.id);
+
+		// Use 'name' instead of 'title' to match dashboard.
```
srlowe, line 66: `name` instead of `title` now

lib/actions-oclif/key/rm.ts
```diff
@@ -0,0 +1,80 @@
...
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			KeyRmCmd,
+		);
+
+		const patterns = await import('../../utils/patterns');
+
+		await patterns.confirm(
+			options.yes ?? false,
+			`Are you sure you want to delete key ${params.id}?`,
```
srlowe, line 75: Added id in confirmation message, to be safe.

lib/actions-oclif/keys.ts
```diff
@@ -0,0 +1,56 @@
...
+	};
+
+	public static authenticated = true;
+
+	public async run() {
+		this.parse<FlagsDef, {}>(KeysCmd);
+
+		const keys = await getBalenaSdk().models.key.getAll();
+
+		// Use 'name' instead of 'title' to match dashboard.
```
srlowe, line 49: `name` instead of `title` now

lib/utils/oclif-validation.ts
```diff
@@ -0,0 +1,31 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { ExpectedError } from '../errors';
+
+export function parseAsInteger(input: string, paramName?: string) {
```
srlowe, line 19: This one doesn't allow leading zeros ;-)

lib/actions-oclif/key/index.ts
```diff
@@ -0,0 +1,79 @@
...
+	help: void;
+}
+
+interface ArgsDef {
+	id: number;
+}
+
+export default class KeyCmd extends Command {
+	public static description = stripIndent`
+		Display an SSH key.
```
srlowe, line 37: 'List' seemed wrong when it's only ever a single item.

pdcastro: Suggestion: `Display an SSH key registered in balenaCloud for the logged in user`
(balenaCloud being relevant because afaik it doesn't work with openBalena)

lib/actions-oclif/key/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+		Show information about a single SSH key.
+	`;
+
+	public static examples = ['$ balena key 17'];
+
+	public static args: Array<IArg<any>> = [
+		{
+			name: 'id',
+			description: 'key id',
```
pdcastro, line 47: Suggestion: `balenaCloud database ID for the SSH key`
(Because one might wonder if 'key id' means the ssh key fingerprint or something...)

lib/actions-oclif/key/rm.ts
```diff
@@ -0,0 +1,80 @@
...
+		Notice this command asks for confirmation interactively.
+		You can avoid this by passing the \`--yes\` boolean option.
+	`;
+
+	public static examples = ['$ balena key rm 17', '$ balena key rm 17 --yes'];
+
+	public static args: Array<IArg<any>> = [
+		{
+			name: 'id',
+			description: 'key id',
```
pdcastro, line 51: Suggestion: `balenaCloud database ID for the SSH key`
(Because one might wonder if 'key id' means the ssh key fingerprint or something...)

lib/actions-oclif/key/rm.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export default class KeyRmCmd extends Command {
+	public static description = stripIndent`
+		Remove an SSH key.
+
+		Remove a single SSH key from balena.
+
+		Notice this command asks for confirmation interactively.
+		You can avoid this by passing the \`--yes\` boolean option.
```
pdcastro, line 43: 'Note' sounds better than 'Notice'? Also personally I prefer to avoid the pronoun "You". Suggestion: instead of the two lines above (from 'Notice' to 'option'):
```
The --yes option may be used to avoid interactive confirmation.
```

lib/utils/oclif-validation.ts
srlowe: Yeah, I saw that parseDbId one too.   Ok sure, can move to that validation.ts file.

lib/actions-oclif/key/rm.ts
```diff
@@ -0,0 +1,80 @@
...
+	`;
+
+	public static examples = ['$ balena key rm 17', '$ balena key rm 17 --yes'];
+
+	public static args: Array<IArg<any>> = [
```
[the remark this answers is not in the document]
srlowe: Ah yeah I guess it could, ok will change.

lib/utils/oclif-validation.ts
```diff
@@ -0,0 +1,31 @@
...
+	// Allow only digits, no leading 0
+	if (!/^[1-9][0-9]*$/.test(input)) {
```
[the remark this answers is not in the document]
srlowe: Good point, will test.

lib/actions-oclif/key/index.ts
srlowe: Hmm yeah.  Not sure I like `balenaCloud database ID` though - seems like we are exposing internal implementation details a bit?

srlowe: What about `balenaCloud ID for the SSH key`

lib/actions-oclif/keys.ts
```diff
@@ -0,0 +1,56 @@
...
+
+interface FlagsDef {
+	help: void;
+}
+
+export default class KeysCmd extends Command {
+	public static description = stripIndent`
+		List all SSH keys.
+
+		List all SSH keys registered with balena.
```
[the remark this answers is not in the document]
srlowe: Do you suggest this for both lines, or just the 3rd?

lib/actions-oclif/key/rm.ts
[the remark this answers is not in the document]
srlowe: This didn't work with the base class unfortunately.  I suppose we could look at generics, but probably better to look at that as a seperate task later.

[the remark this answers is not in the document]
srlowe: (I've added it to my improvements todo list)

lib/actions-oclif/keys.ts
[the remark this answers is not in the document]
pdcastro: hmm I had only thought about the 3rd line. Maybe:
```
List the SSH keys in balenaCloud

List all SSH keys registered in balenaCloud for the logged in user
```

lib/actions-oclif/key/index.ts
pdcastro: Sounds good 👍 (`balenaCloud ID for the SSH key`)

tests/utils/validation.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect } from 'chai';
+import { ExpectedError } from '../../build/errors';
+import { parseAsInteger } from '../../build/utils/validation';
+
+describe('parseAsInteger() function', function() {
```
pdcastro, line 22: 💯

PR #1735, opened 2020-04-18
Unpin balena-sdk and review 'instanceof' usage

pdcastro:
Change-type: patch


lib/app-oclif.ts
```diff
@@ -43,9 +42,11 @@ export async function run(command: string[], options: AppOptions) {
 				return require('@oclif/command/flush');
 			}
 		},
-		(error: Error) => {
+		error => {
 			// oclif sometimes exits with ExitError code 0 (not an error)
-			if (error instanceof ExitError && error.oclif.exit === 0) {
+			// (Avoid `error instanceof ExitError` here for the reasons explained
+			// in the CONTRIBUTING.md file regarding the `instanceof` operator.)
+			if (error.oclif?.exit === 0) {
```
pdcastro, line 49: Here I have not used the `instanceOf` function (from `lib/errors.ts`) because the previous code relied on the plain `instanceof` operator as a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#instanceof-type-guards) that allowed the expression `error.oclif.exit` to be compiled. Without the type guard, typescript would not allow writing `error.oclif`. Shame. The new node does not look too bad, though...

lib/utils/device/live.ts
```diff
@@ -285,7 +286,7 @@ export class LivepushManager {
 			this.logger.logError(
 				`An error occured whilst trying to perform a livepush: `,
 			);
-			if (e instanceof ContainerNotRunningError) {
+			if (instanceOf(e, ContainerNotRunningError)) {
```
pdcastro, line 289: I've tested this, pushing to an RPi and deleting the container.

lib/utils/patterns.ts
```diff
@@ -405,7 +405,7 @@ export async function getOnlineTargetUuid(
 			})),
 		});
 	} catch (err) {
-		if (!(err instanceof BalenaApplicationNotFound)) {
+		if (!instanceOf(err, BalenaApplicationNotFound)) {
```
pdcastro, line 408: * `balena ssh <uuid>` now works, even with `balena-sdk` v12.33.0.
* Another possible version of the fix would have been:
```
if (err.name !== 'BalenaApplicationNotFound') {
```
It would even had allowed removing the `balena-errors` dependency from `package.json`. Very tempting. On balance my choice was to keep the dependency and instanceof, because relying solely on string comparison introduces its own fragility: it's prone to typo errors, the compiler does not complain if one writes `if (err.name !== 'foo_blah')` even if `'foo_blah'` is nonsense.

package.json
```diff
@@ -170,10 +169,10 @@
     "balena-config-json": "^2.1.1",
     "balena-device-init": "^5.0.2",
     "balena-device-status": "^3.2.1",
-    "balena-errors": "^4.2.1",
+    "balena-errors": "^4.3.0",
```
pdcastro, line 172: Aligning with the dependency declared by `balena-sdk` (even though shrinkwrap actually installed v4.3.2 - it's OK).

PR #1740, opened 2020-04-21
Update dependencies to pick up performance improvements

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


package.json
```diff
@@ -142,19 +142,18 @@
...
-    "husky": "^4.2.3",
+    "husky": "^4.2.5",
     "intercept-stdout": "^0.1.2",
-    "mocha": "^6.2.2",
+    "mocha": "^6.2.3",
     "mock-require": "^3.0.3",
     "nock": "^11.9.1",
     "parse-link-header": "~1.0.1",
     "pkg": "^4.4.2",
-    "prettier": "^1.19.1",
```
Page-, line 152: This is unnecessary as the prettifying is handled by balena-lint now (which installs its own version of prettier, and this one could potentially clash)

package.json
```diff
@@ -230,10 +229,10 @@
     "resin-doodles": "^0.1.1",
     "resin-image-fs": "^5.0.9",
     "resin-multibuild": "^4.4.2",
-    "resin-release": "^1.2.0",
+    "resin-release": "^1.2.2",
```
Page-, line 232: This doesn't change the shrinkwrapped version, just updated the package.json to match that actual version that will be installed. This should also be replaced by balena-release @pdcastro / @srlowe which I believe only has the breaking change of requiring node 8+ but I don't really have the capability to test so it'd be great if one of you could handle it

srlowe:
@Page- Nice.  I wonder which dependency update contributed the majority of that?

Page-:
It's a bunch of them combined @srlowe, all of the sdk modules imported lodash via `lodash/$function` which reduces the browser bundle significantly but is much slower to require on nodejs so I went through most of them and significantly reduced/removed lodash usage (since the vast can be done natively in js/ts now):
https://github.com/balena-io-modules/balena-hup-action-utils/pull/21
https://github.com/balena-io-modules/rindle/pull/11
https://github.com/balena-io-modules/balena-request/pull/142
and not included in this PR but will gain some more benefits (on some commands, not every command causes the requires to be hit due to lazy-loading): https://github.com/balena-io/balena-sdk/pull/888

package.json
srlowe: Ok, opened issue for that: https://github.com/balena-io/balena-cli/issues/1741

PR #1742, opened 2020-04-21
Fix left-overs from balena-sdk v13 bump

thgreasi:
Connects-to: #1770 (meta v12 release issue)
Change-type: patch
See: https://github.com/balena-io/open-balena-api/issues/338
See: https://www.flowdock.com/app/rulemotion/i-cli/threads/0Mu_15537VAJ1wxbElpROBA-g5Y

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -185,7 +185,6 @@
     "archiver": "^3.1.1",
     "balena-config-json": "^2.1.1",
     "balena-device-init": "^5.0.2",
-    "balena-device-status": "^3.2.1",
     "balena-errors": "^4.3.0",
     "balena-image-manager": "^6.1.2",
     "balena-preload": "^8.4.0",
--- a/lib/actions-oclif/device/index.ts
+++ b/lib/actions-oclif/device/index.ts
@@ -69,13 +69,19 @@ export default class DeviceCmd extends Command {
 
 		const balena = getBalenaSdk();
 
-		const device: ExtendedDevice = await balena.models.device.get(
-			params.uuid,
-			expandForAppName,
-		);
-
-		const deviceStatus = await balena.models.device.getStatus(device.uuid);
-		device.status = deviceStatus;
+		const [device, overallStatus] = await Promise.all([
+			balena.models.device.get(params.uuid, expandForAppName) as Promise<
+				ExtendedDevice
+			>,
+			// TODO: drop this and add `overall_status` to a $select in the above
+			// pine query once the overall_status field is moved to open-balena-api.
+			// See: https://github.com/balena-io/open-balena-api/issues/338
+			balena.models.device
+				.get(params.uuid, { $select: 'overall_status' })
+				.then(({ overall_status }) => overall_status)
+				.catchReturn(''),
+		]);
+		device.status = overallStatus;
 
 		device.dashboard_url = balena.models.device.getDashboardUrl(device.uuid);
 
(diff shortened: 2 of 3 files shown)


lib/actions/device.coffee
```diff
@@ -87,32 +87,54 @@ exports.info =
...
+						'supervisor_version'
+						'is_web_accessible'
+						'note'
+						'os_version'
+					]
+				}, expandForAppName)
+			)
+			# TODO: combine with the above query once the overall_status field
+			# is moved to open-balena-api
+			balena.models.device.get(params.uuid, { $select: 'overall_status' }).get('overall_status').catchReturn(null)
```
thgreasi, line 112: I opened the PR just in case that the CLI needs to catch-up with any newer SDK feature before the open-balena-api adds this computed field.
On principle this shouldn't be a blocker for the PR, especially since the older balenaDeviceStatus wasn't working properly (as I just found out) since not all the required fields were provided.

INSTALL.md
```diff
@@ -93,7 +93,7 @@ If you are a Node.js developer, you may wish to install the balena CLI via [npm]
 The npm installation involves building native (platform-specific) binary modules, which require
 some additional development tools to be installed first:
 
-* [Node.js](https://nodejs.org/) version 8, 10 or 12
+* [Node.js](https://nodejs.org/) version 10 or 12
```
pdcastro, line 96: I see you've labeled this PR `change-type: major`, resulting in CLI v12.0.0. We've been collecting breaking changes for v12 in this GDoc: https://docs.google.com/document/d/1uHE-_LLu5gJUQZICQJtlo76EQCm90z0CMAwpkfw6s8w/edit

@srlowe, I think pressure is building to release CLI v12... Node v8 is no longer supported, Node 14 is around the corner, Page wanted to take advantage of Node 10+ features, we need to upgrade balena-sdk too, including testing. I wanted to include issue #1032 in v12, but each week there are new hurdles (the latest being balena CI being unhappy with the MIT licence in the migration from `zeit/dockerignore` to `balena-io-modules/dockerignore`). Anyway, we need to put some effort into that GDoc and CLI v12, perhaps with a pause in the oclif/typescript conversion. Maybe something we could also discuss with @OdysLam on Thursday's call.

thgreasi: @pdcastro is there a v12 release branch that you PR against?
Also heads up that there is a pending arch call item wrt whether we should be using release branches at all or not, so it would be great if you could be there  as well.

pdcastro: No release branch yet. My thinking is that, when possible, we should release the "breaking changes" behind flags (command-line options) that enable the new behaviour in the current major release. This is possible, for example, for the new gitignore / dockerignore behaviour (but probably not possible for Node 8 → Node 14 support). In this case, it is not really a breaking change, and it also allows users to try the new feature before upgrading the CLI to the next major version. Then, when all this work is done and already released, a release branch is created for the shortest amount of time possible, and what it does is to change the "default behaviour" (without the enabling flags) to be same as the behaviour with the enabling flags, and the enabling flags become a no-op (redundant). I'll look out for the arch call, thanks for the hint! 👍

package.json
```diff
@@ -74,7 +74,7 @@
   "author": "Juan Cruz Viotti <juan@balena.io>",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=8.0"
+    "node": ">=10.0"
```
pdcastro, line 77: @thgreasi, as discussed in Google chat (balena-io/product), would it be possible to avoid this Node.js bump in this PR, and have this PR's change-type be minor or patch rather than major? And then we merge this PR in CLI v11, and I can add the "balenaCloud backend support/deprecation policy" to the CLI.

thgreasi: @pdcastro you mean, even though the SDK's package.json declares that the minimum supported node version is v10?
I haven't tried it, but I think that while doing a fresh npm install on node 8, npm will complain as soon as it sees that the sdk needs node 10.

thgreasi:
Connects-to: #1770

PR #1752, opened 2020-04-23
Convert app commands to oclif

srlowe:
Convert commands `apps`, `app`, `app create`, `app rm`, `app restart` to oclif and async/await.

Change-type: patch

Depends-on: #1781


tests/commands/app/create.spec.ts
```diff
@@ -53,7 +53,8 @@ describe('balena app create', function() {
 		api.done();
 	});
 
-	it('should print help text with the -h flag', async () => {
+	// Temporarily skipped because of parse/checking order issue with -h
+	it.skip('should print help text with the -h flag', async () => {
```
pdcastro, line 57: Does this still need skipping? This is not the issue you had mentioned about the `app` command and topics, right? -
```
COMMANDS
  app:create   Create an application.
  app:restart  Restart an application.
  app:rm       Remove an application.
```

> because of parse/checking order issue

If relevant (you may already know), there is a "manual sort" function that allows manually tweaking the sort order when convenient:
* https://github.com/balena-io/balena-cli/blob/v11.31.22/lib/actions/help.coffee#L51-L67
* https://github.com/balena-io/balena-cli/blob/v11.31.22/lib/utils/helpers.ts#L228-L254

srlowe: @pdcastro This skip is still needed at the moment unfortunately.  I didn't explain it well, but by `parse/checking order` here I mean the fact that we currently do login/root/etc checks before parsing the command.  This means that the `-h` flag does not work if you're not logged in etc.  I have a fix for this comming, but am on support at the moment.

Also looked into overriding the help output (for the COMMANDS issue).  It doesn't look like this will be possible cleanly.  I guess our options are an npx patch, or forking/duplicating the oclif/help-plugin repo.  Any thoughts here?

pdcastro: > This means that the `-h` flag does not work if you're not logged in etc. I have a fix for this coming

Got it. Nice, we probably want to wait for the fix before merging the PR then.

> I guess our options are an npx patch, or forking/duplicating the oclif/help-plugin repo. Any thoughts here?

My rule of thumb is the amount/extent of changes. If it's just a few lines, I tend to prefer using `npx patch-package`, if it's something more extensive then fork. Either way, ideally we would submit the changes upstream, so we can then remove the fork or patch. We have some oclif patches (to make oclif work with native Node modules on Windows and macOS) that I never submitted upstream because the upstream (ex-)maintainer explained he thought those changes would break oclif's autoupdate, which we are not using yet. Once we start using oclif's autoupdate and sort whatever issue there may be, we can then submit those patches upstream.

srlowe: Re the upstream submission.  I'm don't think this would be accepted upstream because it's a move away from the oclif way of doing things (removing or amending the COMMANDS section, and removing `=` from param section).  I think the only way to do this in an upstreamable way would be to implement some kind of template system, like they suggest here: https://github.com/oclif/oclif/issues/213#issuecomment-471008302  But that would be quite an investment of time (and the two changes would require two sets of templates, since the code is in different places).

pdcastro: OK, not upstreamable. The patch vs fork rule of thumb probably still stands... A forked / duplicated repo requires some additional setting up of the repo (admin permissions and balena CI occur to me), which is probably an additional argument in favour of patching, if the patch is small enough.

Issue #1309, opened 2019-06-14
Get update available notification even if already on that latest version


imrehg:
- **Cli version:** 11.2.1
- **Standalone install:** no
- **Node version:** v11.15.0
- **Npm version:** 6.9.0
- **Operating system and architecture:** ArchLinux 64-bit

Received a warning even if I that latest version is installed for me a while now.

```
~/app> balena deploy --help 


   ╭────────────────────────────────────────────────────────────────────╮
   │                                                                    │
   │                  Update available 11.2.0 → 11.2.1                  │
   │                                                                    │
   │   https://github.com/balena-io/balena-cli/blob/master/INSTALL.md   │
   │                                                                    │
   ╰────────────────────────────────────────────────────────────────────╯

Usage: deploy <appName> [image]
[...snip...]

~/app> balena version
11.2.1
```

---

*Please keep in mind that we try to use the issue tracker of this repository for specific bug reports & CLI feature requests. General & troubleshooting questions are encouraged to be posted to the [balena forums](https://forums.balena.io) where the community can both contribute and benefit from the answers.*

*Before submitting this issue please check that this issue is not a duplicate. If there is another issue describing the same problem or feature please add your information to the existing issue's comments.*

PR #1775, opened 2020-04-28
Only notify of an update if the new version is actually newer

Page-:
Resolves: #1309
Change-type: patch


Page-:
@balena-ci rebase

Page-:
I expect so @pdcastro

PR #1786, opened 2020-04-29
balena deploy: Fix "TypeError: images.push is not iterable"

pdcastro:
Resolves: #1783 
Change-type: patch

Internal discussion: https://www.flowdock.com/app/rulemotion/i-cli/threads/GOcVcJdg2SN3R_6VIIlt1LnPObL

### DIFF
--- a/lib/utils/compose.js
+++ b/lib/utils/compose.js
@@ -530,10 +530,13 @@ const getPreviousRepos = (sdk, docker, logger, appID) =>
 						return registry.imageName;
 					});
 				});
+			} else {
+				return [];
 			}
 		})
 		.catch(e => {
 			logger.logDebug(`Failed to access previously pushed image repo: ${e}`);
+			return [];
 		});
 
 const authorizePush = function(


lib/utils/compose.js
```diff
@@ -519,7 +519,7 @@ const getPreviousRepos = (sdk, docker, logger, appID) =>
 		})
 		.then(function(release) {
 			// grab all images from the latest release, return all image locations in the registry
-			if (release.length > 0) {
+			if (release && release.length > 0) {
```
Page-, line 522: The release && should be unnecessary as the line client should never return null for this

pdcastro: > **should be** unnecessary [...] pine client **should** never

_vs._

> **is** unnecessary [...] pine client **will** never

That's OK, I am happy with the harmless redundancy when it gives me extra confidence. :-)

Page-: If it ever gives you anything other than an array things have gone so badly wrong that bailing is actually the best you can hope for

lib/utils/compose.js
```diff
@@ -530,6 +530,8 @@ const getPreviousRepos = (sdk, docker, logger, appID) =>
 						return registry.imageName;
 					});
 				});
+			} else {
+				return Promise.resolve([]);
 			}
 		})
 		.catch(e => {
```
srlowe, line 537: Wont we get an undefined returned here too?

Page-: Yes

lib/utils/compose.js
```diff
@@ -530,10 +530,13 @@ const getPreviousRepos = (sdk, docker, logger, appID) =>
 						return registry.imageName;
 					});
 				});
+			} else {
+				return Promise.resolve([]);
 			}
```
Page-, line 535: There's no need for the promise.resolve here

lib/utils/compose.js
```diff
@@ -530,10 +530,13 @@ const getPreviousRepos = (sdk, docker, logger, appID) =>
...
 		})
 		.catch(e => {
 			logger.logDebug(`Failed to access previously pushed image repo: ${e}`);
+			return Promise.resolve([]);
```
Page-, line 539: There's no need for the promise.resolve here

nazrhom:
@balena-ci retest

lib/utils/compose.js
thgreasi: I would also say that should be doing
```
.then(function(([release])) {
```

lib/utils/compose.js
pdcastro: Done

Issue #1784, opened 2020-04-29
Integrate new livepush


srlowe:
Relevant links:

https://github.com/balena-io-modules/livepush/pull/79/files

https://github.com/balena-io-modules/livepush/blob/master/lib/dockerfile.ts#L25
https://github.com/balena-io-modules/livepush/blob/master/lib/dockerfile.ts#L99

srlowe:
Reopened as integration in #1787 was not correct.

PR #1787, opened 2020-04-29
Integrate livepush v3 and live directives

srlowe:
Integrate livepush v3 and live directives

Change-type: minor
Resolves: #1784

### DIFF
--- a/lib/utils/device/live.ts
+++ b/lib/utils/device/live.ts
@@ -10,6 +10,7 @@ import { BuildTask } from 'resin-multibuild';
 import { instanceOf } from '../../errors';
 import Logger = require('../logger');
 
+import { Dockerfile } from 'livepush';
 import DeviceAPI, { DeviceInfo, Status } from './api';
 import {
 	DeviceDeployOptions,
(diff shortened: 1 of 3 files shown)


lib/utils/device/live.ts
```diff
@@ -105,12 +106,14 @@ export class LivepushManager {
...
-				const dockerfile = buildTask.dockerfile;
-				if (dockerfile == null) {
+				if (buildTask.dockerfile == null) {
 					throw new Error(
 						`Could not detect dockerfile for service: ${serviceName}`,
 					);
 				}
+				const dockerfile = new Dockerfile(
+					new Dockerfile(buildTask.dockerfile).generateLiveDockerfile(),
+				);
```
srlowe, line 116: Not sure if there is a more elegant way of doing this?

lib/utils/device/live.ts
```diff
@@ -18,6 +18,7 @@ import {
 } from './deploy';
 import { BuildError } from './errors';
 import { getServiceColourFn } from './logs';
+import Dockerfile from 'livepush/build/dockerfile';
```
CameronDiver, line 21:
```suggestion
import { Dockerfile } from 'livepush';
```

lib/utils/device/live.ts
CameronDiver:
```suggestion
const dockerfile = new Dockerfile(buildTask.dockerfile)
```

The generation happens automatically by default, and the internal representation is replaced with the live one. 
`generateLiveDockerfile` is a bit of a misnomer tbh and you only really need it when you want the string representation of the dockerfile

lib/utils/device/live.ts
```diff
@@ -380,13 +383,17 @@ export class LivepushManager {
 			buildLogs[serviceName] = buildLog;
 			const stageImages = LivepushManager.getMultistageImageIDs(buildLogs);
 
-			instance.livepush = await Livepush.init(
-				buildTask.dockerfile!,
-				buildTask.context!,
-				container.containerId,
-				stageImages[serviceName],
-				this.docker,
+			const dockerfile = new Dockerfile(
```
CameronDiver, line 386:
```suggestion
			const dockerfile = new Dockerfile(buildTask.dockerfile!)
```

Issue #1788, opened 2020-04-29
Error: The command '/tmp/qemu-execve -execve /bin/sh -c ...'  returned a non-zero code: 1


pdcastro:
As originally reported at: https://forums.balena.io/t/deploy-failed-when-trying-to-deploy-on-openbalena/99205

This happens because of recent updates to balenalib base images that updated `/bin/sh`, replacing the executable binary with a [hashbang'ed](https://en.wikipedia.org/wiki/Shebang_(Unix)) script: https://github.com/balena-io/contracts/blob/master/contracts/sw.os%2Barch.sw/image-info.tpl#L3-L6

The fix appears to require changes to QEMU, so it better handles hashbangs:
  * https://github.com/balena-io/qemu/pull/17
  * https://github.com/balena-io/qemu/pull/19
  * Internal discussion: https://www.flowdock.com/app/rulemotion/r-beginners/threads/gQUwEH4kxRFKcw1KZJyJliWd9wa

**Known workaround**
Pin the balenalib base image to previous version. For example, the [Dockerfile.template](https://github.com/balena-io-examples/balena-node-hello-world/blob/v1.0.0/Dockerfile.template#L4) file may currently contains a FROM line:
```
FROM balenalib/%%BALENA_MACHINE_NAME%%-node:10-stretch-run
```
Replace it with:
```
FROM balenalib/%%BALENA_MACHINE_NAME%%-node:10-stretch-run-20200327
```
When the `-20200327` suffix is not included, the latest image version is used, which is currently `20200420` (version `20200417` is also affected by the issue). A list of the versions is available on Dockerhub: https://hub.docker.com/r/balenalib/raspberrypi3-node/tags?page=1&name=10-stretch-run

pdcastro:
Fixed in CLI release v11.31.26 and later, which uses a new version of the QEMU emulator ([v4.0.0+balena2](https://github.com/balena-io/qemu/releases/tag/v4.0.0%2Bbalena2)) that supports the latest balenalib base images.

PR #1789, opened 2020-04-29
balena build/deploy: Update QEMU version to support newer balenalib images

pdcastro:
Resolves: #1788 
Change-type: patch

Internal discussion: https://www.flowdock.com/app/rulemotion/r-beginners/threads/gQUwEH4kxRFKcw1KZJyJliWd9wa

### DIFF
--- a/lib/utils/qemu.js
+++ b/lib/utils/qemu.js
@@ -17,7 +17,7 @@
 
 import * as Promise from 'bluebird';
 import { getBalenaSdk } from './lazy';
-export const QEMU_VERSION = 'v4.0.0-balena';
+export const QEMU_VERSION = 'v4.0.0+balena2';
 export const QEMU_BIN_NAME = 'qemu-execve';
 
 export function qemuPathInContext(context) {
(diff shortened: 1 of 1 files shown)


lib/utils/qemu.js
```diff
@@ -84,11 +84,12 @@ export function installQemu(arch) {
...
-				const downloadArchiveName = `qemu-${QEMU_VERSION.replace(
-					/^v/,
-					'',
-				)}-${qemuArch}.tar.gz`;
-				const qemuUrl = `https://github.com/balena-io/qemu/releases/download/${QEMU_VERSION}/${downloadArchiveName}`;
+				const fileVersion = QEMU_VERSION.replace(/^v/, '').replace('+', '.');
+				const urlFile = encodeURIComponent(
+					`qemu-${fileVersion}-${qemuArch}.tar.gz`,
+				);
+				const urlVersion = encodeURIComponent(QEMU_VERSION);
```
pdcastro, line 91: The new QEMU version string uses a `'+'` character instead of `'-'`:
`'v4.0.0-balena'` _vs._
`'v4.0.0+balena2'`
This requires calling `encodeURIComponent`. Also, the GitHub release filename (download) replaces the `'+'` with a `'.'`. These are the reasons for the minor refactoring.

PR #1791, opened 2020-04-30
Errors refactor

srlowe:
This is a small refactor to: 
 - use `checkLoggedIn()` instead of `exitIfNotLoggedIn()`, and remove latter.
 - move `exitWithExpectedError()` and `printErrorMessage()` into errors module.

### DIFF
--- a/lib/errors.ts
+++ b/lib/errors.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2016-2020 Balena
+Copyright 2016-2020 Balena Ltd.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,6 +18,8 @@ import { stripIndent } from 'common-tags';
 import * as _ from 'lodash';
 import * as os from 'os';
 import { TypedError } from 'typed-error';
+import { getChalk } from './utils/lazy';
+import { getHelp } from './utils/messages';
 
 export class ExpectedError extends TypedError {}
 
@@ -130,8 +132,6 @@ const messages: {
 };
 
 export async function handleError(error: any) {
-	const { printErrorMessage } = await import('./utils/patterns');
-
 	process.exitCode =
 		error.exitCode === 0
 			? 0
@@ -145,7 +145,7 @@ export async function handleError(error: any) {
 	const message = [interpret(error)];
 
 	if (process.env.DEBUG && error.stack) {
-		message.push(error.stack);
+		message.push('\n' + error.stack);
 	}
 	printErrorMessage(message.join('\n'));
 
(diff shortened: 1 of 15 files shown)


lib/errors.ts
```diff
@@ -171,3 +170,29 @@ export async function handleError(error: any) {
...
+ * error reporting through Sentry.io's platform (raven.Raven.captureException).
+ * Note that lib/errors.ts provides top-level error handling code to catch any
+ * otherwise uncaught errors, AND to report them through Sentry.io. But many
+ * "expected" errors (say, a JSON parsing error in a file provided by the user)
+ * don't warrant reporting through Sentry.io.  For such mundane errors, catch
+ * them and call this function.
+ *
+ * DEPRECATED: Use `throw new ExpectedError(<message>)` instead.
+ */
+export function exitWithExpectedError(message: string | Error): never {
```
pdcastro, line 191: As this function was moved to a different file and this PR was pending review while other PRs got merged, I suggest a new string search ("exitWithExpectedError") over the CLI codebase to make sure that no usage was left behind.

lib/utils/promote.ts
```diff
@@ -325,7 +325,7 @@ async function createApplication(
 ): Promise<BalenaSdk.Application> {
 	const form = await import('resin-cli-form');
 	const validation = await import('./validation');
-	const patterns = await import('./patterns');
+	const { printErrorMessage } = await import('../errors');
```
pdcastro, line 328: The `errors` module is already imported at the top of the file, so `printErrorMessage` could be imported there too.

lib/app-capitano.coffee
```diff
@@ -20,7 +20,7 @@ actions = require('./actions')
 events = require('./events')
 
 capitano.permission 'user', (done) ->
-	require('./utils/patterns').exitIfNotLoggedIn()
+	require('./utils/patterns').checkLoggedIn()
```
pdcastro, line 23: If you haven't, could you test that this change in particular works? It depends on Capitano correctly handling a rejected promise here. (Previously, `process.exit()` was called so it didn't matter...)

lib/utils/device/deploy.ts
```diff
@@ -74,7 +74,7 @@ async function environmentFromInput(
 	serviceNames: string[],
 	logger: Logger,
 ): Promise<ParsedEnvironment> {
-	const { exitWithExpectedError } = await import('../patterns');
+	const { exitWithExpectedError } = await import('../../errors');
```
pdcastro, line 77: The `errors` module is already imported at the top of the file, so maybe `exitWithExpectedError` could be included there too. (3 occurrences in this file)

lib/utils/promote.ts
srlowe: Good catch.

lib/utils/device/deploy.ts
srlowe: Looks like that's a different errors module (`utils/device/errors.ts`).

lib/app-capitano.coffee
srlowe: Checked.

PR #1799, opened 2020-05-01
Avoid unnecessary api calls in `balena build` and `balena deploy`

Page-:
Change-type: patch

### DIFF
--- a/lib/actions/deploy.js
+++ b/lib/actions/deploy.js
@@ -296,20 +296,8 @@ Examples:
 					});
 				}
 			})
-			.then(() =>
-				Promise.join(
-					helpers.getApplication(appName),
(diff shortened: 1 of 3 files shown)


lib/actions/deploy.js
```diff
@@ -296,20 +296,8 @@ Examples:
 					});
 				}
 			})
-			.then(() =>
-				Promise.join(
-					helpers.getApplication(appName),
```
Page-, line 301: Previously we were calling `getApplication` here as well as in `getArchAndDeviceType`, by combining it all into `getAppWithArch` it allows us to save an api call, reduce duplicate code, and remove some `@ts-ignore`s

Page-:
Tested and looks ok locally

PR #1801, opened 2020-05-04
Configure the sentry command scope earlier

Page-:
This should allow for better sentry info in cases it fails early

### DIFF
--- a/lib/events.ts
+++ b/lib/events.ts
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as Sentry from '@sentry/node';
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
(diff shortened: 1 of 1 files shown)


lib/events.ts
```diff
@@ -44,7 +43,11 @@ const getMixpanel = _.once((balenaUrl: string) => {
  * information in Sentry.io error reporting, for CLI debugging purposes
  * (mainly unexpected/unhandled exceptions -- see also `lib/errors.ts`).
  */
-export function trackCommand(commandSignature: string) {
+export async function trackCommand(commandSignature: string) {
+	const Sentry = await import('@sentry/node');
```
pdcastro, line 47: By the way, `@sentry/node` could perhaps be bumped to v5.15.5 (current latest):
https://github.com/getsentry/sentry-javascript/blob/master/CHANGELOG.md
I have not tested it, though.

Issue #1779, opened 2020-04-28
Improve error boilerplate message


srlowe:
Base on recent feedback and discussions, the error boilerplate should be rewritten more succinctly.

Current message:
```
Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, "$env:DEBUG = 1" on
powershell, or "export DEBUG=1" on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For CLI bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```

srlowe:
@pdcastro, @Page- , @OdysLam 

How about something like this for reduced boilerplate?

```
Additional information may be available with the `--debug` flag.
For help, see our forums: https://forums.balena.io
For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/
```

We could also use the github url shortener (git.io) for the issues page: 
https://git.io/JfYbB

OdysLam:
@srlowe - just to be certain, this is for the unexpected error, correct?
The important part for me is not to be a wall of red text, other than that I am not opinionated.

srlowe:
@OdysLam , That right.  It will only be displayed in unexpected errors, and at the bottom of the help pages.  Expected errors will also not be red.  I think it was mainly Page that wanted to reduce the text.

pdcastro:
> Additional information may be available with the `--debug` flag.
> For help, see our forums: https://forums.balena.io
> For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/

Could we add the word "support" on the second line? Not sure if instead of, or in addition to the word "help". Suggestions:

* `For help, see our support forums: https://forums.balena.io` 👍 
* `For support, see our forums: https://forums.balena.io` 🤔 
* `For help and support, see our forums: https://forums.balena.io` 👎 

> We could also use the github url shortener (git.io) for the issues page

Tempting, but my vote goes to the long/expanded github URL because users can tell where it takes them before clicking on it. Just one vote. :-)

srlowe:
Yeah first line sounds good to me.  Also, perhaps visit might be better than see?
` - For help, visit our support forums: https://forums.balena.io`

OdysLam:
[...]

Issue #1757, opened 2020-04-23
Remove boilterplate and red color from expected errors.


pdcastro:
Feedback from the balena team itself (most recently from @OdysLam but also from others):

- [ ] Remove boilerplate from expected errors
- [ ] Remove use of color in expected errors
- [ ] Add boilerplate to command `balena help` and `balena help -v` output 
- [ ] Only first line should be red in unexpected errors

The suggestion is to shorten:

```
$ balena scan
You need root/admin privileges to run this command

Additional information may be available by setting a DEBUG=1 environment
variable: "set DEBUG=1" on a Windows command prompt, "$env:DEBUG = 1" on
powershell, or "export DEBUG=1" on Linux or macOS.

If you need help, don't hesitate in contacting our support forums at
https://forums.balena.io

For CLI bug reports or feature requests, have a look at the GitHub issues or
create a new one at: https://github.com/balena-io/balena-cli/issues/
```

To: 

```
$ balena scan
You need root/admin privileges to run this command
See 'balena scan --help'
```

And then the output of the `balena help` and `balena help -v`, would include the extra help, probably at the bottom:

```
$ balena help
Usage: balena [COMMAND] [OPTIONS]

Primary commands:

    help [command...]                       show help
    login                                   login to balena
    push <applicationOrDevice>              Start a remote build on the balena cloud build servers or a local mode device
    logs <uuidOrDevice>                     show device logs
    ssh <applicationOrDevice> [serviceName] SSH into the host or application container of a device
    apps                                    list all applications
    app <name>                              list a single application
    devices                                 list all devices
    device <uuid>                           list a single device
    tunnel <deviceOrApplication>            Tunnel local ports to your balenaOS device
    preload <image>                         preload an app on a disk image (or Edison zip archive)
    build [source]                          Build a single image or a multicontainer project locally
    deploy <appName> [image]                Deploy a single image or a multicontainer project to a balena application
    join [deviceiporhostname]               move a local device to an application on another balena server
    leave [deviceiporhostname]              remove a local device from its balena application
    scan                                    scan for balenaOS devices on your local network

Run `balena help --verbose` to list additional commands

Global Options:

    --help, -h
    --version, -v

If you are coming across errors, additional information may be available in
debug mode, by setting a DEBUG=1 environment variable:
"set DEBUG=1" on a Windows command prompt,
"$env:DEBUG = 1" on PowerShell, or
"export DEBUG=1" on Linux or macOS.

[...]

PR #1806, opened 2020-05-05
Improve presentation of errors, help

srlowe:
Improve presentation of errors, help, in line with recent discussions.

- Removed boilerplate from expected errors
- Removed color from expected errors
- Edited boilerplate to be more concise.
- Now only first lines of unexpected errors are red.
- Errors now print first line of stacktrace (if available) even when not in debug mode.
- Added extra expected errors pattern match for oclif parser errors.
- Added extra line after capitano command help to make it clearer, and match oclif.
- Added --debug to the list of global flags in help

Change-type: patch
Resolves: #1779 #1757

### DIFF
--- a/lib/actions/help.js
+++ b/lib/actions/help.js
@@ -73,8 +73,8 @@ const manuallySortedPrimaryCommands = [
 
 const general = function(_params, options, done) {
 	console.log('Usage: balena [COMMAND] [OPTIONS]\n');
-	console.log(messages.reachingOut);
-	console.log('\nPrimary commands:\n');
+
+	console.log('Primary commands:\n');
 
 	// We do not want the wildcard command
 	// to be printed in the help screen.
@@ -123,6 +123,9 @@ const general = function(_params, options, done) {
 				console.log('\nGlobal Options:\n');
 				print(parse(capitano.state.globalOptions).sort());
 			}
+			console.log(indent('--debug\n'));
+
+			console.log(messages.help);
 
 			return done();
 		})
@@ -151,6 +154,8 @@ const commandHelp = (params, _options, done) =>
 			console.log('\nOptions:\n');
 			print(parse(command.options).sort());
 		}
+
+		console.log();
 
 		return done();
 	});
--- a/lib/errors.ts
+++ b/lib/errors.ts
@@ -28,7 +28,7 @@ export class NotLoggedInError extends ExpectedError {}
 export class InsufficientPrivilegesError extends ExpectedError {}
 
 /**
- * instanceOf is a more reliable implemention of the plain `instanceof`
+ * instanceOf is a more reliable implementation of the plain `instanceof`
  * typescript operator, for use with TypedError errors when the error
  * classes may be defined in external packages/dependencies.
  * Sample usage:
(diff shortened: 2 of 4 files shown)


lib/errors.ts
```diff
@@ -126,56 +126,81 @@ const messages: {
...
-	// Report "unexpected" errors via Sentry.io
-	const Sentry = await import('@sentry/node');
-	Sentry.captureException(error);
-	await Sentry.close(1000);
-	// Unhandled/unexpected error: ensure that the process terminates.
-	// The exit error code was set above through `process.exitCode`.
-	process.exit();
+	// Expected?
+	const isExpectedError =
+		instanceOf(error, ExpectedError) ||
```
srlowe, line 169: Using the custom instanceOf function here, as intanceof may fail.

lib/errors.ts
```diff
@@ -126,56 +126,81 @@ const messages: {
...
-		/^Unexpected arguments?:/, // oclif's command line parsing error
-	];
-
-	if (
-		error instanceof ExpectedError ||
-		expectedErrorREs.some(re => re.test(message[0]))
-	) {
-		return;
+	if (error.stack) {
+		if (process.env.DEBUG) {
```
CameronDiver, line 159:
```suggestion
	if (error.stack && process.env.DEBUG) {
```

NIT

lib/errors.ts
```diff
@@ -126,56 +126,81 @@ const messages: {
...
+			message.push('\n' + error.stack);
+		} else {
+			// Include first line of stacktrace
+			message.push('\n' + error.stack.split(`\n`)[0]);
```
CameronDiver, line 163: Is this useful at all? I would assume we'd like to not show this completely without the DEBUG flag (hence my suggestion above)

lib/actions/help.js
```diff
@@ -123,6 +123,9 @@ const general = function(_params, options, done) {
 				console.log('\nGlobal Options:\n');
 				print(parse(capitano.state.globalOptions).sort());
 			}
+			console.log(indent('--debug\n'));
```
pdcastro, line 126: That's cheating! 🚓  But I will join your scheme. 😁

lib/errors.ts
pdcastro: Harmless, but the custom `instanceOf` function is only useful when the class (`ExpectedError`) is defined in an external package/repo (other than `balena-cli`), because of the risk of there being different installed versions of the external package in different `node_modules` folders (each such version listed in the `npm-shrinkwrap.json` file). In this case, `ExpectedError` is defined in `balena-cli` itself, in fact defined in this same file, so no such risk.

lib/utils/messages.ts
```diff
@@ -9,19 +9,16 @@ create a new one at: https://github.com/balena-io/balena-cli/issues/\
 `;
 
 const debugHint = `\
-Additional information may be available by setting a DEBUG=1 environment
-variable: "set DEBUG=1" on a Windows command prompt, "$env:DEBUG = 1" on
-powershell, or "export DEBUG=1" on Linux or macOS.\n
+Additional information may be available with the \`--debug\` flag.
```
pdcastro, line 12: A few lines above, there is a `reachingOut` message that apparently is still printed by the `balena login` command. Removing or shortening that message seems to fit under the objective of issue #1779, and this PR/commit is labeled with "Resolves #1779".

pdcastro: Or not.

lib/errors.ts
srlowe: This was a feature request from @Page- during the discussion about changes.

lib/actions/help.js
srlowe: Haha yeah. I'll consolidate this when we move the help command to oclif.

lib/utils/messages.ts
srlowe: Yeah I left that one in on purpose, because it seemed like some friendly output might be nice  and didn't seem to hurt there.  Perhaps let's leave it for this PR and see what people thing (will be a quick change to take it out later).

lib/errors.ts
srlowe: Ah yeah, will revert that.

Issue #1802, opened 2020-05-04
Should have a deprecation policy


thgreasi:
As a follow-up of adding a deprecation policy to the SDK, the CLI should also have one, so that we are able to eventually drop deprecated API endpoints.

See: https://app.frontapp.com/open/cnv_7d92qx9
HQ: https://github.com/balena-io/balena/issues/2032
See: https://www.flowdock.com/app/rulemotion/r-product/threads/A6-bJBldfUFnhG9vixTz-DLz9iB

thgreasi:
Here is the SDK PR for reference
See: https://github.com/balena-io/balena-sdk/pull/892

PR #1809, opened 2020-05-07
Add a deprecation policy

thgreasi:
Essentially mirroring the section from the SDK. Just changes the referenced name & versions.

Resolves: #1802
Change-type: minor
See: https://app.frontapp.com/open/cnv_7d92qx9
HQ: https://github.com/balena-io/balena/issues/2032
See: https://www.flowdock.com/app/rulemotion/r-product/threads/A6-bJBldfUFnhG9vixTz-DLz9iB

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,22 @@ If you come across any problems or would like to get in touch:
 * For bug reports or feature requests,
   [have a look at the GitHub issues or create a new one](https://github.com/balena-io/balena-cli/issues/).
 
+## Deprecation policy
+
+The balena CLI uses [semver versioning](https://semver.org/), with the concepts
+of major, minor and patch version releases.
+
+The latest release of the previous major version of the balena CLI will remain
+compatible with the balenaCloud backend services for one year from the date when
+the next major version is released. For example, balena CLI v10.17.5, as the
+latest v10 release, would remain compatible with the balenaCloud backend for one
+year from the date when v11.0.0 is released.
+
+At the end of this period, the older major version is considered deprecated and
+some of the functionality that depends on balenaCloud services may stop working
+at any time.
+Users are encouraged to regularly update the balena CLI to the latest version.
+
 ## Contributing (including editing documentation files)
 
 Please have a look at the [CONTRIBUTING.md](./CONTRIBUTING.md) file for some guidance before
--- a/automation/capitanodoc/capitanodoc.ts
+++ b/automation/capitanodoc/capitanodoc.ts
@@ -160,6 +160,7 @@ export async function getCapitanoDoc(): Promise<typeof capitanoDoc> {
 		mdParser.getSectionOfTitle('Installation'),
 		mdParser.getSectionOfTitle('Getting Started'),
 		mdParser.getSectionOfTitle('Support, FAQ and troubleshooting'),
+		mdParser.getSectionOfTitle('Deprecation policy'),
 	]);
 	capitanoDoc.introduction = sections.join('\n');
 	return capitanoDoc;
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -128,6 +128,22 @@ If you come across any problems or would like to get in touch:
 * For bug reports or feature requests,
   [have a look at the GitHub issues or create a new one](https://github.com/balena-io/balena-cli/issues/).
 
+## Deprecation policy
+
+The balena CLI uses [semver versioning](https://semver.org/), with the concepts
+of major, minor and patch version releases.
+
+The latest release of the previous major version of the balena CLI will remain
+compatible with the balenaCloud backend services for one year from the date when
+the next major version is released. For example, balena CLI v10.17.5, as the
+latest v10 release, would remain compatible with the balenaCloud backend for one
+year from the date when v11.0.0 is released.
+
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -140,6 +140,21 @@ If you come across any problems or would like to get in touch:
...
 
+## Deprecation policy
+
+The balena CLI uses [semver versioning](https://semver.org/), with the
+concepts of major, minor and patch version releases.
+
+The latest release of the previous major version of the balena CLI will be
+supported for one year from the date when the next major version is released.
+For example, the balena CLI v10.17.5 (the latest v10 release) will be supported for
+one year from the date when v11.0.0 is released.
```
pdcastro, line 151: > The latest release of the previous major version of the balena CLI will be supported for one year [...]
>> So, as discussed in the arch call, this is all about us keeping the required endpoints working and not about bugfixing.
>> https://chat.google.com/room/AAAAgEE5JmQ/Kiptkj_j6pc

To avoid the interpretation that "supported" means bug fixing, let's be more explicit that we mean balenaCloud backend service compatibility. I suggest:

```
## Deprecation policy

Several of the balena CLI commands make use of balenaCloud backend services, such as the
balena HTTP API and the balenaCloud builder. While a specific balena CLI version may be remain
available for download and installation for a long period of time, the backend services (HTTP
API) evolve independently and may eventually become incompatible with older versions of the
balena CLI. Several commands of such older balena CLI versions would then stop working. This
deprecation policy aims at providing some clarity to balena CLI users as to the period of time
that an older balena CLI version should remain compatible with balenaCloud backend services.

This deprecation policy is expressed in terms of [semver versioning](https://semver.org/),
which is used to identify balena CLI releases with the concepts of major, minor and patch
version releases.

**The latest release of the previous major version of the balena CLI shall remain compatible
with balenaCloud backend services for one year from the date when the next major version is
released.** For example, balena CLI v10.17.5 (the latest v10 release) shall remain compatible
with balenaCloud backend services for one year from the date when v11.0.0 is released.

To be clear, this statement is limited to balenaCloud backend service compatibility and does
not extend to software maintenance such as, for example, bug fixes and security patches. Such
maintenance is normally only provided for the latest major version. For this reason, users are
advised to regularly update their installation of the balena CLI.
```

README.md
```diff
@@ -140,6 +140,21 @@ If you come across any problems or would like to get in touch:
 * For bug reports or feature requests,
   [have a look at the GitHub issues or create a new one](https://github.com/balena-io/balena-cli/issues/).
 
+## Deprecation policy
```
pdcastro, line 143: In addition to appearing in the README file on GitHub and the npm registry, I think we would like this deprecation policy to appear on the docs site too: https://www.balena.io/docs/reference/balena-cli/
This is achieved by automatically adding this README section to the `doc/cli.markdown` file (in this repo), which in turn is achieved by adding the following line of code to the following file:
```
mdParser.getSectionOfTitle('Deprecation policy'),
```
https://github.com/balena-io/balena-cli/blob/v11.32.15/automation/capitanodoc/capitanodoc.ts#L162

README.md
thgreasi: It's a bit too descriptive imo, but I'm fine with it :+1:

thgreasi:
Let's also have @shaunmulligan take a look and maybe we can mirror this in the SDKs as well :)

README.md
thgreasi: @pdcastro here is something simpler, based on your initial suggestion on the SDK:
```
The balena CLI uses [semver versioning](https://semver.org/), with the
concepts of major, minor and patch version releases.

The balenaCloud backend services required to continue supporting the operation
of latest release of the previous major version of the balena CLI, will remain
compatible for one year from the date when the next major version is released.
For example, the balenaCloud backend services which are required to support the
operation of balena CLI v10.17.5 (the latest v10 release) will remain
operational and compatible for one year from the date when v12.0.0 is released.

At the end of this support period, the older major version is considered
deprecated and some of the functionality that depends to balenaCloud backend
services might unexpectedly stop working, and you are highly encouraged to
upgrade to the to latest major version.
```

pdcastro: * Instead of saying that the backend services will remain compatible with the CLI, it may be better to say that the CLI will remain compatible with the backend services. Almost the same, but the former sounds closer to "the backend will remain compatible with CLI v10.17.5", whereas the latter sounds closer to "we can release v10.17.6 so the CLI remains compatible with the backend".
* We can avoid the word "support" altogether.

New suggestion (even shorter):
```
The balena CLI uses [semver versioning](https://semver.org/), with the concepts of major,
minor and patch version releases.

The latest release of the previous major version of the balena CLI will remain compatible
with the balenaCloud backend services for one year from the date when the next major
version is released. For example, balena CLI v10.17.5, as the latest v10 release, would
remain compatible with the balenaCloud backend for one year from the date when v11.0.0 is
released.

At the end of this period, the older major version is considered deprecated and some of
the functionality that depends on balenaCloud services may stop working at any time.
Users are encouraged to regularly update the balena CLI to the latest version.
```

thgreasi: Loved it @pdcastro :heart: Re-pushing in a bit :tada:

thgreasi:
@balena-ci retest

PR #1811, opened 2020-05-08
Update resin-multibuild

srlowe:
Change-type: patch


CameronDiver:
@balena-ci retest

Issue #1148, opened 2019-03-28
Nested .gitignore / .dockerignore files are not interpreted correctly


pdcastro:
One of the CLI users has kindly provided a detailed description of a case where relative paths and wildcards in nested .gitignore and .dockerignore files are not interpreted as expected, with CLI version 8.0.0 and above:

https://github.com/Krismix1/balena_cli_test

This issue is closely related to #1032, but could be addressed independently in relation to the treatment of relative paths (versus the separate matter of whether .ignore should apply to different files or at a different image build stage as .dockerignore).

The known workaround is to move the nested .gitignore or .dockignore files up towards the root of the tree (or at the root of each service in a multi-service app), and modify the matching patterns to include the names of subfolders.

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_ibv7)

wlisac:
I just ran into this while building a Swift-based project with nested package dependencies.

+1 for this issue (and/or ideally https://github.com/balena-io/balena-cli/issues/1032).

jimsynz:
Just got bitten by this too, since I'm working on a multi container app and the `.dockerignore` files are in each service's subdirectories.

bekriebel:
I'm evaluating Balena right now and just hit this issue with a go project. If [this project](https://github.com/sirupsen/logrus) is included in a `vendor directory`, the way the `.gitignore` is being read is making the whole logrus folder not get included and failing the build.

pdcastro:
CLI release v11.34.0 introduces a new flag for the push / build / deploy commands: `--nogitignore` or `-G`. Quoting `balena help push`:
```
$ balena help push
...
DOCKERIGNORE AND GITIGNORE FILES
By default, both '.dockerignore' and '.gitignore' files are taken into account
in order to prevent files from being sent to the balenaCloud builder or Docker
or balenaEngine (balenaOS device).

However, this behavior has been DEPRECATED and will change in an upcoming major
version release. The --nogitignore (-G) option should be used to enable the new
behavior already now. This option will cause the CLI to:

* Disregard all '.gitignore' files at the source directory and subdirectories,
  and consider only the '.dockerignore' file (if any) at the source directory.
* Consequently, allow files to be sent to balenaCloud / Docker / balenaEngine
  even if they are listed in '.gitignore' files (a longstanding feature request).
[...]

PR #1813, opened 2020-05-10
Add --nogitignore flag (new treatment of .gitignore and .dockerignore files)

pdcastro:
Connects-to: #1032
Connects-to: #1148
Change-type: minor

balena team discussion invited on this private thread:
https://www.flowdock.com/app/rulemotion/i-cli/threads/AmVsLKlTc5MA1P9meyk0Oqon0it


pdcastro:
@balena-ci retest

npm-shrinkwrap.json
```diff
@@ -104,6 +104,11 @@
         }
       }
     },
+    "@balena/dockerignore": {
```
pdcastro, line 107: `@balena/dockerignore` does not have any additional dependencies, so it is not a big deal to temporarily have it in addition to `zeit/dockerignore` (until the next major version release).

tests/projects.ts
```diff
@@ -0,0 +1,85 @@
+/**
```
pdcastro, line 1: I refactored some "test data setup code" from `tests/docker-build.ts` to `tests/projects.ts` because I felt the code was more specific to the `tests/test-data/projects/` test data than specific to "docker build".

tests/test-data/projects/no-docker-compose/dockerignore2/symlink-b.txt
```diff
@@ -0,0 +1 @@
+src/src-b.txt
```
pdcastro, line 1: `symlink-a.txt` and `symlink-b.txt` are symbolic links rather than regular files. I found that they checkout as regular files on Windows with MSYS2, but it's OK, the tests are happy with this all the same. On Linux and macOS, the symbolic links are tested as symbolic links.

lib/utils/ignore.ts
```diff
@@ -182,3 +184,92 @@ export class FileIgnorer {
...
+): Promise<FileStats[]> {
+	const files: FileStats[] = [];
+	const dirEntries = await fs.readdir(dir);
+	await Promise.all(
+		dirEntries.map(async entry => {
+			const filePath = path.join(dir, entry);
+			const stats = await fs.stat(filePath);
+			if (stats.isDirectory()) {
+				files.push(...(await listFiles(projectDir, filePath)));
+			} else if (stats.isFile()) {
```
pdcastro, line 213: Here `stats` will not refer to a symbolic link, because `fs.stat(filePath)` (unlike `fs.lstat`) returns stats for the linked file in case `filePath` is a symbolic link. (Unless perhaps it was a symbolic link pointing to another symbolic link... Not sure if `fs.stat` follows links recursively...) The `isFile()` test is to guard against other "special files" like Linux character devices, block devices and such.

pdcastro: I have confirmed that `stats.isFile()` returns `true` even if the `filePath` is a symbolic link pointing to another symbolic link, pointing to a regular file. (Which suggests that `fs.stat()` follows symlinks recursively.)

Issue #1820, opened 2020-05-12
Anti-virus and windows detect balena.exe as rare threat


Biorix:
- **balena CLI version:**  11.33.3
- **Operating system** version: Windows 10 : 64-bit Windows on 64-bit Intel processor
- **Install method:**  zip standalone (balena-cli-v11.33.3-windows-x64-standalone.zip)

---

AVG anti-virus immediately block the .exe and send it to their lab for analysis. It returned as dangerous and has been quarantine.\
Windows is also blocking it itself.

The threat name for AVG is "Custody" and "DS#EvoGEn||algo"

I didn't find a proper log of the analysis within the sub-folders of AVG

I want to add the .exe to exception but since I haven't found any ticket regarding this issue I need to make sure it safe and that my download has not been corrupted somehow.

pdcastro:
@Biorix, thank you for reporting this issue. I have removed the `balena-cli-v11.33.3-windows-x64-standalone.zip` from download in the releases page while we investigate this issue. What I have found so far:

* I installed the AVG antivirus from `avg.com` on a Windows 10 Home (Build 18363) laptop. The laptop also has the built-in Microsoft "Windows Security" application that includes a "Virus & threat protection" scanner.
* I updated both antivirus applications to the latest "virus threat definitions" and "security intelligence" as of approximately 9pm UTC on 12 May 2020.
* I downloaded the balena CLI zip packages and executable installers for Windows for releases v11.33.2 and v11.33.3 and scanned them with both antivirus apps.

Results:

balena-cli asset | AVG antivirus | Microsoft Windows Security
--- | --- | ---
v11.33.2 executable installer | Pass | Pass
v11.33.2 zip file | Pass | Pass
v11.33.2 zip file (unzipped) | Pass | Pass
v11.33.2 "C:\Program Files\balena-cli" | Pass | Pass
v11.33.3 executable installer | Pass | Pass
v11.33.3 zip file | Pass | Pass
v11.33.3 zip file (unzipped) | Fail on `balena.exe` | Pass
v11.33.3 "C:\Program Files\balena-cli" | Pass | Pass

I am now trying to isolate the difference between the versions and installers that could explain the AVG fail, to try and conclude whether this could be a false positive.

pdcastro:
[...]

PR #1821, opened 2020-05-12
Re-create standalone zip package (release asset) for Windows

pdcastro:
Connects-to: #1820 
Change-type: patch

See https://github.com/balena-io/balena-cli/issues/1820#issuecomment-627597915


package.json
```diff
@@ -260,3 +260,4 @@
     "windosu": "^0.3.0"
   }
 }
+
```
pdcastro, line 263: This PR simply adds a blank line at the end of the `package.json` file. The `npm-shrinkwrap.json` file is unchanged and the dependencies are the same as in v11.33.3. I have tested that this is enough to change the hash code of the `balena.exe` file (part of the standalone zip package) so that the AVG antivirus on Windows no longer identifies `balena.exe` as containing malware. I believe this is evidence that the AVG report was a false positive.

Issue #1824, opened 2020-05-13
balena app create "You already have an application with that name!"


pdcastro:
An application named "qemux86" does not exist in the user's account (as correctly reported by `balena apps`), and yet `balena app create` errors with _"You already have an application with that name!"_
```
$ balena apps
ID      APP NAME                       DEVICE TYPE ONLINE DEVICES DEVICE COUNT
1662555 balena-e2e-balena-cloud_com    qemux86-64  0              2
1661450 balena-e2e-k8s_balena-cloud_co qemux86-64  0              2
1662550 balena-e2e-resin_io            qemux86-64  0              2

$ balena app create qemux86
You already have an application with that name!
```
As reported in (private thread) https://www.flowdock.com/app/rulemotion/resin-devices/threads/9rJNFKmsEEn_n1xOMcvL2CN1NyV

Issue #1770, opened 2020-04-27
Release new major version (feature list)


pdcastro:
**Also, a number of "breaking changes" had been discussed to go in the next major release:**

- [x] [pdcastro] Update `balena-sdk` to v13, which requires dropping Node 8 support (#1742)
- [x] [pdcastro] Update `pkg` to v4.4.8 (drops support for Node 8)
- [x] [pdcastro] Update `@balena/lint` to `^5.0.0` (uses prettier v2 which does not work with Node 8)

**Doable with "feature switches":**

- [x] [pdcastro] Have `--nogitignore` be the default for balena push/build/deploy (drop support for `gitignore` - issues #1032 and #1148)
- [x] [srlowe] Have the ‘--all’ option be the default behaviour of ‘balena envs’
- [x] [pdcastro] Have the ALIASES column show by default in the output of `devices supported`
- [x] [pdcastro] Have the SLUG column show by default in the output of `balena apps`
- [x] [pdcastro] Promote project validation check for ‘docker-compose.yml’ in parent folder from warning to error.
- [x] [srlowe] Have `--convert-eol` be the default behaviour of ‘balena push’, ‘build’, ‘deploy’
- [x] [pdcastro] Have “--logs” by default in `balena build` or `balena deploy --build`, because some build errors are hidden without that option, leading to support queries.
- [x] [srlowe]: Remove ID from `tags` result (ID not useful/usable).

- [x] [srlowe] Improve error message when not logged in. Currently it is:
> _You have to log in to continue_
> _Run the following command to go through the login wizard:_
> 	_$ balena login_

I think it would be nicer to have something like:
> _Login required: use the “balena login” command to log in._

**Changes that are not breaking, and don't need to be part of v12 release:**
- [x] [srlowe]: Consolidate all 4 `public-url` commands to one.
- [ ] [srlowe] ~~`env rename` should be changed to `env set`, because it changes a variable’s value rather than its name (ref: #1414)~~ Let's add `env set`, and document `env rename` as an alias to `env set`. This doesn't have to be a breaking change.

**Changes not being made:** (at this time)
- ~~[srlowe]: Consider `tag set` -> `tag add`~~ (doesn't really make sense with `env set` since change was for consistency)
[...]

PR #1825, opened 2020-05-13
`app create`: "You already have an application with that name!"

pdcastro:
Resolves: #1824
Change-type: minor

The PR also adds a `--verbose` option to `balena apps`, to include the application slug in the output (app's "full name").

### DIFF
--- a/lib/actions-oclif/app/create.ts
+++ b/lib/actions-oclif/app/create.ts
@@ -78,22 +78,25 @@ export default class AppCreateCmd extends Command {
 		const balena = getBalenaSdk();
 		const patterns = await import('../../utils/patterns');
 
-		// First make sure they don't already have an app with this name
-		if (await balena.models.application.has(params.name)) {
-			throw new ExpectedError(
-				'You already have an application with that name!',
-			);
-		}
-
 		// Create application
 		const deviceType = options.type || (await patterns.selectDeviceType());
-		const application = await balena.models.application.create({
-			name: params.name,
-			deviceType,
-		});
-
+		let application: import('balena-sdk').Application;
+		try {
+			application = await balena.models.application.create({
+				name: params.name,
+				deviceType,
+			});
+		} catch (err) {
+			// BalenaRequestError: Request error: Unique key constraint violated
+			if ((err.message || '').toLowerCase().includes('unique')) {
+				throw new ExpectedError(
+					`Error: application "${params.name}" already exists`,
+				);
+			}
+			throw err;
+		}
 		console.info(
-			`Application created: ${application.app_name} (${application.device_type}, id ${application.id})`,
+			`Application created: ${application.slug} (${application.device_type}, id ${application.id})`,
 		);
 	}
 }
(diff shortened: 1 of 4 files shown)


lib/actions-oclif/apps.ts
```diff
@@ -46,21 +47,25 @@ export default class AppsCmd extends Command {
 
 	public static flags: flags.Input<FlagsDef> = {
 		help: cf.help,
+		verbose: flags.boolean({
+			char: 'v',
+			description: 'add extra columns in the tabular output (SLUG)',
```
pdcastro, line 52: If we just added the `slug` column in the output without a `--verbose` option, this would technically be a breaking change that required a major version release (because users parse the CLI output in shell scripts with `awk`, `sed`, etc). But we can add the SLUG column by default in the next major version release: see #1770

lib/errors.ts
```diff
@@ -132,6 +132,7 @@ const messages: {
 };
 
 const EXPECTED_ERROR_REGEXES = [
+	/^BalenaAmbiguousApplication:/, // balena-sdk
```
pdcastro, line 135: This error is raised is someone runs, for example:
```
balena app rm qemux86 -t qemux86-64
```
But not if I run:
```
balena app rm gh_paulo_castro/qemux86
```

pdcastro:
@balena-ci retest

lib/actions-oclif/app/create.ts
```diff
@@ -78,22 +78,25 @@ export default class AppCreateCmd extends Command {
...
-
+		let application: import('balena-sdk').Application;
+		try {
+			application = await balena.models.application.create({
+				name: params.name,
+				deviceType,
+			});
+		} catch (err) {
+			// BalenaRequestError: Request error: Unique key constraint violated
+			if ((err.message || '').toLowerCase().includes('unique')) {
```
Page-, line 91:
```suggestion
			if (err.message?.toLowerCase()?.includes('unique')) {
```
or
```suggestion
			if (/unique/i.test(err.message)) {
```

lib/errors.ts
Page-: For these expected errors can they not also be handled via catching the specific error class rather than a regex?

pdcastro: I merged the PR just a minute before you added a review. :-)  Re catching the specific error class, it may be possible by using a custom `instanceOf` function, see:
https://github.com/balena-io/balena-cli/blob/v11.35.0/lib/errors.ts#L53

There was a bug fixed recently where the plain `instanceof` operator would fail to match error classes because there were multiple installations (multiple versions) of `balena-errors` in the `npm-shrinkwrap.json` file, such that, for example, `BalenaAmbiguousApplication` may be raised from such a class from a different npm package version than the npm package version providing `BalenaAmbiguousApplication` for use by `instanceof` in `errors.ts`. In other words, as I understand it, the same class may be treated as a different class if they come from different `node_modules` folders...

PR #1826, opened 2020-05-14
Update typescript to 3.9

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/actions/deploy.js
+++ b/lib/actions/deploy.js
@@ -54,110 +54,113 @@ const deployProject = function(docker, logger, composeOpts, opts) {
 			}
 
 			// find which services use images that already exist locally
-			return Promise.map(project.descriptors, function(d) {
-				// unconditionally build (or pull) if explicitly requested
-				if (opts.shouldPerformBuild) {
-					return d;
-				}
-				return docker
-					.getImage(typeof d.image === 'string' ? d.image : d.image.tag)
-					.inspect()
-					.return(d.serviceName)
-					.catchReturn();
-			})
-				.filter(d => !!d)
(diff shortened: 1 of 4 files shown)


lib/actions/deploy.js
```diff
@@ -54,110 +54,113 @@ const deployProject = function(docker, logger, composeOpts, opts) {
...
+							d =>
+								builtImages[d.serviceName] ?? {
+									serviceName: d.serviceName,
+									name: typeof d.image === 'string' ? d.image : d.image.tag,
+									logs: 'Build skipped; image for service already exists.',
+									props: {},
+								},
+						),
+					)
+					// @ts-ignore slightly different return types of partial vs non-partial release
```
Page-, line 114: This comment triggered a reformat, the rest of the code is unchanged apart from formatting

lib/utils/compose.js
```diff
@@ -195,7 +195,7 @@ function originalTarDirectory(dir, param) {
 
 	const getFiles = () =>
 		// @ts-ignore `klaw` returns a `Walker` which is close enough to a stream to work but ts complains
-		streamToPromise(klaw(dir))
+		Promise.resolve(streamToPromise(klaw(dir)))
```
Page-, line 198: This is to enforce that we have a bluebird promise that we can `.filter`/`.map` on as typescript couldn't infer it fully from the js  (`streamToPromise` can return a native promise depending on the `any-promise` import I believe, so it types it as a native promise by default but in our usage it's actually a bluebird promise)

Issue #1710, opened 2020-04-13
Reword web authorization 


pdcastro:
`balena login`, choose "web authorization", the browser opens and the web page says:

> Success!
> You successfully logged in the balena CLI
> [Go to the dashboard](https://dashboard.balena-cloud.com/)

The _"Go to the dashboard"_ button does not make much sense, does it? It should be more like:  
_"You may now close this page and go back to the command prompt"._

Relate issue: #1404 [Login command may hang for a while after authorizing through the web dashboard](https://github.com/balena-io/balena-cli/issues/1404)

pdcastro:
Implemented in CLI release v11.35.2.

Issue #1404, opened 2019-08-22
Login command may hang for a while after authorizing through the web dashboard


roman-mazur:
- **balena CLI version:** 11.8.3
- **Operating system version:** macOS 10.14.5
- **32/64 bit OS and processor:** 64
- **Install method:** npm
- **If npm install, Node.js and npm version:** Node v10.16.0, npm 6.9.2

I can reproduce this almost every time:
- `balena login`
- Select "Web authorization"
- Click on "Authorize" on the dashboard
- See a page saying authorization has been granted with a button to go to the dashboard

I expect the CLI process to exit at this moment (there is no reason why it should hang). However, in practice, it may hang up to 1 minute until it understands login has succeeded.
If you interrupt the CLI process, you can use balena CLI commands - so the token is actually delivered to the CLI, but for some the process continues to wait.

roman-mazur:
It has been bothering for a couple of months... Every time I promised I will look into the code and find the reason, but it never was a good moment, so finally decided to record the issue (didn't find other ope ticket).

roman-mazur:
This might be related: #455

pdcastro:
Possibly related to https://github.com/balena-io/balena-cli/issues/1436#issuecomment-548401846 (CC: @srlowe)

pdcastro:
Useful data point thanks to an investigation on #1765: replacing the browser with the following curl command line allows web authorization to complete "immediately", while the browser may hang for up to a full minute:

```
curl -i -X POST \
   -H "Content-Type:application/x-www-form-urlencoded" \
   -H "Accept-Language:en-US,en;q=0.9" \
   -H "Cache-Control:max-age=0" \
   -H "Connection:keep-alive" \
   -H "DNT:1" \
   -H "Host:[::1]:8989" \
   -H "Origin:null" \
   -H "Sec-Fetch-Dest:document" \
   -H "Sec-Fetch-Mode:navigate" \
   -H "Sec-Fetch-Site:cross-site" \
   -H "Sec-Fetch-User:?1" \
   -H "Upgrade-Insecure-Requests:1" \
   -H "User-Agent:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.113 Safari/537.36" \
[...]

Issue #1810, opened 2020-05-08
sfdisk: /img/balena.img: does not contain a recognized partition table


curtiskoelling:
- **balena CLI version:** 11.32.15
- **Operating system version:** Ubuntu 18.04
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** npm Node v8.10.0 and npm v6.10.3

`balena preload balena-cloud-[APP_NAME]-intel-nuc-2.47.1+rev1-v10.6.27.img.zip --splash-image [SPLASH_IMG].png --app [MY_APP_ID] --debug`

**ERROR**
```
Traceback (most recent call last):
  File "/usr/src/app/preload.py", line 811, in <module>
    PARTITIONS_CACHE[None] = prepare_global_partitions()
  File "/usr/src/app/preload.py", line 80, in prepare_global_partitions
    return get_partitions(IMAGE)
  File "/usr/src/app/preload.py", line 69, in get_partitions
    return {p.label: p for p in PartitionTable(image).partitions if p.label}
  File "/usr/src/app/preload.py", line 381, in __init__
    sfdisk("--dump", "--json", image).stdout.decode("utf8")
  File "/usr/lib/python3.6/site-packages/sh.py", line 1427, in __call__
    return RunningCommand(cmd, call_args, stdin, stdout, stderr)
  File "/usr/lib/python3.6/site-packages/sh.py", line 774, in __init__
    self.wait()
  File "/usr/lib/python3.6/site-packages/sh.py", line 792, in wait
    self.handle_command_exit_code(exit_code)
  File "/usr/lib/python3.6/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: 

  RAN: /sbin/sfdisk --dump --json /img/balena.img

  STDOUT:


  STDERR:
sfdisk: /img/balena.img: does not contain a recognized partition table
    at container.wait.then (/usr/local/lib/node_modules/balena-cli/node_modules/balena-preload/lib/preload.js:317:19)
    at tryCatcher (/usr/local/lib/node_modules/balena-cli/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/usr/local/lib/node_modules/balena-cli/node_modules/bluebird/js/release/promise.js:547:31)
    at Promise._settlePromise (/usr/local/lib/node_modules/balena-cli/node_modules/bluebird/js/release/promise.js:604:18)
[...]

Issue #1683, opened 2020-03-27
`balena login` using web auth fails silently on a machine with no display attached


anujdeshpande:
- **balena CLI version:** e.g. 11.2.1 
- **Operating system version:** Ubuntu
- **Install method:** npm 
- **If npm install, Node.js and npm version:** e.g. Node v8.16.0 and npm v6.4.1

---

I sometimes use `ssh` from my Windows machine to a Linux box. For such a setup when I do ` balena login`, and use `Web authorization` it fails silently. Most likely because the display variable is not set. Is it possible to detect such a situation where the URL is shown instead of opening the default browser.

The other option would be something that shows verbose output of commands
https://github.com/balena-io/balena-cli/issues/1682

PR #1829, opened 2020-05-15
Fix 'balena login' web authorization hanging with Google Chrome

pdcastro:
Resolves: #1404
Resolves: #1710
Connects-to: #1810 
Connects-to: #1683 
Change-type: patch

### DIFF
--- a/lib/actions/preload.js
+++ b/lib/actions/preload.js
@@ -239,10 +239,11 @@ export const preload = {
 	description: 'preload an app on a disk image (or Edison zip archive)',
 	help: `\
 Preload a balena application release (app images/containers), and optionally
-a balenaOS splash screen, in a previously downloaded balenaOS image file (or
-Edison zip archive) in the local disk. The balenaOS image file can then be
-flashed to a device's SD card.  When the device boots, it will not need to
-download the application, as it was preloaded.
+a balenaOS splash screen, in a previously downloaded '.img' balenaOS image file
+in the local disk (a zip file is only accepted for the Intel Edison device type).
+After preloading, the balenaOS image file can be flashed to a device's SD card.
+When the device boots, it will not need to download the application, as it was
+preloaded.
 
 Warning: "balena preload" requires Docker to be correctly installed in
 your shell environment. For more information (including Windows support)
--- a/lib/auth/index.ts
+++ b/lib/auth/index.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 Balena
+Copyright 2016-2020 Balena
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -15,6 +15,9 @@ limitations under the License.
 */
 
 import { getBalenaSdk } from '../utils/lazy';
+import { awaitForToken, shutdownServer } from './server';
+
+export { shutdownServer };
 
 /**
  * @module auth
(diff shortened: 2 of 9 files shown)


lib/actions/preload.js
```diff
@@ -239,10 +239,11 @@ export const preload = {
...
 Preload a balena application release (app images/containers), and optionally
-a balenaOS splash screen, in a previously downloaded balenaOS image file (or
-Edison zip archive) in the local disk. The balenaOS image file can then be
-flashed to a device's SD card.  When the device boots, it will not need to
-download the application, as it was preloaded.
+a balenaOS splash screen, in a previously downloaded '.img' balenaOS image file
+in the local disk (a zip file is only accepted for the Intel Edison device type).
+After preloading, the balenaOS image file can be flashed to a device's SD card.
+When the device boots, it will not need to download the application, as it was
+preloaded.
```
pdcastro, line 246: Separate commit, partially addressing #1810, trying to make it clearer that `balena preload` requires a `.img` file instead of `.zip` for most device types. Re automatically unzipping for the user, the tricky bit is deciding the device type before unzipping, given that the device type may be determined by inspecting the image -- which requires the image to be in the correct format for that device type to begin with. There is another issue for that though: https://github.com/balena-io-modules/balena-preload/issues/156

lib/auth/index.ts
```diff
@@ -52,14 +55,14 @@ export const login = async () => {
 	// from mixed content warnings (as the target of a form in the result page)
 	const callbackUrl = `http://127.0.0.1:${options.port}${options.path}`;
 	const loginUrl = await utils.getDashboardLoginURL(callbackUrl);
+	console.info(`Opening web browser for URL:\n${loginUrl}`);
```
pdcastro, line 58: This changes the current output, from:
```
$ ./bin/balena login
Logging in to balena-cloud.com
? How would you like to login? Web authorization (recommended)
Connecting to the web dashboard
```
To:
```
$ ./bin/balena login
Logging in to balena-cloud.com
? How would you like to login? Web authorization (recommended)
Opening web browser for URL:
https://dashboard.balena-cloud.com/login/cli/http%253A%252F%252F127.0.0.1%253A8989%252Fauth
```
If you have any reservations about this change, please say so! It partially addresses issue #1683, which requires a URL to be printed so that web authorization can be performed on a different machine than the one the CLI is running on. _"Partially"_ because, as the moment, the printed URL uses a `127.0.0.1` redirect IP address, which is no good for remote authorisation. Still, the user could copy-and-paste the URL and manually replace the redirect IP address on the URL.

lib/auth/server.ts
```diff
@@ -60,77 +82,29 @@ export const awaitForToken = (options: {
...
-			token?: string;
-		}) => {
-			const context = await getContext(viewName);
-			response.status(statusCode).render(viewName, context);
-			request.connection.destroy();
-			closeServer(errorMessage, token);
-		};
-
 		app.post(options.path, async (request, response) => {
+			server.close(); // stop listening for new connections
```
pdcastro, line 86: Previously, `server.close()` was called at the end of `app.post`, now it is called at the beginning. Shouldn't matter much, as all it does is to stop accepting new connections (which frees up the listening port, and important detail for the existing automated tests). Note that the images and CSS used in the web page served by the CLI are inlined (base64 for the images), so the browser does not need to make additional HTTP requests to the CLI server.

PR #1834, opened 2020-05-19
Fix caching by preserving all file stats when pushing to device or cloud

CameronDiver:
This is likely one of the most frustrating bugs for our users, when sometimes cache is invalidated for apparently no reason. To fix this, save all stats from the host, not just size and permissions.

Change-type: patch

---

### DIFF
--- a/lib/utils/compose.js
+++ b/lib/utils/compose.js
@@ -228,7 +228,12 @@ function originalTarDirectory(dir, param) {
 				readFile(file),
 				(filename, stats, data) =>
 					pack.entry(
-						{ name: toPosixPath(filename), size: stats.size, mode: stats.mode },
+						{
+							name: toPosixPath(filename),
+							mtime: stats.mtime,
+							size: stats.size,
+							mode: stats.mode,
+						},
 						data,
 					),
 			);
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -189,8 +189,9 @@ export async function tarDirectory(
 		pack.entry(
 			{
 				name: toPosixPath(fileStats.relPath),
-				size: fileStats.stats.size,
+				mtime: fileStats.stats.mtime,
 				mode: fileStats.stats.mode,
+				size: fileStats.stats.size,
 			},
 			await readFile(fileStats.filePath),
 		);


lib/utils/compose_ts.ts
```diff
@@ -189,8 +189,7 @@ export async function tarDirectory(
 		pack.entry(
 			{
 				name: toPosixPath(fileStats.relPath),
-				size: fileStats.stats.size,
-				mode: fileStats.stats.mode,
+				...fileStats.stats,
```
pdcastro, line 192: This function applies when the `--nogitignore` option is used. When the option is not used, the `originalTarDirectory` function applies, so a similar change would presumably be also needed at:
https://github.com/balena-io/balena-cli/blob/v11.35.3/lib/utils/compose.js#L231

The plan is to de-duplicate the tarDirectory functions in the upcoming major version release, when the `--nogitignore` option would become the default behaviour (no-op).

CameronDiver: oh, done @pdcastro

lib/utils/compose.js
```diff
@@ -227,10 +227,7 @@ function originalTarDirectory(dir, param) {
 				fs.stat(file),
 				readFile(file),
 				(filename, stats, data) =>
-					pack.entry(
-						{ name: toPosixPath(filename), size: stats.size, mode: stats.mode },
-						data,
-					),
+					pack.entry({ ...stats, name: toPosixPath(filename) }, data),
```
pdcastro, line 230: OK. I worry a bit that `...stats` might includes properties that would inadvertently invalidate the cache (the opposite of what we are trying to achieve) if developers push to the same balena app from different machines (whether the same developer of different developers), because it includes lots of properties:
```
Stats {
  dev: 2114,
  ino: 48064969,
  mode: 33188,
  nlink: 1,
  uid: 85,
  gid: 100,
  rdev: 0,
  size: 527,
  blksize: 4096,
  blocks: 8,
  atimeMs: 1318289051000.1,
  mtimeMs: 1318289051000.1,
  ctimeMs: 1318289051000.1,
  birthtimeMs: 1318289051000.1,
  atime: Mon, 10 Oct 2011 23:24:11 GMT,
  mtime: Mon, 10 Oct 2011 23:24:11 GMT,
  ctime: Mon, 10 Oct 2011 23:24:11 GMT,
  birthtime: Mon, 10 Oct 2011 23:24:11 GMT }
```
But if you have tested it, I am OK with it. :-)

lib/utils/compose_ts.ts
```diff
@@ -189,7 +189,7 @@ export async function tarDirectory(
 		pack.entry(
 			{
 				name: toPosixPath(fileStats.relPath),
-				size: fileStats.stats.size,
+				mtime: fileStats.stats.mtime,
```
pdcastro, line 192: Removing `size`? I think that what we saw in the source code for `docker build` is that it gets the full header with "all fields":
* https://github.com/balena-os/balena-engine/blob/v19.03.11/pkg/archive/archive.go#L361
* https://golang.org/pkg/archive/tar/#FileInfoHeader

```
hdr, err := tar.FileInfoHeader(fi, link)
```
And then _modifies_ the full header, adjusting the following fields (and conditionally other fields elsewhere in source code):
https://github.com/balena-os/balena-engine/blob/v19.03.11/pkg/archive/archive.go#L366-L370
```
	hdr.ModTime = hdr.ModTime.Truncate(time.Second)
	hdr.AccessTime = time.Time{}
	hdr.ChangeTime = time.Time{}
	hdr.Mode = fillGo18FileTypeBits(int64(chmodTarEntry(os.FileMode(hdr.Mode))), fi)
	hdr.Name = canonicalTarName(name, fi.IsDir())
```
But, of course, we don't have to do the same that `docker build` does. :-)

CameronDiver: I just went for the smallest change, which was to add the mtime, everything else we can consider later (good catch on the size, missed it when reverting).

Issue #1727, opened 2020-04-15
Remove windows dns workaround in favour of single initial resolution.


srlowe:
There's a windows DNS issue that was worked around as described in the commend below.  We should replace the workaround with option 3 described there for the sake of code clarity.

https://github.com/balena-io/balena-cli/issues/1518#issuecomment-557136983

Issue #1528, opened 2019-12-05
Error pushing image using hostname


ll01:
- **balena CLI version:11.18.3
- **Operating system version: Windows 10
- **32/64 bit OS and processor:64-bit Windows on 64-bit AMD processor
- **Install method: executable installer

---

my pi's hostname is pi4-proxy

If I run the command Balena push pi4-proxy.local I get an error.
However, if I enter the IP address as the command it works fine.
The error It outputs is:

ENOTFOUND: getaddrinfo ENOTFOUND pi4-proxy.local pi4-proxy.local:48484
Error: getaddrinfo ENOTFOUND pi4-proxy.local pi4-proxy.local:48484
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:67:26)
From previous event:
    at Function.promisifiedRequest (C:\Program Files\balena-cli\client\build\utils\device\api.js:128:25)
    at DeviceAPI.getDeviceInformation (C:\Program Files\balena-cli\client\build\utils\device\api.js:42:26)
    at Object.deployToDevice (C:\Program Files\balena-cli\client\build\utils\device\deploy.js:91:34)
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:229:7)

![image](https://user-images.githubusercontent.com/6953881/70274008-3d133b00-17a3-11ea-88c1-53c5e27a1eb2.png)

pdcastro:
@ll01, thank you for reporting this issue. How are your RPi and your laptop/desktop computer connected to the local network? For example:
* Connected to each other by a direct Ethernet cable, with the laptop providing internet access through "internet connection sharing". 
* Or both connected to a WiFi router -- is it the same WiFi router for both?
* Or some other arrangement?

Also, is the `balena scan` command able to list the device?

I wonder if this could be the same issue as #1518 (Livepush using `.local` address fails on Windows (with incorrect error)). @srlowe, do you think this is something that could be fixed by PR #1525 (Introduce workaround for windows dns issue on local `balena push`)?

srlowe:
@pdcastro Yes, this looks very much like the same issue, which #1525 should fix!

srlowe:
@ll01 Please could you try with latest, v11.19.1, and see if that fixes your issue?  
cc @pdcastro

pdcastro:
[...]

Issue #1518, opened 2019-11-19
Livepush using `.local` address fails on Windows (with incorrect error).


srlowe:
Livepush using `xxxxxx.local` address does not work with Windows, and fails with the incorrect error `The supervisor version on this remote device does not support multicontainer local mode...`

It looks like the `api.getVersion()` call is failing with an address resolution error `getaddrinfo ENOTFOUND`.  (The prior call to `api.ping()` completes successfully however, presumably with working address resolution).

This error is then being incorrectly masked with the supervisor version error, by the error handling in `deployToDevice(...)`.

srlowe:
@CameronDiver Do you remember if there is any particular reason why the `try/catch` around `api.getVersion()` is masking errors with the `versionError`?  E.g. Maybe old supervisors not supporting the `getVersion` call, and so the implication that errors there are due to old versions?

https://github.com/balena-io/balena-cli/blob/a8bd5d332acd671002881c55625794bc3bc10106/lib/utils/device/deploy.ts#L142-L161

CameronDiver:
@srlowe that's right, the version string didn't exist until v7.21.0: https://www.balena.io/docs/reference/supervisor/supervisor-api/#supervisor-version-get-v2version

srlowe:
After some sleuthing with @hedss, it turns out to be that repeated DNS resolution calls in quick succession cause the problem, which appears to be a symptom of this windows bug:

https://support.microsoft.com/en-gb/help/4057932/getaddrinfo-failed-with-wsahost-not-found-11001-error

We came up with 3 possible workarounds:

1. (tested) Add the option { family: 4 } to http requests.
2. (not tested) Manually call DNS (once) and replace hostname with IP in subsequent request URLs.
3. (tested) Put a pause between each stop (just for windows I guess).

# 1 May interfere with ip6 networking
# 2 Has not been tested (though should work).
# 3 Works well with shortish delay (e.g. 300ms)

Will implement workaround # 3 (only when CLI host is windows).

srlowe:
Note that this Windows problem may be present in other parts of the CLI.

pdcastro:
@srlowe, FYI, looks like the commit headers did not automatically close this issue. Just a guess: a space may be required after the colon character (`:`) for parsing to succeed as documented in: https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

Issue #1585, opened 2020-01-30
balena push / build / deploy should support IPv6 addresses


pdcastro:
- **balena CLI version:** e.g. 11.23.0
- **Operating system version:** macOS 10.15.2
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** npm
- **If npm install, Node.js and npm version:** Node v10.18.1 and npm v6.13.4

---

Example using IPv4:
```
$ ./bin/balena deploy test-rpi --build -s ~/myapp -h 192.168.0.37 -p 2375
[Info]    Compose file detected
...
[Success] Deploy succeeded!
```

Using IPv6:
```
$ ./bin/balena build ~/myapp -a test-rpi -h fd01::fcc8:d9eb:8481:8971 -p 2375
Docker seems to be unavailable. Is it installed and running?

$ ./bin/balena deploy test-rpi --build -s ~/myapp -h fd01::fcc8:d9eb:8481:8971 -p 2375
Docker seems to be unavailable. Is it installed and running?

$ ./bin/balena push fd01::fcc8:d9eb:8481:8971 -s ~/myapp
Build target not recognised. Please provide either an application name or device address.
```

Yet I can ping the device, and I can ssh to it (though [not balena ssh](https://github.com/balena-io/balena-cli/issues/1540)):
```
$ ping6 fd01::fcc8:d9eb:8481:8971
PING6(56=40+8+8 bytes) fd01::c73:d05d:82d3:5d9f --> fd01::fcc8:d9eb:8481:8971
16 bytes from fd01::fcc8:d9eb:8481:8971, icmp_seq=0 hlim=64 time=7.472 ms
...

$ ssh -p 22222 root@fd01::fcc8:d9eb:8481:8971
Last login: Thu Jan 30 21:30:51 2020 from fd01::c73:d05d:82d3:5d9f
root@da4da23:~#
```

And I can even contact balenaEngine on the device over IPv6, using the standard docker client:
```
$ docker -H fd01::fcc8:d9eb:8481:8971 info
...
Server Version: 18.09.10-dev
Storage Driver: overlay2
 Backing Filesystem: extfs
...
Init Binary: balena-engine-init
Kernel Version: 4.19.71
Operating System: balenaOS 2.46.1+rev3
...
```

pdcastro:
Some notes in case this helps someone (myself in the future?) :-) to test with IPv6, on a local IPv4-only network (i.e. a local network that does not have an active IPv6 router).

[...]

PR #1838, opened 2020-05-22
Replace windows dns workaround with single lookup

srlowe:
This replaces a workaround for a windows dns issue affecting livepush (see https://github.com/balena-io/balena-cli/issues/1528 , https://github.com/balena-io/balena-cli/issues/1518), with an alternative approach that initially does a single dns lookup, and then uses the IP address for subsequent operations.  (Requested by @CameronDiver )

Change-type: patch
Connects-to: #1518
Resolves: #1727

### DIFF
--- a/lib/utils/device/deploy.ts
+++ b/lib/utils/device/deploy.ts
@@ -34,7 +34,6 @@ import {
 	loadProject,
 	makeBuildTasks,
 } from '../compose_ts';
-import { workaroundWindowsDnsIssue } from '../helpers';
 import Logger = require('../logger');
 import { DeviceAPI, DeviceInfo } from './api';
 import * as LocalPushErrors from './errors';
@@ -125,6 +124,18 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 	const { exitWithExpectedError } = await import('../../errors');
 	const { displayDeviceLogs } = await import('./logs');
 
+	// Resolve .local addresses to IP to avoid
+	// issue with Windows and rapid repeat lookups.
+	// see: https://github.com/balena-io/balena-cli/issues/1518
+	if (opts.deviceHost.includes('.local')) {
+		const util = await import('util');
+		const dns = await import('dns');
+		const { address } = await util.promisify(dns.lookup)(opts.deviceHost, {
+			family: 4,
+		});
+		opts.deviceHost = address;
+	}
+
 	const api = new DeviceAPI(globalLogger, opts.deviceHost);
 
 	// First check that we can access the device with a ping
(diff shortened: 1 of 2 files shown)


lib/utils/device/deploy.ts
```diff
@@ -125,6 +124,14 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 	const { exitWithExpectedError } = await import('../../errors');
 	const { displayDeviceLogs } = await import('./logs');
 
+	// Resolve .local addresses to IP to avoid
+	// issue with Windows and rapid repeat lookups.
+	// see: https://github.com/balena-io/balena-cli/issues/1518
+	if (opts.deviceHost.includes('.local')) {
+		const dns = await import('dns');
+		opts.deviceHost = (await dns.promises.lookup(opts.deviceHost)).address;
```
pdcastro, line 132: The `.promises` bit -- was this something that didn't work with Node v8? We need to make progress on #1770 and drop Node 8 as soon as we can...

pdcastro:
```
$ nvm use 8
Now using node v8.17.0 (npm v6.13.4)
$ node
> require('dns').promises
undefined
```
```
$ nvm use 12
Now using node v12.16.3 (npm v6.14.4)
$ node
> require('dns').promises
{ lookup: [Function: lookup],
...
```
Also: I understand that the CLI does not yet support IPv6 (#1585), and looking at the documentation for `dns.lookup(hostname, options)`, it is possible to pass an `options` object to request the IP address resolution to use IPv4 instead of IPv6: `{ family: 4 }`.
https://nodejs.org/docs/latest-v8.x/api/dns.html#dns_dns_lookup_hostname_options_callback
Although it is not strictly necessary "for now" because the documentation also reads:

> verbatim <boolean> When true, the callback receives IPv4 and IPv6 addresses in the order the DNS resolver returned them. When false, IPv4 addresses are placed before IPv6 addresses. Default: currently false (addresses are reordered) **but this is expected to change in the not too distant future.** New code should use { verbatim: true }.

srlowe: Ah yes, I think you're right.  In that case we will need to hold off merging this until after 1770 then I guess.

srlowe: Re ip4, right that might avoid some surprises in the future.

pdcastro: > we will need to hold off merging this until after 1770

`dns.promises` is not available but `util.promisify()` can be used instead (with Node v8):

> https://nodejs.org/docs/latest-v8.x/api/dns.html#dns_dns_lookup_hostname_options_callback
> If this method is invoked as its util.promisify()ed version, and all is not set to true, it returns a Promise for an object with address and family properties.

srlowe:
Tested with node v8

PR #1840, opened 2020-05-25
balena deploy: Fix "access denied" pushing images to registry

pdcastro:
This bug was introduced in the Typescript conversion in #1837 (my bad...)
Kindly reported by @xginn8 in https://www.flowdock.com/app/rulemotion/i-cli/threads/yDJIhV8zk_Iu65rLKVpPJ-cG2V5

Change-type: patch

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -588,9 +588,7 @@ async function getTokenForPreviousRepos(
 	const { authorizePush, getPreviousRepos } = await import('./compose');
 	const sdk = getBalenaSdk();
 	const previousRepos = await getPreviousRepos(sdk, docker, logger, appId);
-	if (!previousRepos || previousRepos.length === 0) {
-		return '';
-	}
+
 	const token = await authorizePush(
 		sdk,
 		apiEndpoint,


lib/utils/compose_ts.ts
```diff
@@ -588,9 +588,6 @@ async function getTokenForPreviousRepos(
 	const { authorizePush, getPreviousRepos } = await import('./compose');
 	const sdk = getBalenaSdk();
 	const previousRepos = await getPreviousRepos(sdk, docker, logger, appId);
-	if (!previousRepos || previousRepos.length === 0) {
-		return '';
```
pdcastro, line 592: I had added this `if` test as an "optimization" but I misread the implementation of `authorizePush()`: it should be called even if `previousRepos` is an empty array.

pdcastro:
Force push to fix CI error `npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/balena-cli - You cannot publish over the previously published versions: 11.35.9-fix-deploy-authorize-token-2d72eb0da11fba16b062bd3c30a9902e796706be.`

PR #1850, opened 2020-05-28
v12 preparations: Add feature switch for '--nogitignore'

pdcastro:
Connects-to: #1770
Change-type: patch

### DIFF
--- a/lib/utils/helpers.ts
+++ b/lib/utils/helpers.ts
@@ -25,9 +25,6 @@ import * as ShellEscape from 'shell-escape';
 import { ExpectedError } from '../errors';
 import { getBalenaSdk, getChalk, getVisuals } from './lazy';
 
-import * as semver from 'semver';
-import { version } from '../../package.json';
-
 export function getGroupDefaults(group: {
 	options: Array<{ name: string; default?: string }>;
 }): { [name: string]: string | number | undefined } {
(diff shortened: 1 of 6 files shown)


lib/utils/version.ts
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as semver from 'semver';
+import { version } from '../../package.json';
+
+export function isVersionGTE(v: string): boolean {
```
pdcastro, line 21: I came across a module import loop when `messages.ts` imported `isVersionGTE` from `helpers.ts`, which imported `errors.ts`, which imported `messages.ts`. Hence moving it to its own `version.ts` module.

lib/utils/helpers.ts
```diff
@@ -468,7 +465,3 @@ export function getProxyConfig(): ProxyConfig | undefined {
 		}
 	}
 }
-
-export function isVersionGTE(v: string) {
```
pdcastro, line 472: See comment in `version.ts`

PR #1851, opened 2020-05-29
v12 preparations: Add feature switch for 'envs --all'

srlowe:
Change-type: patch
Connects-to: #1770

### DIFF
--- a/lib/actions-oclif/envs.ts
+++ b/lib/actions-oclif/envs.ts
@@ -24,9 +24,10 @@ import { ExpectedError } from '../errors';
 import * as cf from '../utils/common-flags';
 import { getBalenaSdk, getVisuals } from '../utils/lazy';
 import { CommandHelp } from '../utils/oclif-utils';
+import { isV12 } from '../utils/version';
 
 interface FlagsDef {
-	all?: boolean; // whether to include application-wide, device-wide variables
+	all?: boolean; // whether to include application-wide, device-wide variables //TODO: REMOVE
 	application?: string; // application name
 	config: boolean;
 	device?: string; // device UUID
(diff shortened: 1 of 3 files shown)


lib/actions-oclif/envs.ts
```diff
@@ -93,28 +129,50 @@ export default class EnvsCmd extends Command {
...
 	).trim();
 
 	public static flags: flags.Input<FlagsDef> = {
-		all: flags.boolean({
-			description: stripIndent`
+		...(isV12()
+			? {
+					all: flags.boolean({
+						description: stripIndent`
+				No-op since balena CLI v12.0.0. See "balena help envs".`,
```
pdcastro, line 165: I think _"See balena help envs"_ could be removed, as `balena help envs` doesn't clarify much regarding the `--all` option. I guess you got inspiration for it from my previous PR (`--nogitignore`), but in the case of that PR, the `balena help push` output included a paragraph that actually mentioned `.gitignore` files and explained that _"Previous balena CLI releases (before v12.0.0) also took '.gitignore' files into account, but this is no longer the case."_

lib/actions-oclif/envs.ts
```diff
@@ -202,6 +266,22 @@ export default class EnvsCmd extends Command {
...
+				});
+
+				fields.push(options.json ? 'appName' : 'appName => APPLICATION');
+				if (options.device) {
+					fields.push(options.json ? 'deviceUUID' : 'deviceUUID => DEVICE');
+				}
+				if (!options.config) {
+					fields.push(options.json ? 'serviceName' : 'serviceName => SERVICE');
+				}
+			}
```
pdcastro, line 284: Wouldn't it be simpler and much less code duplication to have the following? Early on in the `run()` method:
```
options.all = options.all || isV12();
```
Then all code that currently refers to `options.all` can remain unchanged. After v12 is released, we can go around deleting some redundant conditionals.

pdcastro:
Build tests are failing because `cli.markdown` was not updated (`npm run build` or `npm test` needs to be run locally with `BALENA_CLI_VERSION_OVERRIDE` unset).

PR #1854, opened 2020-06-02
Convert 'balena devices public-url' commands to oclif

srlowe:
Change-type: patch


lib/actions-oclif/device/public-url.ts
```diff
@@ -0,0 +1,116 @@
...
+import Command from '../../command';
+import { ExpectedError } from '../../errors';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk } from '../../utils/lazy';
+import { tryAsInteger } from '../../utils/validation';
+
+interface FlagsDef {
+	enable: boolean;
+	disable: boolean;
+	status: boolean;
```
pdcastro, line 30: I'm thinking whether `enable`, `disable` and `status` could be in ArgsDef instead of FlagsDef, i.e. be positional arguments instead of flags/options. This would keep compatibility with the current implementation and avoid the need for a breaking change and feature switches.

lib/actions-oclif/device/public-url.ts
```diff
@@ -0,0 +1,116 @@
...
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			DevicePublicUrlCmd,
+		);
+
+		const balena = getBalenaSdk();
+
+		if (options.enable) {
```
pdcastro, line 87: If `enable` was a positional argument, would the change here be as simple as replacing `options.enable` with `params.enable`?

pdcastro:
@balena-ci retest

doc/cli.markdown
```diff
@@ -309,7 +306,7 @@ Examples:
 
 #### -v, --verbose
 
-add extra columns in the tabular output (SLUG)
+No-op since release v12.0.0
```
pdcastro, line 309: It looks like `npm run build` was run with `BALENA_CLI_VERSION_OVERRIDE=12.0.0`. The plan is to do this for the `change-type: major` PR, but this PR is just `patch`.

lib/actions-oclif/device/public-url.ts
```diff
@@ -0,0 +1,143 @@
...
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			DevicePublicUrlCmd,
+		);
+
+		// Legacy command format support.
+		// Previous this command used the following format
+		// (changed due to oclif technicalities):
+		//   `balena device public-url enable|disable|status <uuid>`
+		if (params.legacyUuid) {
+			const action = params.uuid;
```
pdcastro, line 104: Nice! Perhaps we could add here an extra validation against typos on the command line:
```
if (!['enable', 'disable', 'status'].includes(action)) {
    throw new ExpectedError(`Unexpected arguments: ${params.uuid} ${params.legacyUuid}`);
}
```

doc/cli.markdown
srlowe: Oh damn, sorry!

srlowe:
@balena-ci retest

PR #1856, opened 2020-06-03
v12 preparations: Add feature switch for default eol-converson

srlowe:
Change-type: patch
Connects-to: #1770

### DIFF
--- a/lib/actions/build.js
+++ b/lib/actions/build.js
@@ -60,7 +60,7 @@ const buildProject = function(docker, logger, composeOpts, opts) {
 				opts.buildEmulated,
 				opts.buildOpts,
 				composeOpts.inlineLogs,
-				opts.convertEol,
+				composeOpts.convertEol,
 				composeOpts.dockerfilePath,
 				composeOpts.nogitignore,
 			);
@@ -152,9 +152,6 @@ Examples:
 		}
 		delete params.source;
 
-		options.convertEol = options['convert-eol'] || false;
-		delete options['convert-eol'];
-
 		const { application, arch, deviceType } = options;
 
 		return Promise.try(function() {
@@ -203,7 +200,6 @@ Examples:
 							deviceType: resolvedDeviceType,
 							buildEmulated: !!options.emulated,
 							buildOpts,
-							convertEol: options.convertEol,
 						}),
 				);
 			});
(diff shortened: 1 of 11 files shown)


lib/actions/build.js
```diff
@@ -152,8 +153,13 @@ Examples:
 		}
 		delete params.source;
 
-		options.convertEol = options['convert-eol'] || false;
-		delete options['convert-eol'];
+		if (isV12()) {
+			options.convertEol = options['convert-eol'] || false;
+			delete options['convert-eol'];
+		} else {
```
pdcastro, line 159: Should this `if-else` be the other way around? Swapping the `if` block with the `else` block.

lib/utils/eol-conversion.ts
```diff
@@ -110,7 +111,9 @@ export async function readFileWithEolConversion(
 		);
 		// And summary warning later
 		globalLogger.deferredLog(
-			'Windows-format line endings were detected in some files. Consider using the `--convert-eol` option.',
+			isV12()
+				? 'Windows-format line endings were detected in some files, but were not converted due to `--no-convert-eol` option.'
```
pdcastro, line 115: I didn't dig in the code to check, but could eol conversion have been skipped because the file is too large, instead of because `--no-convert-eol` was provided?

lib/actions/build.js
srlowe: Yep, that's done already but not yet pushed.

lib/utils/eol-conversion.ts
srlowe: Conversion may be skipped if too big, but this message wouldn't be displayed in that case.

lib/actions/deploy.js
```diff
@@ -263,16 +263,6 @@ Examples:
 		appName = appName_raw || appName || options.application;
 		delete options.application;
 
-		options.convertEol = options['convert-eol'] || false;
-		delete options['convert-eol'];
-		if (options.convertEol && !options.build) {
-			return Promise.reject(
-				new ExpectedError(
-					'The --eol-conversion flag is only valid with --build.',
```
pdcastro, line 271: I've removed this check / requirement because I noticed that we also need EOL conversion (or prevention with `--noconvert-eol`) even if `--build` is not specified, because an image will be built even if `--build` is not specified in case the respective image is not found locally. In this case, I also realised that the EOL logic for `balena build` and `balena deploy` become identical, so I moved the flag from `opts.convertEol` to `composeOpts.convertEol` so that the logic was centralised in `compose.js` instead of duplicated in `build.js` and `deploy.js`.

PR #1860, opened 2020-06-07
Release CLI v12

pdcastro:
Resolves: #1770
Change-type: major


pdcastro:
Closed in favor of #1867, with a different branch name (mainly so that I can keep track of the things I've tried). I have had to drop the `net-keepalive` commit because, it turns out, it requires Node.js >= v10.20.0, which is newer than the versions bundled in the GUI installers (Windows, macOS) and the standalone zip packages.

PR #1864, opened 2020-06-11
config: Allow setting the initialDeviceName when configuring an OS image

richbayliss:
See: https://github.com/balena-io/balena-supervisor/pull/1208

Change-type: patch


lib/actions-oclif/os/configure.ts
```diff
@@ -155,6 +156,10 @@ export default class OsConfigureCmd extends Command {
 			description:
 				'device type slug (e.g. "raspberrypi3") to override the application device type',
 		}),
+		'initial-device-name': flags.string({
+			description:
+				'The name the device should use before syncing with balenaCloud',
```
CameronDiver, line 161:
```suggestion
				'This option will set the device name when the device provisions with balenaCloud',
```

PR #1866, opened 2020-06-15
Add tests for standalone executable via proxy server

pdcastro:
This PR mainly extends our existing automated test cases for the standalone zip package (and also adds test cases for the `balena logs` command). This is actually a by-product of our ongoing effort to release CLI v12 with balena SDK v13 and the latest 'pkg' and 'net-keepalive' packages (that improve support for Node v14 as well). Manual testing with those updates revealed a few issues, mostly with the standalone zip package that wasn't automatically tested.

Change-type: patch


pdcastro:
Closed in favor of #1867 because it turns out that the proxy features required by the standalone tests require Node.js >= 10.16.0, which thus requires CLI v12 (which drops support for Node v8).

Issue #1870, opened 2020-06-17
Consider a .dockerignore file per service for docker-compose compatibility


pdcastro:
Since v8.0.0 (2018-10-19), the CLI has taken into account a single `.dockerignore` file in the project source folder: the folder specified by the `--source` option to the `balena deploy` and `balena push` commands, or as the positional argument to the `balena build` command. If the folder is not specified, it defaults to the current working directory. For microservices / multicontainer applications, the source folder is also the folder that contains the `docker-compose.yml` file.

In addition, in CLI v8, v9, v10 and v11, `.gitignore` files (in the source folder or in subfolders) were also taken into account. There was strong feedback against the CLI taking `.gitignore` files into account (#1032), and in CLI v11.34.0 a `--nogitignore` option was added (push/build/deploy commands). This option became the default behavior in CLI v12.0.0.

With the removal of support for `.gitignore` files, the treatment of `.dockerignore` got the spotlight, especially for microservices / multicontainer applications. We have learned that `/usr/local/bin/docker-compose` works differently: it respects a `.dockerignore` file per service, effectively by delegating service build to `/usr/local/bin/docker` (the `/usr/local/bin` prefix is not relevant, only using it to emphasize that I mean "the software tool").

Some microservices / multicontainer projects are structured so that each service comes from a different repository and are even developed / maintained by different engineering teams. In this case, having a `.dockerignore` file per service makes more sense than a single `.dockerignore` file in the source folder (alongside `docker-compose.yml`).

Suggestions:
* Add a `--multi-dockerignore` option to CLI v12 (push/build/deploy commands) so that each service can define its own `.dockerignore` file, in the service folder.
* Only one `.dockerignore` file per service. If multiple `.dockerignore` files exist, for example one alongside `docker-compose.yml` and another at the service folder, then only the `.dockerignore` file in the service folder would be considered. No "merging" of patterns. This aligns with:
  * How `/usr/local/bin/docker-compose` works
  * How `/usr/local/bin/docker` works
  * The differences between the `.gitignore` format and the `.dockerignore` format, summarised [in this README](https://github.com/balena-io-modules/dockerignore/blob/master/README.md#whats-different-from-ignore)

* A `.dockerignore` file in the source folder / project root (alongside `docker-compose.yml`) could still be used to filter files/folders outside of service folders.
* The `--multi-dockerignore` option could become the default behavior in CLI v13.

pdcastro:
[...]

PR #1875, opened 2020-06-22
Add runtime warning for unused .dockerignore files

pdcastro:
Print a warning message at runtime if `.dockerignore` files are found in project subdirectories, as these files will not be used. Motivated by forum thread: [Builds failing overnight w/ cli v12.1.1](https://forums.balena.io/t/builds-failing-overnight-w-cli-v12-1-1/137120/).

Connects-to: #1870 
Connects-to: balena-io/balena-on-balena/issues/200
Change-type: patch

### DIFF
--- a/lib/errors.ts
+++ b/lib/errors.ts
@@ -140,7 +140,7 @@ const EXPECTED_ERROR_REGEXES = [
 	/^BalenaDeviceNotFound/, // balena-sdk
 	/^BalenaExpiredToken/, // balena-sdk
 	/^Missing \w+$/, // Capitano,
-	/^Missing \d required argument/, // oclif parser: RequiredArgsError, RequiredFlagError
+	/^Missing \d+ required arg/, // oclif parser: RequiredArgsError, RequiredFlagError
 	/^Unexpected argument/, // oclif parser: UnexpectedArgsError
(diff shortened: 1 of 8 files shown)


lib/errors.ts
```diff
@@ -140,7 +140,7 @@ const EXPECTED_ERROR_REGEXES = [
 	/^BalenaDeviceNotFound/, // balena-sdk
 	/^BalenaExpiredToken/, // balena-sdk
 	/^Missing \w+$/, // Capitano,
-	/^Missing \d required argument/, // oclif parser: RequiredArgsError, RequiredFlagError
+	/^Missing \d+ required arg/, // oclif parser: RequiredArgsError, RequiredFlagError
```
pdcastro, line 143: This change is not necessary if the CLI installs correctly, but it is still useful in case the user runs `npm install balena-cli` as the root user without adding the `--unsafe-perm` option. In this case, the `postinstall` npm script is skipped, oclif is not patched and the error message reverts to oclif's original `Missing 1 required args:`. While testing on Linux as the root user, I ran `npm install && npm test` on the `balena-cli` dir and the result was failed tests (which is OK: broken install) and error reports sent to Sentry. This code change avoids such Sentry reporting.

PR #1881, opened 2020-06-23
Use native fs promises rather than promisify or mz

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


lib/actions/ssh.ts
```diff
@@ -77,7 +77,7 @@ async function getContainerId(
 		// We need to execute a balena ps command on the device,
 		// and parse the output, looking for a specific
 		// container
-		const { child_process } = await import('mz');
+		const childProcess = await import('child_process');
```
Page-, line 80: Strictly speaking this isn't `fs` but since the only method we use is `spawn` and it's not touched by mz( see https://github.com/normalize/mz/blob/2.7.0/child_process.js#L5-L6) I replaced this with the direct import so we could drop mz altogether

Page-:
@balena-ci retest

Page-:
@balena-ci retest

tests/commands/build.spec.ts
```diff
@@ -193,16 +193,19 @@ describe('balena build', function () {
...
 		const qemuBinPath = await qemuMod.getQemuPath(arch);
 		try {
 			mock(fsModPath, {
 				...fsMod,
-				exists: async (p: string) =>
-					p === qemuBinPath ? true : fsMod.exists(p),
+				promises: {
+					...fsMod.promises,
+					access: async (p: string) =>
+						p === qemuBinPath ? undefined : fsMod.promises.exists(p),
```
pdcastro, line 207: Shouldn't this line call `access` instead of `exists`? -
`p === qemuBinPath ? undefined : fsMod.promises.access(p),`

Page-: You're right, I guess we never hit the real version during tests but it should be fixed

Page-: fixed

Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

Page-:
@balena-ci retest

PR #1883, opened 2020-06-24
Convert command ssh to oclif.

srlowe:
Change-type: patch


lib/actions-oclif/ssh.ts
```diff
@@ -0,0 +1,390 @@
...
+		host OS or service container of the chosen device.
+
+		For local devices, the IP address and .local domain name are supported.
+		If the device is referenced by IP or \`.local\` address, the connection
+		is initiated directly to balenaOS on port \`22222\` via an
+		openssh-compatible client. Otherwise, any connection initiated remotely
+		traverses the balenaCloud VPN.
+
+		Commands may be piped to the shell, for example:
+			\`echo "uptime; exit;" | ./bin/balena ssh f49cefd\`
```
pdcastro, line 61: * `balena` instead of `./bin/balena` because the latter may look alien to many end users :-)
* I suggest removing the markdown backtick around the command line, as they appear verbatim on a text console and some users might actually copy and paste them...
* A dollar sign prefix would be consistent with the "examples" section.
* What about the following?

```
		Commands may be piped to the standard input for remote execution (see examples).
		Note however that remote command execution on service containers (as opposed to
		the host OS) is not currently possible when a device UUID is used (instead of
		an IP address) because of a balenaCloud backend limitation.
...
public static examples = [
		'$ balena ssh MyApp',
		'$ balena ssh f49cefd',
		'$ balena ssh f49cefd my-service',
		'$ balena ssh f49cefd --port <port>',
		'$ balena ssh 192.168.0.1 --verbose',
		'$ balena ssh f49cefd.local my-service',
		'$ echo "uptime; exit;" | balena ssh f49cefd',
		'$ echo "uptime; exit;" | balena ssh 192.168.0.1 myService',
	];
```

lib/actions-oclif/ssh.ts
```diff
@@ -0,0 +1,390 @@
...
+
+	public static args = [
+		{
+			name: 'applicationOrDevice',
+			description: 'application name, device uuid, or address of local device',
+			required: true,
+		},
+		{
+			name: 'serviceName',
+			description: 'service name, if connecting to container',
```
pdcastro, line 86: `"to a container"` would read a bit better to me

lib/actions-oclif/ssh.ts
```diff
@@ -0,0 +1,387 @@
...
+
+	public static usage = 'ssh <applicationOrDevice> [serviceName]';
+
+	public static flags: flags.Input<FlagsDef> = {
+		port: flags.integer({
+			description: stripIndent`
+				SSH server port number (default 22222) if the target is an IP address or .local
+				hostname. Otherwise, port number for the balenaCloud gateway (default 22).`,
+			char: 'p',
+			parse: (p) => parseAsInteger(p, 'port'),
```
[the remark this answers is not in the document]
srlowe: Right.  I would personally consider a port num with leading zeros to be invalid. I've never seen anyone express it that way before.  I guess it's a bit of a breaking change, but very low risk I would say.

lib/actions-oclif/ssh.ts
srlowe: ok sure

srlowe: It's acceptable to drop articles in English when space is at a premium (e.g. in newspaper headlines), but no need here.

lib/actions-oclif/ssh.ts
srlowe: oops re ./bin!

srlowe:
@balena-ci retest

srlowe:
@balena-ci retest

srlowe:
@balena-ci retest

PR #1884, opened 2020-06-24
Use cacheScope for fast-boot2 to allow caching when globally installed

Page-:
Saves ~250ms on subsequent calls to `balena help` when using the global npm install version, not sure about savings for other packaging methods


Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

PR #1888, opened 2020-06-26
Generate/include an oclif.manifest.json when packaging

Page-:
This saves ~220ms for me when running oclif commands

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "patches0/",
     "patches/",
     "*.md",
-    "npm-shrinkwrap.json"
+    "npm-shrinkwrap.json",
+    "oclif.manifest.json"
   ],
   "bin": {
     "balena": "./bin/balena"
@@ -40,7 +41,8 @@
       "node_modules/opn/xdg-open",
       "node_modules/open/xdg-open",
       "node_modules/windosu/*.bat",
-      "node_modules/windosu/*.cmd"
+      "node_modules/windosu/*.cmd",
+      "oclif.manifest.json"
     ]
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
     "watch": "gulp watch",
     "lint": "balena-lint -e ts -e js --typescript --fix automation/ lib/ typings/ tests/ gulpfile.js",
     "update": "ts-node --transpile-only ./automation/update-module.ts",
-    "prepublishOnly": "npm run build"
+    "prepublishOnly": "npm run build && npx oclif-dev manifest"
```
pdcastro, line 70: Nice! 👍   If the manifest is generated in the `prepublishonly` script, I think it would only be available for the `npm install` method, not the standalone zip package and the native installers (Windows and macOS). What if `npx oclif-dev manifest` was added to the `build` npm script? And then the `oclif.manifest.json` file would also need to be added to:
* This list of files used when building the native installers: https://github.com/balena-io/balena-cli/blob/v12.1.16/patches/@oclif+dev-cli+1.22.0.dev.patch#L89-L103
* The `pkg.assets` section of `package.json` so that the file is included in the standalone zip packages.

patches/@oclif+dev-cli+1.22.0.dev.patch
```diff
@@ -100,6 +100,7 @@ index 3e613e0..8c8f698 100644
 +            'patches',
 +            'patches0',
 +            'typings',
++            'oclif.manifest.json',
```
pdcastro, line 103: CI error:
```
**ERROR** Failed to apply patch for package @oclif/dev-cli
  This happened because the patch file patches/@oclif+dev-cli+1.22.0.dev.patch could not be parsed.
```
It rings a bell! `patch-package` requires patch files to be recreated. The way to do this would be:
* Revert the patch file to the `master` branch state:  
`git checkout master -- patches/\@oclif+dev-cli+1.22.0.dev.patch`
* `rm -rf node_modules/@oclif/dev-cli`
* `npm install`  (This should automatically run `patch-package` through the `postinstall` script, and patch `@oclif/dev-cli` according to the master branch patch file.)
* Edit file `node_modules/@oclif/dev-cli/lib/tarballs/build.js` file directly to add the `'oclif.manifest.json'` entry.
* Run `npx patch-package \@oclif/dev-cli` to recreate the patch file.
* `git commit` the new patch file.

PR #1894, opened 2020-06-29
Use @balena/es-version to set the desired es version for modules

Page-:
This allows us to use es2018 versions of downstream modules (or the closest version under) when they're available, which means we'll be using native object spread, native async/await, etc instead of transpiled versions


Page-:
@balena-ci rebase

Issue #1896, opened 2020-06-30
Balena ssh not working since 12.1.15


rickardgranberg:
# Expected Behavior
```
balena ssh <ip-address>
```
should ssh into the device

# Actual Behavior
```
balena ssh <ip-address>
```
> Part of the CLI could not be loaded. This typically means your CLI install is in a broken state.
> You can normally fix this by uninstalling and reinstalling the CLI.
> 
> Additional information may be available with the `--debug` flag.
> For help, visit our support forums: https://forums.balena.io
> For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/

Running with `--debug` gives the following:

> [debug] new argv=[/balena/balena-cli/balena,/snapshot/versioned-source/bin/balena,ssh,<ip-address>] length=4
> Part of the CLI could not be loaded. This typically means your CLI install is in a broken state.
> You can normally fix this by uninstalling and reinstalling the CLI.
> 
> Error: Cannot find module 'lodash/reduce'
> 1) If you want to compile the package/file into executable, please pay attention to compilation warnings and specify a literal in 'require' call. 2) If you don't want to compile the package/file into executable and want to 'require' it from filesystem (likely plugin), specify an absolute path in 'require' call using process.cwd() or process.execPath.
>     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:634:15)
>     at Function.Module._resolveFilename (pkg/prelude/bootstrap.js:1351:46)
>     at Function.resolveFilenameOptimized [as _resolveFilename] (/snapshot/versioned-source/node_modules/fast-boot2/index.js:57:33)
>     at Function.Module._load (internal/modules/cjs/loader.js:560:25)
>     at Module.require (internal/modules/cjs/loader.js:690:17)
>     at Module.require (pkg/prelude/bootstrap.js:1230:31)
>     at require (internal/modules/cjs/helpers.js:25:18)
>     at Promise.resolve.then (/snapshot/versioned-source/build/utils/device/ssh.js:7:55)
> 
> Additional information may be available with the `--debug` flag.
> For help, visit our support forums: https://forums.balena.io
> For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/
> 

# Steps to Reproduce the Problem

1. Install Balena CLI 12.1.15 or newer (I've tested with 12.1.16 and todays 12.3.1, both have the same problem)
1. Run balena ssh <ip-address>

Reverting to 12.1.14 and ssh works again.

[...]

PR #1897, opened 2020-06-30
ssh: fix "Cannot find module 'lodash/reduce'" standalone zip package

pdcastro:
Resolves: #1896
Change-type: patch


pdcastro:
Closing in favour of #1898 and #1909

PR #1900, opened 2020-06-30
Update dependencies still using balena-sdk 10.x

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


lib/actions/config.js
```diff
@@ -18,6 +18,13 @@ import * as commandOptions from './command-options';
 
 import { normalizeUuidProp } from '../utils/normalization';
 import { getBalenaSdk, getVisuals } from '../utils/lazy';
+import * as _ from 'lodash';
```
Page-, line 21: lodash is used for top level command creation so it will always be easily loaded, and it's generally used in so many places that there won't be a benefit to lazy loading it until a lot more of its usage is removed

lib/actions/config.js
```diff
@@ -152,24 +151,22 @@ Examples:
...
-			)
-			.tap(() => {
-				console.info('Done');
-			});
+		const drive =
+			options.drive || getVisuals().drive('Select the device drive');
+		await umountAsync(drive);
+
+		const fs = await import('fs');
+		const configJSON = await JSON.parse(
```
pdcastro, line 162: That's one `await` too many, though I think it's harmless...

lib/actions/config.js
```diff
@@ -152,24 +151,22 @@ Examples:
...
-				require('fs')
-					.promises.readFile(params.file, 'utf8')
-					.then(JSON.parse)
-					.then((configJSON) => config.write(drive, options.type, configJSON)),
-			)
-			.tap(() => {
-				console.info('Done');
-			});
+		const drive =
+			options.drive || getVisuals().drive('Select the device drive');
```
pdcastro, line 158: Is it missing `await` here, before `getVisuals().drive(...)`? I image it would return a promise...

lib/actions/config.js
Page-: Right, it was previously `.then`'d and I just did the equivalent even though it's not necessarily, I'll fix it

PR #1901, opened 2020-06-30
Add typing info to requires wherever possible

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


package.json
```diff
@@ -106,7 +106,7 @@
     "@oclif/config": "^1.14.0",
     "@oclif/dev-cli": "1.22.0",
     "@oclif/parser": "^3.8.4",
-    "@octokit/plugin-throttling": "^2.7.1",
+    "@octokit/plugin-throttling": "^3.3.0",
```
Page-, line 109: Required to get types for the module, changelog: https://github.com/octokit/plugin-throttling.js/releases/tag/v3.0.0
Basically drops node 8 and a slight change to the export which I updated in the one place we use it (typescript was able to catch the need to update it as well :) )

automation/deploy-bin.ts
```diff
@@ -109,8 +110,11 @@ function getPageNumbers(
 	if (!response.headers.link) {
 		return res;
 	}
-	const parse = require('parse-link-header');
+	const parse = require('parse-link-header') as typeof import('parse-link-header');
 	const parsed = parse(response.headers.link);
+	if (parsed == null) {
+		throw new Error(`Failed to parse link header: '${response.headers.link}'`);
+	}
```
Page-, line 117: I had to add this to satisfy the typings, it just means if we do hit this branch we get a much better error message :)

PR #1904, opened 2020-06-30
Enable long stack traces for `--debug` as well as `DEBUG=1`

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/app-common.ts
+++ b/lib/app-common.ts
@@ -218,28 +218,6 @@ function setupBalenaSdkSharedOptions(settings: CliSettings) {
 	});
 }
 
-let BluebirdConfigured = false;
-
-/**
- * Configure Bluebird and assign it as the global promise library.
- * Modules like `stream-to-promise` will otherwise produce native promises,
- * which leads to errors as much of the CLI JavaScript code expects Bluebird
- * promises.
- */
-export function configureBluebird() {
-	if (BluebirdConfigured) {
-		return;
-	}
-	BluebirdConfigured = true;
-	const Bluebird = require('bluebird') as typeof import('bluebird');
-	Bluebird.config({
-		longStackTraces: process.env.DEBUG ? true : false,
-	});
-	if (!(global as any)['@@any-promise/REGISTRATION']) {
-		require('any-promise/register/bluebird');
-	}
-}
-
 /**
  * Addresses the console warning:
  * (node:49500) MaxListenersExceededWarning: Possible EventEmitter memory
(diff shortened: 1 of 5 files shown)


lib/app-common.ts
```diff
@@ -218,28 +218,6 @@ function setupBalenaSdkSharedOptions(settings: CliSettings) {
...
-		return;
-	}
-	BluebirdConfigured = true;
-	const Bluebird = require('bluebird') as typeof import('bluebird');
-	Bluebird.config({
-		longStackTraces: process.env.DEBUG ? true : false,
-	});
-	if (!(global as any)['@@any-promise/REGISTRATION']) {
-		require('any-promise/register/bluebird');
-	}
```
Page-, line 240: We haven't needed this for a while since `stream-to-promise` is the only `any-promise` user we use and the only place we use that it's already manually wrapped

lib/app-common.ts
```diff
@@ -218,28 +218,6 @@ function setupBalenaSdkSharedOptions(settings: CliSettings) {
...
- * Modules like `stream-to-promise` will otherwise produce native promises,
- * which leads to errors as much of the CLI JavaScript code expects Bluebird
- * promises.
- */
-export function configureBluebird() {
-	if (BluebirdConfigured) {
-		return;
-	}
-	BluebirdConfigured = true;
-	const Bluebird = require('bluebird') as typeof import('bluebird');
```
Page-, line 234: Removing this import means we open up the possibility lazy-loading bluebird and the associated startup time savings

lib/app-common.ts
pdcastro: hmm I seem to remember that this was needed for code in dependencies, rather than code in the CLI itself. Maybe we could do some manual testing before merging this PR.

Page-: The full list of usage is:
```
+-- livepush@3.4.0
| `-- mz@2.7.0
|   +-- any-promise@1.3.0  deduped
|   `-- thenify-all@1.6.0
|     `-- thenify@3.3.0
|       `-- any-promise@1.3.0  deduped
`-- stream-to-promise@2.2.0
  +-- any-promise@1.3.0  deduped
  `-- stream-to-array@2.3.0
    `-- any-promise@1.3.0  deduped
```
so the only other thing that uses it is `mz` which is used by livepush, but livepush assumes mz will return native promises anyway because that's what the typings say

Page-:
@balena-ci rebase

PR #1906, opened 2020-07-01
Convert `logs` command to oclif

srlowe:
Change-type: patch


doc/cli.markdown
```diff
@@ -1302,9 +1302,9 @@ answer "yes" to all questions (non interactive use)
 
 # Logs
 
-## logs &#60;uuidOrDevice&#62;
+## logs &#60;device&#62;
```
srlowe, line 1305: I changed the argument name from `uuidOrDevice` to `device` (which I think makes more sense, since it can be one of : device uuid, device IP, device hostname).

PR #1907, opened 2020-07-01
Remove all module level bluebird usage from eagerly loaded modules

Page-:
When combined with https://github.com/balena-io-modules/balena-settings-storage/pull/27/files and #1905 this removes bluebird from the `balena help` path entirely and saves ~50ms, and also means any other actions which do not use bluebird will also get those benefits

### DIFF
--- a/lib/events.ts
+++ b/lib/events.ts
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
 import * as Mixpanel from 'mixpanel';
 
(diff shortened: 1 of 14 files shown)


lib/utils/qemu.ts
```diff
@@ -37,9 +36,14 @@ export function copyQemu(context: string, arch: string) {
...
-		.catch({ code: 'EEXIST' }, function () {
-			// noop
+	return fs.promises
+		.mkdir(binDir)
+		.catch(function (err) {
+			if (err.code === 'EEXIST') {
+				// ignore
+				return;
+			}
+			throw err;
```
pdcastro, line 46: Personal preference:
```
		.catch((err) => {
			if (err.code !== 'EEXIST') {
				throw err;
			}
		})
```

Page-: The reason for this style was 1 to be equivalent to the previous style, and 2 because it works better if you have multiple checks, eg
```
if err.code = A then do X
if err.code = B then do Y
else throw err
```
I'm happen to change it if you still prefer it (in the single check case it doesn't matter either way) but that was the reasoning with going for this style

lib/events.ts
```diff
@@ -78,34 +77,34 @@ export async function trackCommand(commandSignature: string) {
...
+
+		const mixpanel = getMixpanel(balenaUrl);
+
+		Sentry.configureScope((scope) => {
+			scope.setUser({
+				id: username,
+				username,
 			});
-		})
-		.timeout(100)
```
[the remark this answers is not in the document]
Page-: I didn't notice any change by removing this and the mixpanel checking code seems to take > 100ms for me already. I've got a feeling it didn't actually have any effect in reality because if the request is still in progress it'll keep nodejs running until completed even if we've timed out this promise

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

PR #1909, opened 2020-07-01
balena ssh: Refactor error handling and test cases

pdcastro:
Connects-to: #1896 
Change-type: patch


pdcastro:
@balena-ci retest

PR #1911, opened 2020-07-02
Fix cross-framework command calls and convert `device init` to oclif

srlowe:


lib/actions-oclif/device/init.ts
```diff
@@ -0,0 +1,172 @@
...
+		options.application = options.application || options.app;
+		delete options.app;
+
+		// Get application and
+		const application = await balena.models.application.get(
+			options['application'] || (await patterns.selectApplication()),
+		);
+
+		// Register new device
+		const deviceUuid = balena.models.device.generateUniqueKey();
```
srlowe, line 110: I changed the process order here a little bit to improve failed case handling.

lib/actions-oclif/device/init.ts
```diff
@@ -0,0 +1,172 @@
...
+		// Consolidate application options
+		options.application = options.application || options.app;
+		delete options.app;
+
+		// Get application and
+		const application = await balena.models.application.get(
+			options['application'] || (await patterns.selectApplication()),
+		);
+
+		// Register new device
```
srlowe, line 109: Registering a device is now done without a command-command call.

lib/actions-oclif/device/init.ts
```diff
@@ -0,0 +1,172 @@
...
+		const osVersion = options['os-version'] || 'default';
+		await runCommand(
+			`os download ${deviceType} --output '${path}' --version ${osVersion}`,
+		);
+	}
+
+	async configureOsImage(path: string, uuid: string, options: FlagsDef) {
+		let configureCommand = `os configure ${path} --device ${uuid}`;
+		if (options.config) {
+			configureCommand += ` --config ${options.config}`;
```
srlowe, line 155: It seems that the single quotes needed to be removed around the paths here (I think it must be due to oclif and capitano parsing differences).

lib/actions-oclif/device/init.ts
```diff
@@ -0,0 +1,172 @@
...
+		help: cf.help,
+	};
+
+	public static authenticated = true;
+
+	public async run() {
+		const { flags: options } = this.parse<FlagsDef, {}>(DeviceInitCmd);
+
+		// Imports
+		const { promisify } = await import('bluebird');
```
pdcastro, line 89: Maybe `import { promisify } from 'util';` like Page did for example here:
https://github.com/balena-io/balena-cli/pull/1907/files#diff-f49db67e2f5f28913d23492aec7161d2R20

lib/utils/helpers.ts
```diff
@@ -96,9 +97,26 @@ export async function sudo(
 }
 
 export function runCommand<T>(command: string): Promise<T> {
-	const capitano = require('capitano') as typeof import('capitano');
-	const capitanoRunAsync = promisify(capitano.run);
-	return capitanoRunAsync(command);
+	let commandArgs = command.split(' ');
```
pdcastro, line 100: Splitting command lines with spaces is problematic. For example, above you mentioned _"single quotes needed to be removed around the paths here."_ In a command line like `os configure ${path} --device ${uuid}`, the `path` could contain blank spaces, and `command.split(' ')` would split a single path as multiple separate arguments even if the path is quoted (and different shells like bash / cmd.exe / powershell have different quoting rules).

Anyway, my suggestion is to change the `runCommand` function interface to take an array of strings instead of a single string, thus avoiding the need for `command.split()` here:
```
export function runCommand<T>(commandArgs: string[]): Promise<T> {
```
Then in the caller, instead of:
```
	async configureOsImage(path: string, uuid: string, options: FlagsDef) {
		let configureCommand = `os configure ${path} --device ${uuid}`;
		...
		await runCommand(configureCommand);
```
We'd have:
```
	async configureOsImage(path: string, uuid: string, options: FlagsDef) {
		let configureCommand = ['os', 'configure', path, '--device', uuid];
		...
		await runCommand(configureCommand);
```
Unless I'm missing something, I feel that using arrays all along is preferable to using `string.split(' ')`.

By the way, even better would be to avoid `runCommand` altogether -- to avoid "command-command calls" like you described it. Instead of `balena device init` calling `balena os configure`, it should just call a Javascript function that does what `balena os configure` does without having to parse another command line and go multiple times through validation checks, whoami / logged in, mixpanel (?) and all. But this would be a more involved refactoring; perhaps one step at a time.

lib/actions-oclif/device/init.ts
pdcastro: See my other comment in the `runCommand` function: I suggest using arrays all along:
```
let configureCommand = ['os', 'configure', path, '--device', uuid];
if (options.config) {
    configureCommand.push('--config', options.config);
```

lib/utils/helpers.ts
srlowe: Yeah I think the best is to get rid of this entirely, and call isolated logic.   (BTW, mixpanel and probably some of the others you mentioned are not invoked my this).

lib/actions-oclif/device/init.ts
srlowe: Just tried this, but it doesn't seem to work as a drop in replacement - I'm getting typescript compilation errors regarding tmpNameAsync.

PR #1912, opened 2020-07-02
Add explicit dependency on balena-settings-storage

Page-:
This also removes the last bluebird usage from `balena help` saving ~50ms for that specific command, a lot of other commands will still import bluebird but they can be handled on a command by command basis now that the main path is clean :)


Page-:
@balena-ci rebase

Page-:
@balena-ci retest

PR #1913, opened 2020-07-03
Modify runCommand to accept args array instead of string

srlowe:
Change-type: patch

### DIFF
--- a/lib/actions-oclif/device/init.ts
+++ b/lib/actions-oclif/device/init.ts
@@ -144,28 +144,34 @@ export default class DeviceInitCmd extends Command {
 
 	async downloadOsImage(path: string, deviceType: string, options: FlagsDef) {
 		const osVersion = options['os-version'] || 'default';
-		await runCommand(
-			`os download ${deviceType} --output '${path}' --version ${osVersion}`,
-		);
+		await runCommand([
+			'os',
+			'download',
+			deviceType,
+			'--output',
+			path,
+			'--version',
+			osVersion,
+		]);
 	}
 
 	async configureOsImage(path: string, uuid: string, options: FlagsDef) {
-		let configureCommand = `os configure ${path} --device ${uuid}`;
+		const configureCommand = ['os', 'configure', path, '--device', uuid];
 		if (options.config) {
-			configureCommand += ` --config ${options.config}`;
+			configureCommand.push('--config', options.config);
 		} else if (options.advanced) {
-			configureCommand += ' --advanced';
+			configureCommand.push('--advanced');
 		}
 		await runCommand(configureCommand);
 	}
 
 	async writeOsImage(path: string, deviceType: string, options: FlagsDef) {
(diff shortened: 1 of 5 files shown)


lib/utils/helpers.ts
```diff
@@ -115,7 +114,7 @@ export function runCommand<T>(command: string): Promise<T> {
 		require('../app-capitano');
 
 		const capitanoRunAsync = promisify(capitano.run);
-		return capitanoRunAsync(command);
+		return capitanoRunAsync(commandArgs.join(' '));
```
pdcastro, line 117: I was assuming we would be able to avoid `join(' ')` as well. I think we can:
* https://github.com/balena-io/capitano/blob/v1.9.2/lib/capitano.coffee#L59-L60
* https://github.com/balena-io/capitano/blob/v1.9.2/lib/parse.coffee#L11

Hopefully it is sufficient to edit the following typings file in the `balena-cli` repo:
* https://github.com/balena-io/balena-cli/blob/v12.3.12/typings/capitano/index.d.ts#L85

`command: string` → `command: string | string[]`

Then:
```
		return capitanoRunAsync(commandArgs);
```

lib/actions-oclif/device/init.ts
```diff
@@ -144,28 +144,40 @@ export default class DeviceInitCmd extends Command {
...
 		const osVersion = options['os-version'] || 'default';
-		await runCommand(
-			`os download ${deviceType} --output '${path}' --version ${osVersion}`,
-		);
+		await runCommand([
+			'os',
+			'download',
+			deviceType,
+			'--output',
+			`'${path}'`,
```
pdcastro, line 152: Does it not work with simply `path` instead of `'${path}'`? Perhaps after the change that I am suggesting in the other comment, so `capitanoRunAsync` takes an array of strings.

lib/utils/helpers.ts
srlowe: Ah nice, will try that.

lib/actions-oclif/device/init.ts
srlowe: I'm not sure, I was trying to ensure that nothing changed, but will test (would be good if we could avoid oclif/capitano differences).

lib/utils/helpers.ts
srlowe: Yep, appears to work.

lib/actions-oclif/device/init.ts
srlowe: Works!

srlowe:
@balena-ci retest

srlowe:
@balena-ci

pdcastro:
> @balena-ci

Was that meant to be `@balena-ci retest` ?

srlowe:
@pdcastro Ah yes well spotted! (Something else running now though, so I'll wait)

Issue #1916, opened 2020-07-03
tap is not a function / catchReturn is not a function


pdcastro:
```
$ balena version          
12.3.13
```
```
$ DEBUG=1 ./bin/balena login
...
balena.auth.getToken(...).catchReturn is not a function
Error: balena.auth.getToken(...).catchReturn is not a function
```
```
❯ DEBUG=1 balena ssh e655ba7
...
pine.get(...).tap is not a function
TypeError: pine.get(...).tap is not a function
```
This is happening because the `.fast-boot.json` is not being correctly updated in `npm `installations.

**Workaround**
Delete the `.fast-boot.json` file located in the CLI's installation folder. The npm installation varies between systems (nvm, local npm, global npm...). Examples:
```
bin/.fast-boot.json
~/.nvm/versions/node/v12.18.0/lib/node_modules/balena-cli/bin/.fast-boot.json
```

PR #1917, opened 2020-07-03
Publish '.fast-boot.json' to the npm registry to avoid stale caching

pdcastro:
Resolves: #1916 
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   },
   "preferGlobal": true,
   "files": [
+    "bin/.fast-boot.json",
     "bin/",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   },
   "preferGlobal": true,
   "files": [
+    "bin/.fast-boot.json",
```
Page-, line 13: Does this not get covered by the `bin/` below?

pdcastro: It may indeed be redundant!

PR #1920, opened 2020-07-06
Clarify -c flag in balena push

vipulgupta2048:
Add more information to the --no-cache or -c option for balena push CLI command. 

Change-type: patch

See: [Flowdock](https://www.flowdock.com/app/rulemotion/i-cli/threads/XrTz7wOdxS74q8G_iKU8m16Xgzp)
@pdcastro :eyes:

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -2074,7 +2074,11 @@ Alternative Dockerfile name/path, relative to the source folder
 
 #### --nocache, -c
 
-Don't use cache when building this project
+Don't use cached layers of previously built images for this project. This ensures
+that the latest base image and packages are pulled. Note that build logs may still
+display the message _"Pulling previous images for caching purposes" (as the cloud
+builder needs previous images to compute delta updates), but the logs will not
+display the "Using cache" lines for each build step of a Dockerfile.
 
 #### --noparent-check
 
--- a/lib/actions/push.ts
+++ b/lib/actions/push.ts
@@ -192,7 +192,12 @@ export const push: CommandDefinition<
 		{
 			signature: 'nocache',
 			alias: 'c',
-			description: "Don't use cache when building this project",
+			description: stripIndent`
+                Don't use cached layers of previously built images for this project. This ensures
+                that the latest base image and packages are pulled. Note that build logs may still
+                display the message _"Pulling previous images for caching purposes" (as the cloud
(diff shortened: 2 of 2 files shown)


doc/cli.markdown
```diff
@@ -2066,7 +2066,9 @@ Alternative Dockerfile name/path, relative to the source folder
 
 #### --nocache, -c
 
-Don't use cache when building this project
+Instructs balenaEngine (Docker) to initiate a fresh build and not use any cache from previous builds.
+
+Using this option would still lead to the message "pulling previous images for caching purposes" in the build logs. Indicating that balenaCloud builder although pulls previous images for balenaEngine's reuse but won't be using them during a new image build. This will be indicated in the build logs by the absence of "Using cache" lines.
```
pdcastro, line 2071: * The `cli.markdown` file is automatically generated, see note on CONTRIBUTING file: https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#editing-documentation-files-changelog-readme-website
These changes need to go to the `lib/actions/push.ts` file, then run `npm run build`. See this sample diff: https://github.com/balena-io/balena-cli/compare/clarify-flag-nocache-2?expand=1
* I suggest a rewording for further clarity:
```
{
	signature: 'nocache',
	alias: 'c',
	description: stripIndent`
		Don't use cached layers of previously built images for this project. This ensures
		that the latest base image and packages are pulled. Note that build logs may still
		display the message _"Pulling previous images for caching purposes" (as the cloud
		builder needs previous images to compute delta updates), but the logs will not
		display the "Using cache" lines for each build step of a Dockerfile.`,
	boolean: true,
},
```

* It may be a grey area, but I would normally use `Change-type: patch` for a PR like this, as it does not add any new functionally from a user's point of view. (Clarifying a flag description is closer to being a bug fix than a new feature, the way I see it.)

vipulgupta2048: Thanks for the sample diff, should have read the contributing document first. How about the current change, looks good?

vipulgupta2048:
@pdcastro Made all the changes as suggested

doc/cli.markdown
pdcastro: Yes, looks good, thank you! 🙏 😀

pdcastro:
@balena-ci rebase

Issue #1844, opened 2020-05-26
balena build fails to detect docker version when using docker-machine in OS X


pipex:
# Expected Behavior

CLI launches build process

# Actual Behavior

The CLI fails to launch docker and answers `Docker seems to be unavailable`, even though the binary is available and the `DOCKER_HOST` environment variable is set. 

```
[balena-cli-masterclass] balena build --application cliApp --emulated                                          master
Docker seems to be unavailable. Is it installed and running?

Additional information may be available with the `--debug` flag.
For help, visit our support forums: https://forums.balena.io
For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/

[balena-cli-masterclass] balena build --debug --arch aarch64 --deviceType raspberrypi4-64 --emulated
[Debug]   Parsing input...
Docker seems to be unavailable. Is it installed and running?

Additional information may be available with the `--debug` flag.
For help, visit our support forums: https://forums.balena.io
For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/

[balena-cli-masterclass] docker run hello-world                                                                Hello from Docker!
This message shows that your installation appears to be working correctly.
...
```

The problem seems to be that the CLI is not considering the `DOCKER_CERT_PATH` environment variable set by docker-machine to indicate the location of the credentials. 

```
[balena-cli-masterclass] balena build --ca $DOCKER_CERT_PATH/ca.pem --cert $DOCKER_CERT_PATH/cert.pem --key $DOCKER_CERT_PATH/key.pem --application cliApp --emulated
[Info]    No "docker-compose.yml" file found at "/Users/flalanne/Development/balena-cli-masterclass"
[Info]    Creating default composition with source: "/Users/flalanne/Development/balena-cli-masterclass"
[Info]    Building for aarch64/raspberrypi4-64
[Info]    Emulation is enabled
```

# Steps to Reproduce the Problem

This is the most important and helpful part of a bug report. If we cannot reproduce the
problem, it is difficult to tell what the fix should be, or whether code changes have
fixed it.

1. `eval $(docker-machine env default)
2. `git clone https://github.com/balena-io/balena-cli-masterclass.git && cd balena-cli-masterclass`
3. `balena build --debug --arch aarch64 --deviceType raspberrypi4-64 --emulated`

# Specifications

- **balena CLI version:** 11.34.0
- **Operating system version:** macOS 10.15.4
- **Install method:** npm
[...]

PR #1921, opened 2020-07-06
build/deploy/preload: Accept DOCKER_CERT_PATH and DOCKER_TLS_VERIFY env vars

pdcastro:
Resolves: #1844
Change-type: minor

### DIFF
--- a/lib/utils/docker-js.js
+++ b/lib/utils/docker-js.js
@@ -112,28 +112,44 @@ const generateConnectOpts = function (opts) {
 
 	return Bluebird.try(function () {
 		const connectOpts = {};
-		// Firsly need to decide between a local docker socket
-		// and a host available over a host:port combo
+
+		// Start with docker-modem defaults which take several env vars into account,
+		// including DOCKER_HOST, DOCKER_TLS_VERIFY, DOCKER_CERT_PATH, SSH_AUTH_SOCK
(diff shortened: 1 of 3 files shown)


lib/utils/docker-js.js
```diff
@@ -112,28 +112,44 @@ const generateConnectOpts = function (opts) {
...
 	return Bluebird.try(function () {
 		const connectOpts = {};
-		// Firsly need to decide between a local docker socket
-		// and a host available over a host:port combo
+
+		// Start with docker-modem defaults which take several env vars into account,
+		// including DOCKER_HOST, DOCKER_TLS_VERIFY, DOCKER_CERT_PATH, SSH_AUTH_SOCK
+		// https://github.com/apocas/docker-modem/blob/v2.1.3/lib/modem.js#L15-L65
+		const Modem = require('docker-modem');
+		const defaultOpts = new Modem();
```
pdcastro, line 120: Previously, this code would only run if neither `--docker` nor `--dockerHost` were provided in the command line, in which case `docker-modem` was used to process the DOCKER_HOST env var (https://github.com/apocas/docker-modem/blob/v2.1.3/lib/modem.js#L15-L65)
In order to take advantage of `docker-modem`'s parsing of additional env vars like DOCKER_CERT_PATH, I've moved the code here so it always runs, and then any options explicitly provided in the command line simply override any values captured from env vars.

PR #1928, opened 2020-07-08
Be lazier with imports in oclif actions

Page-:
This avoids quite a few cases where we were loading modules in a command even though they were only used in a single branch (and in some cases those branches are very unlikely to be hit)


Page-:
@balena-ci retest

PR #1929, opened 2020-07-09
Convert `util available-drives` to oclif

srlowe:
Change-type: patch


lib/actions-oclif/util/available-drives.ts
```diff
@@ -0,0 +1,75 @@
...
+interface FlagsDef {
+	help: void;
+}
+
+export default class UtilAvailableDrivesCmd extends Command {
+	public static description = stripIndent`
+		List available drives.
+
+		List available drives which are usable for writing an OS image to.
+		Does not list system drives.
```
pdcastro, line 32:
```
System drives are not listed.
```
My problem 😊   was with mixing "List" with "Does not list" (as opposed to "List + Do not list" or "Lists + Does not list").

srlowe: I think either are correct, but I thought this was a little clearer :)

PR #1931, opened 2020-07-09
Test cases: Re-add nock checks for unexpected HTTP requests

pdcastro:
This PR addresses the balena CI test failures that we observed the other day ([flowdock thread](https://www.flowdock.com/app/rulemotion/i-cli/threads/30l3sW2MJn9WYiYW2At5s6MWHO5)) where mixpanel requests would cause tests to fail because those tests were not correctly "declaring" the requests as expected. Sample failure:
```
1) balena devices supported
       should list currently supported devices, with correct filtering:

      AssertionError: expected [ 'NockMock: Unexpected HTTP request: GET https//api.balena-cloud.com:443/mixpanel/track?ip=0&verbose=0&data=eyJldmVudCI6IltDTEldIGRldmljZXMgc3VwcG9ydGVkIiwicHJvcGVydGllcyI6eyJkaXN0aW5jdF9pZCI6ImdoX3VzZXIiLCJ2ZXJzaW9uIjoiMTIuNC4wIiwibm9kZSI6InYxMC4yMS4wIiwiYXJjaCI6Ing2NCIsImJhbGVuYVVybCI6ImJhbGVuYS1jbG91ZC5jb20iLCJwbGF0Zm9ybSI6IndpbjMyIiwidG9rZW4iOiJiYWxlbmEtbWFpbiIsIm1wX2xpYiI6Im5vZGUifX0%3D\n' ] to deeply equal []
      + expected - actual

      -[
      -  "NockMock: Unexpected HTTP request: GET https//api.balena-cloud.com:443/mixpanel/track?ip=0&verbose=0&data=eyJldmVudCI6IltDTEldIGRldmljZXMgc3VwcG9ydGVkIiwicHJvcGVydGllcyI6eyJkaXN0aW5jdF9pZCI6ImdoX3VzZXIiLCJ2ZXJzaW9uIjoiMTIuNC4wIiwibm9kZSI6InYxMC4yMS4wIiwiYXJjaCI6Ing2NCIsImJhbGVuYVVybCI6ImJhbGVuYS1jbG91ZC5jb20iLCJwbGF0Zm9ybSI6IndpbjMyIiwidG9rZW4iOiJiYWxlbmEtbWFpbiIsIm1wX2xpYiI6Im5vZGUifX0%3D\n"
      -]
      +[]
      
      at Context.it (C:\Users\Administrator\concourse\work\volumes\live\835b0111-0972-416c-5f29-321269d74ba3\volume\tests\commands\device\supported.spec.ts:33:31)
```

Connects-to: #1910 
Change-type: patch


pdcastro:
@balena-ci retest

lib/events.ts
```diff
@@ -96,14 +96,17 @@ export async function trackCommand(commandSignature: string) {
...
-		await mixpanel.track(`[CLI] ${commandSignature}`, {
-			distinct_id: username,
-			version: packageJSON.version,
-			node: process.version,
-			arch: process.arch,
-			balenaUrl, // e.g. 'balena-cloud.com' or 'balena-staging.com'
-			platform: process.platform,
-		});
+		// Don't actually call mixpanel.track() while running test cases
+		if (!process.env.BALENA_CLI_TEST_TYPE) {
```
pdcastro, line 100: In the other commit, I added `api.expectGetMixpanel({ optional: true });` to all test cases that execute CLI commands. Alas, this still wasn't enough to avoid the CI errors mentioned in the PR description. Mixpanel requests are special in that the CLI does not `await` for a reply, so my current theory is that a request initiated in a previous test case (that finishes quickly) may only be "seen" by nock during the execution of a subsequent test, such that the previous tests sees no mixpanel request (and succeeds because of `optional: true`), and the next test case sees two mixpanel requests (and fails because of `persist: false`). One way of getting tests to pass might be to write `api.expectGetMixpanel({ optional: true, persist: true });`, but this really means _"I don't care whether or not this request is made, or how many times",_ in which case, I thought, we might as well disable the requests during test execution. Hence coming up with the env var conditional. `BALENA_CLI_TEST_TYPE` is set to either `source` or `standalone` in the npm scripts in `package.json`.

Issue #1878, opened 2020-06-23
Get 65200 illegal hardware instruction when running balena cli commands and the device is in local mode


aifreedom:
# Expected Behavior

When the device is in local mode, I expect I can push image to it with `balena push <device_id>.local`.

# Actual Behavior
See the error in this gist https://gist.github.com/aifreedom/ed715dcf6643ce54fa6c6048eb9d9bc3

The same command can work if the device is not in local mode. 

# Steps to Reproduce the Problem

1. Put the device (mine is a Raspberry 4) in local mode
1. Run `balena push <device_id>.local` from an image I'm working on
1. See the error above. The error also shows when I run other balena commands, like `balena logs <device_id>.local --system`

# Specifications

- **balena CLI version:** balena-cli version "12.1.3"
- **Operating system version:** macOS 10.15.5
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** npm
- **If npm install, Node.js and npm version:** , Node.js version "14.4.0", npm version "6.14.5"

pdcastro:
Thanks for this report @aifreedom. 👍  I wonder if it is the same issue as #1847 - Node.js v14 is not yet fully supported. Does the error happen with Node.js v12? You can also try the graphical installer for macOS or the standalone zip file, which bundle their own Node.js interpreter. Instructions at: https://github.com/balena-io/balena-cli/blob/master/INSTALL.md

aifreedom:
@pdcastro Sorry for late reply. It works without problem with Node.js v12.

pdcastro:
`balena push` now works with Node.js v14 -- since balena CLI version 12.9.0, I believe (just tested with CLI v12.14.4). Therefore I am closing this issue -- but feel free to add comments as needed.

Note however that, for now, the officially supported Node.js versions are still 10 and 12 ([as documented](https://github.com/balena-io/balena-cli/blob/master/INSTALL.md)) because other CLI commands like `os configure` still don't work with Node.js v14 (issue #1990).

Thanks again for reporting this issue! 👍

PR #1932, opened 2020-07-09
Update net-keepalive to 1.3.1

Page-:
Connects-to: #1878 
Change-type: minor

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -93,7 +93,7 @@ If you are a Node.js developer, you may wish to install the balena CLI via [npm]
 The npm installation involves building native (platform-specific) binary modules, which require
 some additional development tools to be installed first:
 
-* [Node.js](https://nodejs.org/) version 10 or 12 (version 14 is not yet fully supported)
+* [Node.js](https://nodejs.org/) version 10 (min **10.20.0**) or 12 (version 14 is not yet fully supported)
   * **Linux, macOS** and **Windows Subsystem for Linux (WSL):**  
     Installing Node via [nvm](https://github.com/nvm-sh/nvm/blob/master/README.md) is recommended.
     When the "system" or "default" Node.js and npm packages are installed with "apt-get" in Linux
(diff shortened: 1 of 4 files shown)


pdcastro:
I'm holding approval of this PR because of the CI error:
```
1) balena logs
       should reach the expected endpoints on a local device:
     AssertionError: expected [ 'This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:\n',

  'Error: No native build was found for platform=darwin arch=x64 runtime=node abi=64 uv=1 libc=glibc\n',
```

It looks like a proper error. If you try running `balena logs <ip-address>`, does it fail too? Both `npm install` and standalone zip (`npm run build:standalone`, `./build-bin/balena logs <ip-address>`) It is one of the commands that use `net-keepalive`. (The other command is `balena push <ip-address>`)

INSTALL.md
```diff
@@ -93,7 +93,7 @@ If you are a Node.js developer, you may wish to install the balena CLI via [npm]
 The npm installation involves building native (platform-specific) binary modules, which require
 some additional development tools to be installed first:
 
-* [Node.js](https://nodejs.org/) version 10 or 12 (version 14 is not yet fully supported)
+* [Node.js](https://nodejs.org/) version 10 (min **10.20.0**) or 12 (version 14 is not yet fully supported)
```
pdcastro, line 96: @Page-, I have added this note to the docs. Feel free to modify it.

Page-:
@balena-ci rebase

Page-:
@balena-ci retest

pdcastro:
@Page-, I've got it working with an alternative PR: #1940 
Re commit message `Update net-keepalive to 1.3.1, adding support for nodejs 14`, I think _"adding support for nodejs 14"_ could be a bit misleading in the Changelog because some commands still don't support Node v14, even after updating `net-keepalive` and `pkg`. For example `os configure`, I think it was because of ext2 related modules.

PR #1934, opened 2020-07-09
Adjustments for balena-preload v10 (SDK v14). Improved error reporting.

pdcastro:
Connects-to: balena-io-modules/balena-preload/pull/219
Depends-on: balena-io-modules/balena-preload/pull/220
Change-type: patch


lib/actions/preload.js
```diff
@@ -20,6 +20,27 @@ import * as dockerUtils from '../utils/docker';
 
 const isCurrent = (commit) => commit === 'latest' || commit === 'current';
 
+/** @type {any} */
+const applicationExpandOptions = {
```
pdcastro, line 24: This is a copy-and-paste from:
https://github.com/balena-io-modules/balena-preload/blob/v10.1.0/lib/preload.js#L778-L796
... in order to avoid `balena-cli` depending on it, given that different SDK versions may need different query data structures.

The `any` type avoids some tsc failures that I struggled to resolve... But it was successfully tested.

lib/actions/preload.js
```diff
@@ -178,11 +208,73 @@ Alternatively you can pass the --pin-device-to-release flag to pin only this dev
...
+ * @param {string | number} appId
+ * @returns {Promise<import('balena-sdk').Application>}
+ */
+async function getAppWithReleases(balenaSdk, appId) {
+	return balenaSdk.models.application.get(appId, {
+		$expand: applicationExpandOptions,
+	});
+}
+
+async function prepareAndPreload(preloader, balenaSdk, options) {
```
pdcastro, line 222: The body of this function was moved/refactored from the `action()` function. `then` chains were replaced with async/await.

lib/actions/preload.js
```diff
@@ -178,11 +208,73 @@ Alternatively you can pass the --pin-device-to-release flag to pin only this dev
...
+	const { ExpectedError } = require('../errors');
+
+	await preloader.prepare();
+	preloader.config = { deviceType: 'intel-nuc' };
+
+	const application = options.appId
+		? await getAppWithReleases(balenaSdk, options.appId)
+		: await selectApplication(preloader.config.deviceType);
```
pdcastro, line 230: The `getAppWithReleases` call is new: previously, `balena-preload` would do the application fetching.

lib/actions/preload.js
```diff
@@ -178,11 +208,73 @@ Alternatively you can pass the --pin-device-to-release flag to pin only this dev
...
+		}
+	} else {
+		// this could have the value 'current'
+		commit = await selectApplicationCommit(application.owns__release);
+	}
+
+	await preloader.setAppIdAndCommit(
+		application.id,
+		isCurrent(commit) ? application.commit : commit,
+	);
```
pdcastro, line 264: `setAppIdAndCommit` is a new function added in balena-io-modules/balena-preload/pull/220

patches/all/balena-preload+9.0.0.patch
```diff
@@ -1,20 +0,0 @@
-diff --git a/node_modules/balena-preload/src/preload.py b/node_modules/balena-preload/src/preload.py
-index 887eaf3..0c5cc7f 100755
---- a/node_modules/balena-preload/src/preload.py
-+++ b/node_modules/balena-preload/src/preload.py
```
pdcastro, line 4: This patch is no longer necessary with `balena-preload` v10.2.0.

lib/actions/preload.js
```diff
@@ -178,11 +208,73 @@ Alternatively you can pass the --pin-device-to-release flag to pin only this dev
...
+				);
+			}
+			commit = release.commit;
+		}
+	} else {
+		// this could have the value 'current'
+		commit = await selectApplicationCommit(application.owns__release);
+	}
+
+	await preloader.setAppIdAndCommit(
```
pdcastro, line 261: `setAppIdAndCommit` is a new function added in balena-io-modules/balena-preload/pull/220. This is a key fix that allows balena-preload v10 (which uses SDK v14) to work with the current CLI (which uses SDK v13).

pdcastro:
@balena-ci rebase

pdcastro:
@balena-ci retest

lib/actions/preload.js
thgreasi: Looks like the sdk typings do not support the `Array<Dictionary<string>>` notation for `$orderby` :(
Let's do the following though, so that we do get a bit better typings:
```
/** @type {import('balena-sdk').PineExpandFor<import('balena-sdk').Application>} */
const applicationExpandOptions = {
	owns__release: {
		$select: ['id', 'commit', 'end_timestamp', 'composition'],
		/** @type {any} */
		$orderby: [{ end_timestamp: 'desc' }, { id: 'desc' }],
```

Page-: I'd rather a `// @ts-ignore $reason` or `// @ts-expect-error $reason` for this rather than a cast to `any` with no explanation, with the `ts-ignore`/`ts-expect-error` it's easier to grep for and with the comment it's also possible to know what's necessary to remove the bypass

Page-:
@balena-ci rebase

Page-:
@balena-ci retest

Issue #1731, opened 2020-04-16
Give a better error message on "balena ssh" when the keys don't match


jviotti:
The error currently looks like this:

```
$ balena ssh e19f
user@ssh.balena-devices.com: Permission denied (publickey).
Warning: ssh process exited with non-zero code "255"
```

Which may confuse some users. Maybe we can expand on that saying that we believe they don't have the right SSH key configured in their balena cloud account, and maybe even provide a link?

See https://forums.balena.io/t/94451

nwneisen:
I'd like to start contributing to Balena and thought I would start here.

From what I can tell, `user@ssh.balena-devices.com: Permission denied (publickey).` is the stdout of ssh and `Warning: ssh process exited with non-zero code "255"` is the logged output from balena-cli. This is a generic message that is printed no matter what error code is returned from ssh.

The desired solution seems to be to watch for code 255 and improve the message with something like:
`Are your SSH keys correctly configured in Balena Cloud? https://dashboard.balena-cloud.com/preferences/sshkeys`. I can open a pull request if this sounds like the desired solution.

jviotti:
I think that will do! Thanks!

nwneisen:
@jviotti Got in a PR. Let me know any adjustments it may need.

PR #1935, opened 2020-07-10
Give a better error message on "balena ssh" when the keys don't match

nwneisen:
Improve the SSH error message depending on the error code returned from SSH.

Resolves: #1731 
Change-type: patch

### DIFF
--- a/lib/utils/ssh.ts
+++ b/lib/utils/ssh.ts
@@ -147,9 +147,22 @@ export async function spawnSshAndThrowOnError(
 		// to return exit code 1. In these cases, print a short one-line warning
 		// message, and exits the CLI process with the same error code.
 		process.exitCode = exitCode;
-		const msg = exitSignal
-			? `Warning: ssh process was terminated with signal "${exitSignal}"`
-			: `Warning: ssh process exited with non-zero code "${exitCode}"`;
-		throw new ExpectedError(msg);
+		throw new ExpectedError(sshErrorMessage(exitSignal, exitCode));
 	}
+}
+
+function sshErrorMessage(exitSignal?: string, exitCode?: number) {
+	const msg: string[] = [];
+	if (exitSignal) {
+		msg.push(`Warning: ssh process was terminated with signal "${exitSignal}"`);
+	} else {
+		msg.push(`Warning: ssh process exited with non-zero code "${exitCode}"`);
+		switch (exitCode) {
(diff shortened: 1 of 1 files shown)


nwneisen:
Thanks for the suggestions @pdcastro! Getting into a new code base and new code review process is always interesting so I appreciate the minor changes to get the practice.

lib/utils/ssh.ts
```diff
@@ -147,9 +147,22 @@ export async function spawnSshAndThrowOnError(
 		// to return exit code 1. In these cases, print a short one-line warning
 		// message, and exits the CLI process with the same error code.
 		process.exitCode = exitCode;
-		const msg = exitSignal
-			? `Warning: ssh process was terminated with signal "${exitSignal}"`
-			: `Warning: ssh process exited with non-zero code "${exitCode}"`;
-		throw new ExpectedError(msg);
+		throw new ExpectedError(sshErrorMessage(exitSignal!, exitCode!));
```
pdcastro, line 150: hmm Now my focus in on those two exclamation marks. They tell the tsc compiler "I know that this value will not be undefined", but in fact, for example `exitSignal` will most of the time actually be undefined, and this could mask errors in the implementation of `sshErrorMessage`. I suggest replacing the exclamation marks here with question marks in the signature of `sshErrorMessage`:
```
    throw new ExpectedError(sshErrorMessage(exitSignal, exitCode));
...

function sshErrorMessage(exitSignal?: string, exitCode?: number) {
...
```
Apparently this would not require any changes in the body of `sshErrorMessage`.

nwneisen: I should've looked into ! further (New to typescript). The ? looks appropriate as undefined will simply pass through the switch statement.

pdcastro:
@balena-ci rebase

pdcastro:
@balena-ci rebase

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci rebase

pdcastro:
@nwneisen, it looks like balena CI is ignoring the `rebase` command, although it understood the `retest` command. Not sure if it is specific to this PR: I'll ask the CI team to have a look. I suggest that you rebase it manually on the master branch, then force-push the change. Thanks!

pdcastro:
@balena-ci retest

Issue #323, opened 2016-03-07
allow move device to accept an array of UUIDs


shaunmulligan:
It would be super useful it the `resin device move` could allow someone to move a batch of devices from one application to another.

shaunmulligan:
`device rm`  as well as `env add` and `env rm` would also be awesome to have as batch actions.

nwneisen:
I would like to take a look at adding this feature.

My fist thought was to change the `uuid` parameter of `move` to be a string array of uuids. This could end up breaking user's scripts and be a change for everyone used to the interface. 

My second thought is to have a second method that allows for multiple devices to be moved using a string array of uuids. I'm thinking `balena devices move [<uuid>, [uuid]]`. 

Let me know which interface is preferred and I will switch my code over accordingly.

pdcastro:
> to change the `uuid` parameter of `move` to be a string array of uuids. This could end up breaking user's scripts [...]

Like `balena device move uuid1,uuid2,uuid3` ? Actually, if no spaces were allowed in the list, it would not necessarily be a breaking change, would it? From oclif's point of view, it would still be a single string argument, and our own code (`run()` function in `move.ts`) could split the string at the commas.

> to have a second method [...] `balena devices move [<uuid>, [uuid]]`

No!! :-) You mean we'd have `device move` and `devices move`, doing the same except for the number of arguments? Background: a common criticism of the CLI is that, as it is, it already has several commands that partially overlap in functionality, and this is considered confusing for users.

Additional option:
```
balena device move uuid1 uuid2 uuid3 -a newApp
```
This looks like backwards compatible with the present `balena device move uuid -a newApp`. I see that oclif mentions the possibility of "variable length arguments":
> For variable length arguments, disable argument validation with `static strict = false` on the command.
> https://oclif.io/docs/args It may require some testing / experimentation to check it works 

I don't think I ever tried it though. Also, we should test whether the position of the `-a` flag matters:
`balena device move uuid1 uuid2 uuid3 -a newApp`
`balena device move -a newApp uuid1 uuid2 uuid3`

Ideally the position of `-a` would not matter; if it does, it may still be acceptable, but we'd have to document it.

@srlowe, do you have any additional thoughts on this?

srlowe:
[...]

PR #1939, opened 2020-07-13
balena device: Add ability to move multiple devices in one command

nwneisen:
Add the ability to move multiple devices in one command

Partially Resolves: #323 
Change-type: minor

### DIFF
--- a/lib/actions-oclif/device/move.ts
+++ b/lib/actions-oclif/device/move.ts
@@ -23,6 +23,7 @@ import * as cf from '../../utils/common-flags';
 import { expandForAppName } from '../../utils/helpers';
 import { getBalenaSdk, stripIndent } from '../../utils/lazy';
 import { tryAsInteger } from '../../utils/validation';
+import { ExpectedError } from '../../errors';
 
 interface ExtendedDevice extends Device {
 	application_name?: string;
@@ -40,28 +41,30 @@ interface ArgsDef {
 
 export default class DeviceMoveCmd extends Command {
 	public static description = stripIndent`
-		Move a device to another application.
+		Move one or more devices to another application.
 
-		Move a device to another application.
+		Move one or more devices to another application.
 
 		Note, if the application option is omitted it will be prompted
 		for interactively.
 		`;
 	public static examples = [
 		'$ balena device move 7cf02a6',
+		'$ balena device move 7cf02a6,dc39e52',
 		'$ balena device move 7cf02a6 --application MyNewApp',
 	];
 
 	public static args: Array<IArg<any>> = [
 		{
 			name: 'uuid',
-			description: 'the uuid of the device to move',
+			description:
+				'comma-separated list (no blank spaces) of device UUIDs to be moved',
 			parse: (dev) => tryAsInteger(dev),
 			required: true,
 		},
 	];
 
-	public static usage = 'device move <uuid>';
+	public static usage = 'device move <uuid(s)>';
 
 	public static flags: flags.Input<FlagsDef> = {
 		application: cf.application,
@@ -82,47 +85,72 @@ export default class DeviceMoveCmd extends Command {
 		options.application = options.application || options.app;
 		delete options.app;
 
-		const device: ExtendedDevice = await balena.models.device.get(
-			params.uuid,
-			expandForAppName,
+		const devices: ExtendedDevice[] = await Promise.all(
+			params.uuid
+				.split(',')
+				.map((uuid) => balena.models.device.get(uuid, expandForAppName)),
 		);
 
-		const belongsToApplication = device.belongs_to__application as Application[];
-		device.application_name = belongsToApplication?.[0]
-			? belongsToApplication[0].app_name
-			: 'N/a';
+		for (const device of devices) {
+			const belongsToApplication = device.belongs_to__application as Application[];
+			device.application_name = belongsToApplication?.[0]
+				? belongsToApplication[0].app_name
+				: 'N/a';
+		}
 
 		// Get destination application
 		let application;
 		if (options.application) {
 			application = options.application;
 		} else {
-			const [deviceDeviceType, deviceTypes] = await Promise.all([
-				balena.models.device.getManifestBySlug(device.device_type),
+			const [deviceDeviceTypes, deviceTypes] = await Promise.all([
+				Promise.all(
+					devices.map((device) =>
+						balena.models.device.getManifestBySlug(device.device_type),
+					),
+				),
 				balena.models.config.getDeviceTypes(),
 			]);
 
-			const compatibleDeviceTypes = deviceTypes.filter(
-				(dt) =>
-					balena.models.os.isArchitectureCompatibleWith(
-						deviceDeviceType.arch,
-						dt.arch,
-					) &&
-					!!dt.isDependent === !!deviceDeviceType.isDependent &&
-					dt.state !== 'DISCONTINUED',
+			const compatibleDeviceTypes = deviceTypes.filter((dt) =>
+				deviceDeviceTypes.every(
+					(deviceDeviceType) =>
+						balena.models.os.isArchitectureCompatibleWith(
+							deviceDeviceType.arch,
+							dt.arch,
+						) &&
+						!!dt.isDependent === !!deviceDeviceType.isDependent &&
+						dt.state !== 'DISCONTINUED',
+				),
 			);
 
 			const patterns = await import('../../utils/patterns');
-			application = await patterns.selectApplication(
-				(app: Application) =>
-					compatibleDeviceTypes.some((dt) => dt.slug === app.device_type) &&
-					// @ts-ignore using the extended device object prop
-					device.application_name !== app.app_name,
-			);
+			try {
+				application = await patterns.selectApplication(
+					(app: Application) =>
+						compatibleDeviceTypes.some((dt) => dt.slug === app.device_type) &&
+						// @ts-ignore using the extended device object prop
+						devices.some((device) => device.application_name !== app.app_name),
+					true,
+				);
+			} catch (err) {
+				if (deviceDeviceTypes.length) {
+					throw new ExpectedError(
+						`${err.message}\nDo all devices have a compatible architecture?`,
+					);
+				}
+				throw err;
+			}
 		}
 
-		await balena.models.device.move(params.uuid, tryAsInteger(application));
-
-		console.info(`${params.uuid} was moved to ${application}`);
+		for (const uuid of params.uuid.split(',')) {
+			try {
+				await balena.models.device.move(uuid, tryAsInteger(application));
+				console.info(`${uuid} was moved to ${application}`);
+			} catch (err) {
+				console.info(`${err.message}, uuid: ${uuid}`);
+				process.exitCode = 1;
+			}
+		}
 	}
 }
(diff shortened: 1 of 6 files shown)


lib/actions-oclif/device/move.ts
```diff
@@ -82,47 +83,51 @@ export default class DeviceMoveCmd extends Command {
...
-			const compatibleDeviceTypes = deviceTypes.filter(
-				(dt) =>
-					balena.models.os.isArchitectureCompatibleWith(
-						deviceDeviceType.arch,
-						dt.arch,
-					) &&
-					!!dt.isDependent === !!deviceDeviceType.isDependent &&
-					dt.state !== 'DISCONTINUED',
+		for (const uuid of params.uuid.split(',')) {
+			const device: ExtendedDevice = await balena.models.device.get(
```
pdcastro, line 88: If I'm reading this right (I may not be!), a network request for `await balena.models.device.get(...)` is made for each `uuid`, but the resulting `device` object is only actually possibly used for the first device, in order to determine the destination application (and even then, only in case `options.application` was not provided). Because ultimately, to actually move the device, all that is needed seems to be uuid string:
```
    await balena.models.device.move(uuid, tryAsInteger(application));
```

So how about the following structure:
* Extract the first `uuid` (first device), before any `for` loop, something like:  
`const uuid = params.uuid.split(',')[0];`
* Then determine the destination application. Only if needed (if `options.application` was not provided), call `await balena.models.device.get()` for that first uuid only.
* Then, knowing the destination application, have a small for loop moving the devices:
```
	for (const uuid of params.uuid.split(',')) {
		await balena.models.device.move(uuid, tryAsInteger(application));
		console.info(`${uuid} was moved to ${application}`);
	}
```

lib/actions-oclif/device/move.ts
```diff
@@ -82,47 +83,51 @@ export default class DeviceMoveCmd extends Command {
...
-					// @ts-ignore using the extended device object prop
-					device.application_name !== app.app_name,
-			);
+			const belongsToApplication = device.belongs_to__application as Application[];
+			device.application_name = belongsToApplication?.[0]
+				? belongsToApplication[0].app_name
+				: 'N/a';
+
+			// Get destination application
+			if (!application) {
```
pdcastro, line 99: My other comment suggests a different overall structure. If that structure was not feasible, then I'm thinking it would easier to read the code if this `if (!application) {...}` block was move outside the `for` loop, perhaps in a function.

lib/actions-oclif/device/move.ts
```diff
@@ -40,28 +40,29 @@ interface ArgsDef {
...
 		'$ balena device move 7cf02a6',
+		'$ balena device move 7cf02a6,dc39e52',
 		'$ balena device move 7cf02a6 --application MyNewApp',
 	];
 
 	public static args: Array<IArg<any>> = [
 		{
 			name: 'uuid',
-			description: 'the uuid of the device to move',
+			description: 'the uuid(s) of the device to move',
```
pdcastro, line 59: Suggestion:
```
description: 'comma-separated list (no blank spaces) of device UUIDs to be moved',
```

lib/actions-oclif/device/move.ts
nwneisen: The one problem I see with this method is that `await balena.models.device.get(...)` is validating the uuids exist. If one of the uuids is invalid then the operation will stop before anything is moved. Without this the operation could throw an error and stop with half of the devices moved and the second half left as is. Perhaps the solution for this would be report the uuid that failed to the user and continue trying to move the remaining uuids.

pdcastro: That's a good point and feature: validating that all UUIDs exist before attempting to move any of them. 👍 
In this case, how about:
* Query all UUIDs, perhaps even in parallel with `Promise.all`:  
`const devices = await Promise.all(params.uuid.split(',').map((uuid) => balena.models.device.get(...)));`
* Then determine the destination application, using `devices[0]` as needed (if `options.application` was not provided).
* Then, knowing the destination application, have a small for loop moving the devices:
```
	for (const uuid of params.uuid.split(',')) {
		await balena.models.device.move(uuid, tryAsInteger(application));
		console.info(`${uuid} was moved to ${application}`);
	}
```

lib/actions-oclif/device/move.ts
```diff
@@ -82,47 +85,67 @@ export default class DeviceMoveCmd extends Command {
...
-			const compatibleDeviceTypes = deviceTypes.filter(
-				(dt) =>
-					balena.models.os.isArchitectureCompatibleWith(
-						deviceDeviceType.arch,
-						dt.arch,
-					) &&
-					!!dt.isDependent === !!deviceDeviceType.isDependent &&
-					dt.state !== 'DISCONTINUED',
+			const compatibleDeviceTypes = deviceTypes.filter((dt) =>
+				deviceDeviceTypes.every(
```
pdcastro, line 116: I am contributing a commit: please review it, test the command and feel free to make any further changes (and let me know when/if you're happy. I was doing some testing myself and, before my commit, I realised that some corner cases existed if the device list mixed devices of incompatible architectures. Example, before my commit:
```
$ ./bin/balena device move b1878c6,dcbff32
? Select an application
❯ test-19282 (qemux86-64)
  test-nuc (intel-nuc)

$ ./bin/balena device move dcbff32,b1878c6
? Select an application
  tests (raspberrypi3)
  test-skx2 (skx2)
  test-tx2 (jetson-tx2)
❯ bbb-test (beaglebone-green-wifi)
  essentials (raspberrypi3)
  iwantlegacy (raspberrypi3)
  raspberrypi2 (raspberry-pi2)
```
Where b1878c6 is an amd64 device, dcbff32 is an ARM device. Note how the list of applications changes depending on the order of the devices in the comma-separated list in the command line. By the way, I realise that it was my own suggestion to use `devices[0]`, then again, the original code before my suggestion was also subject to this issue. :-)
After my commit:
```
$ ./bin/balena device move dcbff32,b1878c6
No suitable applications found for selection
Do all devices have a compatible architecture?

$ ./bin/balena device move dcbff32,b1878c6 -a test-rpi
dcbff32 was moved to test-rpi
BalenaInvalidDeviceType: Invalid device type: Incompatible application: test-rpi
```
The last case above raises `BalenaInvalidDeviceType`, which I had added to `errors.ts` as an "expected error". I think it is acceptable.

lib/actions-oclif/device/move.ts
```diff
@@ -144,8 +144,12 @@ export default class DeviceMoveCmd extends Command {
 		}
 
 		for (const uuid of params.uuid.split(',')) {
-			await balena.models.device.move(uuid, tryAsInteger(application));
-			console.info(`${uuid} was moved to ${application}`);
+			try {
```
nwneisen, line 147: I found an additional error case with both of our code changes. When passing the application with `-a`, valid devices before an invalid arch uuid will be moved and devices after will not.This change will move all devices with a valid arch and report the uuid of invalid arch devices. 

This is different behavior than we previously discussed in which we validate devices before moving them. I think moving valid devices and reporting invalid or not moving any when an invalid is found could both be appropriate behavior as long as it's clear to the user. The down side of the first method is this is calling the move command and expecting an error during normal behavior.

I'd like to change method to use the following process:

1. Acquire and validate(not currently done for `-a`) the application.
2. Acquire valid device arch's for the application.
3. Verify all device arch's are valid for the application. Report those that aren't.
4. Move devices if all device archs are valid.

I looked around but wasn't able to find a method to validate an application exists by string or get a list of valid archs for the application.

lib/actions-oclif/device/move.ts
```diff
@@ -144,8 +144,12 @@ export default class DeviceMoveCmd extends Command {
...
+				await balena.models.device.move(uuid, tryAsInteger(application));
+				console.info(`${uuid} was moved to ${application}`);
+			} catch (err) {
+				console.info(`${err.message}, uuid: ${uuid}`);
```
pdcastro, line 151: Nice! 👍   Additional easy improvement: add `process.exitCode = 1;` here, so that the CLI process exit code reflects the fact that something went wrong and some devices were not moved (in case the CLI is being scripted).

> I'd like to change method to use the following process: [...] Move devices if all device archs are valid.

You're welcome to, but my opinion/suggestion is that it is good enough as it is, and it could always be improved further in the future (in the way you suggest). I think it is already a great feature delivery as it is, and additional PRs to deliver the same to other commands (e.g. `device rm`) (or PRs for other issues) would offer a greater return on investment at this point! :-)

nwneisen: `process.exitCode = 1` added. I've also got a branch started to add the same functionality to `device rm`.

pdcastro: Great! Would you mind squashing the 4 commits in a single one? Probably using the description of your first commit. Generally we follow the principle of "meaningful commits", and the commits in this PRs at the moment are effectively amending each other.

nwneisen: I debated about doing so but waited to see what was typically done for the project.

pdcastro:
@balena-ci retest

lib/actions-oclif/device/move.ts
pdcastro: Oh, sorry, just noticed one more thing: the `Change-type` for the commit should be `minor` rather than `patch`, as it is delivering a new feature (and one that is visible to users!).

pdcastro:
@balena-ci retest

pdcastro:
> _It looks like balena CI is ignoring the rebase command, although it understood the retest command. Not sure if it is specific to this PR: I'll ask the CI team to have a look._

Re comment above (from PR #1935), the CI team thinks that is because the source branch is in a forked repo -- nothing wrong with it, it would be a balena CI bug. An issue was created to get this fixed. I have also noticed something else (another CI bug): when you push or force-push a new commit, tests are not re-triggered automatically. Chances are it is the same cause: forked repo. I'll report to the CI team as well. Until then, it looks like the reliable (?) way for you to get the tests re-triggered is to add the `(at)balena-ci retest` command in a PR comment (I had done it, and then again after the force push).

nwneisen:
> the source branch is in a forked repo
> Chances are it is the same cause: forked repo

Out of curiosity what is the alternative method for making changes? Creating branches directly on the main repo? I had thought it was standard practice on most projects to create a fork and merge branches back into the main repo.

pdcastro:
> Creating branches directly on the main repo?

Yes.

> I had thought it was standard practice on most projects to create a fork and merge branches back into the main repo.

I have also previously worked like that (in another company), but I don't really about what the "standard" (most popular?) practice is.

pdcastro:
By the way, talking of odd things, I've also noticed that the automatically updated Changelog uses your first name only, whereas everybody else has name and surname listed there:
https://github.com/balena-io/balena-cli/blob/master/CHANGELOG.md
```
12.8.0 - 2020-07-15 device move: Add ability to move multiple devices in one command [Nick]
12.7.7 - 2020-07-14 Remove os download command dependency for device init [Scott Lowe]
```
Would this be some configuration in your GitHub account? Yet, your surname is clearly visible on the following page:
https://github.com/nwneisen

Would it be that balena CI doesn't have some permissions? No idea...

nwneisen:
@pdcastro My local machine only had my first name set in the git credentials. I've updated it to be my full name and we will find out on my next commit if that was the reason.

PR #1940, opened 2020-07-13
Update net-keepalive and pkg, Node.js >= v10.20.0 (steps towards Node.js v14)

pdcastro:
Connects-to: #1878
Connects-to: #1932
Change-type: minor


patches/apply-patches.js
```diff
@@ -0,0 +1,92 @@
+/**
```
Page-, line 1: @pdcastro since this file isn't run through the ts transpilation step it needs to be directly compatible with node 10 rather than being able to rely on transpilation - is it definitely compatible?

Page-, line 1: It would be nice to include this file in tsconfig.dev.json so it at least gets type checked

pdcastro: Good point! 👍   I had tested with Node 10.21.0 and it was OK, but thanks to your question I tested it with 10.0.0 and found (as expected) that `const { promises: fs } = require('fs');` was not OK. It is best to have it compatible with 10.0.0 if easy to achieve (even if `engines.node` says `>=10.20.0`) because this allows more graceful error / warning message to be printed when users attempt to run the CLI. So I have made some changes and `apply-patches.js` is now compatible with Node v10.0.0.

pdcastro: Done 👍

patches/apply-patches.js
```diff
@@ -0,0 +1,80 @@
...
+async function applyPatches() {
+	const isUnix = ['linux', 'darwin'].includes(process.platform);
+	const patchDirs = [
+		path.join(patchesDir, 'before-all'),
+		path.join(patchesDir, 'all'),
+		...(isUnix ? [path.join(patchesDir, 'unix')] : []),
+		path.join(patchesDir, process.platform),
+	];
+	for (const patchDir of patchDirs) {
+		if (fs.existsSync(patchDir)) {
```
pdcastro, line 64: Using `existsSync` so it is compatible with Node v10.0.0 without any fuss. Performance doesn't change because `apply-patches` is executed in a child process and would be blocked here anyway by `await asyncFunc(...)`.

patches/apply-patches.js
Page-: Did it fail with 10.0.0 @pdcastro or just throw the experimental warning?

pdcastro: iirc, it was proper fatal failure, not just a warning.

pdcastro:
@balena-ci rebase

pdcastro:
@balena-ci rebase

patches/apply-patches.js
Page-: That's really weird, the docs say they were all added as part of 10.0.0

pdcastro: The Changelog for Node.js mentions that `fs/promises` was moved to `fs.promises` in v10.1.0, and I've found the following comment:
> https://github.com/jsdom/jsdom/issues/2939#issuecomment-612193437
> _It does appear that anything `<10.1` will fail because `fs.promises` doesn't exist until then. Before `10.1` it was `fs/promises`._

Maybe this is what I was coming across when the script had failed.

pdcastro:
@balena-ci rebase

PR #1941, opened 2020-07-14
Convert `os download` to oclif, typescript

srlowe:
- Convert `os download` to oclif, typescript
 - Remove `os download` command dependency for `device init`


lib/utils/cloud.ts
```diff
@@ -106,3 +106,105 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
...
+	const fs = await import('fs');
+	const manager = await import('balena-image-manager');
+
+	console.info(`Getting device operating system for ${deviceType}`);
+
+	if (!OSVersion) {
+		console.warn(`OS version is not specified, using the default version: \
+the newest stable (non-pre-release) version if available, or the newest version otherwise \
+(if all existing versions for the given device type are pre-release).`);
+	}
```
srlowe, line 140: I'll update this later with the new short version.

lib/actions-oclif/os/download.ts
```diff
@@ -0,0 +1,95 @@
...
+	type: string;
+}
+
+export default class OsDownloadCmd extends Command {
+	public static description = stripIndent`
+		Download an unconfigured OS image.
+
+		Download an unconfigured OS image for a certain device type.
+		Check available types with \`balena devices supported\`
+
```
pdcastro, line 39: The Capitano version had these lines in the command help text:
```
> Note: Currently this command only works with balenaCloud, not openBalena.	
> If using openBalena, please download the OS from: https://www.balena.io/os/
```
Accidental or intentional removal?

lib/actions-oclif/os/download.ts
```diff
@@ -0,0 +1,95 @@
...
+				or 'latest' (includes pre-releases),
+				or 'default' (excludes pre-releases if at least one stable version is available),
+				or 'recommended' (excludes pre-releases, will fail if only pre-release versions are available),
+				or 'menu' (will show the interactive menu)
+				`,
+		}),
+		help: cf.help,
+	};
+
+	public static authenticated = true;
```
pdcastro, line 84: `authenticated = true;`  doesn't match the Capitano version, and I think it was a recent change as well (CLI v12.4.0). Given my other comment about the help text, I am guessing that the oclif conversion was based on an old commit / version of the CLI.

lib/actions-oclif/os/download.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			OsDownloadCmd,
+		);
+
+		const { downloadOSImage } = await import('../../utils/cloud');
+
+		return await downloadOSImage(params.type, options.output, options.version);
```
pdcastro, line 93: * Great refactoring! 👍 
* Just `await` instead of `return await`?

lib/utils/cloud.ts
pdcastro: Then I must not approve the PR yet, as bulldozer would merge it if I did... :-)

lib/actions-oclif/device/init.ts
```diff
@@ -116,7 +117,8 @@ export default class DeviceInitCmd extends Command {
 		const tmpPath = (await tmpNameAsync()) as string;
 		try {
 			logger.logDebug(`Downloading OS image...`);
-			await this.downloadOsImage(tmpPath, application.device_type, options);
+			const osVersion = options['os-version'] || 'default';
```
pdcastro, line 120: This defaulting to `'default'` here prevents the printing of the warning message regarding a default version being used.

lib/utils/cloud.ts
```diff
@@ -106,3 +106,105 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
...
+ * If not provided, OSVersion will be set to `default`
+ *
+ * @param deviceType
+ * @param outputPath
+ * @param OSVersion
+ */
+export async function downloadOSImage(
+	deviceType: string,
+	outputPath: string,
+	OSVersion?: string,
```
pdcastro, line 128: How about declaring this parameter with a default value?
```
	OSVersion = 'default',
```

The reason for this suggestion is that I saw the constant `'default'` being defined in `actions-oclif/device/init.ts` and also below:
```
	OSVersion = OSVersion
		? await resolveOSVersion(deviceType, OSVersion)
		: 'default';
```
It's nicer to have the constant in a single place. But we'd need to check that a blank string is not provided as argument, as a blank string currently gets detected by the `if (!OSVersion)` test but would not be replaced by a default value here.

lib/actions-oclif/os/download.ts
srlowe: Oops - accidental.

lib/actions-oclif/os/download.ts
srlowe: Ah ok.  It may have been a bit old yeah (1 week maybe).

lib/actions-oclif/os/download.ts
srlowe: Yeah right.  It was like that before (I believe) because the output what being used by the command-command calls (but it's not anymore, and that kind of value passing doesn't work with oclif anyway).  Will amend

lib/actions-oclif/device/init.ts
srlowe: Yeah, that was the previous behaviour, so I kept it.  Would you prefer it changed?

lib/utils/cloud.ts
srlowe: Pro's and cons I think.  With the current setup the caller (either a user of the os download command, or some part of the CLI), can explicitly specify `default` as the version supress the warning (which seems reasonable).

srlowe: Though considering the message is nice and short now, there's no harm in always printing it I suppose.

lib/actions-oclif/os/download.ts
srlowe: I see that those angle brackets are in the source - are they supposed to be there?

lib/actions-oclif/device/init.ts
pdcastro: Just realised that the help for `os-version` in `balena help device init` is quite detailed:
```
		'os-version': flags.string({
			description: stripIndent`
				exact version number, or a valid semver range,
				or 'latest' (includes pre-releases),
				or 'default' (excludes pre-releases if at least one stable version is available),
				or 'recommended' (excludes pre-releases, will fail if only pre-release versions are available),
				or 'menu' (will show the interactive menu)
				`,
		}),
```
And a user might type `--os-version 'default'`, in which case being warned about a default version being used (if that's what they just typed) is not particularly helpful.  Nonetheless, yes, I think it makes sense to remove ` || 'default'` on this line.

lib/utils/cloud.ts
pdcastro: > can explicitly specify default as the version suppress the warning

True. I think `OSVersion?: string` was better after all, considering that a user might type `'default'` on the command line.

lib/utils/cloud.ts
```diff
@@ -106,3 +106,105 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
...
+) {
+	const unzip = await import('node-unzip-2');
```
Page-, line 130: Can you move this inside the
```ts
	if (stream.mime === 'application/zip') {
```
below, that way we can only require it if we actually need it

lib/utils/cloud.ts
```diff
@@ -106,3 +106,105 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
...
+	const fs = await import('fs');
```
Page-, line 131: Same for moving this inside the
```ts
	} else {
		output = fs.createWriteStream(outputPath);
	}
```
below - if we're going to lazy load it we might as well do it right (I'm not certain if lazy-loading an in-built module can have an effect though)

lib/utils/cloud.ts
```diff
@@ -106,3 +106,105 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
...
+	const manager = await import('balena-image-manager');
```
Page-, line 132: I'd also move this directly before it's used, that way if `resolveOSVersion` fails we can avoid it, it also means we'll print the warning asap which should be better UX rather than having a slight delay

lib/actions-oclif/os/download.ts
pdcastro: They were meant to display nicely when formatted as markdown, but feel free to remove.

PR #1944, opened 2020-07-14
Update cli.md for push dockerfile option to specify compose is supported

phil-d-wilson:
Change-type: patch
Signed-off-by: Phil Wilson <phil@balena.io>

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -2545,7 +2545,7 @@ Run an emulated build using Qemu
 
 #### --dockerfile &#60;Dockerfile&#62;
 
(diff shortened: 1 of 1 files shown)


srlowe:
Thanks for the PR Phil!  That file is auto-generated though, the one you want to update is `lib/actions/push.ts`

phil-d-wilson:
Ah! That's annoying that the "Improve this doc" button makes you do the wrong thing. Updating documents just feels too hard at the moment.

I'll close this PR and open another for the correct file. Thanks Scott!

PR #1948, opened 2020-07-15
Convert `os initialize` to oclif, typescript

srlowe:
Change-type: patch


srlowe:
@balena-ci retest

lib/actions-oclif/os/initialize.ts
```diff
@@ -0,0 +1,137 @@
...
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			OsInitializeCmd,
+		);
+
+		const { promisify } = await import('bluebird');
+		const { umount } = await import('umount');
+		const umountAsync = promisify(umount);
```
srlowe, line 84: I wasn't able to get `util.promisify` working with `umount` (ts complained about wrong number of parameters).

srlowe:
@balena-ci retest

lib/actions-oclif/os/initialize.ts
Page-: It'll work if you change typings/umount/index.d.ts to
```ts
declare module 'umount' {
	export const umount: (
		device: string,
		callback: (err: Error | null, stdout?: any, stderr?: any) => void,
	) => void;
	export const isMounted: (
		device: string,
		callback: (err: Error | null, isMounted?: boolean) => void,
	) => void;
}

Page-: Specifically `util.promisify` needs the first arg typing to be `Error | null`

srlowe: Ah thanks!

PR #1950, opened 2020-07-15
Drop net-keepalive dependency (save 130MB and widen Node.js compatibility)

pdcastro:
Internal discussion: https://www.flowdock.com/app/rulemotion/i-cli/threads/GndT7dxnSblvT2NvykKQch59ATf

Connects-to: hertzg/node-net-keepalive/issues/26
Change-type: patch


Page-:
Updated net-keepalive avoids the install size issue

PR #1960, opened 2020-07-24
Convert command `push` to oclif

srlowe:
Change-type: patch


Page-:
@balena-ci retest

tests/commands/help.spec.ts
```diff
@@ -27,7 +27,7 @@ Primary commands:
 
     help [command...]                       show help
     login                                   login to balena
-    push <applicationOrDevice>              Start a remote build on the balena cloud build servers or a local mode device
+    push <applicationordevice>              start a remote build on the balena cloud build servers or a local mode device
```
Page-, line 30: It'd be nice to be able to either keep the capitalization or switch to `application_or_device` to make it more readable, with no differentiation between the words it loses a lot of clarity

lib/actions-oclif/push.ts
```diff
@@ -261,70 +185,50 @@ export const push: CommandDefinition<
...
+	public static primary = true;
+
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			PushCmd,
+		);
+
 		const sdk = getBalenaSdk();
-		const Bluebird = await import('bluebird');
 		const remote = await import('../utils/remote-build');
```
Page-, line 229: Could you move this import down into the branch where it's actually used, that way we can avoid the import for local builds :)

lib/actions-oclif/push.ts
```diff
@@ -261,70 +185,50 @@ export const push: CommandDefinition<
...
 		const deviceDeploy = await import('../utils/device/deploy');
```
Page-, line 230: Could you move this import down into the branch where it's actually used, that way we can avoid the import for remote builds :)

lib/actions-oclif/push.ts
```diff
@@ -444,5 +337,74 @@ export const push: CommandDefinition<
...
+
+		if (validateIPAddress(appOrDevice) || validateDotLocalUrl(appOrDevice)) {
+			return BuildTarget.Device;
+		}
+
+		return null;
+	}
+
+	async getAppOwner(sdk: BalenaSDK, appName: string) {
+		const { selectFromList } = await import('../utils/patterns');
```
Page-, line 362: Could you move this down so we can avoid the import altogether if the app name is non-ambiguous

tests/commands/help.spec.ts
srlowe: Yeah I agree - this is an issue with the way the capitano based help command displays oclif commands.  This will be fixed when we move the help command to oclif (last command to be converted, about 4 away now).

lib/actions-oclif/push.ts
Page-: This one still stands

Page-:
@balena-ci retest

Issue #1956, opened 2020-07-17
Add an environment variable to multiple applications, devices, or services.


nwneisen:
This is a previously discussed issue that is desired to be added to the cli.

# Expected Behavior
Add an environment variable to multiple applications, devices, or services with one command.

# Actual Behavior
Not currently implemented

# Steps to Reproduce
`balena env add -d <uuid>,<uuid>...`
`balena env add -s <name>,<name>...`
`balena env add -a <name>,<name>...`

# Additional References
Originally mentioned in #323.

nwneisen:
A few clarifications that come to mind with this one:

- The desire is to add one variable to multiple applications, devices, or services. As apposed to multiple variables to one application, device, or service. Or even a many to many possibility..
- Should it be possible to add the variable to a combination of devices, services, and applications? Such as:
`balena env add -d <uuid>,<uuid>... -s <name>,<name>... -a <name>,<name>...`. Or maybe it doesn't make sense to be putting the same variable in the different locations from a user's perspective.

pdcastro:
> The desire is to add one variable to multiple applications, devices, or services

👍   I suspect the most common use case would be to list multiple devices.

> Should it be possible to add the variable to a combination of devices, services, and applications?

The `--service` option can already be used alongside `--device` or `--application`, but the `--device` and `--application` options are mutually exclusive. An explanation is that the variable value would usually be different between device and application: for example, an application variable being overridden (with a different value) for specific devices. I suggest preserving this behavior.

PR #1962, opened 2020-07-28
Add ability to add env var to multiple locations in one command

nwneisen:
Add ability to add env var to multiple locations in one command

Resolves: #1956 
Change-type: minor

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -961,15 +961,16 @@ do not prompt for confirmation before deleting the variable
 
 ## env add NAME [VALUE]
 
-Add an environment or config variable to an application, device or service,
-as selected by the respective command-line options. Either the --application
-or the --device option must be provided, and either may be be used alongside
-the --service option to define a service-specific variable. (A service is an
-application container in a "microservices" application.) When the --service
-option is used in conjunction with the --device option, the service variable
-applies to the selected device only. Otherwise, it applies to all devices of
-the selected application (i.e., the application's fleet). If the --service
-option is omitted, the variable applies to all services.
+Add an environment or config variable to one or more applications, devices 
+or services, as selected by the respective command-line options. Either the 
+--application or the --device option must be provided, and either may be be 
+used alongside the --service option to define a service-specific variable. 
+(A service is an application container in a "microservices" application.) 
+When the --service option is used in conjunction with the --device option, 
+the service variable applies to the selected device only. Otherwise, it 
+applies to all devices of the selected application (i.e., the application's 
+fleet). If the --service option is omitted, the variable applies to all 
+services.
 
 If VALUE is omitted, the CLI will attempt to use the value of the environment
 variable of same name in the CLI process' environment. In this case, a warning
@@ -988,9 +989,13 @@ Examples:
 
 	$ balena env add TERM --application MyApp
 	$ balena env add EDITOR vim --application MyApp
+	$ balena env add EDITOR vim --application MyApp,MyApp2
 	$ balena env add EDITOR vim --application MyApp --service MyService
+	$ balena env add EDITOR vim --application MyApp,MyApp2 --service MyService,MyService2
 	$ balena env add EDITOR vim --device 7cf02a6
+	$ balena env add EDITOR vim --device 7cf02a6,d6f1433
 	$ balena env add EDITOR vim --device 7cf02a6 --service MyService
+	$ balena env add EDITOR vim --device 7cf02a6,d6f1433 --service MyService,MyService2
 
 ### Arguments
 
--- a/lib/actions-oclif/env/add.ts
+++ b/lib/actions-oclif/env/add.ts
@@ -39,17 +39,18 @@ interface ArgsDef {
 
 export default class EnvAddCmd extends Command {
 	public static description = stripIndent`
-		Add an environment or config variable to an application, device or service.
+		Add an environment or config variable to one or more applications, devices or services.
 
-		Add an environment or config variable to an application, device or service,
-		as selected by the respective command-line options. Either the --application
-		or the --device option must be provided, and either may be be used alongside
-		the --service option to define a service-specific variable. (A service is an
-		application container in a "microservices" application.) When the --service
-		option is used in conjunction with the --device option, the service variable
-		applies to the selected device only. Otherwise, it applies to all devices of
-		the selected application (i.e., the application's fleet). If the --service
-		option is omitted, the variable applies to all services.
+		Add an environment or config variable to one or more applications, devices 
+		or services, as selected by the respective command-line options. Either the 
+		--application or the --device option must be provided, and either may be be 
+		used alongside the --service option to define a service-specific variable. 
+		(A service is an application container in a "microservices" application.) 
+		When the --service option is used in conjunction with the --device option, 
+		the service variable applies to the selected device only. Otherwise, it 
+		applies to all devices of the selected application (i.e., the application's 
+		fleet). If the --service option is omitted, the variable applies to all 
+		services.
 
 		If VALUE is omitted, the CLI will attempt to use the value of the environment
 		variable of same name in the CLI process' environment. In this case, a warning
@@ -67,9 +68,13 @@ export default class EnvAddCmd extends Command {
 	public static examples = [
 		'$ balena env add TERM --application MyApp',
 		'$ balena env add EDITOR vim --application MyApp',
+		'$ balena env add EDITOR vim --application MyApp,MyApp2',
 		'$ balena env add EDITOR vim --application MyApp --service MyService',
+		'$ balena env add EDITOR vim --application MyApp,MyApp2 --service MyService,MyService2',
 		'$ balena env add EDITOR vim --device 7cf02a6',
+		'$ balena env add EDITOR vim --device 7cf02a6,d6f1433',
 		'$ balena env add EDITOR vim --device 7cf02a6 --service MyService',
+		'$ balena env add EDITOR vim --device 7cf02a6,d6f1433 --service MyService,MyService2',
 	];
 
 	public static args = [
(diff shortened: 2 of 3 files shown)


Page-:
@balena-ci retest

lib/actions-oclif/env/add.ts
```diff
@@ -147,17 +150,31 @@ export default class EnvAddCmd extends Command {
...
-			);
+			for (const app of options.application.split(',')) {
+				try {
+					await balena.models.application[varType].set(
+						app,
+						params.name,
+						params.value,
+					);
+				} catch (err) {
+					console.info(`${err.message}, app: ${app}`);
```
pdcastro, line 161: `console.error(...)`

lib/actions-oclif/env/add.ts
```diff
@@ -171,31 +188,36 @@ async function setServiceVars(
...
-			options.service!,
-		);
-		await sdk.models.service.var.set(serviceId, params.name, params.value!);
-	} else {
+		for (const app of options.application.split(',')) {
+			try {
+				const serviceId = await getServiceIdForApp(sdk, app, options.service!);
+				await sdk.models.service.var.set(serviceId, params.name, params.value!);
+			} catch (err) {
+				console.info(`${err.message}, application: ${app}`);
```
pdcastro, line 196: `console.error(...)`

doc/cli.markdown
```diff
@@ -983,8 +984,10 @@ Examples:
 
 	$ balena env add TERM --application MyApp
 	$ balena env add EDITOR vim --application MyApp
+	$ balena env add EDITOR vim --application MyApp,MyApp2
 	$ balena env add EDITOR vim --application MyApp --service MyService
```
pdcastro, line 988: What about allowing multiple services to be listed as well? -
```
$ balena env add EDITOR vim --application MyApp,MyApp2 --service MyService,MyService2
```

lib/actions-oclif/env/add.ts
```diff
@@ -171,31 +188,36 @@ async function setServiceVars(
...
-		const serviceId = await getServiceIdForApp(
-			sdk,
-			options.application,
-			options.service!,
-		);
-		await sdk.models.service.var.set(serviceId, params.name, params.value!);
-	} else {
+		for (const app of options.application.split(',')) {
+			try {
+				const serviceId = await getServiceIdForApp(sdk, app, options.service!);
```
pdcastro, line 193: If a comma-separated list of service names was allowed, then there could be an inner loop here over `options.service.split(',')`.

lib/actions-oclif/env/add.ts
```diff
@@ -171,31 +188,36 @@ async function setServiceVars(
...
+			const [device, app] = await getDeviceAndAppFromUUID(
+				sdk,
+				uuid,
+				['id'],
+				['app_name'],
+			);
+			const serviceId = await getServiceIdForApp(
+				sdk,
+				app.app_name,
+				options.service!,
```
pdcastro, line 212: And another inner loop here (over `options.service`).

doc/cli.markdown
nwneisen: I think this makes sense. I'll add it in.

lib/actions-oclif/env/add.ts
```diff
@@ -171,31 +190,40 @@ async function setServiceVars(
...
-		);
-		await sdk.models.device.serviceVar.set(
-			device.id,
-			serviceId,
-			params.name,
-			params.value!,
-		);
+		for (const uuid of options.device.split(',')) {
+			for (const service of options.service!.split(',')) {
+				const [device, app] = await getDeviceAndAppFromUUID(
```
pdcastro, line 212: It looks like this call can be moved to the outer loop, it doesn't need to be re-executed for each service:
```
for (const uuid of options.device.split(',')) {

    const [device, app] = await getDeviceAndAppFromUUID(...);

    for (const service of options.service!.split(',')) {
```

Also, the other loops elsewhere in this PR have try / catch / process.exitCode blocks, but these loops here seem to be missing it.

lib/actions-oclif/env/add.ts
```diff
@@ -171,31 +190,49 @@ async function setServiceVars(
...
+		for (const uuid of options.device.split(',')) {
+			try {
+				const [device, app] = await getDeviceAndAppFromUUID(
+					sdk,
+					uuid,
+					['id'],
+					['app_name'],
+				);
+				for (const service of options.service!.split(',')) {
+					const serviceId = await getServiceIdForApp(
```
pdcastro, line 219: I think it is missing a `try`/`catch`/`process.exitCode = 1` in the inner loop. The outer try/catch would catch any errors, but:
* The service name list comes from the command line, and the user could have mistyped one of the services, e.g. `--service TypoService,GoodService`. As coded, `GoodService` would not be processed as the inner loop breaks when processing `TypoService` (which would throw an error at this line).
* The overall approach in this PR is that `console.error` is printed for items that went wrong, but nothing is printed for items that succeed. Therefore, the user might incorrectly assume that `GoodService` succeeded when in fact it was not even attempted.

If a separate try/catch block is added for the inner loop, then the outer try/catch block could be reduced in scope to include only the call for `getDeviceAndAppFromUUID` (and not the inner loop). This should help a bit with readability.

nwneisen: I had to leave the inner loop within the `try` block. Leaving `[device, app]` as `const` makes them unavailable in the inner loop and the linter doesn't like them being declared using `var` or `let` because their value never changes.

Page-: I'm not sure what the code this refers to looks like but the pattern to expose something inside a try to an outer scope @nwneisen would be like
```ts
let x
try {
	x = ...;
} catch {
	x = ...; // or `throw ...` or deal with the fact x becomes `T | undefined` due to this branch
}

Comment (nwneisen) [reply lib/actions-oclif/env/add.ts]: @Page- Not sure how I missed trying this. I had to add an additional check for undefined variables but otherwise it is working.

File: lib/actions-oclif/env/add.ts [hunk 8]
```diff
@@ -171,31 +190,60 @@ async function setServiceVars(
...
+						);
+						if (device) {
+							await sdk.models.device.serviceVar.set(
+								device.id,
+								serviceId,
+								params.name,
+								params.value!,
+							);
+						}
+					}
```
Page-, line 240: Adding the `return` in the catch will avoid the need for the `if (app)` below and also allow the process to exit straight away
```suggestion
			} catch (err) {
				console.error(`${err.message}, device: ${uuid}`);
				process.exitCode = 1;
				return;
			}
			for (const service of options.service!.split(',')) {
				try {
					const serviceId = await getServiceIdForApp(
						sdk,
						app.app_name,
						service,
					);
					if (device) {
						await sdk.models.device.serviceVar.set(
							device.id,
							serviceId,
							params.name,
							params.value!,
						);
					}
```

nwneisen: I don't think we would want to return at this point. In a previous PR for these bulk commands, @pdcastro and I discussed what happens if there is a problem with one value in a list of devices, services, or apps. If we return due to an error with a value in the middle of the list, then the values before the error value will have been changed and the values after the error would not have been looked at. We decided it was better for the user to continue changing the remaining values and report the values that caused an error.

nwneisen: [Original discussion](https://github.com/balena-io/balena-cli/pull/1939#discussion_r453690066)

Page-: Oh right we're in a loop here, what about using a `continue` to skip just the current uuid that failed:
```suggestion
			} catch (err) {
				console.error(`${err.message}, device: ${uuid}`);
				process.exitCode = 1;
				continue;
			}
			for (const service of options.service!.split(',')) {
				try {
					const serviceId = await getServiceIdForApp(
						sdk,
						app.app_name,
						service,
					);
					if (device) {
						await sdk.models.device.serviceVar.set(
							device.id,
							serviceId,
							params.name,
							params.value!,
						);
					}
```

nwneisen: That would be fine with me, though the behavior is the same either way(based on previous testing). I think the `continue` would explicitly show the behavior when reading the code.

pdcastro: I believe the `continue` statement as suggested by Page is best practice ("almost a bug" not to use it), because there is no purpose in even entering the inner loop if the `device` and `app` variables are not successfully set being set by the outer loop. To me it feels "disrespectful of the CPU and the science of algorithm analysis" 😁  to enter a loop just to _repeatedly_ run the same tests that skip all code execution, as opposed to not entering the loop in the first place if it is practical to avoid it. I know, you would rightly argue that the "performance" difference is negligible in micro or nanoseconds, but it's a matter of principle and style! 😎 

Btw, Page's suggested snippet still leaves behind the `if (device)` test in the inner loop, but it looks redundant and should also be removed when the `continue` statement is added. As a bonus, this change would also improve legibility a little bit, and reduce some 4 lines of code.

nwneisen: I was thinking too much about the inner loop. This does make sense and for cleaner code.

pdcastro: Sorry for my mouthful / storm in a teacup. 😁   It looks better now with the `continue` statement. 👍
Thank you!

pdcastro:
@balena-ci retest

lib/actions-oclif/env/add.ts
nwneisen: No Worries! I was thinking about it wrong. Thanks for the correction! I completely agree after looking closer.

pdcastro:
@nwneisen, the build was failing for an unrelated reason: `Error: connect ECONNREFUSED 127.0.0.1:3000` in tests for the `balena login` command. PR #1972 just got merged, changing the tests (and feature code) to use any free port number instead of hardcoded port numbers 3000 or 8989. If you rebase this PR on master, at least that error should not happen again.

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

pdcastro:
Issued another `retest` because the tests did not appear to have triggered automatically following the rebase. 😞 
Almost there though... 🤞

PR #1964, opened 2020-07-28
Convert `config` commands to typescript, oclif

srlowe:
Change-type: patch


lib/actions-oclif/config/generate.ts
```diff
@@ -0,0 +1,239 @@
...
+			answers.deviceType = deviceType;
+			config = await generateApplicationConfig(
+				resource as Application,
+				answers,
+			);
+		}
+
+		// Output
+		if (options.output != null) {
+			await require('fs').promises.writeFile(
```
Page-, line 194: Can you make this
```suggestion
			const fs = await import('fs');
			await fs.promises.writeFile(
```
so that we benefit from type-checking (`require` returns `any` for `.ts` files)

lib/actions-oclif/config/generate.ts
```diff
@@ -0,0 +1,239 @@
...
+		});
+		answers.version = options.version;
+
+		// Generate config
+		const { generateDeviceConfig, generateApplicationConfig } = await import(
+			'../../utils/config'
+		);
+
+		let config;
+		if ((resource as Device).uuid != null) {
```
Page-, line 178: Does using
```suggestion
		if ('uuid' in resource && resource.uuid != null) {
```
avoid the need for casting?

lib/actions-oclif/config/generate.ts
```diff
@@ -0,0 +1,239 @@
...
+		help: cf.help,
+	};
+
+	public static authenticated = true;
+
+	public async run() {
+		const { flags: options } = this.parse<FlagsDef, {}>(ConfigGenerateCmd);
+
+		const balena = getBalenaSdk();
+		const { tryAsInteger } = await import('../../utils/validation');
```
Page-, line 130: Can you move this inside the single branch that uses it please

lib/actions-oclif/config/generate.ts
srlowe: Ah missed that.

lib/actions-oclif/config/generate.ts
srlowe: It does!

lib/actions-oclif/config/generate.ts
```diff
@@ -0,0 +1,237 @@
...
+
+		// Generate config
+		const { generateDeviceConfig, generateApplicationConfig } = await import(
+			'../../utils/config'
+		);
+
+		let config;
+		if ('uuid' in resource && resource.uuid != null) {
+			config = await generateDeviceConfig(
+				resource as Device & { belongs_to__application: PineDeferred },
```
Page-, line 180: Could you move this above as
```ts
resource = await balena.models.device.get(tryAsInteger(options.device)) as Device & { belongs_to__application: PineDeferred };
```
that way this can be just
```suggestion
				resource,
```
and the actual cast will happen at the point we fetch the device and make it easier to verify that the cast is valid, and also mean that all usage of the `Device` variant of `resource` will be typed more accurately

srlowe: Yeah that's a bit nicer.

PR #1967, opened 2020-07-30
Convert command `build` to typescript, oclif, and refactor

srlowe:
Change-type: patch


lib/actions-oclif/build.ts
```diff
@@ -0,0 +1,273 @@
...
+ * limitations under the License.
+ */
+
+import { flags } from '@oclif/command';
+import Command from '../command';
+import * as cf from '../utils/common-flags';
+import { getBalenaSdk } from '../utils/lazy';
+import * as compose from '../utils/compose';
+import DockerToolbelt = require('docker-toolbelt'); // https://github.com/balena-io-modules/docker-toolbelt/pull/40
+import Logger = require('../utils/logger');
```
srlowe, line 24: Any idea how to remove the need for this line?  We just need the type imported.  The logger import is now done dynamically in the base class. (See below)

lib/actions-oclif/build.ts
```diff
@@ -0,0 +1,271 @@
...
+
+		await Command.checkLoggedInIf(!!options.application);
+
+		// compositions with many services trigger misleading warnings
+		// @ts-ignore editing property that isn't typed but does exist
+		(await import('events')).defaultMaxListeners = 1000;
+
+		const sdk = getBalenaSdk();
+
+		const logger = await Command.getLogger();
```
srlowe, line 117: *logger

lib/actions-oclif/build.ts
```diff
@@ -0,0 +1,271 @@
...
+	 *   buildOpts: arguments to forward to docker build command
+	 *
+	 * @param {import('docker-toolbelt')} docker
+	 * @param {import('../utils/logger')} logger
+	 * @param {import('../utils/compose-types').ComposeOpts} composeOpts
+	 * @param {any} opts
+	 */
+	protected async buildProject(
+		docker: DockerToolbelt,
+		logger: Logger,
```
srlowe, line 228: `Logger` type import required because of this.

lib/command.ts
```diff
@@ -93,6 +107,13 @@ export default abstract class BalenaCommand extends Command {
 		this.stdin = await (await import('get-stdin'))();
 	}
 
+	/**
+	 * Get a logger instance.
+	 */
+	protected static async getLogger() {
+		return (await import('./utils/logger')).getLogger();
```
srlowe, line 114: Logger now imported here.

srlowe:
@Page- @pdcastro The spinner and progress indicator don't seem to be functioning after these changes - do you have any insight into that? (I'm trying to diagnose now).

lib/actions-oclif/build.ts
Page-: The alternative to a module level `import type` is:
```suggestion
		logger: import('../utils/logger'),
```

lib/actions-oclif/build.ts
Page-: To make it a type-only import here you'd to change the logger export to be es module compatible so you can then do eg
```
import type { Logger } from '../utils/logger'
```

lib/actions-oclif/build.ts
```diff
@@ -0,0 +1,271 @@
...
+	 * Opts must be an object with the following keys:
+	 *   app: the app this build is for (optional)
+	 *   arch: the architecture to build for
+	 *   deviceType: the device type to build for
+	 *   buildEmulated
+	 *   buildOpts: arguments to forward to docker build command
+	 *
+	 * @param {import('docker-toolbelt')} docker
+	 * @param {import('../utils/logger')} logger
+	 * @param {import('../utils/compose-types').ComposeOpts} composeOpts
```
Page-, line 223: These can now be changed to refer to typescript types/type only imports rather than needing this in-lined typescript

srlowe: Yeah I did change the logger module to have the class a s default export, which fixes that problem, but then breaks a load of other stuff.  Maybe best to try to do that in a seperate PR.

srlowe: Ah yeah, will remove those jsdoc imports.

srlowe: > The alternative to a module level import type is:

Perhaps we can grab the Logger type from the base Command class (in a way that doesn't affect performance if logger is not used)?

lib/actions-oclif/build.ts
srlowe: Yeah I did change the logger module to have the class a s default export, which fixes that problem, but then breaks a load of other stuff. Maybe best to try to do that in a separate PR.

lib/actions-oclif/build.ts
srlowe: Thanks.  Perhaps we can grab the Logger type from the base Command class (in a way that doesn't affect performance if logger is not used)?

lib/actions-oclif/build.ts
srlowe: Right, will do.

lib/actions-oclif/build.ts
Page-: This import is purely in typings so will only exist for the typescript compiler - at runtime none of the type info will exist (even if it's a non-type only module level import it'll probably still be elided if it's only used for types, but I prefer the unambiguous ways of doing it because it avoids accidents)

srlowe:
> @Page- @pdcastro The spinner and progress indicator don't seem to be functioning after these changes - do you have any insight into that? (I'm trying to diagnose now).

It seems this was already the case with master (I was comparing with an older version), so promoting to full PR.  I'm not sure when the spinner/progress problem got introduced.

Issue #1773, opened 2020-04-27
'balena login' web authorization fails silently if another process is listening on port 8989


pdcastro:
As originally reported in #1765, if another (unrelated, long-lived) process is already listening on port 8989, the web authorization option of `balena login` fails silently. Well, the web browser may eventually display a browser-specific message like "this site can't be reached", but there is no indication to the user of what the problem might be.

More specifically, in #1765, the issue was that the pre-existing long lived process was listening on `127.0.0.1:8989` (IPv4 only), whereas the CLI was listening on `*:8989` (any IPv4 or IPv6 address).

I am not aware of a reason why the CLI should prefer a fixed port number (8989) over a random unallocated port number (which the OS should be able to offer from an unallocated pool). If there is no good reason, the implementation should probably be modified that way. Otherwise, the CLI should add a check on whether `127.0.0.1:8989` specifically is already taken, and if so, fail immediately with an error.

pdcastro:
Connects-to: #1683 [`balena login` using web auth fails silently on a machine with no display attached](https://github.com/balena-io/balena-cli/issues/1683)

pdcastro:
From the forums: ([comment link](https://forums.balena.io/t/balena-cli-version-11-35-16-wont-login-on-on-window-10/125395/6))
> It turns out that: https://sonarr.tv/ runs on that port [8989].
> I shut it down and the login process worked without issue.

PR #1972, opened 2020-08-01
login: Use any free port number instead of 8989 for web authentication

pdcastro:
In addition to addressing issue #1773, this PR should avoid the CI tests errors we've been occasionally getting about TCP port 3000 being already in use.

Resolves: #1773 
Change-type: minor


pdcastro:
@balena-ci rebase

lib/actions-oclif/login.ts
```diff
@@ -101,6 +105,12 @@ export default class LoginCmd extends Command {
 			description: 'password',
 			dependsOn: ['credentials'],
 		}),
+		port: flags.integer({
+			char: 'P',
+			description:
+				'TCP port number of local HTTP login server (--web auth only)',
```
pdcastro, line 111: The objective of this PR is to allow the OS to automatically allocate a free TCP port number for the login server, instead of using the pre-defined port number 8989. I am not aware of reasons why this change would affect anyone, or indeed reasons why anyone would want to specify `--port 8989` on the command line. However, just in case someone decided to complain that:
> _"I was relying on the login server running on port 8989 because of X, now it doesn't work anymore!"_ 

our answer could then be along the lines of:
> _"I am sorry to hear that. You can preserve the old behaviour with the command line `balena login -w -P 8989`"_

lib/auth/server.ts
```diff
@@ -15,75 +15,63 @@ limitations under the License.
...
- * any reason. The only reliable way around it seems to be to explicitly unref the
- * sockets, so the event loop stops waiting for it. See:
- * https://github.com/nodejs/node/issues/2642
- * https://github.com/nodejs/node-v0.x-archive/issues/9066
- */
-export function shutdownServer() {
-	serverSockets.forEach((s) => s.unref());
-	serverSockets.splice(0);
-}
+export class LoginServer extends EventEmitter {
```
pdcastro, line 26: Previously there wasn't a class, just the functions `createServer()` and `awaitForToken()` in `server.ts`, where `awaitForToken` would call `createServer`.

`createServer` expected to be told what port number to use, which was hardcoded elsewhere as 8989. Code elsewhere also printed a URL to the console (in case of web auth), which I think is nice (allows me to copy and paste to the right browser window, when I have multiple browsers/profiles logged in to different balena user accounts) and which was actually a user request in a past CLI issue:
```
$ ./bin/balena login
...
Logging in to balena-cloud.com
? How would you like to login? Web authorization (recommended)
Opening web browser for URL:
https://dashboard.balena-cloud.com/login/cli/http%253A%252F%252F127.0.0.1%253A56628%252Fauth
```

So `createServer` or `awaitForToken` had to return the port number that was allocated to the OS, but `awaitForToken` was an async function that awaited for the user to click a button on the web browser (and the URL containing the port number had to be printed before the web browser was opened). Anyway, a refactoring was needed and I that's how I eventually came up with LoginServer class as an EventEmitter.

The majority of the code is reorganised copy-and-paste.

package.json
```diff
@@ -57,11 +57,11 @@
     "release": "ts-node --transpile-only automation/run.ts release",
     "pretest": "npm run build",
     "test": "npm run test:source && npm run test:standalone",
-    "test:source": "cross-env BALENA_CLI_TEST_TYPE=source mocha --timeout 6000 -r ts-node/register/transpile-only \"tests/**/*.spec.ts\"",
+    "test:source": "cross-env BALENA_CLI_TEST_TYPE=source mocha --timeout 12000 -r ts-node/register/transpile-only \"tests/**/*.spec.ts\"",
```
pdcastro, line 60: Increasing the timeout from 6s to 12s in a separate commit. I don't know if this will actually solve the relatively frequent timeout errors in CI builds in recent PRs; I though it was worth trying.

.github/ISSUE_TEMPLATE.md
```diff
@@ -64,6 +64,7 @@ fixed it.
 # Specifications
 
 - **balena CLI version:** e.g. 1.2.3 (output of the `"balena version -a"` command)
+- **Cloud backend: openBalena or balenaCloud?** If unsure, it will be balenaCloud
```
pdcastro, line 67: This change is in a separate commit. Incidental to this PR.

lib/actions-oclif/login.ts
```diff
@@ -166,7 +171,7 @@ ${messages.reachingOut}`);
 		// Web
 		else if (loginOptions.web) {
 			const auth = await import('../auth');
-			await auth.login();
+			await auth.login({ port: loginOptions.port || 0 });
```
Page-, line 174: Is this `|| 0` actually necessary? Also is it better as `?? 0` if you're specifically targetting nullish

lib/auth/server.ts
```diff
@@ -93,18 +81,69 @@ export const awaitForToken = (options: {
...
+		// Setting `server.timeout` to some value like 3 seconds works, but then the CLI
+		// process hangs for "only" 3 seconds (not good enough). Reducing the timeout to 1
+		// second may cause authentication failure if the laptop or CI server are slow for
+		// any reason. The only reliable way around it seems to be to explicitly unref the
+		// sockets, so the event loop stops waiting for it. See:
+		// https://github.com/nodejs/node/issues/2642
+		// https://github.com/nodejs/node-v0.x-archive/issues/9066
+		//
+		this.serverSockets.forEach((s) => s.unref());
+		this.serverSockets.splice(0);
```
Page-, line 128: You can replace all the code for this with https://www.npmjs.com/package/server-destroy

lib/actions-oclif/login.ts
pdcastro: Removed -- it was not necessary. 👍

lib/auth/server.ts
pdcastro: The `server-destroy` package calls `socket.destroy()` on each socket, whereas this code calls `socket.unref()`. I am almost sure that I tried calling `socket.destroy()` and found that it was not useful / sufficient to fix issue #1404 (months ago), whereas calling `unref()` did the trick. You're welcome to try it. :-)
Also, _"all the code for this"_ is about 3 lines of code (excluding comments), and adding a dependency to `package.json` and importing and calling it would also be about 3 lines of code...

PR #1974, opened 2020-08-03
Update balena-sdk to 15.x

Page-:


package.json
```diff
@@ -229,7 +229,7 @@
     "js-yaml": "^3.13.1",
     "klaw": "^3.0.0",
     "livepush": "^3.5.0",
-    "lodash": "^4.17.15",
+    "lodash": "^4.17.19",
```
Page-, line 232: Updated this to dedupe deps but it also includes a security issue, I do not believe the security issue affects us but there's no reason not to update

package.json
```diff
@@ -187,17 +187,17 @@
...
     "balena-release": "^3.0.0",
-    "balena-sdk": "^14.8.0",
-    "balena-semver": "^2.2.0",
+    "balena-sdk": "^15.2.1",
+    "balena-semver": "^2.3.0",
     "balena-settings-client": "^4.0.5",
-    "balena-settings-storage": "^6.0.0",
-    "balena-sync": "^11.0.0",
+    "balena-settings-storage": "^6.0.1",
+    "balena-sync": "^11.0.2",
```
Page-, line 200: Updated to a version that uses sdk 15.x to be able to dedupe

package.json
```diff
@@ -187,17 +187,17 @@
...
     "archiver": "^3.1.1",
-    "balena-config-json": "^4.0.0",
+    "balena-config-json": "^4.1.0",
     "balena-device-init": "^5.0.2",
-    "balena-errors": "^4.3.0",
-    "balena-image-manager": "^7.0.1",
-    "balena-preload": "^10.2.0",
+    "balena-errors": "^4.4.1",
+    "balena-image-manager": "^7.0.3",
+    "balena-preload": "^10.2.3",
```
Page-, line 194: Updated to a version that uses sdk 15.x to be able to dedupe

package.json
```diff
@@ -187,17 +187,17 @@
...
     "JSONStream": "^1.0.3",
     "archiver": "^3.1.1",
-    "balena-config-json": "^4.0.0",
+    "balena-config-json": "^4.1.0",
     "balena-device-init": "^5.0.2",
-    "balena-errors": "^4.3.0",
-    "balena-image-manager": "^7.0.1",
-    "balena-preload": "^10.2.0",
+    "balena-errors": "^4.4.1",
+    "balena-image-manager": "^7.0.3",
```
Page-, line 193: Updated to a version that uses sdk 15.x to be able to dedupe

package.json
```diff
@@ -187,17 +187,17 @@
     "@zeit/dockerignore": "0.0.3",
     "JSONStream": "^1.0.3",
     "archiver": "^3.1.1",
-    "balena-config-json": "^4.0.0",
+    "balena-config-json": "^4.1.0",
     "balena-device-init": "^5.0.2",
-    "balena-errors": "^4.3.0",
-    "balena-image-manager": "^7.0.1",
-    "balena-preload": "^10.2.0",
+    "balena-errors": "^4.4.1",
```
Page-, line 192: Updated to dedupe with other packages

package.json
```diff
@@ -187,17 +187,17 @@
     "@zeit/dockerignore": "0.0.3",
     "JSONStream": "^1.0.3",
     "archiver": "^3.1.1",
-    "balena-config-json": "^4.0.0",
+    "balena-config-json": "^4.1.0",
```
Page-, line 190: Updated to a version that uses sdk 15.x to be able to dedupe

package.json
```diff
@@ -187,17 +187,17 @@
...
+    "balena-preload": "^10.2.3",
     "balena-release": "^3.0.0",
-    "balena-sdk": "^14.8.0",
-    "balena-semver": "^2.2.0",
+    "balena-sdk": "^15.2.1",
+    "balena-semver": "^2.3.0",
     "balena-settings-client": "^4.0.5",
-    "balena-settings-storage": "^6.0.0",
-    "balena-sync": "^11.0.0",
+    "balena-settings-storage": "^6.0.1",
```
Page-, line 199: Small fix to add a missing dep, not strictly required but will avoid any issues if for some reason we don't install the dep/install a different version of it

package.json
```diff
@@ -262,7 +262,7 @@
     "tar-utils": "^2.1.0",
     "through2": "^2.0.3",
     "tmp": "^0.2.1",
-    "typed-error": "^3.2.0",
+    "typed-error": "^3.2.1",
```
Page-, line 265: Updates its dev deps, just doing the safe update to reduce the noise from `npm outdated`

package.json
```diff
@@ -127,7 +127,7 @@
     "@types/is-root": "^2.1.2",
     "@types/js-yaml": "^3.12.3",
     "@types/klaw": "^3.0.1",
-    "@types/lodash": "^4.14.150",
+    "@types/lodash": "^4.14.159",
```
Page-, line 130: Updated in line with the main package

Page-:
Main code changes are to handle the fact that updated sdk/deps now return native promises instead of bluebird

package.json
```diff
@@ -187,17 +187,17 @@
...
     "archiver": "^3.1.1",
-    "balena-config-json": "^4.0.0",
+    "balena-config-json": "^4.1.0",
     "balena-device-init": "^5.0.2",
-    "balena-errors": "^4.3.0",
-    "balena-image-manager": "^7.0.1",
-    "balena-preload": "^10.2.0",
+    "balena-errors": "^4.4.1",
+    "balena-image-manager": "^7.0.3",
+    "balena-preload": "^10.2.4",
```
Page-, line 194: Updated to a version that uses sdk 15.x to be able to dedupe

tests/auth/tokens.ts
```diff
@@ -1,20 +1,39 @@
...
 export default {
 	johndoe: {
-		token:
-			'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImpvaG5kb2UxIiwiZW1haWwiOiJqb2huZG9lQGpvaG5kb2UuY29tIiwiZ2l0bGFiX2lkIjoxMzI1LCJzb2NpYWxfc2VydmljZV9hY2NvdW50IjpudWxsLCJoYXNQYXNzd29yZFNldCI6dHJ1ZSwibmVlZHNQYXNzd29yZFJlc2V0IjpmYWxzZSwicHVibGljX2tleSI6ZmFsc2UsImZlYXR1cmVzIjpbXSwiaWQiOjEzNDQsImludGVyY29tVXNlckhhc2giOiJlMDM3NzhkZDI5ZTE1NzQ0NWYyNzJhY2M5MjExNzBjZjI4MTBiNjJmNTAyNjQ1MjY1Y2MzNDlkNmRlZGEzNTI0IiwicGVybWlzc2lvbnMiOltdLCJpYXQiOjE0MjY3ODMzMTJ9.v5bmh9HwyUZu8zhh1rA79mTL-1jzDOO8eUr_lVaBwhg',
-		data: {
-			email: 'johndoe@johndoe.com',
-			username: 'johndoe1',
-			id: 1344,
-		},
+		token: jsonwebtoken.sign(johnDoe, 'very-secret'),
```
Page-, line 34: Due to a bug fix in balena-request the token refresh now works properly in cases it was previously broken and would not be called, to match the previous behavior for testing I've switched these tokens to be generated on the fly so they are new and do not trigger a refresh (previous tokens had an `iat` some time in 2015)

Page-:
@balena-ci rebase

Page-:
@balena-ci retest

package.json
```diff
@@ -187,17 +189,17 @@
...
     "balena-release": "^3.0.0",
-    "balena-sdk": "^14.8.0",
-    "balena-semver": "^2.2.0",
+    "balena-sdk": "^15.2.3",
```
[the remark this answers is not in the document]
Page-: Not by default, it would only cause it in the case that a submodule updates balena-sdk and we update that submodule without also updating the top level balena-sdk and/or running npm dedupe.

lib/auth/utils.ts
```diff
@@ -63,30 +63,27 @@ export const getDashboardLoginURL = (callbackUrl: string) => {
...
-				.loginWithToken(token)
-				.return(token)
-				.then(balena.auth.isLoggedIn)
-				.tap((isLoggedIn) => {
-					if (isLoggedIn) {
-						return;
-					}
+	let currentToken;
+	try {
+		currentToken = await balena.auth.getToken();
```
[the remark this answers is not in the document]
Page-: It can't be moved there because
```ts
await balena.auth.loginWithToken(token);
```
will switch to the new token and the "current"/previous token will be lost at that point

PR #1980, opened 2020-08-04
Dont set the preloader config device type to intel-nuc

zvin:
Change-type: patch


lib/actions/preload.js
```diff
@@ -221,7 +221,6 @@ async function prepareAndPreload(preloader, balenaSdk, options) {
 	const { ExpectedError } = require('../errors');
 
 	await preloader.prepare();
-	preloader.config = { deviceType: 'intel-nuc' };
```
pdcastro, line 224: Ooops, my bad. Yes, leftover from debugging. 🤦

PR #1982, opened 2020-08-05
Allow rebasing of npm-shrinkwrap

Page-:
`@balena-ci rebase` already complete ignores this attempt to block rebasing and all it does for humans is cause additional red tape and frustration and end up with weird shrinkwraps if the steps are missed at all. If you want the npm shrinkwrap to be generated in a specific way then handle it automatically via a git hook or add a CI step to verify it's as you want - this approach does not work and just wastes time


pdcastro:
> `(at)balena-ci rebase` already complete ignores this attempt to block rebasing

Really? Does it not use `git` for the rebase, is the `.gitattributes` file ignored? I had assumed that in the event of a merge being required for `npm-shrinkwrap.json`, the CI auto rebase attempt would fail and simply not take place.

> this approach does not work and just wastes time

@Page-, check this comment of yours a few months ago, in the PR that modified the `.gitattributes` file:
> https://github.com/balena-io/balena-cli/pull/1653#discussion_r391007578
> _I think it makes sense to avoid auto-merges for the shrinkwrap since they could be fairly dangerous, especially as npm doesn't seem to care if the shrinkwrap makes sense or not and just installs it regardless (see the junk in net-keepalives shrinkwrap that gets installed even though it's not in the package.json for the module)_

It is supposed to be an improvement to the CLI, preventing broken installations (thus improving user experience) and also saving developer time having to debug what went wrong with an incorrectly updated (merged) shrinkwrap file (iirc, it happened at least once before). Another motivation to add this auto rebase protection was to prevent damages to the great effort you had put a few months ago in tidying up the CLI's shrinkwrap file, to optimise dependencies, avoid duplicates, save disk space and probably improve performance too.

However if the original reasoning was incorrect / incomplete and git merge of the shrinkwrap file is actually safe and OK, then yes we should remove the restriction.

> then handle it automatically via a git hook or add a CI step to verify it's as you want 

All in favour of automating these things! 👍   I've just found this [npm-merge-driver](https://www.npmjs.com/package/npm-merge-driver) that configures git to automatically run `npm install --package-lock-only`, but it seems to rely on a change to `.git/config`, which is local to each cloned repo, isn't? Ideally the solution would be committed to the repo and not require additional manual steps to enable it (though a `postinstall` npm script would probably be OK - in fact it is suggested in this [npm-merge-driver issue](https://github.com/npm/npm-merge-driver/issues/17#issuecomment-626815527)). Git hooks are also local to each cloned repo, but perhaps [husky](https://www.npmjs.com/package/husky) can help?

Do you remember where the `(at)balena-ci rebase` code is located? We should consider what the CI is doing and test that any hooks solution will be used by the CI too.

Page-:
>> `(at)balena-ci rebase` already complete ignores this attempt to block rebasing
>
> Really? Does it not use `git` for the rebase, is the `.gitattributes` file ignored? I had assumed that in the event of a merge being required for `npm-shrinkwrap.json`, the CI auto rebase attempt would fail and simply not take place.

Honestly I have no idea how it ignores it but I've never once had it complain about npm-shrinkwrap even when I'm pretty certain a local rebase would have caused it.

> @Page-, check this comment of yours a few months ago, in the PR that modified the `.gitattributes` file:

Yep, I still think junk in the shrinkwrap is a problem but I've realized that this is not the solution, between the nightmare it makes of rebasing it also doesn't actually prevent that junk getting in - in some ways it makes it more likely because having to regenerate the shrinkwrap so often means it's less likely to be paid attention/more likely to be rushed.

> Do you remember where the `(at)balena-ci rebase` code is located? We should consider what the CI is doing and test that any hooks solution will be used by the CI too.

I've never seen the code but my assumption is that it wouldn't run `npm install` at all and so wouldn't run hooks

package.json
```diff
@@ -57,7 +57,8 @@
     "package": "npm run build:fast && npm run build:standalone && npm run build:installer",
     "release": "ts-node --transpile-only automation/run.ts release",
     "pretest": "npm run build",
-    "test": "npm run test:source && npm run test:standalone",
+    "test": "npm test:shrinkwrap && npm run test:source && npm run test:standalone",
+    "test:shrinkwrap": "npm i && npm dedupe && npm i && npm run catch-uncommitted",
```
Page-, line 61: I've added this which should help to keep junk out of the shrinkwrap by having the CI check

pdcastro: But `npm run catch-uncommitted` specifically ignores `npm-shrinkwrap.json` -- I don't remember why, maybe you do:
* `"catch-uncommitted": "ts-node --transpile-only automation/run.ts catch-uncommitted"`
in `package.json`
* `npx catch-uncommitted --catch-no-git --skip-node-versionbot-changes --ignore-space-at-eol`
at https://github.com/balena-io/balena-cli/blob/v12.12.1/automation/build-bin.ts#L310-L314
* `":(exclude)npm-shrinkwrap.json"`
https://github.com/balena-io-modules/catch-uncommitted/blob/v1.5.0/catch-uncommitted.sh#L16

pdcastro: Oh, I think the reason why `catch-uncommitted` ignores the `npm-shrinkwrap.json` file is related to the fact that the file includes the CLI version number, which versionbot automatically updates alongside the Changelog file. Then again, the change is committed to repo, or is it? If it was committed, then `catch-uncommitted` should be OK with it...

.gitattributes
```diff
@@ -10,6 +10,3 @@ doc/cli.markdown text eol=lf
 # crlf for the eol conversion test files
 tests/test-data/projects/docker-compose/basic/service2/file2-crlf.sh eol=crlf
 tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh eol=crlf
-
-# Prevent auto merging of the npm-shrinkwrap.json file: see notes in CONTRIBUTING.md
-/npm-shrinkwrap.json merge=binary
```
pdcastro, line 15: [As discussed in flowdock](https://www.flowdock.com/app/rulemotion/d-build-pipelines/threads/UbtLhou_pLyZQENwZMbWXL5cywQ), we have learned that balena CI's rebase script is this [rebase.sh](https://github.com/product-os/scripts/blob/master/meta/rebase.sh), which uses plain `git fetch`, `git checkout` and `git rebase` to do the rebase, and takes the `.gitattributes` file into account so that the `npm-shrinkwrap.json` will not be merged automatically. Here's a sample output of the CI's rebase job, for a test pull request (#1984):

https://ci.balena-dev.com/teams/main/pipelines/github-events-balena-io/jobs/rebase-event/builds/264
```
Using index info to reconstruct a base tree...
M    npm-shrinkwrap.json
Falling back to patching base and 3-way merge...
warning: Cannot merge binary files: npm-shrinkwrap.json (HEAD vs. test-rebase)
Auto-merging npm-shrinkwrap.json
CONFLICT (content): Merge conflict in npm-shrinkwrap.json
error: Failed to merge in the changes.
Patch failed at 0001 test-rebase
hint: Use 'git am --show-current-patch' to see the failed patch
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
```
When the auto merge of `npm-shrinkwrap.json` fails as above (because of the `.gittattributes` file), the `(at)balena-ci rebase` command fails and the PR is not automatically rebased.

package.json
Page-: Reworked this so it should now work

package.json
```diff
@@ -58,7 +58,8 @@
     "package": "npm run build:fast && npm run build:standalone && npm run build:installer",
     "release": "ts-node --transpile-only automation/run.ts release",
     "pretest": "npm run build",
-    "test": "npm run test:source && npm run test:standalone",
+    "test": "npm test:shrinkwrap && npm run test:source && npm run test:standalone",
```
pdcastro, line 61: `npm test:shrinkwrap` → `npm run test:shrinkwrap`

automation/test-lock-deduplicated.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
pdcastro, line 1: This file needs executable permissions:
```
$ npm test
...
Error: "test-shrinkwrap": Error: '/Users/paulo/balena/balena-io/balena-cli/automation/test-lock-deduplicated.sh' program not found. Is it installed?
```

Page-: urg right, git and windows fun

Page-: it should be executable now

npm-shrinkwrap.json
```diff
@@ -8898,6 +8898,59 @@
         "strict-event-emitter-types": "^2.0.0",
         "ts-node": "^8.6.2",
         "typed-error": "^3.2.0"
+      },
+      "dependencies": {
```
pdcastro, line 8902: Interesting that this PR includes changes to `npm-shrinkwrap.json`, given that it does not make changes to dependencies in `package.json`. As if this PR was fixing the master branch's shrinkwrap, but running `npm i && npm dedupe && npm i` on the master branch does not produce any changes. 🤔

Page-: Yeah this makes no sense, they don't get deduped and when I reset the changes and rerun those commands they don't come back, maybe I need to throw a `prune` in

Page-: An npm prune doesn't solve it either :S

PR #1983, opened 2020-08-06
Tests: add verification of 'pkg' warnings against saved output

pdcastro:
Change-type: patch

When `pkg` runs, it produces warnings such as:
```
> Warning Cannot resolve ''./' + target'
  /Users/paulo/balena/balena-io/balena-cli/node_modules/balena-sync/build/sync/index.js
  Dynamic require may fail at run time, because the requested file
  is unknown at compilation time and not included into executable.
  Use a string literal as an argument for 'require', or leave it
  as is and specify the resolved file name in 'scripts' option.
> Warning Cannot include file %1 into executable.
  %1: node_modules/open/xdg-open
> Warning Cannot include file %1 into executable.
  %1: node_modules/open/xdg-open
> Warning Cannot include file %1 into executable.
  %1: node_modules/drivelist/build/Release/drivelist.node
  %2: path-to-executable/drivelist.node
...
```
They are usually harmless, until they aren't: when changes are made to dependencies and `pkg` produces new warnings for actually problematic issues. This PR commits the known harmless output to a text file, then compares the `pkg` output against it. If the output doesn't match, a diff is printed and the CLI build fails.

### DIFF
--- a/automation/utils.ts
+++ b/automation/utils.ts
@@ -23,6 +23,69 @@ import * as shellEscape from 'shell-escape';
 export const MSYS2_BASH = 'C:\\msys64\\usr\\bin\\bash.exe';
 export const ROOT = path.join(__dirname, '..');
 
+/** Tap and buffer this process' stdout and stderr */
+export class StdOutTap {
+	public stdoutBuf: string[] = [];
+	public stderrBuf: string[] = [];
+	public allBuf: string[] = []; // both stdout and stderr
+
+	protected origStdoutWrite: typeof process.stdout.write;
+	protected origStderrWrite: typeof process.stdout.write;
+
+	constructor(protected printDots = false) {}
+
+	tap() {
+		this.origStdoutWrite = process.stdout.write;
+		this.origStderrWrite = process.stderr.write;
+
+		process.stdout.write = (chunk: string, ...args: any[]): boolean => {
+			this.stdoutBuf.push(chunk);
+			this.allBuf.push(chunk);
+			const str = this.printDots ? '.' : chunk;
+			return this.origStdoutWrite.call(process.stdout, str, ...args);
+		};
+
+		process.stderr.write = (chunk: string, ...args: any[]): boolean => {
+			this.stderrBuf.push(chunk);
+			this.allBuf.push(chunk);
+			const str = this.printDots ? '.' : chunk;
+			return this.origStderrWrite.call(process.stderr, str, ...args);
+		};
+	}
+
+	untap() {
+		process.stdout.write = this.origStdoutWrite;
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -156,6 +156,7 @@
     "chai-as-promised": "^7.1.1",
     "cross-env": "^7.0.2",
     "deep-object-diff": "^1.1.0",
+    "diff": "^4.0.2",
```
pdcastro, line 159: `diff` was actually already an indirect dependency found in the shrinkwrap file. This makes it a direct dev dependency.

automation/utils.ts
```diff
@@ -23,6 +23,69 @@ import * as shellEscape from 'shell-escape';
...
+/** Tap and buffer this process' stdout and stderr */
+export class StdOutTap {
+	public stdoutBuf: string[] = [];
+	public stderrBuf: string[] = [];
+	public allBuf: string[] = [];
+
+	protected origStdoutWrite: typeof process.stdout.write;
+	protected origStderrWrite: typeof process.stdout.write;
+
+	constructor(protected printDots = false) {}
```
pdcastro, line 35: This class captures `stdout` and `stderr` for comparison against `pkg`'s saved output. If `printDots` is `true`, then the output actually printed to the console is replaced with dots, like in the long line of dots below:
```
$ npm run build
...
Building standalone zip package for CLI v12.12.1
...
...............................................................................
```
The output capture is the real output rather than the dots, of course.
I thought that it would be nicer to print dots than the `pkg` warnings that make people think that something is wrong:
```
> Warning Cannot include file %1 into executable.
  %1: node_modules/drivelist/build/Release/drivelist.node
  %2: path-to-executable/drivelist.node
```
This is OK now that the known safe output is saved to a file and automatically checked.

srlowe: Replacing the known errors with dots is very nice.

srlowe:
@pdcastro What are the implications of a separate file per platform?  E.g. perhaps we would not see the error locally when we build, but might see one in CI for another platform?

pdcastro:
> What are the implications of a separate file per platform? E.g. perhaps we would not see the error locally when we build, but might see one in CI for another platform?

That's right, unless we run `npm test` locally for all platforms. Separate files are needed because each platform has a different set of native node modules, so `pkg` prints different warnings on each platform. Having a single file with hardcoded exceptions would still face this issue.

pdcastro:
> perhaps we would not see the error locally when we build

Also, `npm test` or `npm run build` don't run `pkg` anyway: the developer would have to explicitly run `npm run build:standalone`, which I don't think most developers do anyway, even for their development machine's platform.

PR #1989, opened 2020-08-11
Improve tests

Page-:
Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
   "mocha": {
     "reporter": "spec",
     "require": "ts-node/register/transpile-only",
+    "file": "./tests/config-tests",
     "timeout": 12000,
     "_": "tests/**/*.spec.ts"
   },
--- a/package.json
+++ b/package.json
@@ -59,11 +59,11 @@
     "release": "ts-node --transpile-only automation/run.ts release",
     "pretest": "npm run build",
     "test": "npm run test:source && npm run test:standalone",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -91,6 +91,13 @@
       "pre-commit": "node automation/check-npm-version.js && node automation/check-doc.js"
     }
   },
+  "mocha": {
+    "reporter": "spec",
+    "require": "ts-node/register/transpile-only",
+    "file": "./tests/config-tests",
+    "timeout": 12000,
+    "_": "tests/**/*.spec.ts"
```
Page-, line 99: These are just defaults, for instance this positional arg is still overridden in `test:only` above

tests/config-tests.ts
```diff
@@ -15,18 +15,21 @@
...
+setEsVersion('es2018');
 
-setMaxListeners(35); // it appears that 'nock' adds a bunch of listeners - bug?
+import * as tmp from 'tmp';
+tmp.setGracefulCleanup();
+// Use a temporary dir for tests data
+process.env.BALENARC_DATA_DIRECTORY = tmp.dirSync().name;
+
+import { EventEmitter } from 'events';
+EventEmitter.defaultMaxListeners = 35; // it appears that 'nock' adds a bunch of listeners - bug?
```
Page-, line 28: I changed this to editing `events` directly so we don't load a bunch of the cli in here since in general this file is to set up the test environment prior to actually running any tests/cli code

Page-:
@balena-ci retest

PR #1994, opened 2020-08-13
Update dependencies

Page-:
Update balena-sdk from 15.2.5 to 15.3.0

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


patches/all/@oclif+plugin-help+3.2.0.patch
```diff
@@ -1,37 +1,37 @@
...
                  description = `[default: ${a.default}] ${description}`;
-@@ -130,8 +130,7 @@ class CommandHelp {
+@@ -131,8 +131,7 @@ class CommandHelp {
                      value = flag.options.join('|');
                  }
                  if (!value.includes('|'))
 -                    value = underline(value);
 -                left += `=${value}`;
-+                left += ` <${value}>`;
++                    left += ` <${value}>`;
```
Page-, line 20: @srlowe I fixed the indentation here to signify it's part of the `if` statement. I think that's a bug but as I don't know what it's doing I've left the behavior as is and I'll let you take a look

package.json
```diff
@@ -279,7 +278,7 @@
     "window-size": "^1.1.0"
   },
   "optionalDependencies": {
-    "net-keepalive": "^1.3.3",
+    "net-keepalive": "^1.3.6",
```
Page-, line 281: 1.3.6 is actually the version already in the shrinkwrap, this just updates package.json to reflect that

package.json
```diff
@@ -232,43 +232,42 @@
...
     "is-elevated": "^3.0.0",
     "is-root": "^2.1.0",
-    "js-yaml": "^3.13.1",
+    "js-yaml": "^3.14.0",
     "klaw": "^3.0.0",
     "livepush": "^3.5.0",
     "lodash": "^4.17.19",
     "minimatch": "^3.0.4",
     "mixpanel": "^0.10.3",
-    "mkdirp": "^0.5.1",
```
Page-, line 245: Moved this to dev deps because it's only used in the package.json scripts

patches/all/@oclif+dev-cli+1.22.2.patch
```diff
@@ -248,14 +248,13 @@ index 343eb00..5521e2d 100644
 +    console.log(`fetchNodeBinary using tmp="${tmp}`);
 +
      if (arch === 'arm')
-         arch = 'armv6l';
+         arch = 'armv7l';
```
pdcastro, line 251: Okay... We don't currently use this, right? Oclif installers for ARM. Just checking.

Page-: It's an upstream bugfix for the urls to download from, from https://nodejs.org/en/blog/release/v12.18.3/ they look like 
```
ARMv7 32-bit Binary: https://nodejs.org/dist/v12.18.3/node-v12.18.3-linux-armv7l.tar.xz
```

PR #1996, opened 2020-08-13
build / push: Fix sudden process crash (update tar-utils dependency)

pdcastro:
Change-type: patch

As reported in: https://www.flowdock.com/app/rulemotion/i-cli/threads/Q00bMYsp8tA-3_cJlE0VolnI5Yv

### DIFF
--- a/package.json
+++ b/package.json
@@ -268,7 +268,7 @@
     "string-width": "^4.2.0",
     "strip-ansi-stream": "^1.0.0",
     "tar-stream": "^2.1.3",
-    "tar-utils": "^2.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -268,7 +268,7 @@
     "string-width": "^4.2.0",
     "strip-ansi-stream": "^1.0.0",
     "tar-stream": "^2.1.3",
```
pdcastro, line 270: Maybe we should pin `tar-stream` here as well, however it was not needed during my testing. 🤔

pdcastro:
@balena-ci I self-certify!

Issue #1992, opened 2020-08-13
Empty build args in v12.14.6


magnucha:
# Expected Behavior
When running `balena build -a myApp -B SOME_VAR=$SOME_VAR` I expect the SOME_VAR build arg to match the content of my host machines SOME_VAR environment variable.

# Actual Behavior
After updating to balena-cli v12.14.6 the variables are not set during docker build, and contain empty strings.
This behavior does not occur with my previous balena-cli v12.6.5, when running the exact same command in the same repository.

# Steps to Reproduce the Problem
Update to balena-cli v12.14.6 and try using build arguments. During build, they should appear as empty strings.

# Specifications

- **balena CLI version:** 12.14.6
- **Cloud backend:** balenaCloud
- **Operating system version:** Ubuntu 18.04
- **32/64 bit OS and processor:** 64 bit
- **Install method:** zip package

pdcastro:
Thank you for reporting this issue @magnucha 👍   I have found that it was introduced in CLI v12.9.9 during the conversion of the `build` command from [Capitano](https://github.com/balena-io/capitano/) to [oclif](https://oclif.io/). I am working on a fix, and will add more test code as well, so it doesn't regress. Meanwhile, please use CLI v12.9.8.

pdcastro:
This issue was fixed in CLI release v12.14.10. Test code was added to prevent future regressions. Sorry for trouble it caused you, and thank you again for reporting it! 👍

PR #1997, opened 2020-08-14
build: Fix --buildArg and --cache-from options (broken since v12.9.9)

pdcastro:
Resolves: #1992 
Change-type: patch

### DIFF
--- a/lib/actions-oclif/build.ts
+++ b/lib/actions-oclif/build.ts
@@ -115,7 +115,10 @@ ${dockerignoreHelp}
 		const logger = await Command.getLogger();
 		logger.logDebug('Parsing input...');
 
-		this.translateParams(params, options);
+		// `build` accepts `source` as a parameter, but compose expects it as an option
+		options.source = params.source;
+		delete params.source;
+
 		await this.validateOptions(options, sdk);
 
 		const app = await this.getAppAndResolveArch(options);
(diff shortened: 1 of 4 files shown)


lib/actions-oclif/build.ts
```diff
@@ -139,21 +142,6 @@ ${dockerignoreHelp}
 		logger.logSuccess('Build succeeded!');
 	}
 
-	protected translateParams(params: ArgsDef, options: FlagsDef) {
-		// Copy flags to those expected by other modules
-		options.arg = options.buildArg;
-		delete options.buildArg;
-		options['image-list'] = options['cache-from'];
-		delete options['cache-from'];
```
pdcastro, line 147: @srlowe, FYI, I guess that this "mapping" of `buildArg` to `arg` and from `cache-from` to `image-list` originated from a misunderstanding of Capitano's `signature` and `parameter` properties:

https://github.com/balena-io/balena-cli/blob/v12.9.8/lib/utils/docker-js.js#L83-L84
```
			signature: 'buildArg',
			parameter: 'arg',

			signature: 'cache-from',
			parameter: 'image-list',
```
The `parameter` property is not the "new name" of the flag, but merely what Capitano prints in the help output as follows:
```
$ ./balena-cli-12.9.8/balena help build
...
    --buildArg, -B <arg>
...
    --cache-from <image-list>
```
The flags are still referred to in the code as `options.buildArg` and `options['cache-from']`, despite the `parameter` property.

pdcastro:
@balena-ci I self-certify!

PR #2004, opened 2020-08-25
Don't try to parse deviceOrApplication as an integer in the tunnel action

### LINKED ISSUE #2003: `balena tunnel <short uuid>` fails if the uuid can be parsed as an integer
While `balena ssh <same short uuid>` works.
`balena tunnel` tries to parse the uuid as an integer before passing it to `getOnlineTargetUuid`, this is wrong as it expects a string.

zvin:
Fixes #2003

### DIFF
--- a/lib/actions-oclif/tunnel.ts
+++ b/lib/actions-oclif/tunnel.ts
@@ -28,7 +28,6 @@ import { getOnlineTargetUuid } from '../utils/patterns';
 import * as _ from 'lodash';
 import { tunnelConnectionToDevice } from '../utils/tunnel';
 import { createServer, Server, Socket } from 'net';
-import { tryAsInteger } from '../utils/validation';
 import { IArg } from '@oclif/parser/lib/args';
 
 interface FlagsDef {
@@ -78,7 +77,6 @@ export default class TunnelCmd extends Command {
 		{
 			name: 'deviceOrApplication',
 			description: 'device uuid or application name/id',
-			parse: (x) => tryAsInteger(x),
 			required: true,
 		},
 	];
--- a/lib/utils/patterns.ts
+++ b/lib/utils/patterns.ts
@@ -359,10 +359,31 @@ export function inferOrSelectDevice(preferredUuid: string) {
 	});
 }
 
+async function getApplicationByIdOrName(
+	sdk: BalenaSdk.BalenaSDK,
+	idOrName: string,
+) {
+	if (validation.looksLikeInteger(idOrName)) {
+		try {
+			return await sdk.models.application.get(Number(idOrName));
+		} catch (error) {
+			const { BalenaApplicationNotFound } = await import('balena-errors');
+			if (!instanceOf(error, BalenaApplicationNotFound)) {
+				throw error;
+			}
+		}
+	}
+	return await sdk.models.application.get(idOrName);
+}
+
 export async function getOnlineTargetUuid(
 	sdk: BalenaSdk.BalenaSDK,
 	applicationOrDevice: string,
 ) {
+	// applicationOrDevice can be:
+	// * an application name
+	// * an application ID (integer)
+	// * a device uuid
 	const Logger = await import('../utils/logger');
 	const logger = Logger.getLogger();
 	const appTest = validation.validateApplicationName(applicationOrDevice);
@@ -390,9 +411,9 @@ export async function getOnlineTargetUuid(
 	// otherwise, it may be a device OR an application...
 	try {
 		logger.logDebug(
-			`Fetching application by name ${applicationOrDevice} (${typeof applicationOrDevice})`,
+			`Fetching application by ID or name ${applicationOrDevice} (${typeof applicationOrDevice})`,
 		);
-		const app = await sdk.models.application.get(applicationOrDevice);
+		const app = await getApplicationByIdOrName(sdk, applicationOrDevice);
 		const devices = await sdk.models.device.getAllByApplication(app.id, {
 			$filter: { is_online: true },
 		});
--- a/lib/utils/validation.ts
+++ b/lib/utils/validation.ts
@@ -75,9 +75,13 @@ export function validateUuid(input: string): boolean {
 	return validateLongUuid(input) || validateShortUuid(input);
 }
 
-export function parseAsInteger(input: string, paramName?: string) {
+export function looksLikeInteger(input: string) {
 	// Allow only digits, no leading 0
-	if (!/^(0|[1-9][0-9]*)$/.test(input)) {
+	return /^(?:0|[1-9][0-9]*)$/.test(input);
+}
+
+export function parseAsInteger(input: string, paramName?: string) {
+	if (!looksLikeInteger(input)) {
 		const message =
 			paramName == null
 				? 'The parameter must be an integer.'


lib/actions-oclif/tunnel.ts
```diff
@@ -78,7 +77,6 @@ export default class TunnelCmd extends Command {
 		{
 			name: 'deviceOrApplication',
 			description: 'device uuid or application name/id',
-			parse: (x) => tryAsInteger(x),
```
pdcastro, line 81: This looks like it is going in the right direction, but there is still a corner case. The documentation says that a (numeric) application ID is also accepted:
```
$ balena help tunnel
...
ARGUMENTS
  <deviceOrApplication>  device uuid or application name/id
```
And indeed, the following works where `1301645` is an app ID, not a device UUID:
```
$ balena tunnel 1301645 -p 22222
? Select a device (Use arrow keys)
❯ frosty-snow (dcbff32)
```
This however relies on the code line above that is being deleted:
```
    parse: (x) => tryAsInteger(x),
```
I propose that this line be deleted, but the `tryAsInteger()` call moved to the beginning of this block:
https://github.com/balena-io/balena-cli/blob/v12.14.12/lib/utils/patterns.ts#L392-L395
```
                appOrDevice = tryAsInteger(appOrDevice);
		logger.logDebug(
			`Fetching application by name ${appOrDevice} (${typeof appOrDevice})`,
		);
		const app = await sdk.models.application.get(appOrDevice);
```
Note also how the `getOnlineTargetUuid` function already had several occurrences of `typeof applicationOrDevice` because the code assumed that `applicationOrDevice` might be of type `number` or `string` (and indeed it could and would), despite the argument being declared as a string (which was a type declaration bug).

Note also that `sdk.models.application.get(applicationOrDevice)` behaves differently depending on the type of `applicationOrDevice` being `number` or `string` -- the difference being numeric app ID and an app name, e.g. `1234` _vs._ `'1234'`. For this reason, I think the code should actually be calling `sdk.models.application.get(applicationOrDevice)` twice, once with a numeric argument (if it parses successfully as an integer) and then again with a string argument (if the first call was not successful in retrieving an application).

pdcastro (changes requested):
Given that the commit line goes in the Changelog file that some end users will be reading, I suggest changing it:
_From:_ `Don't try to parse deviceOrApplication as an integer in the tunnel action`
_To:_ `Fix device UUID parsing for 'balena tunnel'`

lib/utils/patterns.ts
```diff
@@ -359,10 +359,31 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+		try {
+			return await sdk.models.application.get(Number(idOrName));
+		} catch (error) {
+			const { BalenaApplicationNotFound } = await import('balena-errors');
+			if (!(error instanceof BalenaApplicationNotFound)) {
+				throw error;
+			}
+		}
+	}
+	return await sdk.models.application.get(Number(idOrName));
```
pdcastro, line 376: Looks like this should be just `idOrName` rather than `Number(idOrName)`.

lib/utils/patterns.ts
```diff
@@ -359,10 +359,31 @@ export function inferOrSelectDevice(preferredUuid: string) {
...
+async function getApplicationByIdOrName(
+	sdk: BalenaSdk.BalenaSDK,
+	idOrName: string,
+) {
+	if (validation.looksLikeInteger(idOrName)) {
+		try {
+			return await sdk.models.application.get(Number(idOrName));
+		} catch (error) {
+			const { BalenaApplicationNotFound } = await import('balena-errors');
+			if (!(error instanceof BalenaApplicationNotFound)) {
```
pdcastro, line 371: There is a function called `instanceOf` in `lib/errors.ts` that may be safer to use than plain `instanceof` in this case, in the event that dependencies (including `balena-sdk`) end up importing a different version of `balena-errors` (because of incompatible requirements in each dependency's own `package.json` file). (In the past, this led to bugs reported by users.)
```js
    const { instanceOf } = await import('../errors');
    const { BalenaApplicationNotFound } = await import('balena-errors');
    if (!(instanceOf(error, BalenaApplicationNotFound))) {
        ...
```
See https://github.com/balena-io/balena-cli/blob/v12.14.13/lib/errors.ts#L65

lib/utils/patterns.ts
```diff
@@ -390,9 +411,9 @@ export async function getOnlineTargetUuid(
 	// otherwise, it may be a device OR an application...
 	try {
 		logger.logDebug(
-			`Fetching application by name ${applicationOrDevice} (${typeof applicationOrDevice})`,
+			`Fetching application by id or name ${applicationOrDevice} (${typeof applicationOrDevice})`,
```
pdcastro, line 414: Upper case 'ID' would look better?

pdcastro (changes requested):

lib/utils/patterns.ts
zvin: Updated.

lib/utils/patterns.ts
zvin: Good catch, fixed.

lib/utils/patterns.ts
zvin: I prefer it lowercase but I don't mind, updated.

lib/utils/validation.ts
```diff
@@ -75,9 +75,13 @@ export function validateUuid(input: string): boolean {
 	return validateLongUuid(input) || validateShortUuid(input);
 }
 
-export function parseAsInteger(input: string, paramName?: string) {
+export function looksLikeInteger(input: string) {
 	// Allow only digits, no leading 0
-	if (!/^(0|[1-9][0-9]*)$/.test(input)) {
+	return !/^(0|[1-9][0-9]*)$/.test(input);
```
pdcastro, line 80: The exclamation mark is probably incorrect here. It may be the reason why the tests are failing with the error:
```
[Debug]   Fetching application by id or name abc1234 (string)
Not a valid value: number

Error: Not a valid value: number
```

lib/utils/validation.ts
```diff
@@ -75,9 +75,13 @@ export function validateUuid(input: string): boolean {
...
+}
+
+export function parseAsInteger(input: string, paramName?: string) {
+	if (looksLikeInteger(input)) {
```
pdcastro, line 84: Missing exclamation mark: `if (!looksLikeInteger(input)) {`

pdcastro (changes requested):

lib/utils/patterns.ts
pdcastro: You wrote here it's updated, but it isn't... :-)

pdcastro:
Not sure if you had seen this comment of mine:
> Given that the commit line goes in the Changelog file that some end users will be reading, I suggest changing it:
> _From:_ `Don't try to parse deviceOrApplication as an integer in the tunnel action`
> _To:_ `Fix device UUID parsing for 'balena tunnel'`

pdcastro:
Btw, while fixing the test failures: the PR also needs rebasing.

lib/utils/validation.ts
Page-, line 80:
```suggestion
	return !/^(?:0|[1-9][0-9]*)$/.test(input);
```

pdcastro: `?:` introduces a [non-capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges) as opposed to a capturing group. Out of interest, what would be the advantage of using it here, with `.test()`?

Page-: There's a performance cost to a capturing group (you have to track the match, store it in an array, etc), it's possible in a `.test` it makes no difference (I'd hope so) but I use non-capturing groups as a best practice wherever we don't actually care about the contents

zvin: Updated

lib/utils/validation.ts
zvin: Good catch !

lib/utils/patterns.ts
zvin: Sorry, I've updated in the comment above. Updating here too.

pdcastro (approved):

PR #2007, opened 2020-08-26
Convert command preload to oclif, typescript

srlowe:
Change-type: patch


lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,549 @@
...
+
+		// Check image file exists
+		try {
+			const fs = await import('fs');
+			await fs.promises.access(params.image);
+		} catch (error) {
+			throw new ExpectedError(
+				`The provided image path does not exist: ${params.image}`,
+			);
+		}
```
srlowe, line 160: Added check to prevent cryptic non-expected error later in process.

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,549 @@
...
+			$expand: this.applicationExpandOptions,
+		}) as Promise<Application & { should_be_running__release: [Release?] }>;
+	}
+
+	async prepareAndPreload(
+		preloader: Preloader,
+		balenaSdk: BalenaSDK,
+		options: FlagsDef,
+	) {
+		const { ExpectedError } = require('../errors');
```
Page-, line 494: Can you use `import` so we get typings:
```suggestion
		const { ExpectedError } = await import('../errors');
```

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,549 @@
...
+				.value();
+		});
+	}
+
+	getApplicationsWithSuccessfulBuilds(deviceTypeSlug: string) {
+		const balena = getBalenaSdk();
+
+		return this.getDeviceTypesWithSameArch(deviceTypeSlug).then(
+			(deviceTypes) => {
+				/** @type {import('balena-sdk').PineOptions<ApplicationWithDeviceType & { should_be_running__release: [import('balena-sdk').Release?] }>} */
```
Page-, line 342: This comment can be removed since the typing is now done via typescript syntax

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,549 @@
...
+				throw err;
+			}
+		} finally {
+			if (!gotSignal) {
+				await preloader.cleanup();
+			}
+		}
+	}
+
+	readonly applicationExpandOptions: any = {
```
Page-, line 280: You should be able to type this as
```suggestion
	readonly applicationExpandOptions: PinejsClientCore.ResourceExpand = {
```

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,549 @@
...
+			let progressBar = progressBars[event.name];
+			if (!progressBar) {
+				progressBar = progressBars[event.name] = new visuals.Progress(
+					event.name,
+				);
+			}
+			return progressBar.update({ percentage: event.percentage });
+		};
+
+		const spinners: any = {};
```
Page-, line 174:
```suggestion
		const spinners: {[key: string]: visuals.Spinner}= {};
```

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,549 @@
...
+
+		const spinnerHandler = function (event: any) {
```
Page-, line 176:
```suggestion
		const spinnerHandler = function (event: {name: string}) {
```

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,549 @@
...
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			PreloadCmd,
+		);
+
+		const balena = getBalenaSdk();
+		const balenaPreload = await import('balena-preload');
+		const visuals = getVisuals();
+		const nodeCleanup = await import('node-cleanup');
+		const { ExpectedError, instanceOf } = require('../errors');
```
Page-, line 150:
```suggestion
		const { ExpectedError, instanceOf } = await import('../errors');
```

lib/actions-oclif/preload.ts
srlowe: Ah sorry, missed that.

lib/actions-oclif/preload.ts
srlowe: Ahgr another

srlowe: Actually, I've just realised that we have a bunch of `exitWithExpectedError()`s, which should changed to `throw new ExpectedError()`s.  So probably the `ExpectedError` import should then be moved to the top.

lib/actions-oclif/preload.ts
srlowe: How should that be done when `visuals = getVisuals()`?  Do we need a seperate `import type`?

lib/actions-oclif/preload.ts
Page-: Sure, if there's any performance impact to that I'll catch/deal with it later - finishing the conversion and removing capitano will more than offset it anyway :)

lib/actions-oclif/preload.ts
srlowe: Do you know where I can find that namespace/type?  IDE not helping.

lib/actions-oclif/preload.ts
Page-: Yeah it should be possibly to do this via an `import type`, worst case it can be done with something along the lines of:
```suggestion
		const spinners: {[key: string]: ReturnType<typeof getVisuals>['Spinner']}= {};
```

lib/actions-oclif/preload.ts
Page-: It's in the `pinejs-client-core` module

Page-: Although balena-sdk uses some typings customized to the specific api model so it might be possible to use that - @thgreasi would know how to do that

srlowe: Hmm, Looks like that `ResourceExpand` type is generic.  If I use `ResourceExpand<any>` I get :
```
Type '{ end_timestamp: string; }' is not assignable to type 'OrderByValues'.   Type '{ end_timestamp: string; }' is not assignable to type '"desc"'.
```

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,552 @@
...
+
+		const spinners: {
+			[key: string]: ReturnType<typeof getVisuals>['Spinner'];
+		} = {};
+
+		const spinnerHandler = function (event: { name: string; action: string }) {
+			let spinner = spinners[event.name];
+			if (!spinner) {
+				spinner = spinners[event.name] = new visuals.Spinner(event.name);
+			}
```
Page-, line 189: This can be done as
```suggestion
			const spinner = spinners[event.name] ??= new visuals.Spinner(event.name);
```

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,552 @@
...
+		const progressHandler = function (event: {
+			name: string;
+			percentage: number;
+		}) {
+			let progressBar = progressBars[event.name];
+			if (!progressBar) {
+				progressBar = progressBars[event.name] = new visuals.Progress(
+					event.name,
+				);
+			}
```
Page-, line 177: This can be done as
```suggestion
			const progressBar = progressBars[event.name] ??= new visuals.Progress(
				event.name,
			);
```

Page- (approved):

lib/actions-oclif/preload.ts
Page-: Handled these in https://github.com/balena-io/balena-cli/pull/2010/files# as it requires typescript 4

lib/actions-oclif/preload.ts
```diff
@@ -0,0 +1,552 @@
...
+	}
+
+	getApplicationsWithSuccessfulBuilds(deviceTypeSlug: string) {
+		const balena = getBalenaSdk();
+
+		return this.getDeviceTypesWithSameArch(deviceTypeSlug).then(
+			(deviceTypes) => {
+				const options: PineOptions<
+					ApplicationWithDeviceType & { should_be_running__release: [Release?] }
+				> = {
```
thgreasi, line 353: I think that this should have been:
```ts
const options: PineOptions<Application> = {
```
for which to work you should also type `applicationExpandOptions` as `BalenaSdk.PineExpand<Application>`

srlowe: Ok thanks, will add that in later PR.

srlowe: Ah, I can't use `BalenaSdk.PineExpand<Application>` at the moment because yesterday page added a fix to the typings for the `$orderby` format used here.

thgreasi: I'm already on it @srlowe

thgreasi: @srlowe 
See: https://github.com/balena-io/balena-cli/pull/2011

This is my current proposal on how to type everything with the current SDK typings.

PR #2009, opened 2020-08-26
Drop no longer used pkg dependency config

thgreasi:
There is no longer a balena-pine folder in that path.

Change-type: patch

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "assets": [
       "build/auth/pages/*.ejs",
       "node_modules/resin-discoverable-services/services/**/*",
-      "node_modules/balena-sdk/node_modules/balena-pine/**/*",
       "node_modules/balena-pine/**/*",
       "node_modules/pinejs-client-core/**/*",
       "node_modules/opn/xdg-open",


package.json
```diff
@@ -35,7 +35,6 @@
     "assets": [
       "build/auth/pages/*.ejs",
       "node_modules/resin-discoverable-services/services/**/*",
-      "node_modules/balena-sdk/node_modules/balena-pine/**/*",
```
pdcastro, line 38: The thing is... At the moment this line is not useful -- but also harmless to have there. At one point (not too long ago), CLI dependencies including but not limited to `balena-sdk` declared requirements for different versions of `balena-pine`. When this happened, `npm install` resolved the conflict by bundling a separate copy of `balena-pine` under `node_modules/balena-sdk/node_modules/`, recording this in the `npm-shrinkwrap.json` file. This broke the CLI's standalone executable and required time-consuming investigation to fix -- the fix being this line that this PR proposes to delete. This situation could happen again at any time (when CLI dependencies are updated), and we would find ourselves again having to investigate and fix the problem... I am not confident that the CLI has an automated test suite comprehensive enough to at least automatically point out the problem, in which case it could be end users filing bug issues. It would rightly be argued that the proper fix is to improve CLI testing and/or add automated checks for multiple versions of `balena-pine` in `npm-shrinkwrap.json`, rather than leaving this ghost / dead line in `package.json` "just in case". I agree! But... A counter argument might be that if this PR added such a comprehensive test suite and/or the checks for multiple versions of `balena-pine`, then it would be easier to see the benefits of deleting this line. :-)

thgreasi: I was amazed that `pkg` didn't show a single warning about the config which was instructing it to bundle a folder that doesn't exist. I don't atm have the bandwidth to add such test cases, but I agree that running the test suite against the standalone build sounds like a great idea.

thgreasi: There are also configs for nested pinej-client-core files so the possibility matrix grows larger quickly.
I'm also not entirely sure how npm will decide to deduplicate balena-pine or pinejs-client-core, the next time we will do a major, so that's yet an extra reason that the dead config might not prove helpful in the future.

pdcastro: @srlowe, @Page-, what is your 3rd opinion here? It's not a big deal to me, but this PR should be either approved or closed, because balena-ci keeps rebasing it and wasting CI resources. :-)

srlowe: I think we should probably leave it in if it's not doing any harm, and there's the possibility of it catching an issue in the future.

PR #2010, opened 2020-08-27
Update to typescript 4.0

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/actions-oclif/local/configure.ts
+++ b/lib/actions-oclif/local/configure.ts
@@ -119,9 +119,7 @@ export default class LocalConfigureCmd extends Command {
 	readonly CONNECTIONS_FOLDER = '/system-connections';
 
 	getConfigurationSchema(connectionFileName?: string) {
-		if (connectionFileName == null) {
-			connectionFileName = 'resin-wifi';
-		}
+		connectionFileName ??= 'resin-wifi';
 		return {
 			mapper: [
 				{
--- a/lib/actions-oclif/preload.ts
+++ b/lib/actions-oclif/preload.ts
@@ -40,17 +40,11 @@ import type { ResourceExpand } from 'pinejs-client-core';
 
 interface FlagsDef extends DockerConnectionCliFlags {
 	app?: string;
-	appId?: string; // Internal use.
 	commit?: string;
 	'splash-image'?: string;
-	splashImage?: string; // Internal use.
 	'dont-check-arch': boolean;
-	dontCheckArch?: boolean; // Internal use.
 	'pin-device-to-release': boolean;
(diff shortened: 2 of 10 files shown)


tests/nock-mock.ts
```diff
@@ -29,7 +29,7 @@ export interface ScopeOpts {
 export class NockMock {
 	public readonly scope: nock.Scope;
 	// Expose `scope` as `expect` to allow for better semantics in tests
-	public readonly expect = this.scope;
+	public readonly expect;
```
Page-, line 32: This errored with `Property 'scope' is used before its initialization.`

package.json
```diff
@@ -111,7 +111,7 @@
     }
   },
   "devDependencies": {
-    "@balena/lint": "^5.1.0",
+    "@balena/lint": "^5.2.0",
```
Page-, line 114: This guarantees typescript 4 support in the linting

lib/actions-oclif/local/configure.ts
```diff
@@ -119,9 +119,7 @@ export default class LocalConfigureCmd extends Command {
 	readonly CONNECTIONS_FOLDER = '/system-connections';
 
 	getConfigurationSchema(connectionFileName?: string) {
-		if (connectionFileName == null) {
-			connectionFileName = 'resin-wifi';
-		}
+		connectionFileName ??= 'resin-wifi';
```
Page-, line 122: I switched to this pattern for coffeescript style defaults in a bunch of places, but when we're fully in typescript/know null cannot be passed then it can be changed to use
```ts
	getConfigurationSchema(connectionFileName = 'resin-wifi') {
```
instead

lib/actions-oclif/preload.ts
```diff
@@ -195,27 +185,23 @@ Can be repeated to add multiple certificates.\
...
 			: options.commit;
-		options.image = params.image;
-		options.appId = options.app;
-		delete options.app;
+		const image = params.image;
+		const appId = options.app;
 
-		options.splashImage = options['splash-image'];
-		delete options['splash-image'];
+		const splashImage = options['splash-image'];
```
Page-, line 194: These deletes were causing problems as they're not optional and so deleting would violate the type. I changed to assign to variables and remove the "internal only" parts from FlagsDef as it better types what we actually receive and avoids this rename/delete props pattern which is harder to follow imo than straight assignments

srlowe: Right, that's much tidier.

srlowe (approved):

Issue #2005, opened 2020-08-25
`os versions` and `os download` not listing balenaOS ESR releases


pdcastro:
At least for [Production and Enterprise plan](https://www.balena.io/pricing/) users (and for some device types like the Raspberry Pi 3, also for a few other plans?), the `os versions` and `os download --version menu` commands should list balenaOS ESR versions such as `v2020.07.1`, instead of the "usual" version format such as `v2.29.2+rev2`.

Balena support agents in the "Team member plan and standard support" should also be able to list ESR releases, but currently are not.

Originally reported in [this flowdock thread](https://www.flowdock.com/app/rulemotion/resin-devices/threads/iHHXQlkB5Dt-m2_EBzO3WlDBn1G).

# Expected Behavior

```
$ balena os versions
...
v2020.07.1
...
```

# Actual Behavior

```
$ balena os versions raspberrypi3
v2.53.9+rev1.prod (recommended)
v2.53.9+rev1.dev
v2.51.1+rev1.prod
...
```

See: https://www.flowdock.com/app/rulemotion/resin-devices/threads/BtGUlpKl9fyhDGpI1OJ4W32EfAD

shawaj:
Any progress with this? We need this quite urgently

pdcastro:
@shawaj, support for ESR versions was added to CLI v12.52.0 (or later). Check `balena help os download` and `balena help os versions` for notes regarding ESR support (options and limitations). Let us know if it works as you expected.

PR #2012, opened 2020-08-28
`os versions`: Add balenaOS ESR releases to the list

pdcastro:
Resolves: #2005 
Change-type: minor


lib/utils/os.ts
```diff
@@ -0,0 +1,439 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This module is a copy-and-paste (with minor amendments) from:
+ * https://github.com/balena-io/balena-ui/blob/v2.497.9/src/services/os.ts
```
pdcastro, line 20: This PR is still a draft, for discussion. This whole file was copied from:
https://github.com/balena-io/balena-ui/blob/v2.497.9/src/services/os.ts

As kindly pointed out by @JSReds in [this flowdock thread](https://www.flowdock.com/app/rulemotion/resin-devices/threads/iHHXQlkB5Dt-m2_EBzO3WlDBn1G).

I would be able to trim it down a bit (not all exported functions are used by the `balena os versions` command - in fact only `getAllOsVersions` is used - but some other function may still be useful for other CLI commands like `balena os download` - pending investigation).

As this file proved to be useful for both the web dashboard and the CLI, it should probably be part of the balena SDK, no? Thoughts?

lib/utils/os.ts
```diff
@@ -0,0 +1,439 @@
...
+	primitive: true,
+	promise: true,
+});
+const memoizedGetOsVersions = memoizee(getOsVersions, {
+	maxAge: 10 * 60 * 1000,
+	primitive: true,
+	promise: true,
+});
+
+export const getAllOsVersions = (
```
pdcastro, line 331: This is the function used by the `os versions` command.

lib/actions-oclif/os/versions.ts
```diff
@@ -54,13 +59,41 @@ export default class OsVersionsCmd extends Command {
...
-		vs.forEach((v) => {
-			console.log(`v${v}` + (v === recommended ? ' (recommended)' : ''));
+	protected async getOsVersionsForDeviceType(
+		deviceType: string,
+	): Promise<string[]> {
+		const osUtils = await import('../../utils/os');
+		const allVersions = await osUtils.getAllOsVersions([deviceType]);
+		const versionObjects = allVersions[deviceType] || [];
+		return versionObjects.map((vObj) => {
+			let fv = vObj.formattedVersion;
```
pdcastro, line 79: The modifications of `vObj.formattedVersion` in this function are quite hacky. It would make sense to modify `os.ts` (if that file was indeed to become part of the CLI), but I wanted to avoid it in order to preserve the ability to copy-and-paste `os.ts` from `balena-io/balena-ui`.

lib/actions-oclif/os/versions.ts
```diff
@@ -54,13 +59,41 @@ export default class OsVersionsCmd extends Command {
 	public async run() {
 		const { args: params } = this.parse<FlagsDef, ArgsDef>(OsVersionsCmd);
 
-		const {
-			versions: vs,
-			recommended,
-		} = await getBalenaSdk().models.os.getSupportedVersions(params.type);
+		let versions = await this.getOsVersionsForDeviceType(params.type);
+		if (versions.length === 0) {
```
pdcastro, line 63: If a device type "alias" such as `'nuc'` is used, `getOsVersionsForDeviceType` returns an empty list. If this is the case, fetch the aliases, resolve the actual device type name (e.g. `'intel-nuc'`), and call `getOsVersionsForDeviceType` again for the actual device type name. Querying and computing the list of aliases is relatively slow/expensive, which is why I thought it should only be done if the list of versions is empty. A note advising against the use of aliases was added to the help text.

package.json
```diff
@@ -241,6 +242,7 @@
     "klaw": "^3.0.0",
     "livepush": "^3.5.0",
     "lodash": "^4.17.20",
+    "memoizee": "^0.4.14",
```
pdcastro, line 245: `memoizee` and `@types/memoizee` are used in the added `os.ts`, but the dependencies were already included in the `npm-shrinkwrap.json` file, which was thus not modified.

lib/utils/os.ts
sradevski: @pdcastro it indeed should be in the SDK, and I opened a draft PR [here](https://github.com/balena-io/balena-sdk/pull/876). The problem is that we currently don't have a good way to preload hostapps on open balena and other environments, so I switched to working on that before I update the OS methods, so we don't have to have `os.ts` and `newOs.ts` distinction. If this is urgent, however, I can push on that next week and get it merged in the SDK.

pdcastro:
Closing in favour of #2334, as functionality was added to the balena SDK.

PR #2014, opened 2020-09-03
Add support for new device metrics to device command

srlowe:
The CLI part of the release party feature.

Change-type: minor

### DIFF
--- a/lib/actions-oclif/device/index.ts
+++ b/lib/actions-oclif/device/index.ts
@@ -30,6 +30,15 @@ interface ExtendedDevice extends DeviceWithDeviceType {
 	device_type?: string;
 	commit?: string;
 	last_seen?: string;
+	memory_usage_mb: number | null;
+	memory_total_mb: number | null;
+	memory_usage_percent?: number;
+	storage_usage_mb: number | null;
+	storage_total_mb: number | null;
+	storage_usage_percent?: number;
+	cpu_temp_c: number | null;
+	cpu_usage_percent: number | null;
+	undervoltage_detected?: boolean;
 }
 
 interface FlagsDef {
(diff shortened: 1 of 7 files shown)


pdcastro:
`Change-type` (PR and commit) should be `minor`, no?

lib/actions-oclif/device/index.ts
```diff
@@ -30,6 +30,15 @@ interface ExtendedDevice extends DeviceWithDeviceType {
 	device_type?: string;
 	commit?: string;
 	last_seen?: string;
+	memory_usage_mb?: number;
```
pdcastro, line 33: `mb` → `mib`
I gather it is [mib](https://en.wikipedia.org/wiki/Mebibyte) rather than [mb](https://en.wikipedia.org/wiki/Megabyte), right? That's what I heard in the release party live stream. :-) It's very good to have a unit of measurement as part of the field name, 👍   but having an incorrect unit is less helpful.

lib/actions-oclif/device/index.ts
```diff
@@ -85,9 +94,20 @@ export default class DeviceCmd extends Command {
...
+				'memory_usage',
+				'memory_total',
+				'storage_block_device',
+				'storage_usage',
+				'storage_total',
+				'cpu_usage',
+				'cpu_temp',
+				'cpu_id',
+				'is_undervolted',
+				 */
```
pdcastro, line 107: Curious comment block

lib/actions-oclif/device/index.ts
```diff
@@ -119,6 +172,17 @@ export default class DeviceCmd extends Command {
 				'note',
 				'os_version',
 				'dashboard_url',
+				'cpu_usage_percent',
+				'cpu_temp_c',
+				'cpu_id',
+				'memory_usage_mb', // Use MiB, when we support lowercase in headers
```
pdcastro, line 178: I see. But having `MEMORY USAGE MIB` is probably better than `MEMORY USAGE MB`, no? I mean, at least users don't get confused with 1,000 vs 1,024...

lib/actions-oclif/device/index.ts
srlowe: These vars have been created just so we can rename the headers in the table. Unfortunately we are stuck with caps, so it will be displayed as MIB, which people may not recognise as a unit.

lib/actions-oclif/device/index.ts
pdcastro: The YouTube poll has settled it as MB.

pdcastro (approved):

srlowe:
@balena-ci retest

PR #2032, opened 2020-09-10
Misc bugfixes

srlowe:
- Fix numerical id support in device move
- Fix numerical id support in device rm
- Fix required status of param in os versions

---

### DIFF
--- a/lib/actions-oclif/device/move.ts
+++ b/lib/actions-oclif/device/move.ts
@@ -17,7 +17,7 @@
 
 import { flags } from '@oclif/command';
 import type { IArg } from '@oclif/parser/lib/args';
-import type { Application } from 'balena-sdk';
+import type { Application, BalenaSDK } from 'balena-sdk';
 import Command from '../../command';
 import * as cf from '../../utils/common-flags';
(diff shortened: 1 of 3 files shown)


lib/actions-oclif/device/move.ts
```diff
@@ -162,4 +122,53 @@ export default class DeviceMoveCmd extends Command {
 			}
 		}
 	}
+
+	async interactivelySelectApplication(
```
pdcastro, line 126: Much better in a separate function 👍

lib/actions-oclif/device/rm.ts
```diff
@@ -85,7 +84,7 @@ export default class DeviceRmCmd extends Command {
 		// Remove
 		for (const uuid of params.uuid.split(',')) {
```
pdcastro, line 85: Looks like this could be `uuids` instead of `params.uuid.split(',')` (duplicate call to `split()`).

pdcastro (approved):

lib/actions-oclif/device/rm.ts
srlowe: oops yes

PR #2033, opened 2020-09-10
Improve command suggestions, add topic help

srlowe:
```
$ balena scann
scann is not a recognized balena command.

Did you mean: ? 
  scan

Run balena help -v for a list of available commands,
 or balena help <command> for detailed help on a specific command.
```
```
$ balena env renamee

env commands include:
  env add <name> [value]   add env or config variable to application(s), device(s) or service(s)
  env rename <id> <value>  change the value of a config or env var for an app, device or service
  env rm <id>              remove a config or env var from an application, device or service

Run balena help -v for a list of all available commands,
 or balena help <command> for detailed help on a specific command.
```

```
$ balena enve rename

enve is not a recognized balena command.

Did you mean: ? 
  env rename
  envs

Run balena help -v for a list of available commands,
 or balena help <command> for detailed help on a specific command.
```

### DIFF
--- a/lib/help.ts
+++ b/lib/help.ts
@@ -40,6 +40,27 @@ export default class BalenaHelp extends Help {
 			return;
 		}
 
+		// If they've typed a topic (e.g. `balena os`) that isn't also a command (e.g. `balena device`)
+		// then list the associated commands.
+		const topicCommands = this.config.commands.filter((c) => {
+			return c.id.startsWith(`${subject}:`);
+		});
+		if (topicCommands.length > 0) {
+			console.log(`${chalk.yellow(subject)} commands include:`);
+			console.log(this.formatCommands(topicCommands));
+			console.log(
+				`\nRun ${chalk.cyan.bold(
+					'balena help -v',
+				)} for a list of all available commands,`,
+			);
+			console.log(
+				` or ${chalk.cyan.bold(
+					'balena help <command>',
+				)} for detailed help on a specific command.`,
+			);
+			return;
+		}
+
 		throw new ExpectedError(`command ${chalk.cyan.bold(subject)} not found`);
 	}
 
--- a/lib/hooks/command-not-found/suggest.ts
+++ b/lib/hooks/command-not-found/suggest.ts
@@ -22,10 +22,12 @@ import type { IConfig } from '@oclif/config';
  A modified version of the command-not-found plugin logic,
  which deals with spaces separators stead of colons, and
  prints suggested commands instead of prompting interactively.
+
+ Also see help.ts showHelp() for handling of topics.
  */
 
 const hook: Hook<'command-not-found'> = async function (
-	opts: object & { config: IConfig; id?: string },
+	opts: object & { config: IConfig; id?: string; argv?: string[] },
 ) {
 	const Levenshtein = await import('fast-levenshtein');
 	const _ = await import('lodash');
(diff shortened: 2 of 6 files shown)


lib/help.ts
```diff
@@ -40,6 +40,27 @@ export default class BalenaHelp extends Help {
...
+		});
+		if (topicCommands.length > 0) {
+			console.log(`${chalk.yellow(subject)} commands include:`);
+			console.log(this.formatCommands(topicCommands));
+			console.log(
+				`\nRun ${chalk.cyan.bold(
+					'balena help -v',
+				)} for a list of all available commands,`,
+			);
+			console.log(
```
pdcastro, line 56: Should all these `console.log` messages be `console.error` instead? I.e. using `stderr` instead of `stdout`. One "school of thought" is that any messages (including info, warn, error) are not part of the intended output of the command should go to stderr. For example, if the user typed:

```
$ balena envss > env_vars.txt
```
The intention would be for the list of env vars (printed to `stdout`) to be saved to a text file, while any info / warn / error messages would not be redirected. Using `console.error` above would mean that the messages would not be redirected to the text file. If the user wanted both stderr and stdout to go to the text file, they could use:
```
$ balena envss &> env_vars.txt
```
(At least with `bash`)

lib/hooks/command-not-found/suggest.ts
```diff
@@ -44,16 +46,36 @@ const hook: Hook<'command-not-found'> = async function (
...
+	// if there's something there, also test if it might be a double
+	// word command spelt wrongly, rather than command args.
+	if (opts.argv?.[0]) {
+		suggestions.unshift(
+			closest(`${commandId}: + ${opts.argv[0]}`).replace(':', ' ') || '',
+		);
+	}
+
+	// Output suggestions
 	console.error(
```
pdcastro, line 62: Here we're using `console.error`, but below using `console.log`. One could reason that only "proper errors" (as opposed to "info messages") should go to `stderr`, but my preferred reasoning :-) is that any error/warn/info messages should go to stderr (see other comment). But feel free to disagree! :-)

patches/all/@oclif+config+1.17.0.patch
```diff
@@ -0,0 +1,15 @@
...
+index aba1da9..163e439 100644
+--- a/node_modules/@oclif/config/lib/config.js
++++ b/node_modules/@oclif/config/lib/config.js
+@@ -165,7 +165,9 @@ class Config {
+         debug('runCommand %s %o', id, argv);
+         const c = this.findCommand(id);
+         if (!c) {
+-            await this.runHook('command_not_found', { id });
++			// argv added to command_not_found hook
++			// We should try to upstream this change
```
pdcastro, line 11: Minor thing: looks like tabs were mixed with spaces here.

lib/utils/messages.ts
```diff
@@ -45,7 +45,7 @@ export const balenaAsciiArt = `\
 `;
 
 export const registrySecretsHelp = `\
-REGISTRY SECRETS  
+REGISTRY SECRETS
```
pdcastro, line 48: The double space at the end of a line means a line break in markdown syntax, which gets converted to `<br>` (or similar) in HTML on this page: https://www.balena.io/docs/reference/balena-cli/#push-applicationordevice

Without the double space, it looks a bit broken when converted to HTML, a bit like:
```
REGISTRY SECRETS The --registry-secrets option specifies ...
```
Similar issue when viewed with the GitHub preview for files like README.md.

pdcastro (changes requested):

lib/utils/messages.ts
srlowe: That's strange - I didn't knowingly remove that.

pdcastro (approved):

srlowe:
@balena-ci retest

srlowe:
Getting an error from appveyor:
```
Build started
git config --global core.autocrlf input
git clone -q https://github.com/balena-io/balena-cli.git C:\projects\resin-cli
fatal: unable to access 'https://github.com/balena-io/balena-cli.git/': The requested URL returned error: 504
Command exited with code 128
```

Issue #2036, opened 2020-09-16
scan: Part of the CLI could not be loaded. This typically means your CLI install is in a broken state.


pdcastro:
```
$ sudo ./balena-cli-12.21.0/balena scan
Password:
Part of the CLI could not be loaded. This typically means your CLI install is in a broken state.
You can normally fix this by uninstalling and reinstalling the CLI.

Additional information may be available with the `--debug` flag.
For help, visit our support forums: https://forums.balena.io
For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/
```

Applies to the standalone zip installer only, on all platforms (Linux, Windows, macOS).

Appears to have started manifesting in CLI v12.17.0, although I suspect that the root cause can be traced to an earlier pull request: https://github.com/balena-io-modules/balena-sync/pull/163

pdcastro:
Fixed in CLI release v12.21.1.

PR #2037, opened 2020-09-16
scan: Fix "CLI could not be loaded" with the standalone zip installer

pdcastro:
Resolves: #2036
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "scripts": [
       "build/**/*.js",
       "node_modules/balena-sdk/es2018/index.js",
-      "node_modules/balena-sync/build/sync/*.js",
+      "node_modules/balena-sync/build/**/*.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "scripts": [
       "build/**/*.js",
       "node_modules/balena-sdk/es2018/index.js",
-      "node_modules/balena-sync/build/sync/*.js",
+      "node_modules/balena-sync/build/**/*.js",
```
pdcastro, line 30: I suspect (not fully confirmed) that the need for this change comes from this `balena-sync` PR:
https://github.com/balena-io-modules/balena-sync/pull/163

Curiously though, the error started manifesting with the removal of Capitano in CLI release v12.17.0.

pdcastro:
@balena-ci I self-certify!

Issue #1542, opened 2019-12-23
Add `balena device restart`


shaunmulligan:
Currently we only have `balena app restart` and `balena device reboot`. There is no `balena device restart`

pdcastro:
Related to #583 [Implement new device commands](https://github.com/balena-io/balena-cli/issues/583)

Issue #1378, opened 2019-08-06
Usage of isTTY isn't always foolproof when deciding whether to use spinners etc


CameronDiver:
CIs like jenkins for example can force a tty emulation, so we should also provide a flag to override any spinners or progress bars when displaying output. This flag should be supported on any command which uses cli visuals.

---

*Please keep in mind that we try to use the issue tracker of this repository for specific bug
reports & CLI feature requests. General & troubleshooting questions are encouraged to be posted to
the [balena forums](https://forums.balena.io), which are monitored by balena's support team and
where the community can both contribute and benefit from the answers.*

*Before submitting this issue please check that this issue is not a duplicate. If there is another
issue describing the same problem or feature please add your information to the existing issue's
comments.*

pdcastro:
Related to issue #1089 [Implement scriptable (non-interactive) mode](https://github.com/balena-io/balena-cli/issues/1089)

Issue #1172, opened 2019-04-18
Flag to remove coloring from output


mcamou:
- **9.15.6**
- **Standalone install:** no
- **Node version:** 8.11.4
- **Npm version:** 5.6.0
- **Operating system and architecture:** Ubuntu Linux 18.10 amd64

---
For CI and automation purposes, we would like to be able to get the release commit and ID after a `balena push` from a script. In this case the output coloring makes parsing this data extremely inconvenient.

After looking at the sources we've seen that you use `chalk` for coloring. According to the [`chalk` README](https://github.com/chalk/chalk), setting `FORCE_COLOR=0` should disable coloring. That does not work. We have also tried setting `TERM=dumb` but that does not work either.

pdcastro:
Related to issue #1089 (Implement scriptable (non-interactive) mode)

Issue #1089, opened 2019-02-01
Implement scriptable (non-interactive) mode


pdcastro:
***Placeholder feature issue -- requires further discussion***

This is a generalisation of several specific feature requests from CLI end users:
#89, #462, #695, #789, #837, #1001 

Implement a consistent flag through all CLI commands to enable "scriptable/CI/robot mode", as opposed to "human/interactive mode":

*Interactive mode assumes a human reading the output:*
* Interactive menus
* Confirmation prompts
* Spinning animations, ASCII control codes that move the cursor
* Condensed messages (balena-io/balena/issues/1653)
* Suggested alternatives to what the user actually requested (@imrehg's case: balena ssh: inexistent device UUID causes all devices to be listed, good for humans but makes it harder to write scripts)

*Scriptable/CI/robot mode produces parseable output like JSON / CSV / TSV:*
* Places greater importance on backwards compatibility of printed output
* Never presents a selection menu nor asks for confirmation (either succeeds in delivering what was requested through command-line options and configuration files, or fails with process exit status codes and parseable error messages)
* Behind the scenes, interactive mode calls scriptable mode then parses and reformats the output

*CLI can automatically detect whether stdin is a tty terminal (@CameronDiver's proposal)*
* Explicit flags like -S (for “scriptable”) and -i (for “interactive”) allows bypassing auto detection
  * Compatibility with existing customers' (and @imrehg's) :-) scripts that parse interactive output
  * Easier to write automated end-to-end tests

imrehg:
This would be very useful to have, as otherwise e.g. `balena ssh` in scripting mode can connect to the completely wrong device, if used in a script and given an invalid UUID or an UUID that the user cannot connect to. In that case the a menu is popped up, and in interactive mode the first device is connected to.

Thus for SSH at least could work as well:

* if given an UUID, do not prompt the menu  (i.e. assume that the user wanted to connect to that device and nothing else)
* prompt if no UUID is given, since then it should be logical that something needs to be selected

pdcastro:
Finding related to this issue: @CameronDiver [pointed out](https://www.flowdock.com/app/rulemotion/r-resinos/threads/8PJpiyl6CKmXSzICWZ1gk3y5sRn) that oclif has a [cli-ux.table function](https://oclif.io/docs/table) that adds several command-line options for data formatting (csv, json, yaml) that could be handy to have in a scriptable / parseable mode:

https://oclif.io/docs/table
```
OPTIONS
[...]

PR #2038, opened 2020-09-17
Add new command `device restart`

srlowe:
Add new command `device restart`, for feature parity with dash.

Change-type: minor
Resolves: #1542

### DIFF
--- a/lib/actions-oclif/device/restart.ts
+++ b/lib/actions-oclif/device/restart.ts
@@ -1,0 +1,197 @@
+/**
+ * @license
+ * Copyright 2016-2020 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { flags } from '@oclif/command';
+import type { IArg } from '@oclif/parser/lib/args';
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, getCliUx, stripIndent } from '../../utils/lazy';
+import type {
+	BalenaSDK,
+	DeviceWithServiceDetails,
+	CurrentServiceWithCommit,
+} from 'balena-sdk';
+
+interface FlagsDef {
+	help: void;
+	service?: string;
+}
+
+interface ArgsDef {
+	uuid: string;
+}
+
+export default class DeviceRestartCmd extends Command {
+	public static description = stripIndent`
+		Restart containers on a device.
+
+		Restart containers on a device.
+		If the --service flag is provided, then only those services' containers
+		will be restarted, otherwise all containers on the device will be restarted.
+
+		Multiple devices and services may be specified with a comma-separated list
+		of values (no spaces).
+
+		Note this does not reboot the device, to do so use instead \`balena device reboot\`.
+		`;
+	public static examples = [
+		'$ balena device restart 23c73a1',
+		'$ balena device restart 55d43b3,23c73a1',
+		'$ balena device restart 23c73a1 --service myService',
+		'$ balena device restart 23c73a1 -s myService1,myService2',
+	];
+
+	public static args: Array<IArg<any>> = [
+		{
+			name: 'uuid',
+			description:
+				'comma-separated list (no blank spaces) of device UUIDs to restart',
+			required: true,
+		},
+	];
+
+	public static usage = 'device restart <uuid>';
+
+	public static flags: flags.Input<FlagsDef> = {
+		service: flags.string({
+			description:
+				'comma-separated list (no blank spaces) of service names to restart',
+			char: 's',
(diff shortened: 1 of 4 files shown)


lib/actions-oclif/device/restart.ts
```diff
@@ -0,0 +1,183 @@
...
+
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			DeviceRestartCmd,
+		);
+
+		const balena = getBalenaSdk();
+		const ux = getCliUx();
+
+		ux.action.start('Restarting services', '', { stdout: true });
```
pdcastro, line 85: * Would this be better printed to stderr? If the command was to print a JSON object with the list of device UUIDs that failed / succeeded to restart, then the JSON object could be printed to stdout, and these animations printed to stderr.
* Thinking about non-interactive usage (scripts), it should be possible to disable "ASCII animations" like this. Related issues:
  * https://github.com/balena-io/balena-cli/issues/1378
  * https://github.com/balena-io/balena-cli/issues/1172
  * https://github.com/balena-io/balena-cli/issues/1089
  * https://github.com/balena-io/balena-cli/issues/1001

  I think we should have TTY auto detection, but also a flag that overrides detection.

lib/actions-oclif/device/restart.ts
```diff
@@ -0,0 +1,183 @@
...
+		if (options.service) {
+			await this.restartServices(balena, params.uuid, options.service);
+		} else {
+			await this.restartAllServices(balena, params.uuid);
+		}
+		ux.action.stop();
```
pdcastro, line 91: Better have `action.stop()` in a finally block, so it always runs in case the restart methods throw an error.

srlowe: AFAIK `action.stop()` simply changes the spinner status to`done` (by default) and removes the spinner.  If you don't stop it then it prints a red `!` next the the last status (and error is output under it) which is what we want I think.

lib/actions-oclif/device/restart.ts
srlowe: > Would this be better printed to stderr?

Hmm perhaps.  I purposefully printed this to stout, based on our previous discussion about stdout/stderr, but perhaps I missunderstood at the time, or misremembered.  We should probably come up with a short policy, about output and add it to the contributing doc.

> Thinking about non-interactive usage (scripts), it should be possible to disable "ASCII animations" like this. 

According to the docs this gracefully handles ttl/non-ttl.  I don't think it supports an override flag, but I think it would be good to stick to standard implementations if we can.

https://github.com/oclif/cli-ux#cliaction

lib/actions-oclif/device/restart.ts
```diff
@@ -0,0 +1,183 @@
...
+	public static examples = [
+		'$ balena device restart 23c73a1',
+		'$ balena device restart 23c73a1 --service myService',
+		'$ balena device restart 23c73a1 -s myService1 -s myService2',
+	];
+
+	public static args: Array<IArg<any>> = [
+		{
+			name: 'uuid',
+			description: 'the uuid of the device to restart',
```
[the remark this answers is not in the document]
srlowe: following on from conversation in FD...

Latest push accepts comma-separated values for both device and service.

lib/actions-oclif/device/restart.ts
srlowe: Latest push prints spinner output to stderr.

lib/actions-oclif/device/restart.ts
pdcastro: Interesting, just tested it with an explicit `throw new Error('oops')` line between spinner start/stop, and indeed it shows the red exclamation mark ❗️  _before_ 'oops' is printed to the console. So it must be somehow intercepting thrown errors, in a more "intrusive" way than I expected. Hopefully doesn't conflict with Sentry. It seems to work well though. 👍

lib/actions-oclif/device/restart.ts
```diff
@@ -0,0 +1,197 @@
...
+		'$ balena device restart 55d43b3,23c73a1',
+		'$ balena device restart 23c73a1 --service myService',
+		'$ balena device restart 23c73a1 -s myService1,myService2',
+	];
+
+	public static args: Array<IArg<any>> = [
+		{
+			name: 'uuid',
+			description:
+				'comma-separated list (no blank spaces) of device UUIDs to restart',
```
pdcastro, line 62: I'd remove `'to restart'` because it sounds like "the UUIDs would be restarted", or the devices would be restarted, when actually it is the (selected) containers that would be restarted.
```
description:
	'comma-separated list (no blank spaces) of device UUIDs',
```

lib/actions-oclif/device/restart.ts
```diff
@@ -0,0 +1,197 @@
...
+		});
+		const serviceNames = options.service?.split(',');
+
+		// Iterate sequentially through deviceIds.
+		// We may later want to add a batching feature,
+		// so that n devices are processed in parallel
+		for (const deviceId of deviceIds) {
+			ux.action.start(`Restarting services on device ${deviceId}`);
+			if (serviceNames) {
+				await this.restartServices(balena, deviceId, serviceNames);
```
pdcastro, line 100: This will be restarting devices sequentially, rather than in parallel. Even for just 10 devices, this could be quite slow. I think we should have some `Promise.all()` call that initiate the `restartServices` calls in parallel. Ideally with a concurrency limit, and a report of the devices that failed, and devices that succeeded (like I mentioned in another comment), so that users don't need to repeat the command line for all devices if the `restartServices` called failed for just one of the devices.

pdcastro (approved):

Issue #2040, opened 2020-09-19
balena deploy exits with "Everything is up to date" and does not deploy


pdcastro:
Bug introduced in CLI v12.14.18 / PR #2013 
```
$ balena deploy test-rpi
[Info]    Everything is up to date (use --build to force a rebuild)
```

Expected behavior:
```
$ balena deploy test-rpi
[Info]    Everything is up to date (use --build to force a rebuild)
[Info]    Creating release...
[Info]    Pushing images to registry...
[Info]    Saving release...
[Success] Deploy succeeded!
...
```

pdcastro:
Fixed in CLI v12.22.2.

PR #2041, opened 2020-09-19
deploy: Fix unexpected exit with "Everything is up to date"

pdcastro:
A `return` statement was accidentally left behind during the conversion from promise chains to async/away in PR #2013.

Resolves: #2040
Change-type: patch

### DIFF
--- a/lib/actions-oclif/deploy.ts
+++ b/lib/actions-oclif/deploy.ts
@@ -243,30 +243,30 @@ ${dockerignoreHelp}
 				compositionToBuild.services,
 				servicesToSkip,
 			);
+			let builtImagesByService: Dictionary<any> = {};
 			if (_.size(compositionToBuild.services) === 0) {
 				logger.logInfo(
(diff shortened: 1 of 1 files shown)


lib/actions-oclif/deploy.ts
```diff
@@ -243,30 +243,30 @@ ${dockerignoreHelp}
 				compositionToBuild.services,
 				servicesToSkip,
 			);
+			let builtImagesByService: Dictionary<any> = {};
 			if (_.size(compositionToBuild.services) === 0) {
 				logger.logInfo(
 					'Everything is up to date (use --build to force a rebuild)',
 				);
-				return {};
```
pdcastro, line 250: This `return` statement made sense when this block was in a promise chain, but should have been removed in the conversion to async/away. Check this line in PR #2013:
https://github.com/balena-io/balena-cli/pull/2013/files#diff-cd710231a0e1573249414ab3ebf325c7L253

pdcastro:
@balena-ci I self-certify!

Issue #2035, opened 2020-09-16
login: Sentry reports a large number of BalenaRequestError Request error: Unauthorized


pdcastro:
Sentry reports a large number of `BalenaRequestError` `Request error: Unauthorized` errors:
* https://sentry.io/organizations/balena/issues/747313108/?project=149239
* https://sentry.io/organizations/balena/issues/1836332147/?project=149239

They come from the `login` command. I believe these are "harmless" and "expected" errors resulting from expired tokens, in which case the CLI should dismiss them without reporting to Sentry.

srlowe:
@pdcastro Not sure if I'm looking in the wrong place, but both of those links you added point me to an incidence of:
```
BalenaRequestError

Request error: tunneling socket could not be established, statusCode=500
```
Triggered by `balena ssh`

srlowe:
Ah, I found them : https://sentry.io/organizations/balena/issues/747313108/?project=149239

PR #2047, opened 2020-09-25
Fix issue with subclassed errors and instanceof

srlowe:
This works around a limitation of typescript regarding `instanceof` and extended builtins (like errors).  This was causing the `is(error instanceof ExpectedError)` test to fail leading to errors being submitted to sentry which should not have been.  

Have also added `Request error: Unauthorized` to the list of error patterns to match and ignore.

Change-type: patch
Resolves: #2035

### DIFF
--- a/lib/errors.ts
+++ b/lib/errors.ts
@@ -23,19 +23,36 @@ import { getHelp } from './utils/messages';
 
 export class ExpectedError extends TypedError {}
 
-export class NotLoggedInError extends ExpectedError {}
+// Note: Because of a nasty TS limitation re extending builtins, instanceof was not working correctly with
+// subclassed errors.  The workaround is to set their prototypes manually.
+// https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+// https://github.com/microsoft/TypeScript/issues/13965
 
-export class InsufficientPrivilegesError extends ExpectedError {}
+export class NotLoggedInError extends ExpectedError {
+	constructor(m: string) {
+		super(m);
+		Object.setPrototypeOf(this, ExpectedError.prototype);
+	}
+}
+
+export class InsufficientPrivilegesError extends ExpectedError {
+	constructor(m: string) {
+		super(m);
+		Object.setPrototypeOf(this, ExpectedError.prototype);
+	}
+}
 
 export class InvalidPortMappingError extends ExpectedError {
 	constructor(mapping: string) {
 		super(`'${mapping}' is not a valid port mapping.`);
+		Object.setPrototypeOf(this, ExpectedError.prototype);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


lib/errors.ts
```diff
@@ -23,19 +23,36 @@ import { getHelp } from './utils/messages';
 
 export class ExpectedError extends TypedError {}
 
-export class NotLoggedInError extends ExpectedError {}
+// Note: Because of a nasty TS limitation re extending builtins, instanceof was not working correctly with
+// subclassed errors.  The workaround is to set their prototypes manually.
+// https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+// https://github.com/microsoft/TypeScript/issues/13965
```
pdcastro, line 29: Ouch. Could we at least do the `setPrototypeOf` trick to the `ExpectedError` class, and then avoid having to do it for subclasses of `ExpectedError`?

Also, I wonder if there is something that [our own `instanceOf` function](https://github.com/balena-io/balena-cli/blob/v12.22.2/lib/errors.ts#L65) could do work better than the built-in `instanceof` in this regard. (Even if there was though, we might still want/need the `setPrototypeOf` trick.)

srlowe: Do you mean just set it once in ExpectedError? No I don't think that will work according to MS advisory.

srlowe: "However, any subclass of FooError will have to manually set the prototype as well."
https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work

srlowe: I thought about adding to the instanceOf function, but I don't think there's anything it can test. (I guess if there were, then the compiler could implement this, and fix the situation)

pdcastro: Double ouch. Does `TypedErorr` (npm `typed-error` package) implement these `setPrototypeOf` workarounds? Because the functionality provided by `TypedError` might be broken too, and presumably we rely on something they provide. It's even tempting not to subclass the built-in classes, though we might end up with `instanceof Error` returning `false`. :-/

srlowe: I had a quick look before, and I'm not sure - it does set the prototype but not sure if it's to this effect.

@Page- I think you were a contributor to typed-error.  Could you chime in?

srlowe: It's doing `Object.setPrototypeOf(this, new.target.prototype)`

Page-: `typed-error` isn't affected by this because it's either an old version that did manual extension (via a somewhat convoluted es5 method) or it's a modern version that requires es6 and doesn't rely on typescript downleveling so these `Object.setPrototypeOf` changes are unnecessary

Page- (changes requested):

Issue #2060, opened 2020-10-11
RangeError: Maximum call stack size exceeded (@sentry/node/dist/integrations/console.js)


pdcastro:
```
$ balena push 192.168.1.149 -m
...
[Build]   [main] Removing intermediate container 4612f12b4ac1
[Build]   [main]  ---> 7dedd168ac81
[Build]   [main] Successfully built 7dedd168ac81
[Build]   [main] Successfully tagged local_image_main:latest
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
RangeError: Maximum call stack size exceeded
    at Object.error (/snapshot/balena-cli/node_modules/@sentry/node/dist/integrations/console.js:72:36)
    at Object.printErrorMessage (/snapshot/balena-cli/build/errors.js:140:13)
    at Object.handleError (/snapshot/balena-cli/build/errors.js:122:17)
    at Object.run (/snapshot/balena-cli/build/app-oclif.js:34:73)
    at async Object.run (/snapshot/balena-cli/build/app.js:17:5)
The terminal process "/bin/bash '-c', 'balena push 192.168.1.149 -m'" terminated with exit code: 1.
```

Originally reported at: https://www.flowdock.com/app/rulemotion/i-cli/threads/NDIXJP3vrEsNHSz57tuO9UneDNw

pdcastro:
This issue is understood to have been fixed in CLI release v12.25.0 (PR #2061).

PR #2061, opened 2020-10-11
Update Sentry, add BALENARC_NO_SENTRY var, refactor CLI initialization

pdcastro:
This PR updates the Sentry package, introduces the BALENARC_NO_SENTRY env var to disable Sentry, and does a refactoring of CLI initialization code. The refactoring consists of:
* Deleting the `app-common.ts` and `app-oclif.ts` files, which are inheritance from the time when we had both Capitano and oclif at the same time (there used to be an `app-capitano.ts` file that was deleted a while ago, when Capitano was removed).
* Moving proxy-specific code from `app-common.ts` to a new `utils/proxy.ts` file.
* Moving some code that runs very early on (even before the preparser) to a new `utils/bootstrap.ts` file.
* Moving the remaining code to the pre-existing `app.ts` file. Both `app.ts` and `app-oclif.ts` contained an entrypoint function named `run()`. The `run` function in `app.ts` was preserved, and the one from `app-oclif` was renamed `oclifRun`.

Connects-to: #2060
Change-type: minor


lib/utils/bootstrap.ts
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param modFunc Path to a JS module that will be executed via require().
+ * The modFunc argument may optionally contain a function name separated
+ * by '::', for example '::main' in:
+ * 'C:\\snapshot\\balena-cli\\node_modules\\windosu\\lib\\pipe.js::main'
+ * in which case that function is executed in the require'd module.
+ * @param args Optional arguments to passed through process.argv and as
+ * arguments to the function specified via modFunc.
+ */
+export async function pkgExec(modFunc: string, args: string[]) {
```
pdcastro, line 84: The `os initialize` command uses this `pkgExec` function on Windows via a patched `windosu` dependency. I've tested it and it turns out to be broken, but not because of this PR -- broken since CLI v12.14.7 / PR #1994. I have created issue #2062 for further investigation.

lib/utils/oclif-utils.ts
```diff
@@ -48,6 +49,17 @@ export class CommandHelp {
 	}
 }
 
+export class CustomMain extends Main {
```
pdcastro, line 52: Moved here from `app-common.ts`, which was deleted.

lib/utils/proxy.ts
```diff
@@ -104,7 +44,7 @@ type ProxyConfig = string | GlobalTunnelNgConfig;
  * 'localhost' and '127.0.0.1' are always excluded. If NO_PROXY is not defined,
  * default exclusion patterns are added for all private IPv4 address ranges.
  */
-async function setupGlobalHttpProxy(settings: CliSettings) {
+export async function setupGlobalHttpProxy(settings: CliSettings) {
```
pdcastro, line 47: `git` thinks that `app-common.ts` was "renamed" `utils/proxy.ts`, because so much of the content is identical. Only the proxy-specific code was moved to `utils/proxy.ts` though. It was some 150 lines of code, which felt tidier in its own `proxy.ts` file rather than in `app.ts`.

package.json
```diff
@@ -197,7 +197,7 @@
     "@balena/es-version": "^1.0.0",
     "@oclif/command": "^1.8.0",
     "@resin.io/valid-email": "^0.1.0",
-    "@sentry/node": "^5.21.1",
+    "@sentry/node": "^5.25.0",
```
pdcastro, line 200: I have not actually seen any evidence that this version upgrade fixes issue #2060, but one can always hope... Even if it doesn't, the introduction of the `BALENARC_NO_SENTRY` env var should aid in debugging.

srlowe (approved):
Nice to have the app-xxx.ts files consolidated!

Issue #1547, opened 2019-12-23
Purge data not available


shaunmulligan:
One should be able to purge data on devices from the cli

pdcastro:
Related to #583 [Implement new device commands](https://github.com/balena-io/balena-cli/issues/583)

srlowe:
@pdcastro @shaunmulligan I guess the two obvious options are:

    balena device purge <uuid> (new command)
    balena device --purge <uuid> (expand responsibilities of existing command)

One benefit of the first, is that we can more simply implement support for multiple devices. Also commands with multiple responsibilities will be harder to align with the concept of auto-generation.

srlowe:
@pdcastro Given your answer on the `device deactivate` issue, I guess you agree that #1 is best?

pdcastro:
Yes, it would be consistent. :-)

PR #2066, opened 2020-10-14
Add command device purge

srlowe:
Change-type: minor
Resolves: #1547
 

---


doc/cli.markdown
```diff
@@ -174,20 +174,21 @@ Users are encouraged to regularly update balenaCLI to the latest version.
...
 - Device
 
+	- [device &#60;uuid&#62;](#device-uuid)
 	- [device identify &#60;uuid&#62;](#device-identify-uuid)
 	- [device init](#device-init)
-	- [device &#60;uuid&#62;](#device-uuid)
 	- [device move &#60;uuid(s)&#62;](#device-move-uuid-s)
+	- [device os-update &#60;uuid&#62;](#device-os-update-uuid)
+	- [device public-url &#60;uuid&#62;](#device-public-url-uuid)
+	- [](#)
```
pdcastro, line 183: `device purge` is missing here, for some reason.

lib/preparser.ts
```diff
@@ -162,6 +162,7 @@ export const oclifCommandIds = [
 	'device:move',
 	'device:os-update',
 	'device:public-url',
+	`device:purge`,
```
pdcastro, line 165: Not for this PR, but a thought: `oclif.manifest.json` contains this list of commands, and we already generate it during the CLI build (we should double check that it is available in all installation types, e.g. for the standalone zip, but I think it is). In this case, we could delete this command list from `preparser.ts` and use the json file - perhaps patching oclif to avoid re-loading and parsing the json file twice.

pdcastro (changes requested):

lib/preparser.ts
srlowe: Yeah I had the same thought.  The downside would be that of course the app architecture would not function at all without the manifest (which technically is optional), but the tradeoff is probably worth it.  Note that this is needed because of our use of spaces instead of colons (this issue crops up a lot, and has to be hacked - e.g. in help).   There's some talk about oclif supporting this in the future (which would be great, as we could remove all these hacks), but I wouldn't bet on it.

pdcastro (approved):

Issue #1842, opened 2020-05-26
Resource names that look like numbers are treated as IDs / UUIDs


pdcastro:
Some CLI commands currently distinguish between a resource name or ID or short UUID on the basis of whether it "looks like an integer". This used to be a side effect of the Capitano framework, and we thought that moving to oclif would solve it. Alas, after the migration to oclif, we still found ourselves applying a [`tryAsInteger()`](https://github.com/balena-io/balena-cli/blob/v11.35.9/lib/utils/validation.ts#L92) function (#1815 → #1817, #1841) because the affected commands do not currently have a way of distinguishing between a resource ***name*** (string) like `123456` and a resource ***ID*** (number) like `123456`. 

```
balena app 123456
balena tag set myTag --release 123456
```

The balena SDK, e.g. `application.get(nameOrSlugOrId, [options])` has arguments like `nameOrSlugOrId` that distinguish between a name or ID on the basis or the argument type being a number (ID) or string (name or slug). This is OK, as it is up to the CLI (or web dashboard) to determine whether the user input is for a name or an ID. The CLI can then apply the required data type (`123456` may be provided to the SDK as either a string or a number).

https://www.balena.io/docs/reference/sdk/node-sdk/#balena.models.application.get

I think the solution is to add command-line options to explicitly distinguish between ID / UUID / name, even if they are "optional options" in which absence the CLI applies `tryAsInteger()`. For example:

```
balena app 123456  # assumes number with `tryAsInteger()`
balena app --id 123456  # applies `parseInt()` (or similar)
balena app --name 123456  # do not convert to number
```

srlowe:
@pdcastro  Another possible approach: We could have a convention where we use quotes to force the parser to treat them as strings in these ambiguous cases.  e.g. `balena app "12345"`

pdcastro:
> _use quotes to force the parser to treat them as strings_

One downside of this approach is that the quotes would have to be escaped on a command line (so they get through to the CLI process), and different shells have different conventions on how to escape quotes (e.g. bash vs cmd.exe vs PowerShell):
```
# bash:
$ balena app \"123456\"
$ balena app \'123456\'
$ balena app '"123456"'

$ cmd.exe uses the caret ^
$ balena app ^"123456^"

$ PowerShell uses the backtick `
$ balena app `"123456`"
$ balena app '"123456"'
```

[...]

PR #2067, opened 2020-10-14
Add command app purge

srlowe:
Change-type: minor

### DIFF
--- a/lib/commands/app/purge.ts
+++ b/lib/commands/app/purge.ts
@@ -31,11 +31,12 @@ interface ArgsDef {
 
 export default class AppRestartCmd extends Command {
 	public static description = stripIndent`
-		Restart an application.
+		Purge data from an application.
 
-		Restart all devices belonging to an application.
+		Purge data from all devices belonging to an application.
+		This will clear the application's /data directory.
 `;
-	public static examples = ['$ balena app restart MyApp'];
+	public static examples = ['$ balena app purge MyApp'];
 
 	public static args = [
 		{
@@ -45,7 +46,7 @@ export default class AppRestartCmd extends Command {
 		},
 	];
 
-	public static usage = 'app restart <name>';
+	public static usage = 'app purge <name>';
 
 	public static flags: flags.Input<FlagsDef> = {
 		help: cf.help,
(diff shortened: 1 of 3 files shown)


lib/commands/app/purge.ts
```diff
@@ -0,0 +1,82 @@
...
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(AppRestartCmd);
+
+		const balena = getBalenaSdk();
+
+		// balena.models.application.purge only accepts a numeric id
+		// so we must first fetch the app to get it's id, if we have been given a name
+		let nameOrId = tryAsInteger(params.name);
+
+		if (typeof nameOrId === 'string') {
```
pdcastro, line 66: As currently coded, I think this will fail for an app **name** like `123`.  The probability of failure is reduced by trying to query `application.get(nameOrId)` twice, first assuming that `123` is a numeric ID, having it fail, and then querying a second time casting `123` to a string, thus assuming it is a name. But even this approach is not guaranteed to succeed, because someone could have named an app with a number that coincides with an actual app ID (7 digits long?).

lib/commands/app/purge.ts
```diff
@@ -0,0 +1,82 @@
...
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(AppRestartCmd);
+
+		const balena = getBalenaSdk();
+
+		// balena.models.application.purge only accepts a numeric id
+		// so we must first fetch the app to get it's id, if we have been given a name
+		let nameOrId = tryAsInteger(params.name);
```
pdcastro, line 64: There is this issue: https://github.com/balena-io/balena-cli/issues/1842

With pre-existing commands, we have been using `tryAsInteger` as a way of maintaining backwards compatibility. But this PR introduces a new command, so it is an opportunity of "doing it right", starting by deciding what this means. :-)

By the way, I believe the API recently introduced app UUIDs:
https://github.com/balena-io/open-balena-api/pull/491

Not sure if app UUIDs are in production yet, but if they are, this PR is also an opportunity of introducing their usage in the CLI.

srlowe: > But this PR introduces a new command, so it is an opportunity of "doing it right", starting by deciding what this means. :-)

(just copying what I said in FD) I agree that we need to address that, but this command doesn't introduce a new problem (we already have it in multiple commands) so I think it should be merged.  I think we should then have a wider discussion about how to address this at the framework level (what conventions to adopt), and get consensus from @shaunmulligan and others.

> Not sure if app UUIDs are in production yet, but if they are, this PR is also an opportunity of introducing their usage in the CLI.

There's no mention of app UUIDs in the SDK documention, so I don't think so.

BTW - we are having to 'manually' support appNames in this command, because purge only supports IDs - not sure what the situation will be with app UUIDs, but I hope that they will be implemented across the board in the SDK (for the sake of autogeneration of the CLI).

pdcastro (approved):

Issue #1053, opened 2018-12-26
--buildArg flag doesn't work for docker-compose.yml projects


shaunmulligan:
- **Cli version:** v9.6.0
- **Standalone install:** no
- **Node version:** v10.13.0
- **Npm version:** v6.4.1
- **Operating system and architecture:** macOS 10.13.6 

For both `build` and `deploy` the `--buildArg` flag which allows you to define `ARG` variables in your dockerfile only applies when you have a single container application. If one specifies a docker-compose.yml the --buildArg never propagates to any of the services. Below is a simple project to reproduce the issue. 
[testBuildArgs.zip](https://github.com/balena-io/balena-cli/files/2709118/testBuildArgs.zip)

If one runs `balena deploy nuc-staging --nocache --build --buildArg ACCESS_KEY=mySecret --logs` from the root of `testBuildArgs` folder, the `mySecret` variable will not be echo'd in the build logs, however, if the same command is run from `testBuildArgs/module1` folder the the `mySecret` is echo'd as expected.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_g9bv)

pdcastro:
@Wachiwi, I haven't started working on this issue yet, so I'd be very happy for you to fix it! I've myself only recently started getting familiar with the relevant codebase. The hints I can think of at the moment are simply links to some lines of code where the build args are manipulated.

The `balena build` and `balena deploy` commands will call `compose.buildProject` at the following lines of code:

* https://github.com/balena-io/balena-cli/blob/master/lib/actions/build.coffee#L30
* https://github.com/balena-io/balena-cli/blob/master/lib/actions/deploy.coffee#L49

Those calls land here:

* https://github.com/balena-io/balena-cli/blob/master/lib/utils/compose.coffee#L162

And eventually reach:

* https://github.com/balena-io/balena-cli/blob/master/lib/utils/compose.coffee#L222
```js
	if d.image.context?.args?
		task.dockerOpts.buildargs ?= {}
		_.merge(task.dockerOpts.buildargs, d.image.context.args)
```

And eventually call `builder.performBuilds`:

* https://github.com/balena-io/balena-cli/blob/master/lib/utils/compose.coffee#L270

Now, that call lands in an imported repo/package:

* https://www.npmjs.com/package/resin-multibuild
[...]

PR #2069, opened 2020-10-20
Fix `--buildArg` with compose projects; Convert `buildProject` to Typescript

pdcastro:
Resolves: #1053 
Change-type: patch


lib/utils/compose.js
```diff
@@ -641,8 +365,7 @@ export const pushAndUpdateServiceImages = function (
 		Promise.all(
 			images.map(({ serviceImage, localImage, props, logs }, index) =>
 				Promise.all([
-					// @ts-ignore
-					localImage.inspect().get('Size'),
+					localImage.inspect().then((img) => img.Size),
```
pdcastro, line 368: Typing info for `docker-toolbelt` declares native promises, but apparently it actually uses Bluebird promises and the old CLI code was relying on Bluebird being used -- the `.get('Size')` call was Bluebird's `.get`:
http://bluebirdjs.com/docs/api/get.html
This type _vs._ usage incompatibility was ignored with a `"// @ts-ignore"` comment. I've changed the code here and elsewhere not to assume Bluebird, and removed the `ts-ignore` comment.

lib/utils/compose.js
```diff
@@ -85,23 +85,6 @@ export function createProject(composePath, composeStr, projectName = null) {
 	};
 }
 
-/**
```
pdcastro, line 88: The deleted code in `compose.js` was converted to Typescript and moved to the pre-existing `compose_ts.ts` file.

lib/utils/compose_ts.ts
```diff
@@ -131,6 +130,372 @@ async function resolveProject(
...
+	arch: string;
+	deviceType: string;
+	emulated: boolean;
+	buildOpts: import('./docker').BuildOpts;
+	inlineLogs?: boolean;
+	convertEol: boolean;
+	dockerfilePath?: string;
+	nogitignore: boolean;
+	multiDockerignore: boolean;
+}): Promise<BuiltImage[]> {
```
pdcastro, line 158: `buildProject` was very large single function written in a `.then` promise chain. Such large functions are harder to test (if we wanted to...) and harder to make sense of, in my opinion. I've split it in some 7 thematic smaller functions, mostly mapping to `.then()` blocks of the old promise chain. Even then, what was left in `buildProject` itself is still of a significant size!

lib/utils/compose_ts.ts
```diff
@@ -313,14 +313,23 @@ function setTaskAttributes({
...
-			if (d.image.args) {
-				task.dockerOpts.buildargs ??= {};
-				_.merge(task.dockerOpts.buildargs, d.image.args);
-			}
+		}
+		// reassign task.args so that the `--buildArg` flag takes precedence
+		// over assignments in the docker-compose.yml file (service.build.args)
+		task.args = {
+			...task.args,
+			...buildOpts.buildargs,
```
pdcastro, line 323: This fixes issue #1053. Previously, `d.image.args` overwrote the `--buildArg` command-line option(s). Now `--buildArg` takes precedence, in line with the behavior of the standard `docker-compose` command-line tool.

tests/test-data/projects/docker-compose/basic/docker-compose.yml
```diff
@@ -12,3 +12,5 @@ services:
     build:
       context: ./service2
       dockerfile: Dockerfile-alt
+      args:
+        - 'COMPOSE_ARG=an argument defined in the docker-compose.yml file'
```
pdcastro, line 16: This added variable in this test `docker-compose.yml` file allows us to test that the `--buildArg` command-line argument takes precedence.

srlowe (approved):

Issue #2068, opened 2020-10-19
Failed to preload image on openBalena


openedhardware:
# Specifications
- **balenaCLI version:** e.g. 1.2.3 (output of the `"balena version -a"` command)
```shell script
balena-cli version "12.11.0"
Node.js version "10.21.0"
```
- **Cloud backend: openBalena or balenaCloud?** If unsure, it will be balenaCloud
openBalena v2.0.0
- **Operating system version:** e.g. Windows 10, Ubuntu 18.04, macOS 10.14.5
Ubuntu 20.04
- **32/64 bit OS and processor:** e.g. 32-bit Windows on 64-bit Intel processor
64 bit
- **Install method:** npm or zip package or executable installer
executable installer

I was able to deploy an application by using `balena deploy` command.

But cannot preload the deployed application to an img file:
```shell script
ubuntu@ubuntu-desktop:/media$ balena apps
ID APP NAME  SLUG      DEVICE TYPE  ONLINE DEVICES DEVICE COUNT
1  yokosoApp yokosoapp raspberry-pi 0              0
ubuntu@ubuntu-desktop:/media$ balena preload ../balena-rpi-configured.img --app 1 --commit current --splash-image splash.jpg --debug
Building Docker preloader image. [===                     ] 12%
Step 1/7 : FROM docker:17.12.0-ce-dind
Building Docker preloader image. [======                  ] 25%
Step 2/7 : RUN apk update && apk add --no-cache python3 parted btrfs-progs util-linux sfdisk file coreutils sgdisk
 ---> Using cache
Building Docker preloader image. [=========               ] 37%
Step 3/7 : COPY ./requirements.txt /tmp/
 ---> Using cache
Building Docker preloader image. [============            ] 50%
Step 4/7 : RUN pip3 install -r /tmp/requirements.txt
 ---> Using cache
Building Docker preloader image. [===============         ] 62%
Step 5/7 : COPY ./src /usr/src/app
 ---> Using cache
Building Docker preloader image. [==================      ] 75%
Step 6/7 : WORKDIR /usr/src/app
 ---> Using cache
Building Docker preloader image. [=====================   ] 87%
Step 7/7 : CMD ["python3", "/usr/src/app/preload.py"]
 ---> Using cache
 ---> 95c91d1abee6
Successfully built 95c91d1abee6
Building Docker preloader image. [========================] 100%
| Checking that the image is a writable file
| Finding a free tcp port and getting balena settings
[...]

Issue #2063, opened 2020-10-13
BalenaApplicationNotFound for valid app id - preload cmd


nslocum:
The `preload` cmd returns error `BalenaApplicationNotFound` when i pass an application id:

```shell
balena preload $IMAGE_FILENAME \
    --app $APP_ID \
    --commit current \
    --splash-image "$DIR/logo.png"
```

Returns error: 

`Application not found: 1234567 (BalenaApplicationNotFound)`

It works fine when I use the app name instead of its id:

```shell
balena preload $IMAGE_FILENAME \
    --app $APP_NAME \
    --commit current \
    --splash-image "$DIR/logo.png"
```

I get the app id using:

```shell
APP_ID=$(balena apps | grep " $APP_NAME " | awk '{ print $1 }')
```

This error really isn't an issue in my case, using the app name is actually easier. As an aside, I'd even prefer that the app ids be internal only and not exposed to any customers or customer APIs. By itself this isn't a serious security issue, but it needlessly exposes info that an attacker could easily leverage.

pdcastro:
Fixed in CLI v12.33.1

pdcastro:
> As an aside, I'd even prefer that the app ids be internal only and not exposed to any customers or customer APIs. By itself this isn't a serious security issue, but it needlessly exposes info that an attacker could easily leverage.

👍   This issue has been closed re preload command, but another related generic issue re app IDs remains open: #1843

PR #2070, opened 2020-10-20
Fix preload --add-certificate, amend help for `version`, and revert `balenaCLI` styling in docs

pdcastro:
* Revert `'balenaCLI'` styling (without space) as decided in:  
https://www.flowdock.com/app/rulemotion/r-product/threads/xrtkAC3uWDMJ4-4GkJL_Hmn8Jvb
* The update to `preload` help (`--app`) connects to issue #2068 
* The update to `version` help (Node.js version) connects to issue #2063 

Change-type: patch

### DIFF
--- a/lib/commands/preload.ts
+++ b/lib/commands/preload.ts
@@ -43,7 +43,7 @@ interface FlagsDef extends DockerConnectionCliFlags {
 	'splash-image'?: string;
 	'dont-check-arch': boolean;
 	'pin-device-to-release': boolean;
-	'add-certificate'?: string;
+	'add-certificate'?: string[];
 	help: void;
 }
 
@@ -84,7 +84,7 @@ export default class PreloadCmd extends Command {
 
 	public static flags: flags.Input<FlagsDef> = {
 		app: flags.string({
-			description: 'name, slug or numeric ID of the application to preload',
+			description: 'name of the application to preload',
 			char: 'a',
 		}),
 		commit: flags.string({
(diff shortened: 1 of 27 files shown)


lib/commands/preload.ts
```diff
@@ -115,6 +115,7 @@ The file name must end with '.crt' and must not be already contained in the prel
 /etc/ssl/certs folder.
 Can be repeated to add multiple certificates.\
 `,
+			multiple: true,
```
pdcastro, line 118: I think the `--add-certificate` option got broken in the conversion to oclif. `preload --help` and the forums specify that it _"can be repeated to add multiple certificates,"_ and presumably this worked with Capitano, but oclif requires `multiple: true` to be specified.

lib/commands/preload.ts
```diff
@@ -201,14 +202,7 @@ Can be repeated to add multiple certificates.\
...
 
-		let certificates: string[];
-		if (Array.isArray(options['add-certificate'])) {
-			certificates = options['add-certificate'];
-		} else if (options['add-certificate'] === undefined) {
-			certificates = [];
-		} else {
-			certificates = [options['add-certificate']];
-		}
+		const certificates: string[] = options['add-certificate'] || [];
```
pdcastro, line 205: With oclif `multiple: true`, `options['add-certificate']` will be either `undefined` or an array (never a string).

pdcastro:
@balena-ci retest

srlowe (approved):

PR #2073, opened 2020-10-26
Remove need for hardcoded list of command ids

srlowe:
This should also help solve the issues with capitanodoc/oclif integration.

Change-type: patch

### DIFF
--- a/lib/utils/oclif-utils.ts
+++ b/lib/utils/oclif-utils.ts
@@ -69,3 +69,17 @@ export function capitanoizeOclifUsage(
 		.replace(/(?<=\s)[A-Z]+(?=(\s|$))/g, (match) => `<${match}>`)
 		.toLowerCase();
 }
+
+export async function getCommandsFromManifest() {
+	const manifest = require('../../oclif.manifest.json');
+
+	if (manifest.commands == null) {
+		throw new Error('Commands section not found in manifest.');
+	}
+	return manifest.commands;
+}
+
+export async function getCommandIdsFromManifest() {
+	const commands = await getCommandsFromManifest();
+	return Object.keys(commands);
+}
--- a/lib/preparser.ts
+++ b/lib/preparser.ts
@@ -70,7 +70,7 @@ export async function preparseArgs(argv: string[]): Promise<string[]> {
 	let args = cmdSlice;
 
 	// Convert space separated subcommands (e.g. `end add`), to colon-separated format (e.g. `env:add`)
-	if (isSubcommand(cmdSlice)) {
+	if (await isSubcommand(cmdSlice)) {
 		// convert space-separated commands to oclif's topic:command syntax
 		args = [cmdSlice[0] + ':' + cmdSlice[1], ...cmdSlice.slice(2)];
 		Logger.command = `${cmdSlice[0]} ${cmdSlice[1]}`;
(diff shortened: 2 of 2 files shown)


lib/utils/oclif-utils.ts
```diff
@@ -69,3 +69,28 @@ export function capitanoizeOclifUsage(
...
+			throw e;
+		}
+	}
+	return manifest.commands;
+}
+
+export async function getCommandIdsFromManifest() {
+	const commands = await getCommandsFromManifest();
+	if (commands == null) {
+		throw new Error('Commands second not found in manifest.');
```
pdcastro, line 93: What does _"Commands second"_ mean? Would this error be raised if a developer just added a new oclif command and ran the CLI before running `npm run build` (i.e. the manifest hasn't been updated for the new command). Maybe this error message could also suggest the fix:
```
throw new Error(stripIndent`
    Commands second not found in ${manifestPath}.
    If you are developing the CLI, run \`npm run build\` to fix it.`);
```

pdcastro (changes requested):

lib/utils/oclif-utils.ts
```diff
@@ -69,3 +69,28 @@ export function capitanoizeOclifUsage(
 		.replace(/(?<=\s)[A-Z]+(?=(\s|$))/g, (match) => `<${match}>`)
 		.toLowerCase();
 }
+
+export async function getCommandsFromManifest() {
+	const fs = await import('fs');
+	let manifest;
+	try {
+		manifest = JSON.parse(
+			await fs.promises.readFile('./oclif.manifest.json', 'utf8'),
```
pdcastro, line 78: Ah, somehow one of my review comments (the reason for requesting changes!) got lost. Hang on, I am writing it again...

pdcastro: I downloaded the standalone zip file from the draft release page and got:
```
$ ./balena-cli-12.25.5/balena version -a
Oclif manifest not found.

Additional information may be available with the `--debug` flag.
For help, visit our support forums: https://forums.balena.io
For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/
```
The solution might be to test for `process.pkg` and conditionally determine the path for `oclif.manifest.json`, along the lines of:
https://github.com/balena-io/balena-cli/blob/v12.25.4/patches/all/open+7.1.0.patch#L10-L12

The standalone executable tests may have passed because they run on the same folder as the regular tests (I guess), so `./oclif.manifest.json` happened to exist on the current working dir.

lib/utils/oclif-utils.ts
```diff
@@ -69,3 +69,28 @@ export function capitanoizeOclifUsage(
...
+		);
+	} catch (e) {
+		if (e.code === 'ENOENT') {
+			throw new Error('Oclif manifest not found.');
```
pdcastro, line 82: Although this should never happen in normal conditions, I suggest rewording the error message a bit:
```
throw new Error(stripIndent`
    ${manifestPath} not found.
    If you are developing the CLI, run \`npm run build\` to fix it.`);
```
... where `manifestPath` is the conditionally determined path for `oclif.manifest.json`.

pdcastro (changes requested):

lib/utils/oclif-utils.ts
srlowe: Ah, that should be `section`.   I don't think that scenario would cause it though.  It would only throw if the manifest file was present, and valid JSON, but just missing that section.  I'm not sure how that might happen, but I'm feeling a bit cautious relying on this manifest, so want to cover all bases.

lib/utils/oclif-utils.ts
srlowe: Ah! Yeah got caught out there.

lib/utils/oclif-utils.ts
```diff
@@ -69,3 +69,17 @@ export function capitanoizeOclifUsage(
 		.replace(/(?<=\s)[A-Z]+(?=(\s|$))/g, (match) => `<${match}>`)
 		.toLowerCase();
 }
+
+export async function getCommandsFromManifest() {
```
pdcastro, line 73: NIT: These functions (`getCommandsFromManifest`, `getCommandIdsFromManifest`) are no longer async (`require` is synchronous). The processing of promises with async/await has a small performance penalty.

pdcastro (approved):

Issue #2075, opened 2020-10-27
ssh: Found more than one container with a service name xxx


pdcastro:
```
$ ./bin/balena-dev ssh 192.168.1.36 foo
Found more than one container with a service name foo.
This state is not supported, please contact support.
```

This happens if two containers exist with names like "foo" and "freefoo", because of a regex that is too loose:
https://github.com/balena-io/balena-cli/blob/v12.25.5/lib/utils/device/ssh.ts#L52

PR #2076, opened 2020-10-28
ssh: Fix "Found more than one container with a service name <name>"

pdcastro:
Resolves: #2075 
Change-type: patch

### DIFF
--- a/lib/utils/device/ssh.ts
+++ b/lib/utils/device/ssh.ts
@@ -49,7 +49,7 @@ export async function performLocalDeviceSSH(
 			port: 2375,
 		});
 
-		const regex = new RegExp(`\\/?${escapeRegExp(opts.service)}_\\d+_\\d+`);
+		const regex = new RegExp(`(^|\\/)${escapeRegExp(opts.service)}_\\d+_\\d+`);
 		const nameRegex = /\/?([a-zA-Z0-9_]+)_\d+_\d+/;
 		let allContainers: ContainerInfo[];
 		try {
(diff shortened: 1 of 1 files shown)


lib/utils/device/ssh.ts
```diff
@@ -49,7 +49,7 @@ export async function performLocalDeviceSSH(
 			port: 2375,
 		});
 
-		const regex = new RegExp(`\\/?${escapeRegExp(opts.service)}_\\d+_\\d+`);
+		const regex = new RegExp(`(^|\\/)${escapeRegExp(opts.service)}_\\d+_\\d+`);
```
pdcastro, line 52: This is the key fix. Previously, for a command like:
```
balena ssh 192.168.1.36 foo
```
... the regex would match for containers named `/freefoo_1_1` and `/foo_2_1` (services named "freefoo" and "foo"), but we want it to match the latter only.

**While my tests were successful for a recent version of balenaOS and balenaEngine, I don't know if different / older versions would return container names in different formats and cause the regex to fail, in particular the caret character to indicate "beginning of string".** The previous regex had a question mark that suggests that the forward slash is optional. Was the forward slash absent for other versions of balenaEngine? If it was absent, was there anything else in its place?  Reviewer feedback is welcome - @CameronDiver ? :-)

lib/utils/device/ssh.ts
```diff
@@ -61,21 +61,20 @@ export async function performLocalDeviceSSH(
 		}
 
 		const serviceNames: string[] = [];
-		const containers = allContainers
-			.map((container) => {
-				for (const name of container.Names) {
-					if (regex.test(name)) {
-						return { id: container.Id, name };
-					}
+		const containers: Array<{ id: string; name: string }> = [];
```
pdcastro, line 64: This change was to avoid `undefined` being part of the type of the `containers` array. Before this change, the type inferred by Typescript was: `Array<{ id: string; name: string } |  undefined>`.  Avoiding `undefined` allows avoiding the exclamation mark elsewhere, and simplified some other code I added.

CameronDiver (approved):

Issue #1843, opened 2020-05-26
Resource identification with database IDs should not be allowed


pdcastro:
```
$ balena app 1234

$ balena tag rm 1234

$ balena envs -a test-rpi
ID     NAME VALUE
133339 a    20
```

The problem is that database IDs are not "portable" in case the API backend needed to make a database migration. "Portable" UUIDs are available to most (all?) resources, and the CLI should list and allow the usage of UUIDs (including 7-char short UUIDs).

srlowe:
Totally agree.  IMO these DB IDs are internal implementation details, and should not be used as external references in the balena:user interface.  They lead to lots of ambiguous cases, which to cover properly require complex disambiguation logic (which potentially call the API multiple times, even before the action happens).

pdcastro:
Connects-to: #2063 - specifically this comment: 
> [...] As an aside, I'd even prefer that the app ids be internal only and not exposed to any customers or customer APIs. By itself this isn't a serious security issue, but it needlessly exposes info that an attacker could easily leverage.

PR #2080, opened 2020-11-02
device move: Add '-name', '-uuid' and '-id' application and device options

pdcastro:
Proof of concept on how to avoid ambiguity when specifying a device or application given its UUID, database ID, or name. See:

* https://www.flowdock.com/app/rulemotion/i-cli/threads/g26bVg0Nc518ZZknqxtX0GmWhwb
* #1842 
* #1843

Change-type: minor


pdcastro:
@balena-ci retest

Issue #2077, opened 2020-10-28
BalenaApplicationNotFound: Application not found: 1234


pdcastro:
`balena app 1234` should have worked in the example below, but it fails.

```
$ balena app create 1234 --type raspberrypi3
Application created: gh_paulo_castro/1234 (raspberrypi3, id 1758993)
```

```
$ balena apps
ID      APP NAME         SLUG                            DEVICE TYPE           ONLINE DEVICES DEVICE COUNT
1758993 1234             gh_paulo_castro/1234            raspberrypi3          0              0
1301645 test-rpi         gh_paulo_castro/test-rpi        raspberrypi3          0              3
...
```

```
$ balena app 1234
BalenaApplicationNotFound: Application not found: 1234
```

```
$ balena app test-rpi
== TEST RPI
ID:          1301645
DEVICE TYPE: raspberrypi3
SLUG:        gh_paulo_castro/test-rpi
COMMIT:      e3b2bb404a385ed8af1dac8b9043b8da
```

The reason is explained in issue #1842 [Resource names that look like numbers are treated as IDs / short UUIDs](https://github.com/balena-io/balena-cli/issues/1842).

PR #2081, opened 2020-11-02
Improve application-identifier disambiguation

srlowe:
This fixes the app id/name ambiguity issues.  E.g. now:

```
$ balena app 1760161
== 1234
ID:          1760161
DEVICE TYPE: aio-3288c
SLUG:        sl/1234

$ balena app 1234
== 1234
ID:          1760161
DEVICE TYPE: aio-3288c
SLUG:        sl/1234
```

Change-type: patch
Resolves: #2077

### DIFF
--- a/lib/utils/normalization.ts
+++ b/lib/utils/normalization.ts
@@ -1,18 +1,19 @@
-/*
-Copyright 2016-2018 Balena
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-   http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
+/**
+ * @license
+ * Copyright 2016-2020 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
 import type { BalenaSDK } from 'balena-sdk';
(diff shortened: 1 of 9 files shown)


lib/utils/normalization.ts
```diff
@@ -15,17 +15,31 @@ limitations under the License.
...
+		} catch (e) {
+			if (e.name !== 'BalenaApplicationNotFound') {
+				throw e;
+			}
+			// No numerical-name app exists, so return id cast to number.
+			return Number(appIdentifier);
+		}
 	}
+
+	return appIdentifier;
```
pdcastro, line 42: I suggest not discarding a database query that actually found the application -- as another similar call to `balena.models.application.get()` would follow this return statement, for the same `appIdentifier`.

Instead of having a function like `disambiguateAppParam` whose purpose is just to decide whether the app identifier should be a number or a string, we could have a CLI's own `getApplication` method (perhaps in `patterns.ts`) whose purpose was to get and return the application (with any given options) by first attempting to fetch a `looksLikeInteger()` identifier as a name string, and if this failed, then tried again as a database ID number. Along the lines of:

```javascript
// lib/utils/patterns.ts

export async function getApplication(
	sdk: BalenaSDK,
	options?: PineOptions<Application>,
): Promise<Application> {
	if (looksLikeInteger(nameOrSlugOrId)) {
		try {
			return sdk.models.application.get(nameOrSlugOrId, options);
		} catch (err) {
			if (instanceof(err, BalenaApplicationNotFound)) {
				nameOrSlugOrId = parseInt(nameOrSlugOrId, 10);
			} else {
				throw err;
			}
		}
	}
	return sdk.models.application.get(nameOrSlugOrId, options);
}
```
Then in `lib/commands/app/index.ts`, we'd call this `getApplication` function instead of `balena.models.application.get()`.

pdcastro (changes requested):

lib/utils/normalization.ts
srlowe: Yeah this is a better approach.   

I've pushed a variation that, if looksLikeInteger, it tests for the integer ID first, then the name.   The benefit of this is that in situations where we have e.g.:  
NAME   ID 
1234     11111
test        1234
We can still access them both via ID, (otherwise the second would not be accessible by ID), and we can still access the second by name, using the slug `sl/1234`.

Let me know what you think about this inversion, I think it's slightly better, but don't have strong feelings.

pdcastro (commented):
Looks good 👍    I would approve it if it wasn't a draft PR - versionbot won't let me!

lib/utils/sdk.ts
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright 2016-2020 Balena Ltd.
```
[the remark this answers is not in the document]
srlowe: Right, me too - I'd blindly copied it from normalization.tz

pdcastro (approved):

PR #2084, opened 2020-11-04
Modify `os download` help to mention `-dev` suffix

codewithcheese:


srlowe (approved):

lib/commands/os/download.ts
```diff
@@ -46,10 +46,14 @@ export default class OsDownloadCmd extends Command {
...
 		You can pass \`--version menu\` to pick the OS version from the interactive menu
 		of all available versions.
+
+		To download a development image append \`-dev\` to the version or select from
+		the interactive menu.
 `;
 	public static examples = [
 		'$ balena os download raspberrypi3 -o ../foo/bar/raspberry-pi.img',
 		'$ balena os download raspberrypi3 -o ../foo/bar/raspberry-pi.img --version 1.24.1',
+		'$ balena os download raspberrypi3 -o ../foo/bar/raspberry-pi.img --version 1.24.1-dev',
```
pdcastro, line 56: Placing this PR on hold for a moment as I am testing `-dev` _vs._ `.dev`. All the following seems to cause the CLI to download something, just not sure what it is downloading...
```
$ balena os download raspberrypi3 -o test.img --version 2.60.1+rev1-dev
$ balena os download raspberrypi3 -o test.img --version 2.60.1+rev1.dev
$ balena os download raspberrypi3 -o test.img --version 2.60.1+rev1-nodev
$ balena os download raspberrypi3 -o test.img --version 2.60.1+rev1-foo
$ balena os download raspberrypi3 -o test.img --version 2.60.1+rev1.prod
```

Adding a `console.error` for what `--version menu` produces, it produces `.dev` rather than `-dev`:
```
$ ./bin/balena-dev os download raspberrypi3 -o /dev/null --version menu
Getting device operating system for raspberrypi3
? Select the OS version: v2.60.1+rev1.dev
OSVersion=2.60.1+rev1.dev displayVersion= 2.60.1+rev1.dev
```

pdcastro (changes requested):

lib/commands/os/download.ts
pdcastro: The boot partition of the downloaded image file contains a `os-release` file with contents like:
```
VARIANT="Development"
VARIANT_ID="dev"
```
or:
```
VARIANT="Production"
VARIANT_ID="prod"
```
Checking it this way, I have confirmed that `-dev` downloads a production image, while `.dev` downloads a development image.

Given that we were ourselves about to advise users of an incorrect way of downloading development images, :-) it would be nice if the CLI had some validation of versions. I have also checked that the CLI downloads a production image if an inexistent revision is given:

```
$ balena os download raspberrypi3 -o test.img --version 2.60.1+rev9751
```
It only seems to error if the main version is incorrect:
```
$ balena os download raspberrypi3 -o test.img --version 2.60.9
No such version for the device type
```

codewithcheese: @pdcastro hmm wow nice catch. What do you recommend going forward? Should I update this PR to `.dev` and then create a separate issue for version validation?

pdcastro: > Should I update this PR to .dev and then create a separate issue for version validation?

That's the minimum to allow this PR to be merged. :-)
Also, just thought of a test I didn't make: could you confirm that both `2.60.1.dev` and  `2.60.1+rev1.dev` download development images? I think I tested only the latter. (They'd be different image versions, but both should be dev images.) The former is the one that this PR includes in the examples section.

pdcastro: > and then create a separate issue for version validation?

I have opened this balena-sdk PR: https://github.com/balena-io/balena-sdk/pull/1017

> [...] could you confirm that both `2.60.1.dev` and `2.60.1+rev1.dev` download development images?

I tried testing this, but actually I couldn't find any existing non-rev version (like `2.60.1.dev`) to test with! To be on the safe side, I suggest amending the example in this PR to read `2.60.1+rev1.dev` instead of `1.24.1-dev`.

Then we can merge this CLI PR. There is no need to wait for https://github.com/balena-io/balena-sdk/pull/1017, unless you want to. The thing is that the CLI is currently on SDK `15.6.0` while the latest SDK is `15.17.0`, so we'd have to test the CLI in general with the SDK changes.

codewithcheese: @pdcastro thanks for following up on this. I have pushed an update.

pdcastro (approved):

pdcastro:
@balena-ci rebase

PR #2092, opened 2020-11-09
Patch oclif/dev-cli to fix permission issues with .deb packaging

srlowe:
Small patch-package patch to fix permission issues with oclif's .deb package builder script.  Note there's already an upstream PR for this (over one year old), but the dev-cli repo does not seem to be actively maintained.

Change-type: patch

### DIFF
--- a/patches/all/@oclif+dev-cli+1.22.2.patch
+++ b/patches/all/@oclif+dev-cli+1.22.2.patch
@@ -1,3 +1,18 @@
+diff --git a/node_modules/@oclif/dev-cli/lib/commands/pack/deb.js b/node_modules/@oclif/dev-cli/lib/commands/pack/deb.js
+index 51bce6e..a38f00f 100644
+--- a/node_modules/@oclif/dev-cli/lib/commands/pack/deb.js
++++ b/node_modules/@oclif/dev-cli/lib/commands/pack/deb.js
+@@ -37,8 +37,8 @@ class PackDeb extends command_1.Command {
+             await qq.write([workspace, 'DEBIAN/control'], scripts.control(buildConfig, debArch(arch)));
+             await qq.chmod([workspace, 'usr/lib', config.dirname, 'bin', config.bin], 0o755);
+             await qq.x(`ln -s "../lib/${config.dirname}/bin/${config.bin}" "${workspace}/usr/bin/${config.bin}"`);
+-            await qq.x(`chown -R root "${workspace}"`);
+-            await qq.x(`chgrp -R root "${workspace}"`);
++            await qq.x(`sudo chown -R root "${workspace}"`);
++            await qq.x(`sudo chgrp -R root "${workspace}"`);
+             await qq.x(`dpkg --build "${workspace}" "${qq.join(dist, `${versionedDebBase}.deb`)}"`);
+         };
+         const arches = _.uniq(buildConfig.targets
 diff --git a/node_modules/@oclif/dev-cli/lib/commands/pack/macos.js b/node_modules/@oclif/dev-cli/lib/commands/pack/macos.js
 index cd771cd..4a66939 100644
 --- a/node_modules/@oclif/dev-cli/lib/commands/pack/macos.js
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "resin.io",
     "git"
   ],
-  "author": "Juan Cruz Viotti <juan@balena.io>",
   "license": "Apache-2.0",
   "engines": {
     "node": ">=10.20.0"


patches/all/@oclif+dev-cli+1.22.2.patch
```diff
@@ -1,3 +1,18 @@
...
+--- a/node_modules/@oclif/dev-cli/lib/commands/pack/deb.js
++++ b/node_modules/@oclif/dev-cli/lib/commands/pack/deb.js
+@@ -37,8 +37,8 @@ class PackDeb extends command_1.Command {
+             await qq.write([workspace, 'DEBIAN/control'], scripts.control(buildConfig, debArch(arch)));
+             await qq.chmod([workspace, 'usr/lib', config.dirname, 'bin', config.bin], 0o755);
+             await qq.x(`ln -s "../lib/${config.dirname}/bin/${config.bin}" "${workspace}/usr/bin/${config.bin}"`);
+-            await qq.x(`chown -R root "${workspace}"`);
+-            await qq.x(`chgrp -R root "${workspace}"`);
++            await qq.x(`sudo chown -R root "${workspace}"`);
++            await qq.x(`sudo chgrp -R root "${workspace}"`);
```
pdcastro, line 12: Just wondering:
* If `npm run package` was executed as the root user, then I think that adding `sudo` here is a no-op.
* It it was not executed as the root user, then couldn't this cause `sudo` to interactively ask for a password? Thus blocking the build...

pdcastro, line 12: Just wondering:
* If `npm run package` was executed as the root user, then I think that adding `sudo` here is a no-op.
* It it was not executed as the root user, then couldn't this cause `sudo` to interactively ask for a password? Thus blocking the build...

srlowe: Hmm yes good point.  Do we have any other cases of sudo use on the builders?  I was not able to get this to work when running the whole thing as root.  (It seems at least two other oclif users have had to use this sudo approach.)

pdcastro: Re sudo on the builders (understanding builders to mean the balena CI workers that run the CLI tests / build tasks like [npm run package](https://github.com/balena-io/balena-cli/blob/v12.27.2/package.json#L57)), what comes to my mind is this `run_as` function:
* https://github.com/product-os/balena-concourse/blob/v6.3.1/pipelines/github-events/template.yml#L1012-L1018
* https://github.com/product-os/balena-concourse/blob/v6.3.1/pipelines/github-events/template.yml#L1070-L1071

I understand that the Linux and macOS CI workers run as the root user. (On Window I cannot remember.) The `run_as` function mentioned above only makes a difference on macOS, and it actually drops root privileges, so kind of the opposite as `sudo`. I vaguely remember that dropping privileges was necessary to get some `npm` commands to run successfully, in addition to npm flags like `--unsafe-perm`.

> I was not able to get this to work when running the whole thing as root.

Do you mean on your laptop? Ideally we would be in a position where `npm run package` runs successfully in our laptops when running as a regular user (root/sudo not needed), and also runs successfully on CI workers running as the root user. Commands like `chown -R root` and `chgrp -R root` normally need root privileges indeed. But if we proceeded with this patch, then I expect that when running `npm run package` on my macOS laptop as a regular user, the build would eventually be interrupted to ask me to type the root user password -- less than ideal.

Question: are these `chown` and `chgrp` commands really needed? If the patch was to comment out these two lines instead of adding `sudo`, where / how would it fail? Could we then fix that failure in some other way?  

Reminder: as you have been experimenting with sudo or the root user, there may be cached files in several places that are now owned by root and need to be deleted before there is a chance of dev-cli working as a regular user.

I am not trying to prevent patching, nor trying to be difficult, :-) just trying to preserve my ability of running `npm run package` on my laptop as a regular user, without having to wait for an interactive prompt to type a root password...

srlowe: Right. Probably the best way forward with this then is to try to do without chown/chgrp root, and deal with any consequences of that.  I'm currently looking into the signing and publishing process, so will come back to this.  For now I will convert PR to draft.

pdcastro: Just checked that the command that follows `chown` and `chgrp` is `dpkg --build`, so I guess the reason for `chown` and `chgrp` is to store the permissions of a standard common user like `root` inside the `.deb` file that gets generated (perhaps `apt install` will preserve those permission on the user's machine). I wonder if there is an alternative way of setting the permissions inside the `.deb` file that does not require `chown` and `chgrp` on the source filesystem.

srlowe: > so I guess the reason for chown and chgrp is to store the permissions of a standard common user like root inside the .deb file that gets generated

Yes, I believe so  The deb format wraps a couple of tar archives, so perhaps we could simple override ownership when creating the tar (e.g. --owner, --group).

Issue #2091, opened 2020-11-09
Livepush should not watch files that are ignored with .dockerignore


pdcastro:
As originally reported in [this internal / private thread](https://www.flowdock.com/app/rulemotion/r-beginners/threads/vCBixRDgJFO9MYCiRnJ31t6zQBg).

Currently, livepush watches the context folder of application services and ignores a single hardcoded `'.git'` pattern:
* https://github.com/balena-io/balena-cli/blob/v12.27.1/lib/utils/device/live.ts#L166-L170
```
		const monitor = chokidar.watch('.', {
			cwd: context,
			ignoreInitial: true,
			ignored: '.git',
		});
```

Meanwhile, when creating the build context tar stream sent to balenaEngine, livepush uses the same `tarDirectory` function that the CLI uses for the `balena build` and `balena push myApp` commands:
* https://github.com/balena-io/balena-cli/blob/v12.27.1/lib/utils/device/deploy.ts#L199
* https://github.com/balena-io/balena-cli/blob/v12.27.1/lib/utils/device/deploy.ts#L417

This `tarDirectory` function respects the `.dockerignore` file(s) and the `--multi-dockerignore (-m)` flag, which is "correct" but inconsistent with the files that are watched.

I think the correct behaviour would be for livepush to take the `.dockerignore` file(s) into account when deciding what local filesystem files to watch.

pdcastro:
Implemented in CLI release 12.28.0.

Issue #2079, opened 2020-10-30
Should support Node 14 and higher


pdcastro:
The balena CLI currently supports Node 10 ([>= 10.20.0](https://github.com/balena-io/balena-cli/blob/v12.26.0/package.json#L87)) and Node 12 only, as documented in the [installation instructions](https://github.com/balena-io/balena-cli/blob/master/INSTALL-ADVANCED.md#npm-installation).

With Node 14, it crashes as mentioned in issue #1990 [Illegal instruction: 4](https://github.com/balena-io/balena-cli/issues/1990)

What is currently missing in order to support Node 14 is solving a problem with the `ext2fs` dependency: https://github.com/balena-io-modules/node-ext2fs/issues/76 , which in turn blocks the Etcher SDK from supporting Node 14: https://github.com/balena-io-modules/etcher-sdk/issues/183

pdcastro:
It looks like [etcher-sdk](https://www.npmjs.com/package/etcher-sdk) and [ext2fs](https://www.npmjs.com/package/ext2fs) were updated to support Node v14. We see below that the CLI is using (either directly or via old dependencies) old versions of `etcher-sdk` and `ext2fs`. Updating them might be what it takes to support Node v14. It may require pull requests and testing in other repos like `balena-image-fs`, `resin-image-fs` (same repo?), `balena-config-json`, `resin-cli-visuals`. (This last one might be replaceable with `cli-ux`.) Because they are major version bumps, we need to be careful with any API compatibility issues.
```
$ npm list
...
├─┬ balena-config-json@4.1.0
│ ├─┬ balena-image-fs@7.0.1
│ │ ├─┬ ext2fs@2.0.1
...
├─┬ resin-image-fs@5.0.9
│ ├─┬ ext2fs@1.0.31
...
├─┬ etcher-sdk@2.0.20
...
├─┬ resin-cli-visuals@1.7.0
│ ├── etcher-sdk@2.0.20 deduped
```

pdcastro:
Also recently reported: #2165 [livepush: CLI exits unexpectedly (no error) with Node.js v14](https://github.com/balena-io/balena-cli/issues/2165)

mbalamat:
Modules to fix:
* https://github.com/balena-io-modules/balena-device-init
* https://github.com/balena-io-modules/resin-device-operations
* https://github.com/balena-io-modules/resin-cli-visuals

myarmolinsky:
Closing as we currently support node >= 20.6.0

PR #2093, opened 2020-11-10
Livepush: Ignore paths set in .dockerignore files

joshbwlng:
Resolves: #2091 
Change-type: minor


lib/utils/device/live.ts
```diff
@@ -162,11 +165,43 @@ export class LivepushManager {
...
+					} catch (err) {
+						this.logger.logError(
+							`Error reading .dockerignore file "${file.filePath}":\n${err}`,
+						);
+						throw err;
+					}
+					for (const line of dockerignoreContent.split(/\r?\n/)) {
+						ignorePath = line.trim();
+						if (!_.isEmpty(ignorePath)) {
+							ignored.push(path.join(path.dirname(file.relPath), ignorePath));
```
pdcastro, line 195: Unfortunately it may not be this simple because `chokidar` uses `anymatch` as the file pattern matcher:
https://github.com/micromatch/anymatch (See also: https://www.npmjs.com/package/chokidar)

... whereas the `.dockerignore` files use Docker's own pattern matcher, which you can read more about here:
https://github.com/balena-io-modules/dockerignore/#dockerignore

For example, the dockerignore format allows adding pattern match exceptions with the exclamation mark, whereas anymatch has no such concept. The anymatch documentation says it has "bash parity", so presumably (?) it understands bash "wildcards" like `?` and ranges like `a{b,c}`, which are not understood by dockerignore. Also, the anymatch documentation says it does not support the backslash as path separator on Windows, whereas dockerignore does.

I suspect it would be properly painful to try and map the differences between pattern matchers.

Therefore, instead of this approach, what comes to mind is two alternatives:
* Provide an exhaustive, pre-filtered list of files to watch to chokidar. I haven't used chokidar much, but [its documentation] reads:
```
// Initialize watcher.
const watcher = chokidar.watch('file, dir, glob, or array', ...
```
So the idea here would be, instead of `chokidar.watch('.', ...` as the CLI currently does, to use `chokidar.watch(fileList, ...` where `fileList` is a list of every single file to watch, pre-filtered with dockerignore.

* Another approach would to be provide a function as the value of the `ignored` option of chokidar. Again, from chokidar's documentation:
```
`ignored` [...] If a function with two arguments is provided, it gets called twice per path - once with a single argument (the path), second time with two arguments (the path and the fs.Stats object of that path).
```
then that function would use the dockeringore pattern matcher to tell chokidar whether or not to ignore the path.

It is not currently clear to me which approach would be more efficient -- consider both RAM and CPU. Keep in mind that a node_modules folder may contain tens of thousands of files (the CLI has some 50,000 files there!), so the an exhaustive list of files to watch could in some cases be smaller than the list of files to ignore...

pdcastro: Oh, my first alternative approach would not work, because files newly created by the user (while livepush is running) would not be known to chokidar. I guess this leaves us with the second alternative approach - passing a function as the value of `ignored`.

joshbwlng: @pdcastro Thanks for the quick feedback, didn't think about the differences between the two pattern matchers. I'll try those ideas out and see how it goes. And JF now has multiple `node_modules` directories each with thousands of files so I agree setting watch instead of calling a func for ignored is probably going to be the cheaper solution.

lib/utils/device/live.ts
```diff
@@ -162,11 +164,31 @@ export class LivepushManager {
...
+				const { ignoreByDir, serviceDirs } = await getDockerignoreByDir(
+					this.deployOpts.source,
+					serviceDirsByService,
+				);
+
 				// TODO: Memoize this for containers which share a context
 				const monitor = chokidar.watch('.', {
 					cwd: context,
 					ignoreInitial: true,
-					ignored: '.git',
```
joshbwlng, line 169: Should we keep `.git` hard-coded to be ignored? What about adding others, such as `node_modules`, `.balena`, etc?

lib/utils/device/live.ts
joshbwlng: @pdcastro Updated the modifications to couple some existing logic to work with chokidar's ignored function. Let me know what you think.

joshbwlng: Have been using the build from this branch locally to keep working on getting Jellyfish to work with Livepush and have noticed that `balena push`ing to my local mode device sometimes exits without error once builds complete. I've added logs in an attempt to discern if this is somehow caused by my edits, but have not been able to figure out the root problem.

lib/utils/ignore.ts
```diff
@@ -295,11 +296,44 @@ export async function filterFilesWithDockerignore(
...
+	return { filteredFileList, dockerignoreFiles };
+}
+
+/**
+ * Get dockerignore instances for root and all service directories in the project.
+ * Also return the list of service directories found.
+ * @param projectDir Source directory to
+ * @param serviceDirsByService Map of service names to their subdirectories.
+ */
+export async function getDockerignoreByDir(
```
pdcastro, line 325: I was thinking about this function -- its input and output -- and I think it could be made more programmer friendly. Specially the return values: `ingoreByDir` and `serviceDirs`: how is one supposed to use these? The javadoc itself would be complex, to cover directory format requirements. One can figure it all out by carefully reading source code in multiple places, but it could be simpler. It made sense when these data structures were variable in the body of a function, but as an exported function for use in different files, I thought it could be clearer to index by service name rather than by directory. For one reason, service names are widely used and do not vary: a service name like "foo" or "main" is all there is to it, whereas a service directory may be represented in multiple ways, e.g.: `foo`, `./foo`, `./foo/`, `C:\Users\paulo\balena\other\simple_multicontainer\foo'`...

With this in mind, I will be shortly contributing a commit where I propose some changes.

lib/utils/device/live.ts
```diff
@@ -162,11 +164,31 @@ export class LivepushManager {
...
+				const serviceDirsByService = this.deployOpts.multiDockerignore
+					? await getServiceDirsFromComposition(
+							this.deployOpts.source,
+							this.composition,
+					  )
+					: {};
+				const { ignoreByDir, serviceDirs } = await getDockerignoreByDir(
+					this.deployOpts.source,
+					serviceDirsByService,
+				);
```
pdcastro, line 178: The `getServiceDirsFromComposition` and `getDockerignoreByDir` calls are not dependent on each service, so could be moved out of the `for` loop on the grounds of performance.

lib/utils/device/live.ts
pdcastro: The `getDockerIgnoreInstance` function already hardcodes `.git` and other patterns:
https://github.com/balena-io/balena-cli/blob/v12.27.2/lib/utils/ignore.ts#L264-L274

So when using dockerignore in the implementation of the chokidar `ignored` callback, those hardcoded patterns are automatically used. Actually, for the better or the worse (probably the worse), the following files or folders will not be ignored if present under your `node_modules` folders... :-/
```
		'!**/.balena',
		'!**/.resin',
		'!**/Dockerfile',
		'!**/Dockerfile.*',
		'!**/docker-compose.yml',
```

pdcastro (commented):
@joshbwlng, I've contributed a commit - please review and/or test it if you can. (I've already done some testing myself.) Before we can merge the PR, I think we should add some test code, covering some corner cases, to take advantage of the fact that the CI will run the code on Windows, macOS and Linux for us. (The differences between Windows and Linux/macOS are always a pain point...)

Tomorrow I'll git it some more thought on how to write some test cases (and if I cannot come up with something reasonable, I'll test it manually).

joshbwlng:
@pdcastro Thanks for all of the testing and the new commit. I'll pull the changes and try them out today

pdcastro:
> I seem to have issues getting `balena push`es to get past the build stage. I managed to get to livepush output once, but most times the process exits without error once image builds complete. I'm using `bin/balena` after running `npm run build`.

Sound the same as you had [already reported] before I added my commits, right? Such that, if it is related to this PR, then the changes in the first commit are sufficient to trigger the issue - sounds right? Does it happen if you run the unmodified balena CLI v12.27.2, perhaps using the standalone zip package? I mention this specific version because it is the version on which this PR is based, and I mention the standalone zip package because it does not involve the system's Node.js installation which tends to help isolate issues.

And you face this issue when pushing Jellyfish to a local mode device (NUC) - do you have "steps to reproduce" - like, a pointer to a repo that I could clone and livepush. Luckily I do **not** need to get JF working, all I need is to reproduce the CLI crash. :-)

joshbwlng:
@pdcastro Not sure what is wrong exactly, but I can reproduce the issue noted above if I use the `bin/balena` built locally with `npm run build` from the [v12.27.2 commit](https://github.com/balena-io/balena-cli/commit/20e0810d2a4d4cdbf4c92951b94724f006c587c2). If, however, I use the binary provided in the [v12.27.2 official release](https://github.com/balena-io/balena-cli/releases/download/v12.27.2/balena-cli-v12.27.2-linux-x64-standalone.zip) the problem no longer occurs. This makes me think that I am most likely doing something wrong and is a problem that probably won't occur in an official build?

pdcastro:
> _[...] If, however, I use the binary provided in the v12.27.2 official release the problem no longer occurs._

@joshbwlng, in this case, you could also test with the draft release binaries produced for this PR:
* https://github.com/balena-io/balena-cli/releases
* https://github.com/balena-io/balena-cli/releases/tag/untagged-522022c0f7f642ebccb8

Whenever tests pass (even for a draft PR), installers / binaries are produced. Ignore the fact that the draft binaries may be labeled with incorrect CLI version numbers (e.g. currently 12.27.3 for this PR). The version numbers are what would have been produced it the PR was merged as is, but currently it is behind the master branch (not a problem).

By the way:
* Out of curiosity, what is your OS? (Windows, macOS, Linux) The answer probably won't make any difference, anyway. :-)
* The last commit I added includes some test cases, but some are failing. Therefore, the draft release binaries currently regard my previous commit (the last to build successfully).

joshbwlng:
@pdcastro Just checked with the draft release binary you linked to and the problem does not occur, which I guess means something is wrong with `npm run build` on my machine. I'm running Arch Linux with Node v14, if that matters at all.

pdcastro:
> I'm running Arch Linux with Node v14, if that matters at all.

Node v14 may matter, as it is not yet fully supported by balena CLI and it is known to crash the CLI for some commands:
* [Should support Node 14 and higher #2079](https://github.com/balena-io/balena-cli/issues/2079)
* https://github.com/balena-io/balena-cli/blob/master/INSTALL-ADVANCED.md#npm-installation

I wasn't aware of Node v14 crashes for `balena push`, though. Perhaps try with Node v12.

joshbwlng:
@pdcastro Been testing this branch today and it looks like there might be another problem that causes `filterFilesWithDockerignore()` to fail when too many files exist under the project root. Jellyfish currently has multiple `node_modules` directories, five when all `npm install`s have been ran (one for the project root along with one for each app). After adding a few debug logs, I was able to track down the problem to the `listFiles()` function in `lib/utils/ignore.ts`. This runs before dockerignore patterns are taken into account, also causing non-livepush `push`es to fail during `newTarDirectory()`.

pdcastro:
> Been testing this branch today and it looks like there might be another problem that causes `filterFilesWithDockerignore()` to fail when too many files exist under the project root. Jellyfish currently has multiple `node_modules` directories, five when all `npm install`s have been ran (one for the project root along with one for each app). After adding a few debug logs, I was able to track down the problem to the `listFiles()` function in `lib/utils/ignore.ts`. This runs before dockerignore patterns are taken into account, also causing non-livepush `push`es to fail during `newTarDirectory()`.

Thanks for testing it @joshbwlng 👍   This problem with too many files sounds like something that already existed before this pull request, and is not related to extending dockerignore filtering to cover livepush - does this sound right? If so, I suggest we create a another CLI issue and fix the problem in another pull request. I am not reducing the importance of the too-many-files issue - in fact, it even sounds like a more serious issue than the one we are addressing in this PR - but I think it would be less complicated if we tracked and solved it separately.

pdcastro:
> _I was able to track down the problem to the `listFiles()` function in `lib/utils/ignore.ts`_

Two issues come to mind:
* The output of the `listFiles()` function is an array of `FileStats` objects. One issue might be that the number of files is so large, that this array does not reasonably fit in memory. Can you count the number of files, perhaps with a command such as `"find . | wc -l"`? Actually, if this command is able to complete, it would in itself be some evidence that the list of files fits in memory. :-)

* Perhaps there is enough memory for the array of `FileStats` objects, but there isn't enough memory for the recursive implementation of `listFiles()` in its current form. In this case, we might get away with a fix that consisted simply of rewriting `listFiles()` in a non-recursive implementation -- producing the same output, but implemented in a different way.

Do you have a feeling as to whether the second point above might work? Error messages?

pdcastro:
Ok, looking more closely at the implementation of `listFiles` and how it is only used by `filterFilesWithDockerignore`, we could make a cleverer, non-recursive implementation that filtered files as they are discovered, so that a full list of non-filtered files is never kept in memory.

lib/utils/ignore.ts
```diff
@@ -322,19 +291,44 @@ export async function filterFilesWithDockerignore(
...
-			dockerignoreFiles.push(file);
-		}
-		for (const dir of dockerignoreServiceDirs) {
-			if (file.relPath.startsWith(dir)) {
-				return !ignoreByDir[dir].ignores(file.relPath.substring(dir.length));
-			}
-		}
-		return !ignoreByDir[root].ignores(file.relPath);
+	const filteredFileList: FileStats[] = [];
+	const klaw = await import('klaw');
```
pdcastro, line 296: `klaw` was [already being used in the implementation of `originalTarDirectory`](https://github.com/balena-io/balena-cli/blob/v12.27.4/lib/utils/compose.js#L124), although not in a way that saves memory, as a full list of files was being computed before filtering, much like `listFiles()` did. However, `klaw` provides a stream interface to a non-recursive directory traversal that allows gradual filtering out of files as they are traversed. This is what I've done in this commit. It should mean a lower peak memory usage when traversing thousands of files that should be ignored (several `node_modules` folders).

pdcastro:
@balena-ci retest

joshbwlng:
@pdcastro Thanks for looking into this new problem as well, I'll try the latest changes today

pdcastro:
@balena-ci retest

pdcastro:
> Thanks for looking into this new problem as well, I'll try the latest changes today

Thanks @joshbwlng. Build tests have passed, so binaries for the latest changes are also available:
https://github.com/balena-io/balena-cli/releases/tag/untagged-e459539640197e8e4e09

I have done some testing of corner cases on Windows, ✅   but I have not tested a large number of files. I'll leave this stress testing to you. :-)  If your testing is successful, I think we could make this PR non-draft, then I'll squash a couple of my commits, will drop the "bump" commit, and we could get it merged.

joshbwlng:
@pdcastro Ran a couple quick tests with the build you linked and the error caused by having too many files no longer occurs! 🎉 
However, I'm not sure everything that should be ignored is actually being ignored by the livepush file watcher. After `balena push`ing and waiting for the device state to settle, I ran `npm install` inside of one of our directories and it caused a series of errors like this:
```
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
Error: ENOENT: no such file or directory, lstat '/home/josh/git/jellyfish/.libs/jellyfish-worker/node_modules/.staging/es-abstract-caf3933b/.github'
    at Object.lstatSync (fs.js:1000:3)
    at Object.fs.lstatSync (pkg/prelude/bootstrap.js:1015:33)
    at ignored (/snapshot/versioned-source/build/utils/device/live.js:126:32)
    at matchPatterns (/snapshot/versioned-source/node_modules/anymatch/index.js:62:18)
    at FSWatcher._userIgnored (/snapshot/versioned-source/node_modules/anymatch/index.js:94:14)
    at FSWatcher._isIgnored (/snapshot/versioned-source/node_modules/chokidar/index.js:773:15)
    at NodeFsHandler._addToNodeFs (/snapshot/versioned-source/node_modules/chokidar/lib/nodefs-handler.js:581:16)
    at ReaddirpStream.<anonymous> (/snapshot/versioned-source/node_modules/chokidar/lib/nodefs-handler.js:490:12)
    at ReaddirpStream.emit (events.js:315:20)
    at ReaddirpStream.EventEmitter.emit (domain.js:482:12)
    at addChunk (_stream_readable.js:295:12)
    at readableAddChunk (_stream_readable.js:271:9)
    at ReaddirpStream.Readable.push (_stream_readable.js:212:10)
    at ReaddirpStream._read (/snapshot/versioned-source/node_modules/chokidar/node_modules/readdirp/index.js:135:22)
...
```

I have these included in the root `.dockerignore` and believe they should cause `.libs/jellyfish-worker/node_modules` to be ignored:
```
node_modules
**/node_modules
.libs/*/node_modules
```
The last one is probably superfluous, but added it in for testing purposes.

pdcastro:
The following errors should be fixed in my latest commit:
```
Error: ENOENT: no such file or directory, lstat '/home/josh/git/jellyfish/.libs/jellyfish-worker/node_modules/.staging/es-abstract-caf3933b/.github'
```
It was a bug - thanks for testing!

Re some files not being ignored as expected - still pending investigation.

pdcastro:
> I have these included in the root `.dockerignore` and believe they should cause `.libs/jellyfish-worker/node_modules` to be ignored

Yes, and I think it was indeed being ignored, but in its own way. :-) I think that what was happening was that:

1. Chokidar detected a create and/or delete event for file `'/home/josh/git/jellyfish/.libs/jellyfish-worker/node_modules/.staging/es-abstract-caf3933b/.github'`. As it happens, this is a temporary file/folder created during `npm install`.
1. Eventually, subject to livepush [debouncing timers](https://github.com/balena-io/balena-cli/blob/v12.27.4/lib/utils/device/live.ts#L28), chokidar called the `ignored:` handler to ask it the question: _"should this file be ignored?"_ (This file being `'/home/josh/git/jellyfish/.libs/jellyfish-worker/node_modules/.staging/es-abstract-caf3933b/.github'`)
1. But by the time the `ignored:` handler executed, the file had already been deleted. The handler executed `lstatSync` on the file, which failed with `ENOENT` because the file had been deleted, and there wasn't a try/catch block (a bug fixed in my last commit). After the bug fix, the handler should reply: yes, the file should be ignored.

Now, I hear you ask: _"But why was the `ignored:` handler even being called for files under the `node_modules` folder, given that I added `**/node_modules` to my `.dockerignore` file?"_ The answer is that directories are never ignored and every file (even under the `node_modules` folder) is always individually asked the question of whether it should be ignored. The `dockerignore` library requires this behaviour in order to support exclamation-mark exceptions to ignore patterns, for example:
```
**/node_modules
!libs/modA/node_modules/foo/bar.txt
```
_"But my `.dockerignore` file didn't have any exclamation-mark exception patterns!"_ Yes, I know, but the `dockerignore` library is not clever enough to perform parent directory pruning. Every time that the library is asked a question about a file path, it checks the file path against every rule in the `.dockerignore` file. This is how Docker itself is implemented (Docker is written in Go rather than Javascript, but the implementations map nearly line-by-line to each other). It is inefficient, but it works...

Please try again with the latest commit, and thanks again for testing! 👍

joshbwlng:
@pdcastro Thanks for the clarification, tried out the latest build and everything looks good! 💪

pdcastro:
> tried out the latest build and everything looks good! 💪

Nice! 👍   @joshbwlng, are you OK with me rebasing this PR on master and force pushing? It will change all of the commit hashes, but it should preserve your commit details. I will also squash one or two of my commits.

joshbwlng:
@pdcastro Sounds good to me! :+1:

pdcastro:
Rebased and squashed 2 commits. Also, I've changed the change-type of the first commit from patch to minor. I think this PR in a grey area between new feature and bug fix (livepush never had the dockerignore feature before, which can be considered a bug, but also a new feature for livepush). Given the extent of the changes an the relative risk of something getting broken, I chose to tip the balance in favour of 'minor'.

pdcastro:
@balena-ci retest

pdcastro (approved):

pdcastro:
Released in version 12.28.0 🎉

Issue #1100, opened 2019-02-14
balena-cli unusable if `.balenarc.yml` is invalid


xginn8:
- **Cli version:** 9.12.2
- **Standalone install:** no

If `.balenarc.yml` becomes invalid yaml (specifically tested with the global `$HOME/.balenarc.yml`), the CLI exits immediately with only `null`. This lack of information in turn makes debugging the CLI's failure very difficult. Likely this error comes from balena-cli's settings setup [here](https://github.com/balena-io/balena-cli/blob/master/lib/app.coffee#L67), and in turn how `balena-settings-config` returns an error.

To reproduce:

```sh
❯ cp $HOME/.balenarc.yml $HOME/.balenarc.yml.bak
❯ echo "invalid yaml" > $HOME/.balenarc.yml
❯ balena version
null
```

ffissore:
Stumbled into this one just now. I was playing with the [advanced masterclass](https://github.com/balena-io-projects/balena-cli-advanced-masterclass), added the entry `balenaUrl: 'balena-staging.com'` and then commented it, to see if balena-cli could still talk to the production env

balena-cli stopped working (even `balena version` prints nothing)

Problem is [balena settings client - readConfigFile](https://github.com/balena-io-modules/balena-settings-client/blob/master/lib/settings.ts#L86), where an invalid yaml will throw an error. That function is called by [getSettings](https://github.com/balena-io-modules/balena-settings-client/blob/master/lib/settings.ts#L96) which is wrapped by a `_.once`. `getSettings` is called by [app-common setupGlobalProxy](https://github.com/balena-io/balena-cli/blob/master/lib/app-common.ts#L58) which **handles** the error by silently setting `proxy = null`.

So:
- reading settings throws an error
- first calling function handles it and does not report it
- subsequent calls to reading settings will return `undefined` because of `_.once`

My 2 cents: reading the yaml file should return an empty object, so that other settings sources will be used. Error should be returned to users as a warning

srlowe:
In latest, behaviour is:

```
❯ cp $HOME/.balenarc.yml $HOME/.balenarc.yml.bak
❯ echo "invalid yaml" > $HOME/.balenarc.yml
❯ balena version
Error parsing config file /home/scott/.balenarc.yml: Invalid YAML: invalid yaml

Additional information may be available with the `--debug` flag.
For help, visit our support forums: https://forums.balena.io
For bug reports or feature requests, see: https://github.com/balena-io/balena-cli/issues/
[...]

PR #2094, opened 2020-11-11
Avoid reporting balenarc parsing errors

srlowe:
Change-type: patch
Connects-to: #1100

### DIFF
--- a/lib/errors.ts
+++ b/lib/errors.ts
@@ -156,6 +156,7 @@ const EXPECTED_ERROR_REGEXES = [
 	/must also be provided when using/, // oclif parser (depends-on)
 	/^Expected an integer/, // oclif parser (flags.integer)
 	/^Flag .* expects a value/, // oclif parser
+	/^Error parsing config file.*balenarc\.yml/,
 ];
 
 // Support unit testing of handleError


lib/errors.ts
```diff
@@ -156,6 +156,7 @@ const EXPECTED_ERROR_REGEXES = [
 	/must also be provided when using/, // oclif parser (depends-on)
 	/^Expected an integer/, // oclif parser (flags.integer)
 	/^Flag .* expects a value/, // oclif parser
+	/^Error parsing config file.*.balenarc.yml/,
```
pdcastro, line 159: The two dots in the substring `.balenarc.yml` are interpreted in the regex sense of "any character" and I guess this was not intentional, although it works. Note that:
* The homedir configuration file is named `_balenarc.yml` on Windows and `.balenarc.yml` on Linux/macOS.
* A file named `balenarc.yml` (without a `_` or `.` as prefixes) on the current working directory takes precedence over the homedir configuration file ([documentation](https://github.com/balena-io/balena-cli-advanced-masterclass/blob/master/README.md#13-separate-environment-configurations)).

The regex works anyway because `.*.` means the same as `.*`, but I think it would be more... "canonical" to write:
```
/^Error parsing config file.*balenarc\.yml/,
```

pdcastro (approved):

Issue #1479, opened 2019-10-19
macOS: balena-cli can't be opened because Apple cannot check it for malicious software


pdcastro:
- **balena CLI version:** any (e.g. 12.27.3)
- **Operating system version:** macOS 10.15 (Catalina) and 11.0 (Big Sur)
- **32/64 bit OS and processor:** any
- **Install method:** executable (native) installer and standalone zip

Apple has introduced the requirement of "[software notarization](https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution)" for macOS, but the balena CLI installers have not yet been updated to include this step. As a result, macOS will raise warning pop-ups before the CLI can be installed or executed. Until the CLI installers are notarized, please follow the workaround steps below.

<img width="532" alt="native-error" src="https://user-images.githubusercontent.com/15091591/67145203-5570da00-f277-11e9-8823-176e5fd29012.png">

**Workaround steps for the graphical macOS installer (.pkg)**

1. After downloading the executable (native) installer (`.pkg` file), [right-click](https://support.apple.com/en-gb/HT207700) (2-finger tap or CTRL-click) the file and select "Open" in the pop-up menu:

<img width="559" alt="native-open-menu" src="https://user-images.githubusercontent.com/15091591/67145315-78e85480-f278-11e9-8fe2-5e0f8b281b46.png">

&nbsp;
2. A warning pop-up will be raised, including an **Open** button. **Click the Open button:**

<img width="532" alt="native-open-button" src="https://user-images.githubusercontent.com/15091591/67145465-c618f600-f279-11e9-898e-dec441f92df6.png">

3. The executable (native) installer will open: proceed with the installation, [as usual](https://github.com/balena-io/balena-cli/blob/master/INSTALL.md).

&nbsp;  
If you are using the standalone zip package, check the alternative instructions in the next comment.

pdcastro:
Connects to #1376 (CLI package for Mac needs to be notarized)

pdcastro:
Hints for the implementation:
* https://github.com/mitchellh/gon
* https://www.flowdock.com/app/rulemotion/r-etcherprv/threads/VvGI_yrIc-5f9ScEYeovzVWwrkO

jesperbagge:
I've run into additional issues with the standalone package (v11.19.1) on macOS Catalina that can't be resolved with this workaround. macOS also complains on the file **bindings.node**, which I can't get around by right-clicking.

pdcastro:
[...]

PR #2095, opened 2020-11-15
Test code optimization: avoid running ~70 test cases twice

pdcastro:
Change-type: patch

### DIFF
--- a/INSTALL-ADVANCED.md
+++ b/INSTALL-ADVANCED.md
@@ -59,7 +59,7 @@ macOS:   | `/usr/local/lib/balena-cli/` <br> `/usr/local/bin/balena`
    [macOS](https://www.architectryan.com/2012/10/02/add-to-the-path-on-mac-os-x-mountain-lion/#.Uydjga1dXDg) |
    [Windows](https://www.computerhope.com/issues/ch000549.htm)
 
-> * If you are using macOS Catalina (10.15), [check this known issue and
+> * If you are using macOS 10.15 or later (Catalina, Big Sur), [check this known issue and
 >   workaround](https://github.com/balena-io/balena-cli/issues/1479).
 > * **Linux Alpine** and **Busybox:** the standalone zip package is not currently compatible with
 >   these "compact" Linux distributions, because of the alternative C libraries they ship with.
--- a/package.json
+++ b/package.json
@@ -61,13 +61,13 @@
     "test:shrinkwrap": "ts-node --transpile-only automation/run.ts test-shrinkwrap",
     "test:source": "cross-env BALENA_CLI_TEST_TYPE=source mocha",
     "test:standalone": "npm run build:standalone && npm run test:standalone:fast",
-    "test:standalone:fast": "cross-env BALENA_CLI_TEST_TYPE=standalone mocha",
+    "test:standalone:fast": "cross-env BALENA_CLI_TEST_TYPE=standalone mocha --config .mocharc-standalone.js",
     "test:fast": "npm run build:fast && npm run test:source",
     "test:only": "npm run build:fast && cross-env BALENA_CLI_TEST_TYPE=source mocha \"tests/**/${npm_config_test}.spec.ts\"",
     "catch-uncommitted": "ts-node --transpile-only automation/run.ts catch-uncommitted",
     "ci": "npm run test && npm run catch-uncommitted",
     "watch": "gulp watch",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -61,13 +61,13 @@
     "test:shrinkwrap": "ts-node --transpile-only automation/run.ts test-shrinkwrap",
     "test:source": "cross-env BALENA_CLI_TEST_TYPE=source mocha",
     "test:standalone": "npm run build:standalone && npm run test:standalone:fast",
-    "test:standalone:fast": "cross-env BALENA_CLI_TEST_TYPE=standalone mocha",
+    "test:standalone:fast": "cross-env BALENA_CLI_TEST_TYPE=standalone mocha --config .mocharc-standalone.js",
```
pdcastro, line 64: The objective of this commit is to avoid running a sizeable fraction of the test cases twice, redundantly. The CLI still intentionally runs many tests cases twice, first using the Node.js interpreter available in the PATH, and then using the built `pkg` binary (standalone zip package), in order to test the `pkg` binary too. However, this only makes sense for the end-to-end test cases that invoke CLI commands, using the [runCommand](https://github.com/balena-io/balena-cli/blob/v12.27.3/tests/helpers.ts#L183) function in `tests/helpers.ts`. The `runCommand` function decides whether to execute the built `pkg` binary as a child process, or whether to call the CLI entry point in the same mocha process, depending on the value of the BALENA_CLI_TEST_TYPE env var. Other test cases, typically unit tests, don't call the `runCommand` function, and always run in the mocha process, regardless of the value of BALENA_CLI_TEST_TYPE. It is these other test cases that were being run twice, unnecessarily. This commit avoids running these other tests cases for the `test:standalone[:fast]` script.

Before I came up with the solution of adding the `.mocharc.js` and `.mocharc-standalone.js` files, I tried a smaller change:

```
"test:standalone:fast": "cross-env BALENA_CLI_TEST_TYPE=standalone mocha tests/commands/**/*.ts",
```

... i.e., I tried specifying the desired set of test cases on the mocha command line itself. It didn't work: it would always run more tests than specified on the command line. I tried enclosing the pattern with single quotes or double quotes: no difference. I tried different patterns for different folders, I tried explicitly specifying files instead of using wildcard characters, but it would either not work for other reasons, or it would not run the test cases I wanted. Then I tried upgrading mocha from v6 to v8: no difference. (I decided the keep the mocha upgrade anyway, because it seems like a good thing.) Then I finally tested with separate configuration files and the `--config` option, and it worked fine!

package.json
```diff
@@ -61,13 +61,13 @@
...
     "test:fast": "npm run build:fast && npm run test:source",
     "test:only": "npm run build:fast && cross-env BALENA_CLI_TEST_TYPE=source mocha \"tests/**/${npm_config_test}.spec.ts\"",
     "catch-uncommitted": "ts-node --transpile-only automation/run.ts catch-uncommitted",
     "ci": "npm run test && npm run catch-uncommitted",
     "watch": "gulp watch",
-    "lint": "balena-lint -e ts -e js --typescript --fix automation/ lib/ typings/ tests/ bin/balena bin/balena-dev gulpfile.js",
+    "lint": "balena-lint -e ts -e js --typescript --fix automation/ lib/ typings/ tests/ bin/balena bin/balena-dev gulpfile.js .mocharc.js .mocharc-standalone.js",
```
pdcastro, line 70: `.mocha*.js` works on Unix, but not on Windows, hence spelling out the names in full.

INSTALL-ADVANCED.md
```diff
@@ -59,7 +59,7 @@ macOS:   | `/usr/local/lib/balena-cli/` <br> `/usr/local/bin/balena`
    [macOS](https://www.architectryan.com/2012/10/02/add-to-the-path-on-mac-os-x-mountain-lion/#.Uydjga1dXDg) |
    [Windows](https://www.computerhope.com/issues/ch000549.htm)
 
-> * If you are using macOS Catalina (10.15), [check this known issue and
+> * If you are using macOS 10.15 or later (Catalina, Big Sur), [check this known issue and
```
pdcastro, line 62: This change is in its own commit. I have also updated issue #1479 to cover both Catalina and Big Sur.

srlowe (approved):
Good stuff!  Interested to see what effect it has on test times.

pdcastro:
> Good stuff! Interested to see what effect it has on test times.

Sadly the impact is minor - seconds rather than minutes. Still, unnecessarily running half of test cases twice seemed like something to avoid...

Issue #2098, opened 2020-11-19
balena scan --json output has terminal escape sequences before JSON content


sgrimm:
# Expected Behavior

I should be able to pipe `balena scan --json` to something that expects JSON input. For example, `sudo balena scan --json | jq .` should output a jq-formatted copy of the scan results.

# Actual Behavior

```
$ sudo balena scan --json | jq .
parse error: Invalid numeric literal at line 1, column 2
```

This appears to be because `balena scan --json` outputs some terminal escape sequences before the JSON.

```
$ sudo balena scan --json | od -tc | head -1
0000000  033   [   2   K 033   [   1   G  \n   [  \n                   {
```

# Steps to Reproduce the Problem

With at least one device on the local network in local mode, run `sudo balena scan --json | jq .`

# Specifications

- **balena CLI version:** balena-cli version "12.25.0", Node.js version "12.16.2"
- **Cloud backend:** balenaCloud
- **Operating system version:** macOS 10.15.7
- **32/64 bit OS and processor:** 64-bit Intel
- **Install method:** Homebrew

pdcastro:
Thanks for reporting this issue @sgrimm 👍 
Pull request #2099 should fix it (after it is merged, pending review).

pdcastro:
Fixed in CLI v12.28.1. (Spinner animation goes to stderr, JSON output goes to stdout.)

Issue #1713, opened 2020-04-13
`balena scan` should not filter out production devices


pdcastro:
As per product feedback ([gsheet](https://docs.google.com/spreadsheets/d/1V6URepXBhqU7WhFU1FwUrn-LdQi95ZF4je49dlM2YWQ/))

srlowe:
I had a quick look at this, and I can't see anywhere where we are explicitly filtering out production devices.

pdcastro:
>  I can't see anywhere where we are explicitly filtering out production devices.

And did you actually see any production device being listed by `balena scan`? I've just done a test here with three devices: two development devices were listed, one production device was not. Maybe production devices don't advertise themselves? Which would then be a balenaOS matter.

srlowe:
Turns that the CLI is not filtering production devices (nor are module dependencies), but this is a feature of the production OS (see info below).  Will update gsheet and close.

https://www.flowdock.com/app/rulemotion/r-resinos/threads/q8xugwUW-Q3oLDAlcqkxwU16PDQ

> I believe balena scan will only work on development images. The supervisor can set the discoverability on/off using a HOST_DISCOVERABILITY config option that starts/stops the avahi daemon. You can check whether the service is running in the hostOS and that will tell you whether it is discoverable.

pdcastro:
> Turns that the CLI is not filtering production devices

Or maybe it kind of is... :-)  The discussion continued on that FD thread. I have found that `avahi-daemon` was actually running on a prod device, asked @richbayliss about it who pointed out:
> balena scan is looking for a device SSH service advertised over mDNS
> then it connects to that SSH service to pull device info
> it cannot do that in prod mode due to keys (I think)
> it's not just looking for mDNS/avah

So the CLI, or more precisely the `balena-sync` module:
https://github.com/balena-io-modules/balena-sync/blob/v11.0.2/lib/discover.coffee#L26
... would be potentially finding some info being advertised by the prod devices, but then not being able to collect "additional info" (e.g. ssh service) that a dev device would provide. I actually see this `docker.pingAsync()` call in CLI code:
https://github.com/balena-io/balena-cli/blob/v12.26.0/lib/commands/scan.ts#L93-L97
... which might mean that the CLI gives up if balenaEngine is not exposed on the device. I know that balenaEngine is reachable on port 2375 on dev-image devices, but not on prod-image devices (possibly because of iptables rules).

[...]

PR #2099, opened 2020-11-20
scan: Prevent spinner animation output to stdout when --json is used

pdcastro:
Resolves: #2098 
Change-type: patch

### DIFF
--- a/lib/commands/scan.ts
+++ b/lib/commands/scan.ts
@@ -19,7 +19,7 @@ import { flags } from '@oclif/command';
 import type { LocalBalenaOsDevice } from 'balena-sync';
 import Command from '../command';
 import * as cf from '../utils/common-flags';
-import { getVisuals, stripIndent } from '../utils/lazy';
+import { getCliUx, stripIndent } from '../utils/lazy';
 
 interface FlagsDef {
 	json?: boolean;
@@ -65,10 +65,8 @@ export default class ScanCmd extends Command {
 
 	public async run() {
 		const _ = await import('lodash');
-		const { SpinnerPromise } = getVisuals();
 		const { discover } = await import('balena-sync');
 		const prettyjson = await import('prettyjson');
-		const { ExpectedError } = await import('../errors');
 		const dockerUtils = await import('../utils/docker');
 
 		const dockerPort = 2375;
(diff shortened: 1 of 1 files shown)


lib/commands/scan.ts
```diff
@@ -19,7 +19,7 @@ import { flags } from '@oclif/command';
 import type { LocalBalenaOsDevice } from 'balena-sync';
 import Command from '../command';
 import * as cf from '../utils/common-flags';
-import { getVisuals, stripIndent } from '../utils/lazy';
+import { getCliUx, stripIndent } from '../utils/lazy';
```
pdcastro, line 22: The main reason for swapping `resin-cli-visuals` with `cli-ux` in this PR is that the latter allows sending spinner output to stderr instead of stdout (and does it by default), which thus avoids mixing spinner output with `--json` output which is sent to stdout. This is sufficient to allow piping stdout to `jq`, as requested in #2098. Another option would have been to disable the spinner altogether when `--json` is used. I understand that `cli-ux` is also clever enough to auto detect whether the terminal is a TTY, and prevent animation if it is not (I didn't test this, though).

lib/commands/scan.ts
```diff
@@ -80,33 +78,30 @@ export default class ScanCmd extends Command {
...
-				await docker.pingAsync();
-				return true;
-			} catch (err) {
-				return false;
-			}
-		});
-
-		// Exit with message if no devices found
-		if (_.isEmpty(activeLocalDevices)) {
-			// TODO: Consider whether this should really be an error
```
pdcastro, line 103: I've acted on this TODO and stopped treating an empty output as an error.

lib/commands/scan.ts
```diff
@@ -80,33 +78,30 @@ export default class ScanCmd extends Command {
...
+					timeout: dockerTimeout,
+				}) as any;
+				try {
+					await docker.pingAsync();
+					return true;
+				} catch (err) {
+					return false;
+				}
+			}),
+		);
```
pdcastro, line 101: This coding with separate arrays for `localDevices`, `engineReachableDevices` and `activeLocalDevices` makes it clearer what we could do to resolve #1713 -- I suspect that production devices would be listed in `localDevices` but not in `engineReachableDevices`. If so, the solution for #1713 would then be to print information for all devices in `localDevices`, only with less data available about them. More details in https://github.com/balena-io/balena-cli/issues/1713#issuecomment-718926440.

That's for another PR, though. :-)

srlowe (approved):

Issue #918, opened 2018-07-06
Dockerfile ARG support for local push (build arguments, --buildArg)


zrzka:
## Description

I have a project where `Dockerfile.template` builds artifacts in release mode. It defaults to this behavior, so, whever I `git push resin master`, production build is deployed to all devices.

To make it simple, let's focus on one command only (it's more complicated, but this one is enough to demonstrate the problem). Build is done via:

```Dockerfile
RUN cargo build --release
```

In case of `resin local push`, I'd like to have debug build (without `--release`). What can I do now?

## ENV

```Dockerfile
ENV CARGO_BUILD_COMMAND cargo build --release
```

`resin local push --env "CARGO_BUILD_COMMAND=cargo build"` doesn't work. It prints `cargo build --release` if I add `RUN echo $CARGO_BUILD_COMMAND`.

## Replacement

I can do fancy stuff with [sed](https://github.com/resin-io-projects/resin-rust-hello-world/blob/master/scripts/local-push.sh#L30). But it's not ideal.

## ARG support

What I'd like to do is ...

```Dockerfile
ARG cargo_build_command=cargo build --release
```

... and then ...

```bash
resin local push --build-arg "cargo_build_command=cargo build"
```

pdcastro:
Vaguely related to #1053

pdcastro:
For the benefit of anyone reaching this issue: while a `--buildArg` flag is still not available, note that it is currently possible to specify build variables and secrets using a `.balena/balena.yml` file, as documented on this page: https://www.balena.io/docs/learn/deploy/deployment/#build-time-secrets-and-variables

Issue #1705, opened 2020-04-09
balena 


CameronDiver:
- **balena CLI version:** e.g. 11.2.1 (output of the `"balena version"` command)
- **Operating system version:** e.g. Windows 10, Ubuntu 18.04, macOS 10.14.5
- **32/64 bit OS and processor:** e.g. 32-bit Windows on 64-bit Intel processor
- **Install method:** npm or zip or executable installer
- **If npm install, Node.js and npm version:** e.g. Node v8.16.0 and npm v6.4.1

---

*Please note that this issue tracker is used for specific bug reports and feature requests.
General and troubleshooting questions are encouraged to be posted to the [balena
forums](https://forums.balena.io), which are monitored by balena's support team and
where the community can both contribute and benefit from the answers.*

*Before submitting this issue please check that this issue is not a duplicate. If there is another
issue describing the same problem or feature please add your information to the existing issue's
comments.*

russellseymour:
Is there any movement on this at all?

I am running in local mode and trying to send in a DSN as a runtime environment variable on the command line using:

```
balena push x.x.x.x --env CONNSTR="db://user:password@host/name"
```

However if the password contains an `=` symbol it causes issue when the container is pushed to the Raspberry Pi.

![image](https://user-images.githubusercontent.com/791658/159993306-e9ee8184-c0c1-4cd6-9cd1-acc942f5f522.png)

The situation in which the password contains the `=` is when using SaaS based database, such as InfluxDB. The password is a token, which is Base64 encoded, and that contains the `=` character.

If I remove the `=` from the password (which of course makes it invalid) the service starts properly:

![image](https://user-images.githubusercontent.com/791658/159995981-c2fe4fbe-7203-4ac1-a002-0d234d16897a.png)

I do not want to put this in as a `buildArg` because I do not want to bake this into the container. Is there a workaround I can use?

PR #2110, opened 2020-12-01
Add --buildArg support for local push

joshbwlng:
Change-type: minor

Resolves: #918
Change-type: minor
See: https://www.flowdock.com/app/rulemotion/i-cli/threads/rOh4VXCB7_pQhqbHKK3afU-RCBW

---

This PR adds support for the `--buildArg` flag for `balena push`ing to local devices

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -53,20 +53,15 @@ import Logger = require('./logger');
  * @param buildargs --buildArg flag input in dictionary format
  */
 async function parseMetadataFromFlags(
-	dir: string,
+	projectRoot: string,
 	buildargs: Dictionary<string>,
 	composition?: Composition,
 ): Promise<ParsedBuildArguments> {
-	const yaml = await import('js-yaml');
-	const mergedComposition: Composition = yaml.load(
-		await mergeDevComposeOverlay(
-			Logger.getLogger(),
-			yaml.dump(composition),
-			dir,
-		),
-	);
+	const mergedComposition: Partial<Composition> = await mergeDevComposeOverlay<
+		Partial<Composition>
+	>(Logger.getLogger(), composition || {}, projectRoot);
 
-	const serviceNames = Object.keys(mergedComposition?.services || {});
+	const serviceNames = Object.keys(mergedComposition.services || {});
 	const ret: ParsedBuildArguments = {
 		global: {},
 		services: {},
@@ -213,14 +208,17 @@ export async function loadProject(
 }
 
 /**
- * Check for existence of docker-compose dev overlay file
- * and merge in services definitions.
+ * Check for existence of docker-compose dev overlay file and merge in service
+ * definitions. The given composition may be a JS object (Composition) or a
+ * string representing a Yaml file (docker-compose.yml). The return type
+ * matches the input type: either a JS object (Composition) or a string
+ * (docker-compose.yml).
  */
-async function mergeDevComposeOverlay(
+async function mergeDevComposeOverlay<T extends Partial<Composition> | string>(
 	logger: Logger,
-	composeStr: string,
+	composition: T,
 	projectRoot: string,
-) {
+): Promise<T> {
 	const devOverlayFilename = 'docker-compose.dev.yml';
 	const devOverlayPath = path.join(projectRoot, devOverlayFilename);
 
@@ -228,25 +226,52 @@ async function mergeDevComposeOverlay(
 		logger.logInfo(
 			`Docker compose dev overlay detected (${devOverlayFilename}) - merging.`,
 		);
-		interface ComposeObj {
-			services?: object;
-		}
-		const yaml = await import('js-yaml');
-		const loadObj = (inputStr: string): ComposeObj =>
-			(yaml.load(inputStr) || {}) as ComposeObj;
 		try {
-			const compose = loadObj(composeStr);
-			const devOverlay = loadObj(await fs.readFile(devOverlayPath, 'utf8'));
+			const compose: Partial<Composition> =
+				typeof composition === 'string'
+					? loadComposition(composition, 'probably docker-compose.yml')
+					: { ...(composition as Partial<Composition>) };
+			const devOverlay = loadComposition(
+				await fs.readFile(devOverlayPath, 'utf8'),
+				devOverlayPath,
+			);
 			// We only want to merge the services section
 			compose.services = { ...compose.services, ...devOverlay.services };
-			composeStr = yaml.dump(compose, { styles: { '!!null': 'empty' } });
+			return (typeof composition === 'string'
+				? dumpComposition(compose)
+				: compose) as T;
 		} catch (err) {
 			err.message = `Error merging docker compose dev overlay file "${devOverlayPath}":\n${err.message}`;
 			throw err;
 		}
 	}
 
-	return composeStr;
+	return composition;
+}
+
+/** Call `yaml.load` and check that the result is an object. */
+function loadComposition(composeStr: string, hint = ''): Partial<Composition> {
+	const yaml = require('js-yaml') as typeof import('js-yaml');
+	const parsed = yaml.load(composeStr);
+	if (!parsed) {
+		return {};
+	} else if (typeof parsed === 'object') {
+		return parsed;
+	}
+	throw new ExpectedError(
+		'Invalid composition string' + (hint ? ` (${hint})` : ''),
+	);
+}
+
+/**
+ * Call `yaml.dump` with the special `{'!!null': 'empty'}` style. This style
+ * choice is important to avoid an error in the `resin-compose-parse` module.
+ */
+function dumpComposition(composeObj: Partial<Composition>): string {
+	const yaml = require('js-yaml') as typeof import('js-yaml');
+	// The need for `styles: { '!!null': 'empty' }` is explained in:
+	// https://github.com/balena-io/balena-cli/pull/2227
+	return yaml.dump(composeObj, { styles: { '!!null': 'empty' } });
 }
 
 /**
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -813,67 +813,43 @@ function addBuildVariablesToBuildMetadata(
 	metadata: MultiBuild.ParsedBalenaYml,
 	buildArgs: ParsedBuildArguments,
 ): string {
+	const { parseBuildArgs } = require('./docker') as typeof import('./docker');
 	metadata['build-variables'] ||= {};
-	metadata['build-variables'].services ||= {};
+	const bvars = metadata['build-variables'];
 
-	// Convert build variable sets to dictionaries if necessary
-	if (_.isArray(metadata['build-variables'].global)) {
-		const cloned = _.clone(metadata['build-variables'].global);
-		metadata['build-variables'].global = {};
-		for (const buildVar of cloned) {
-			const [key, value] = buildVar.split('=');
-			if (metadata['build-variables'] && metadata['build-variables'].global) {
-				metadata['build-variables'].global[key] = value;
-			}
-		}
+	// Convert build variable arrays to dictionaries if necessary
+	if (Array.isArray(bvars.global)) {
+		bvars.global = parseBuildArgs(bvars.global);
 	}
-	if (metadata['build-variables'] && metadata['build-variables'].services) {
-		Object.keys(metadata['build-variables'].services).forEach(
-			(service: string) => {
-				if (
-					metadata['build-variables'] &&
-					metadata['build-variables'].services &&
-					_.isArray(metadata['build-variables'].services[service])
-				) {
-					const cloned = _.clone(metadata['build-variables'].services[service]);
-					metadata['build-variables'].services[service] = {};
-					_.forEach(cloned, (buildVar: string) => {
-						if (
-							metadata['build-variables'] &&
-							metadata['build-variables'].services
-						) {
-							const [key, value] = buildVar.split('=');
-							metadata['build-variables'].services[service][key] = value;
-						}
-					});
-				}
-			},
-		);
+	for (const [service, vars] of Object.entries(bvars.services || {})) {
+		if (Array.isArray(vars)) {
+			bvars.services![service] = parseBuildArgs(vars);
+		}
 	}
 
 	// Set build variables if specified
 	if (!_.isEmpty(buildArgs.global)) {
-		metadata['build-variables'].global = {
-			...metadata['build-variables'].global,
+		bvars.global = {
+			...bvars.global,
 			...buildArgs.global,
 		};
 	}
 	if (!_.isEmpty(buildArgs.services)) {
-		metadata['build-variables'].services ||= {};
-		Object.keys(buildArgs.services).forEach((service: string) => {
-			if (metadata['build-variables'] && metadata['build-variables'].services) {
-				metadata['build-variables'].services[service] = {
-					...metadata['build-variables'].services[service],
-					...buildArgs.services[service],
-				};
-			}
-		});
+		bvars.services ||= {};
+		for (const [service, vars] of Object.entries(bvars.services)) {
+			bvars.services[service] = {
+				...vars,
+				...buildArgs.services[service],
+			};
+		}
 	}
 
 	if (metadataPath.endsWith('json')) {
-		return JSON.stringify(metadata);
+		return JSON.stringify(metadata, null, 4);
 	} else {
-		return require('js-yaml').dump(metadata);
+		return (require('js-yaml') as typeof import('js-yaml')).dump(metadata, {
+			indent: 4,
+		});
 	}
 }
 
@@ -923,9 +899,22 @@ async function newTarDirectory(
 		filteredFileList,
 		dockerignoreFiles,
 	} = await filterFilesWithDockerignore(dir, multiDockerignore, serviceDirs);
+
 	printDockerignoreWarn(dockerignoreFiles, serviceDirs, multiDockerignore);
+
 	let foundConfig = false;
 	for (const fileStats of filteredFileList) {
+		let buf: Buffer | string;
+		if (metadataPath && fileStats.relPath === metadataPath) {
+			buf = addBuildVariablesToBuildMetadata(
+				metadataPath,
+				metadata,
+				parsedBuildArgs,
+			);
+			foundConfig = true;
+		} else {
+			buf = await readFile(fileStats.filePath);
+		}
 		pack.entry(
 			{
 				name: toPosixPath(fileStats.relPath),
@@ -933,17 +922,8 @@ async function newTarDirectory(
 				mode: fileStats.stats.mode,
 				size: fileStats.stats.size,
 			},
-			metadataPath && fileStats.relPath === metadataPath
-				? addBuildVariablesToBuildMetadata(
-						metadataPath,
-						metadata,
-						parsedBuildArgs,
-				  )
-				: await readFile(fileStats.filePath),
+			buf,
 		);
-		if (metadataPath && fileStats.relPath === metadataPath) {
-			foundConfig = true;
-		}
 	}
 	if (
 		!foundConfig &&
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -86,15 +86,15 @@ async function buildArgumentsFromInput(
 		if (match[1]) {
 			// This is for a service, we check that it actually exists
 			if (!(match[1] in ret.services)) {
-				Logger.getLogger().logWarn(
-					`Warning: Cannot find a service with name ${match[1]}. Treating the string as part of the build argument variable name.`,
+				throw new ExpectedError(
+					`Cannot find a service with name ${match[1]} specified in build argument: ${buildArg}`,
 				);
-				match[2] = `${match[1]}:${match[2]}`;
 			} else {
 				service = match[1];
 			}
 		}
 
+		// Set build argument either as global or service specific
 		if (service != null) {
 			ret.services[service][match[2]] = match[3];
 		} else {
@@ -803,43 +803,77 @@ export async function tarDirectory(
 	}
 }
 
-function setConfig(
-	configPath: string,
-	configData: MultiBuild.ParsedBalenaYml,
+/**
+ * Add or override a `build-variables` section to the `.balena/balena.yml`
+ * build metadata, with the contents of any `--buildArgs` flags provided
+ * on the command line.
+ */
+function addBuildVariablesToBuildMetadata(
+	metadataPath: string,
+	metadata: MultiBuild.ParsedBalenaYml,
 	buildArgs: ParsedBuildArguments,
 ): string {
-	const config = configData as any;
-	config['build-variables'] ||= {};
+	metadata['build-variables'] ||= {};
+	metadata['build-variables'].services ||= {};
 
-	// Set global build variables if specified
-	if (!_.isEmpty(buildArgs.global)) {
-		config['build-variables'].global ||= [];
-		config['build-variables'].global = [
-			...config['build-variables'].global,
-			..._.map(buildArgs.global, (value: string, key: string) => {
-				return `${key}=${value}`;
-			}),
-		];
+	// Convert build variable sets to dictionaries if necessary
+	if (_.isArray(metadata['build-variables'].global)) {
+		const cloned = _.clone(metadata['build-variables'].global);
+		metadata['build-variables'].global = {};
+		for (const buildVar of cloned) {
+			const [key, value] = buildVar.split('=');
+			if (metadata['build-variables'] && metadata['build-variables'].global) {
+				metadata['build-variables'].global[key] = value;
+			}
+		}
+	}
+	if (metadata['build-variables'] && metadata['build-variables'].services) {
+		Object.keys(metadata['build-variables'].services).forEach(
+			(service: string) => {
+				if (
+					metadata['build-variables'] &&
+					metadata['build-variables'].services &&
+					_.isArray(metadata['build-variables'].services[service])
+				) {
+					const cloned = _.clone(metadata['build-variables'].services[service]);
+					metadata['build-variables'].services[service] = {};
+					_.forEach(cloned, (buildVar: string) => {
+						if (
+							metadata['build-variables'] &&
+							metadata['build-variables'].services
+						) {
+							const [key, value] = buildVar.split('=');
+							metadata['build-variables'].services[service][key] = value;
+						}
+					});
+				}
+			},
+		);
 	}
 
-	// Set service-level build variables if specified
+	// Set build variables if specified
+	if (!_.isEmpty(buildArgs.global)) {
+		metadata['build-variables'].global = {
+			...metadata['build-variables'].global,
+			...buildArgs.global,
+		};
+	}
 	if (!_.isEmpty(buildArgs.services)) {
-		config['build-variables'].services ||= {};
+		metadata['build-variables'].services ||= {};
 		Object.keys(buildArgs.services).forEach((service: string) => {
-			config['build-variables'].services[service] ||= [];
-			config['build-variables'].services[service] = [
-				...config['build-variables'].services[service],
-				..._.map(buildArgs.services[service], (value: string, key: string) => {
-					return `${key}=${value}`;
-				}),
-			];
+			if (metadata['build-variables'] && metadata['build-variables'].services) {
+				metadata['build-variables'].services[service] = {
+					...metadata['build-variables'].services[service],
+					...buildArgs.services[service],
+				};
+			}
 		});
 	}
 
-	if (configPath.endsWith('json')) {
-		return JSON.stringify(config);
+	if (metadataPath.endsWith('json')) {
+		return JSON.stringify(metadata);
 	} else {
-		return require('js-yaml').dump(config);
+		return require('js-yaml').dump(metadata);
 	}
 }
 
@@ -876,7 +910,11 @@ async function newTarDirectory(
 
 	// Get balena config so we can set extra build variables if provided
 	const parsedBuildArgs = await buildArgumentsFromInput(buildArgs, composition);
-	const [config, configPath] = await loadBuildMetadata(dir);
+	const buildMetadata = await loadBuildMetadata(dir);
+	const metadata = buildMetadata[0];
+
+	// dir and metadataPath are either both relative or both absolute
+	const metadataPath = path.normalize(path.relative(dir, buildMetadata[1]));
 
 	const tar = await import('tar-stream');
 	const pack = tar.pack();
@@ -886,6 +924,7 @@ async function newTarDirectory(
 		dockerignoreFiles,
 	} = await filterFilesWithDockerignore(dir, multiDockerignore, serviceDirs);
 	printDockerignoreWarn(dockerignoreFiles, serviceDirs, multiDockerignore);
+	let foundConfig = false;
 	for (const fileStats of filteredFileList) {
 		pack.entry(
 			{
@@ -894,22 +933,34 @@ async function newTarDirectory(
 				mode: fileStats.stats.mode,
 				size: fileStats.stats.size,
 			},
-			fileStats.relPath === configPath
-				? setConfig(configPath, config, parsedBuildArgs)
+			metadataPath && fileStats.relPath === metadataPath
+				? addBuildVariablesToBuildMetadata(
+						metadataPath,
+						metadata,
+						parsedBuildArgs,
+				  )
 				: await readFile(fileStats.filePath),
 		);
+		if (metadataPath && fileStats.relPath === metadataPath) {
+			foundConfig = true;
+		}
 	}
 	if (
-		configPath === '' &&
+		!foundConfig &&
 		(!_.isEmpty(parsedBuildArgs.global) || !_.isEmpty(parsedBuildArgs.services))
 	) {
-		const configData = setConfig('.balena/balena.yml', {}, parsedBuildArgs);
+		const configData = addBuildVariablesToBuildMetadata(
+			'.balena/balena.yml',
+			{},
+			parsedBuildArgs,
+		);
+		const buf = Buffer.from(configData);
 		pack.entry(
 			{
-				name: toPosixPath('.balena/balena.yml'),
-				size: configData.length,
+				name: '.balena/balena.yml',
+				size: buf.length,
 			},
-			configData,
+			buf,
 		);
 	}
 	if (preFinalizeCallback) {
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -46,66 +46,37 @@ import { getBalenaSdk, getChalk, stripIndent } from './lazy';
 import Logger = require('./logger');
 
 /**
- * Return a set of build arguments passed in as flag options.
- *
- * @param buildArgs List of raw build argument strings
- * @param composition Optional previously parsed composition object
+ * Parse build argument input from flags like `--buildArg service1:var1=value1`
+ * to produce an object in a similar format as the build-variables section of
+ * the `./.balena/balena.yml` file, separating service-specific build arguments
+ * from global build arguments.
+ * @param buildargs --buildArg flag input in dictionary format
  */
-async function buildArgumentsFromInput(
-	buildArgs: string[],
+async function parseMetadataFromFlags(
+	buildargs: Dictionary<string>,
 	composition?: Composition,
 ): Promise<ParsedBuildArguments> {
-	const serviceNames: string[] =
-		composition && composition.services
-			? Object.keys(composition.services)
-			: [];
-
-	// A normal variable regex, with an added part
-	// to find a colon followed service name at the start
-	const varRegex = /^(?:([^\s:]+):)?([^\s]+?)=(.*)$/;
-
+	const serviceNames = Object.keys(composition?.services || {});
 	const ret: ParsedBuildArguments = {
 		global: {},
 		services: {},
 	};
-	// Populate the object with the service names, as it also
-	// means that we can do a fast lookup of whether a service exists
-	for (const service of serviceNames) {
-		ret.services[service] = {};
-	}
 
-	for (const buildArg of buildArgs) {
-		const maybeMatch = buildArg.match(varRegex);
-		if (maybeMatch == null) {
+	for (const [argSpec, argValue] of Object.entries(buildargs)) {
+		const i = argSpec.indexOf(':', 1);
+		const service = i > 0 ? argSpec.substring(0, i) : '';
+		const argName = i > 0 ? argSpec.substring(i + 1) : argSpec;
+		if (service && !serviceNames.includes(service)) {
 			throw new ExpectedError(
-				`Unable to parse build argument variable: ${buildArg}`,
+				`Cannot find a service with name ${service} specified in build argument: ${argSpec}`,
 			);
 		}
-		const match = maybeMatch!;
-		let service: string | undefined;
-		if (match[1]) {
-			// This is for a service, we check that it actually exists
-			if (!(match[1] in ret.services)) {
-				throw new ExpectedError(
-					`Cannot find a service with name ${match[1]} specified in build argument: ${buildArg}`,
-				);
-			} else {
-				service = match[1];
-			}
-		}
-
 		// Set build argument either as global or service specific
-		if (service != null) {
-			ret.services[service][match[2]] = match[3];
+		if (service) {
+			ret.services[service] ||= {};
+			ret.services[service][argName] = argValue;
 		} else {
-			ret.global[match[2]] = match[3];
-		}
-	}
-
-	// Remove services that have no build variables.
-	for (const service of serviceNames) {
-		if (_.isEmpty(ret.services[service])) {
-			Reflect.deleteProperty(ret.services, service);
+			ret.global[argName] = argValue;
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


lib/utils/device/deploy.ts
```diff
@@ -121,6 +126,58 @@ async function environmentFromInput(
 	return ret;
 }
 
+async function buildArgumentsFromInput(
```
joshbwlng, line 129: Basically a copy paste of `environmentFromInput()`, done under the assumption that we should support setting build arguments for specific services. If this is the way to go, we should probably break out the shared logic.

lib/utils/device/deploy.ts
```diff
@@ -121,6 +126,58 @@ async function environmentFromInput(
...
+	buildArgs: string[],
+	serviceNames: string[],
+	logger: Logger,
+): Promise<ParsedBuildArguments> {
+	const { exitWithExpectedError } = await import('../../errors');
```
srlowe, line 134: `exitWithExpectedError`  is deprecated, in favour of throwing an `ExpectedError`:

```
	const { ExpectedError } = await import('../../errors');
        throw new ExpectedError('Some error');
```

This is then caught and handled by the error handler, rather than exiting ungracefully, which might cause problems.

lib/utils/device/live.ts
```diff
@@ -179,8 +179,9 @@ export class LivepushManager {
 					stageImages: this.imageIds[serviceName],
 					docker: this.docker,
 				});
-				const buildVars = buildTask.buildMetadata.getBuildVarsForService(
-					buildTask.serviceName,
+				const buildVars = _.merge(
+					buildTask.buildMetadata.getBuildVarsForService(buildTask.serviceName),
+					buildTask.args,
```
pdcastro, line 184: Here's what I am thinking. Users can already specify build variables through the `balena.yml` file, which can be used with both livepush (`push <ip-address>`) and cloud push (`push myApp`). The `getBuildVarsForService()` call above takes the build variables from the `balena.yml` file in the build tar stream. The solution currently proposed in this PR goes around that mechanism, and implements something that only works for livepush, not cloud push.

The `--buildArg` flag documentation in this PR, as it is, does not even specify that the feature only works for `balena push <ip-address>`, and not for `balena push myApp`.

We could modify the documentation of the `--buildArg` flag to clarify the restriction, but I am thinking that we could change the implementation so that it works for both local and cloud push.

I do **not** propose that the CLI should to modify any files on disk (the `./.balena/balena.yml` file). However, it could modify the tar stream on the fly like we already do for other purposes, such as converting line endings on Windows:

https://github.com/balena-io/balena-cli/blob/v12.30.4/lib/utils/compose_ts.ts#L658-L664

If we find the `.balena/balena.yml` file in the tar stream, we update it to add the build variables specified with `--buildArg` on the command line (such that the command line takes precedence over any variables of same name that might exist in the file). If we don't find `.balena/balena.yml` in the tar stream, we add a new one.

We should also accept the variations of the file name:
https://github.com/balena-io-modules/resin-multibuild/blob/v4.10.0/lib/build-metadata.ts#L92-L98
```
			{ name: 'balena.yml', type: MetadataFileType.Yaml },
			{ name: 'balena.yaml', type: MetadataFileType.Yaml },
			{ name: 'balena.json', type: MetadataFileType.Json },
			{ name: 'resin.yml', type: MetadataFileType.Yaml },
			{ name: 'resin.yaml', type: MetadataFileType.Yaml },
			{ name: 'resin.json', type: MetadataFileType.Json },
```

Editing the tar stream in this way is a bit more work than the solution currently proposed in this PR, but it is a solution that would work with both local push and cloud push, and would avoid code duplication overall.

`balena.yml` reference:
https://www.balena.io/docs/learn/deploy/deployment/#build-time-secrets-and-variables

joshbwlng: @pdcastro Thanks, I'll take a look at trying that out next week.

iamhosseindhv: This will be very useful and exactly what we're missing in our development process.

pdcastro: @iamhosseindhv, if helpful, note that currently it is possible to specify build variables and secrets using a `.balena/balena.yml` file, as documented on this page: https://www.balena.io/docs/learn/deploy/deployment/#build-time-secrets-and-variables
The `--buildArg` flag to be added in this PR would provide an alternative, more convenient way of doing the same.

Hades32:
@balena-ci retest

lib/utils/compose_ts.ts
```diff
@@ -730,6 +798,46 @@ export async function tarDirectory(
...
+
+	// Set global build variables if specified
+	if (!_.isEmpty(buildArgs.global)) {
+		config['build-variables'].global ||= [];
+		config['build-variables'].global = [
+			...config['build-variables'].global,
+			..._.map(buildArgs.global, (value: string, key: string) => {
+				return `${key}=${value}`;
+			}),
+		];
```
pdcastro, line 817: Am I right that the use of array spread syntax here would allow multiple variables of the same name to remain in the array? For example, I reckon that the returned `config` object might end up with:
```
build-variables:
    global:
        - MY_VAR="Value defined in an existing '.balena/balena.yml' file"
        - MY_VAR="Overriding value provided with --buildArgs"
```
So two values for a single variable. In which case, the ultimate behavior would depend on how the cloud builders or balenaEngine or Docker v19 or Docker v20 or Docker Desktop vs Linux Docker choose to interpret it. This statement may be a bit exaggerated, but is there some truth in it? I am thinking that we are better off "resolving" the overriding behavior in the CLI, in this function, rather than passing the buck to builders / Docker. Unless you are really confident of Docker's / builder behavior, and could perhaps add a link to Docker API documentation that discusses it, which could then be added to the source code as a comment.

Otherwise, here's a suggestion: at least for the purpose of this overriding (but possibly also all the way through to stringification / serialization, given that `resin-multibuild` seems to be prepared to cope with it), instead of using `global: string[]`, we could adopt the `VarList` type of `resin-multibuild`, which I think is simply `{ [key: string]: string; }`, converting from `string[]` to `VarList` as needed, then use object spread syntax instead of array spread syntax:
```
global = { ...{ }, ...{ } }
```
instead of:
```
global = [ ...[ ], ...[ ] ]
```
because object spread syntax would take care of variable overriding for us in a way that array spread syntax doesn't:
```
global = {
    ...{ MY_VAR: "Value defined in an existing '.balena/balena.yml' file" },
    ...{ MY_VAR: "Overriding value provided with --buildArgs" },
}
```

lib/utils/compose_ts.ts
```diff
@@ -775,7 +889,22 @@ async function newTarDirectory(
...
 			},
-			await readFile(fileStats.filePath),
+			fileStats.relPath === configPath
+				? setConfig(configPath, config, parsedBuildArgs)
+				: await readFile(fileStats.filePath),
+		);
+	}
+	if (
+		configPath === '' &&
+		(!_.isEmpty(parsedBuildArgs.global) || !_.isEmpty(parsedBuildArgs.services))
```
pdcastro, line 899: I believe this logic is correct -- if configPath is empty, then the for loop `for (fileStats of filteredFileList)` ought to have found a matching file and called `setConfig`. But I had to stop and think about it for a moment. Would it be clearer to have a `foundConfig` boolean variable that was set to true just before calling `setConfig` if it is ever called? Then here having the test be:
```
	if (
		!foundConfig &&
		(!_.isEmpty(parsedBuildArgs.global) || !_.isEmpty(parsedBuildArgs.services))
```

lib/utils/compose_ts.ts
```diff
@@ -775,7 +889,22 @@ async function newTarDirectory(
...
+	) {
+		const configData = setConfig('.balena/balena.yml', {}, parsedBuildArgs);
+		pack.entry(
+			{
+				name: toPosixPath('.balena/balena.yml'),
```
pdcastro, line 904: `toPosixPath` is not needed here, because the given path is a constant with forward slashes.
The function is defined in `resin-multibuild` and its purpose is to replace backslashes with forward slashes as the path separator:
```
export function toPosixPath(p: string): string {
	return p.replace(nativeSepRE, '/');
}
```

lib/utils/compose_ts.ts
```diff
@@ -775,7 +889,22 @@ async function newTarDirectory(
...
+				size: configData.length,
```
pdcastro, line 905: Not 100% sure, but I think this size is meant to be in bytes, so `configData.length` would only be correct for ASCII strings. Suppose the value of a variable was UTF-8, then:
```
$ node
> '€'.length
1
> Buffer.from('€').length
3
```
Suggestion:
```
const buf = Buffer.from(configData);
pack.entry(
	{
		name: toPosixPath('.balena/balena.yml'),
		size: buf.length,
	},
	buf,
);
```
Or alternatively, dig the `pack` source code and prove to me that it's OK to have a string length as the size, then also add a comment to the source code so no else wonders if it's correct. :-)

lib/utils/compose_ts.ts
```diff
@@ -918,7 +1047,7 @@ export async function checkBuildSecretsRequirements(
 	docker: Dockerode,
 	sourceDir: string,
 ) {
-	const [metaObj, metaFilename] = await loadBuildMetatada(sourceDir);
```
pdcastro, line 921: 😁

pdcastro (changes requested):

lib/utils/compose_ts.ts
```diff
@@ -730,6 +798,46 @@ export async function tarDirectory(
 	}
 }
 
+function setConfig(
+	configPath: string,
+	configData: MultiBuild.ParsedBalenaYml,
+	buildArgs: ParsedBuildArguments,
+): string {
+	const config = configData as any;
```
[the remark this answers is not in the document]
pdcastro: By the way @joshbwlng, I wanted to mention the following, just so you are aware, but there is no need to change course on this PR, unless you really wanted to. In a product call last week, one of the topics was _"what to do about the potential for confusion of a balena app having two `balena.yml` files with different purposes:"_

```
./balena.yml - contracts / Deploy With Balena
./.balena/balena.yml - build secrets and variables
```

Check the **Info** and the **Timeline** tabs of this JF brainstorm topic:
https://jel.ly.fish/90f5a591-db7e-4a1a-8f72-1a38f203e3f2

A proposal was to move the build-variables section from `./.balena/balena.yml` to `./balena.yml`, which already has an `applicationEnvironmentVariables` section. But we would want to maintain backwards compatibility, so I envision a pre-processing step whereby any `build-variables` section of `./.balena/balena.yml` (in `PemissiveVarList` format) is translated to a less permissive format† and moved to `./balena.yml`. The `--buildArgs` logic could then operate on the less permissive format from `./balena.yml`. But there'd still be the complexity of the translation to be implemented. Also, if went that route, we would probably have to address issue #2183 (in relation to `balena push <ip-addr>`) before this PR could be merged. :-/

† Personally I prefer the dictionary format over the array format because it avoids the need of splitting strings on `=` is multiple places, which means spreading the logic of what to do if there are multiple `=` characters to multiple places. Also, as found in this PR, the dictionary format makes it easier to override variables of the same name with spread syntax for objects.

Anyway - I hadn't mentioned the contracts matter before because I thought it would just complicate and delay this PR, and it would be better to address the contract issues in other PRs. And I think it's indeed the case, but I leave it up to you. At least you are now aware of it. :-)

lib/utils/compose_ts.ts
```diff
@@ -44,6 +45,73 @@ import type { DeviceInfo } from './device/api';
 import { getBalenaSdk, getChalk, stripIndent } from './lazy';
 import Logger = require('./logger');
 
+/**
+ * Return a set of build arguments passed in as flag options.
+ *
+ * @param buildArgs List of raw build argument strings
+ * @param composition Optional previously parsed composition object
+ */
+async function buildArgumentsFromInput(
```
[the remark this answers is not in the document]
joshbwlng: Throwing an error if the specified service does not exist seems like a sensible approach with the added benefit of possibly catching environment variable value confusion before it happens. In regards to colons in variable names, from what I understand it is considered an illegal character for both [Unix](https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names#:~:text=So%20names%20may%20contain%20any,not%20begin%20with%20a%20digit.) and [Windows](https://stackoverflow.com/questions/20635695/what-are-valid-characters-for-windows-environment-variable-names-and-values/20635858) environments. Therefore, I'm not sure we should worry about supporting such bad practices in balenaCLI.

And introducing this new behavior initially for `--buildArgs` alone, and then bringing it in to `--env` as a breaking change sounds reasonable. Unless you have any other concerns, I'll go ahead and update this function to throw errors if the specified service isn't defined in the project's composition.

lib/utils/compose_ts.ts
joshbwlng: No, I think you are correct. An oversight on my end.

lib/utils/compose_ts.ts
joshbwlng: This was driving my spellcheck plugin crazy :laughing:

lib/utils/compose_ts.ts
joshbwlng: This is correct, updating the code use object spread syntax for overwrite.

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
joshbwlng: If possible, I'd rather have those issues handled in separate PRs. I see the connection, but it feels like a bit too broad of a scope for what I'm trying to do here. And I'm on board with going the dictionary route, but still doing some debugging on how to properly handle both formats. When setting a global build variable `FIRST_VAR=0` in `.balena/balena.yml`, the parsed object becomes the following (kept some debug logs in for context):
```
initial global value: [
    "FIRST_VAR=0"
]

_.isArray(metadata['build-variables'].global): true

// Add in `SECOND_VAR` from buildArg

metadata: {
    "build-variables": {
        "global": {
            "0": "FIRST_VAR=0",
            "SECOND_VAR": "1234"
        },
        ....
    }
}
```

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
pdcastro: > In regards to colons in variable names, from what I understand it is considered an illegal character for both Unix and Windows environments

Hmm, the problem is that we got bitten by this before. I believe that bash and PowerShell don't allow the dot character in environment variable names, and yet a user needed to set an env var named `INTEGRATION.MQTT.AUTH.GENERIC.SERVER` in a docker-compose file and the supervisor code had to be fixed to allow it: https://github.com/balena-io/balena-supervisor/issues/1104

PowerShell on Windows seems happy to set env vars with a colon in the name, e.g. `MY:VAR`:
```
> $Env:MY:VAR=2

> get-item Env:*
...
MY:VAR        2
```

While `cmd.exe` on Windows is happy with colons and dots:

```
> set MY.VAR=2
> set MY:VAR=2
> set
...
MY.VAR=2
MY:VAR=2
```

And even if most _shells_ (rather than OS kernels) didn't support setting env vars with the dot or colon characters in the name, this doesn't prevent applications (written in most programming languages) setting and using such variables, as the OS kernel (or user space C libraries?) allow them. Node.js is quite happy with dots and colons:

```
> process.env['MY:VAR'] = 2
2
> process.env['INTEGRATION.MQTT.AUTH.GENERIC.SERVER'] = 'tcp://mosquitto:1883'
'tcp://mosquitto:1883'
> process.env
{
  ...
  'INTEGRATION.MQTT.AUTH.GENERIC.SERVER': 'tcp://mosquitto:1883',
  'MY:VAR': '2'
}
```

A Node.js application could set an env var containing colons and dots in the name, and directly pass these variables over to another application executed as a child process (the other app having been written in any other programming language), without getting shells involved.

So I am not sure about simply ignoring the possibility of the colon in environment variable names. It's tempting to say _"bash doesn't support it, we'll be fine",_ but then balena-io/balena-supervisor/issues/1104 is an actual issue that a balena user brought to support, not a hypothetical issue.

By the way, I've remembered about an open CLI issue regarding the `--env` parsing of the '=' character, which we could try to avoid or fix when parsing `--buildArg`: https://github.com/balena-io/balena-cli/issues/1705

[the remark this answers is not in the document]
pdcastro: Oh wait. Build variables / args are not environment variables. `--env` needs to deal with environment variables, but `--buildArgs` only needs to accept whatever Docker / balenaEngine accepts:
* As the `buildargs` JSON object of the Docker API `ImageBuild` endpoint:  
https://docs.docker.com/engine/api/v1.41/#operation/ImageBuild
* As the ARG instruction in a Dockerfile:  
https://docs.docker.com/engine/reference/builder/#arg
* As the `--build-arg` flag of the docker CLI `docker build` command:  
https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg

So if we could determine that the colon character is not acceptable for some of the above, it would have more weight for `--buildArgs` than what an environment variable name may be!

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
pdcastro: > I'd rather have those issues handled in separate PRs

That's fine, I agree it's better. 👍 

> the parsed object becomes the following

`"0": "FIRST_VAR=0"` is a bug that you are debugging, right? It looks like a bug. :-)

> What I'm thinking is first converting to a dictionary if its an array, and then append/overwrite using values set with `--buildArgs`. What do you think?

Sounds like a good idea! 👍  Then, it may not even be necessary to convert back to array format (before adding the entry to the tar stream), even if the original format was an array. This is on the assumption that `resin-multibuild` accepts either format (array or dictionary). We should test this assumption with `balena push` to cloud and to device (or read `resin-multibuild` source code and satisfy ourselves that it is the case).

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
joshbwlng: Tried adding in a colon in the name of a build arg and it doesn't look like it's properly supported? Still looking for something in the documentation to back this up:
```
$ docker build --build-arg MY_VAR=my_value --build-arg NPM:TOKEN=1234 -f Dockerfile .
...
Step 4/17 : ARG MY_VAR
 ---> Running in c7cc8ff2fadb
Removing intermediate container c7cc8ff2fadb
 ---> 211b07ad5d78
Step 5/17 : ARG NPM:TOKEN
 ---> Running in 7a414dae3c98
Removing intermediate container 7a414dae3c98
 ---> fac082db7be5
Step 6/17 : RUN echo "MY_VAR: $MY_VAR"
 ---> Running in 1da221eb092f
MY_VAR: my_value
Removing intermediate container 1da221eb092f
 ---> d2d072226751
Step 7/17 : RUN echo "NPM_TOKEN: $NPM:TOKEN"
 ---> Running in ee6819e2dd11
NPM_TOKEN: :TOKEN
...
```

```
$ docker build --build-arg "MY_VAR=my_value" --build-arg "NPM:TOKEN=1234" -f Dockerfile .
...
Step 4/17 : ARG MY_VAR
 ---> Running in b86167282979
Removing intermediate container b86167282979
 ---> b7bb783ebb0d
Step 5/17 : ARG NPM:TOKEN
 ---> Running in 43eb439b05f7
Removing intermediate container 43eb439b05f7
 ---> 93031a4872c1
Step 6/17 : RUN echo "MY_VAR: $MY_VAR"
 ---> Running in 0fefe6e2bbd2
MY_VAR: my_value
Removing intermediate container 0fefe6e2bbd2
 ---> d62579067df0
Step 7/17 : RUN echo "NPM_TOKEN: $NPM:TOKEN"
 ---> Running in 3e0de989512e
NPM_TOKEN: :TOKEN
...
```

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
joshbwlng: @pdcastro Pushed a commit that seems to be working as expected - converting the metadata build variable sets into dictionaries if arrays, then appending/overwriting with `--buildArg` values. Tested with Livepush, trying cloud push out next.

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
joshbwlng: Also updated `buildArgumentsFromInput` to throw an error if an unknown service name was specified in a build argument instead of assigning it as a global build argument.

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
joshbwlng: Cloud push worked as expected:
- valid NPM_TOKEN passed with `--buildArg` => was able to access private packages
- invalid NPM_TOKEN passed with `--buildArg` => unable to access private packages

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
pdcastro: Neither of the following would work:
```
RUN echo "NPM_TOKEN: $NPM:TOKEN"
RUN echo "NPM_TOKEN: ${NPM:TOKEN}"
```
because they use the "shell form" of RUN, where the line is executed through `/bin/sh -c` on Linux and `cmd /S /C` on Windows: https://docs.docker.com/engine/reference/builder/#run

And `sh` or `bash` don't like variables with a colon in the name. However, if use the "exec form" of RUN and, for example, execute `node` instead of `sh`, then `node` is OK with a variable with a colon in the name:

`Dockerfile`
```
FROM balenalib/intel-nuc-node
ARG MY_VAR
ARG NPM:TOKEN
RUN ["/usr/local/bin/node", "-e", "console.log(process.env)"]
RUN ["/usr/local/bin/node", "-e", "console.log(`NPM:TOKEN: ${process.env['NPM:TOKEN']}`)"]
CMD /bin/sh
```
```
$ docker build --build-arg 'NPM:TOKEN=1234' --no-cache .
...
Step 4/6 : RUN ["/usr/local/bin/node", "-e", "console.log(process.env)"]
 ---> Running in 87c47b8e3c01
{
  PATH: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
  HOSTNAME: '2c99176dfd86',
  LC_ALL: 'C.UTF-8',
  DEBIAN_FRONTEND: 'noninteractive',
  UDEV: 'off',
  NODE_VERSION: '15.7.0',
  YARN_VERSION: '1.22.4',
  'NPM:TOKEN': '1234',
  HOME: '/root'
}

Step 5/6 : RUN ["/usr/local/bin/node", "-e", "console.log(`NPM:TOKEN: ${process.env['NPM:TOKEN']}`)"]

'NPM:TOKEN': 1234
```

Btw, while running this experiment, I've also found this interesting bit of Dockerfile documentation:
> https://docs.docker.com/engine/reference/builder/#using-arg-variables
> _You can use an ARG or an ENV instruction to specify variables that are available to the RUN instruction. Environment variables defined using the ENV instruction always override an ARG instruction of the same name._

So I think that this experiment has shown us a couple of things:
* `docker build --build-arg`, Dockerfile ARG, the Docker API and Docker Engine are happy with build ARG variables with a colon character in the name.
* My previous statement that _"Build variables / args are not environment variables"_ is actually not so true! Build ARG variables are actually provided as environment variables to processes spawned by RUN instructions in exec form, and ENV instructions simply override ARG instructions.

pdcastro:
@joshbwlng, I am contributing a commit that will clarify why you were getting the build error:
`TypeError: Cannot create property 'build-variables' on string 'balena.yml'`
But tests still fail, and I could look into further tomorrow. Also, I think (not sure) we may need to check whether the function `rebuildSingleTask()`, which is called by livepush when certain changes are made to the filesystem, also needs to keep track of build arguments (we wouldn't want build arg flags to be "lost" when livepush rebuilds user containers):
https://github.com/balena-io/balena-cli/blob/v12.40.0/lib/utils/device/deploy.ts#L425

Maybe we should revisit all calls to tarDirectory. Also, I will give some more thought on whether we should unify the several occurrences of `buildargs` (all lowercase) and `buildArgs` (camelCase) in the code (docker options vs flags), and the use of `string[]` in some places and dictionary in others. It was already confusing and complicated before this PR, and adding a separate `buildArgs` property doesn't help much with that. Or maybe not, I will give it some more thought.

And we also need to make a decision on whether to simply disallow the colon character in build arg variable names. I wanted to allow colon characters without resorting to escape characters. Maybe the solution would be additional flags -- needs more thought. Or maybe we simply do the same that was done for `--env`.

lib/utils/docker.ts
```diff
@@ -110,7 +110,7 @@ export interface BuildOpts {
 	t?: string;
 }
 
-function parseBuildArgs(args: string[]): Dictionary<string> {
+export function parseBuildArgs(args: string[]): Dictionary<string> {
```
pdcastro, line 113: I don't think I was the one who originally implemented this function, though I amended it at some point. If I was to reimplement it today (which I actually did before I remembered about it), 😁  I would have coded it without a regex, something like:
```ts
/**
 * Convert an array of variable assignment strings such as:
 *     [ "var1=value1", "var1=value2", "var+3=value3=something", "var4" ]
 * to a dictionary such as:
 *     { var1: "value2", "var+3": "value3=something", var4: "" }
 */
function varArrayToDictionary(varArray: string[]): Dictionary<string> {
	const dict: Dictionary<string> = {};
	for (const buildVar of varArray) {
		let key = buildVar;
		let value = '';
		const i = buildVar.indexOf('=');
		if (i > 0) {
			key = buildVar.substring(0, i);
			value = buildVar.substring(i + 1);
		}
		dict[key] = value;
	}
	return dict;
}
```

pdcastro: Btw, the issue with `split('=')` is that it doesn't handle well the case where the variable value contains the equal character, which is the subject of bug #1705. Sample problematic command line:
```
balena push --env "POSTGRESQL_DSN=postgres://loraserver_ns:59XTH2995@postgresql/loraserver_ns?sslmode=disable" d19baeb.local
```
We should similarly handle such build arguments:
```
balena push --buildArg "POSTGRESQL_DSN=postgres://loraserver_ns:59XTH2995@postgresql/loraserver_ns?sslmode=disable" d19baeb.local
```

joshbwlng: Right, I kind of assumed this split on '=' would be a problem, but left it in for now as something to work from. Not sure how to handle it though off the top of my head.

joshbwlng: And as I'm guessing we can use `parseBuildArgs()` to convert build argument arrays to dictionaries as I mentioned earlier, refactoring it as a part of this PR wouldn't necessarily be out of scope.

pdcastro: > [...] split on '=' would be a problem [...] Not sure how to handle it though
> we can use `parseBuildArgs()` to convert build argument arrays to dictionaries

But I have already done these two things in my commit! It handles the split on `=` and it uses `parseBuildArgs()` to convert build argument arrays to dictionaries. :-)

pdcastro:
I've updated my commit and I think the build tests will now pass (they pass on my Mac and Windows laptops), but the other matters mentioned in my [previous comment](https://github.com/balena-io/balena-cli/pull/2110#issuecomment-781722873) are still pending.

joshbwlng:
@pdcastro Thanks for helping out with this. I'll be able to come back and work on this sometime next week so let me know if there's anything I can do to speed this along. The Jellyfish team is really getting tired of having to constantly edit `.balena/balena.yml` and sometimes accidentally pushing their `NPM_TOKEN`s! :sweat_smile:

pdcastro:
> let me know if there's anything I can do to speed this along

@joshbwlng, we have to decide what to do about build variables with a colon character in the variable name. If we don't have any better ideas, I suggest reverting to the same logic used by the `--env` flag. It allows a variable named `MY:VAR` as long as there isn't a service named `MY` in `docker-compose.yml`. If there is such a service, users at least have the workaround of renaming the service. I don't like it, but I think it is better than disallowing the colon character or using escape characters.

An alternative half baked idea I had was to use different flags:

```
# this sets a variable named literally `svc:MY:VAR` for all services,
# and never depends on the names of any services
--build-arg 'svc:MY:VAR=value'

# this sets a variable named `MY:VAR` for the service named `svc` only,
# that must exist in `docker-compose.yml` (it's an error otherwise)
--service-build-arg 'svc:MY:VAR=value'
```
Or the following, that relies on oclif's `multiple: true` property that produces and array of values for the flag, and has a limitation/bug in that the flag must appear at the end of the command line:
```
# this sets a variable named `MY:VAR` for the service named `svc` only,
# that must exist in `docker-compose.yml` (it's an error otherwise)
--service-build-arg svc 'MY:VAR=value'
```

lib/utils/compose_ts.ts
```diff
@@ -799,18 +771,18 @@ export async function tarDirectory(
...
+ * metadata file, with the contents of any `--buildArgs` flags provided
+ * on the command line (so that the command line flags take precedence).
  */
-function addBuildVariablesToBuildMetadata(
+function mergeMetadataFromFlags(
 	metadataPath: string,
-	metadata: MultiBuild.ParsedBalenaYml,
-	buildArgs: ParsedBuildArguments,
+	metadataFromFile: MultiBuild.ParsedBalenaYml,
+	metadataFromFlags: ParsedBuildArguments,
```
pdcastro, line 781: Renamed to try to make it clearer: override "metadata from file" with "metadata from flags". This leaves the name `buildargs` to mostly refer to a simple dictionary that maps arg names to arg values, while "metadata" refers to the more complex object that includes the `services` and `global` properties.

lib/utils/device/deploy.ts
```diff
@@ -429,6 +429,7 @@ export async function rebuildSingleTask(
 		convertEol: opts.convertEol,
 		multiDockerignore: opts.multiDockerignore,
 		nogitignore: opts.nogitignore,
+		buildargs: opts.buildargs,
```
pdcastro, line 432: This addresses a [previous comment](https://github.com/balena-io/balena-cli/pull/2110#issuecomment-781722873) of mine:

> I think (not sure) we may need to check whether the function `rebuildSingleTask()`, which is called by livepush when certain changes are made to the filesystem, also needs to keep track of build arguments (we wouldn't want build arg flags to be "lost" when livepush rebuilds user containers):
https://github.com/balena-io/balena-cli/blob/v12.40.0/lib/utils/device/deploy.ts#L425
> Maybe we should revisit all calls to tarDirectory.

pdcastro:
I have added a commit that addresses some points raised previously, but it still implements the solution by which the colon character is not allowed in variable names. The colon is used as service name separator and, if the service name is not found in the `docker-compose.yml` file, an error is thrown. I am currently of the view that we should allow the colon in variable names, one way or another, even if through the original "`--env` solution" that allows users to use the colon character by renaming services.

I think the next step for this PR is to address [comment 785485641](https://github.com/balena-io/balena-cli/pull/2110#issuecomment-785485641).

joshbwlng:
@pdcastro After taking another look at https://github.com/balena-io/balena-cli/pull/2110#issuecomment-785485641, I would normally prefer `--service-build-arg svc 'MY:VAR=value'`, but seeing as how that has a bug, `--service-build-arg 'svc:MY:VAR=value'` would be what I would go with. It may make the command a bit longer, but has the benefit of forcing build argument definitions to be much more explicit, which would also help the user in debugging problems. In this case, would `--build-arg` refer to global build arguments? Or would we want to be even more explicit with `--service-build-arg` and `--global-build-arg` options?

pdcastro:
> I would normally prefer `--service-build-arg svc 'MY:VAR=value'`, but seeing as how that has a bug [...]

@joshbwlng, your comment gave me another idea - thanks! 😀 
The bug/limitation I mentioned (that the flag must appear at the end of the command line) applies to any flag that uses `multiple: true`, which is already the case of `--buildArg` in this PR:

```
		buildArg: flags.string({
			description: `...`,
			multiple: true,
```

But it also means that, as currently coded, multiple build variables can be set with **either of:**
```
balena push myApp --buildArg arg1=value1 --buildArg arg2=value2 --buildArg arg3=value3 ...
```
```
balena push myApp --buildArg  arg1=value1  arg2=value2  arg3=value3 ...
```
(Double blank spaces are for clarity only, no special meaning.)
The oclif framework treats both command lines above as equivalent. Looking in particular at the second form, something that every `arg=value` "item" has in common is the presence of the equal character. So, how about we allowed the following:

```
balena push myApp --buildArg  svc1  arg1=value1  svc2  arg2=value2  arg3=value3 ...
```

... where an "item" that does not contain the equal character, like `svc1`, is taken to be a service name that qualifies the next "item" (only). Therefore, in the command line above:
* `arg1=value1` applies to service `svc1` only.
* `arg2=value2` applies to service `svc2` only.
* `arg3=value3` applies to all services (global argument) because it's not immediately preceded by a service name.

This would mean that **service names** are not allowed to contain the '=' character, but I assume this is a reasonable limitation.

What do you think? If in doubt, we also have the option of asking more people for their opinions in Flowdock, e.g. `i/cli` flow.

joshbwlng:
@pdcastro Sure, I'd like to get others opinions on this matter as well, especially since this looks like decisions made for `--buildArg` may affect `--env` in the future. With your last example though, that means this would also be valid correct?
```
balena push myApp
  --buildArg svc1 arg1=value1
  --buildArg svc2 arg2=value2
  --buildArg arg3=value3 ...
```

Personally I think I would organize my command in this manner in order to keep things a bit more legible. If this would also work I quite like this approach. Naming a service scopes the build argument to that service, while no service name scopes the argument to the global level.

lib/utils/compose_ts.ts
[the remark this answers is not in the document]
pdcastro: Also:
* https://forums.balena.io/t/restrictions-for-keys-in-device-device-service-environment-variables/220331  
_"I want to use an application that uses environment variables in a similar form like this: foo:bar:key=somevalue"_

Hades32:
> ```
> balena push myApp
>   --buildArg svc1 arg1=value1
>   --buildArg svc2 arg2=value2
>   --buildArg arg3=value3 ...
> ```
> 
> Personally I think I would organize my command in this manner in order to keep things a bit more legible. 

Please don't go that route. A `--flag` should never have multiple parameters. That's totally unexpected.

I don't know if this is possible with our framework, but I'd like something like this: 
```
--buildArg.svc1 key=val
or
--buildArg[svc1] key=val
```

pdcastro:
> Please don't go that route. A `--flag` should never have multiple parameters. That's totally unexpected.

@Hades32, the balena CLI already has several flags with multiple parameters, including `--buildArg` itself in the `build` and `deploy` commands and the closely related `balena push --env` flag. The full list is:

* `balena build --buildArg`
* `balena deploy --buildArg`
* `balena deploy --release-tag`  
* `balena logs --service`
* `balena preload --add-certificate`
* `balena push --service`
* `balena push --env`
* `balena push --release-tag`
* `balena tunnel --port`
* `balena configure --system-connection`

In the past, when I asked people about their preference of flags to solve specific problems, some people had a clear preference for flags with multiple parameters. For example, in the following Flowdock thread, we were discussing how to implement the `--release-tag` flag and we were considering options including:

https://www.flowdock.com/app/rulemotion/i-cli/threads/XCyAW9_JBlef1rQJdGi21gEpPbX

1. Separate `-key` and `-value` flags that get "zipped".
```
$ balena push MyApp --release-tag-key key1 --release-tag-value value1 \
    --release-tag-key key2 --release-tag-value value2

$ balena push MyApp --release-tag-key key1 --release-tag-value "value1 with spaces"
```
2. Using the blank space as key-value separator (the only character that is not allowed in the key itself).
```
$ balena push MyApp --release-tag "key1 value1" --release-tag "key2 value2"

$ balena push MyApp --release-tag "key1 value1 with spaces"
```
3. Using the colon character as the default key-value separator, but having it be configurable with a `--release-tag-separator` flag.
```
$ balena push MyApp --release-tag key1:value1 --release-tag key2:value2

$ balena push MyApp --release-tag-separator ++
    --release-tag key1:with:colons++value1:with:colons \
    --release-tag "key2++value2 with spaces"
```
4. Using `multiple: true` (multiple parameters)
```
$ balena push MyApp --release-tag key1 --release-tag value1 --release-tag key2 --release-tag value2

$ balena push MyApp --release-tag key1 --release-tag "value1 with spaces" \
    --release-tag key2 --release-tag "value2 with spaces"
```
or with the shorter syntax
```
$ balena push MyApp --release-tag key1 value1 key2 value2

$ balena push MyApp --release-tag key1 "value1 with spaces" key2 "value2 with spaces"
```
and for empty values:
```
$ balena push MyApp --release-tag key1 "" key2 value2

$ balena push MyApp --release-tag key1 "" key2 "value2 with spaces"
```

Among the 4 options above, option 4 (multiple parameters) got the most votes from the "focus group". :-)

Before users make use of any flag, whether it has multiple arguments or not, they have to look at examples or documentation (like `balena help`). Once they see the examples or docs of multiple parameters, it cannot be said to be "totally unexpected" that the flag works like that. For example:

```
$ balena help deploy
...
 --release-tag <release-tag>
        Set release tags if the image deployment is successful. Multiple
        arguments may be provided, alternating tag keys and values (see examples).
        Hint: Empty values may be specified with "" (bash, cmd.exe) or '""' (PowerShell).
...
EXAMPLES
  $ balena deploy myApp myApp/myImage --release-tag key1 "" key2 "value2 with spaces"
```

I doubt that balena CLI is that much isolated in using flags with multiple parameters.

> I don't know if this is possible with our framework, but I'd like something like this:
> `--buildArg.svc1 key=val`
> or
> `--buildArg[svc1] key=val`

It would be nice to have this option in the toolset, but as far as I am aware, oclif does not have support for it. Perhaps @srlowe or @mbalamat came across it before.

Hades32:
Thank you @pdcastro for the VERY elaborate explanation!

I wasn't aware that there are already multi-param flags in the CLI, which definitely makes adding another one less of an issue for me.

I'm still not a fan of this syntax though, and I haven't seen this anywhere else in the wild. Anyway, I find consistency much more important than style, so better not introduce something else, unless we want to change all of them...

BTW, I think you could even pull this off with OCLIF. You would need to parse the compose file very early, so you have the list of services, and then dynamically create the list of flags...

pdcastro:
> BTW, I think you could even pull this off with OCLIF. You would need to parse the compose file very early, so you have the list of services, and then dynamically create the list of flags...

It must be possible to patch oclif to do this one way or another, but a reason why it may not be as easy as we'd like is that oclif uses **static** properties of classes (rather than regular non-static properties) to represent flags: https://oclif.io/docs/flags  
It's not the first time that we regret this design decision of oclif, as it also makes life harder for CLI autogeneration from something like a SBVR data model (or the Node SDK). Last time we visited this topic, iirc our conclusion was that we would have to automate the editing of source code templates, as opposed to the otherwise preferable (more flexible) option of dynamically extending command prototypes. cc: @srlowe 

Also, the year-long pain of switching the balena CLI from Capitano to oclif -- a very large effort with little to show in terms of features and usability to end users -- is still too fresh in our memories. The main motivation for the switch was so that balena would spend less effort maintaining our own CLI framework, and more time in features and bug fixes that balena users care about. The more we patch oclif to behave differently (whether it's a PR that we submit upstream or a patch file in the CLI repo), the more we betray the spirit of that effort. So I'd say: try and convince the oclif maintainers to add the `buildArg.svc1` feature, then we can consider adopting it. :-)

lib/utils/compose_ts.ts
```diff
@@ -44,6 +45,44 @@ import type { DeviceInfo } from './device/api';
...
+		services: {},
+	};
+
+	for (const [argSpec, argValue] of Object.entries(buildargs)) {
+		const i = argSpec.indexOf(':', 1);
+		const service = i > 0 ? argSpec.substring(0, i) : '';
+		const argName = i > 0 ? argSpec.substring(i + 1) : argSpec;
+		if (service && !serviceNames.includes(service)) {
+			throw new ExpectedError(
+				`Cannot find a service with name ${service} specified in build argument: ${argSpec}`,
```
pdcastro, line 71: @srlowe pointed out that the set of service names to check against (before erring out) should be the union of the services in `docker-compose.yml` and the services in `docker-compose.dev.yml`, in case the user added the latter to override the former:
https://www.flowdock.com/app/rulemotion/i-cli/threads/C9VgXn2zKz8Tdea_a1ShX7Hu3Rp

joshbwlng: @pdcastro Added in support for also checking `docker-compose.dev.yml` services in my latest commit.

Hades32:
@balena-ci rebase

lib/utils/compose_ts.ts
```diff
@@ -53,10 +53,20 @@ import Logger = require('./logger');
  * @param buildargs --buildArg flag input in dictionary format
  */
 async function parseMetadataFromFlags(
+	dir: string,
 	buildargs: Dictionary<string>,
 	composition?: Composition,
 ): Promise<ParsedBuildArguments> {
-	const serviceNames = Object.keys(composition?.services || {});
+	const yaml = await import('js-yaml');
+	const mergedComposition: Composition = yaml.load(
```
pdcastro: The build was failing on this line with the error:
```
lib/utils/compose_ts.ts(61,8): error TS2322: Type 'string | number | object | null | undefined' is not assignable to type 'Composition'.
  Type 'undefined' is not assignable to type 'Composition'.
```
I've added a commit that fixes it, and also refactors the code to avoid the `yaml.load` and `yaml.dump` in this function (I was a bit worried about `yaml.dump` because of #2226 and #2227, although I think it would be fine here).

pdcastro:
Closing in favor of #2255 
Internal discussion:
* https://jel.ly.fish/fdaef189-6fa4-4d0c-bbc7-f7bcd4bb95ed
* https://www.flowdock.com/app/rulemotion/i-cli/threads/lqsgv7orNqQAinI5hwKYDs66nLw

PR #2112, opened 2020-12-01
device: Add deactivate command

mbalamat:
Depends-on: balena-io/balena-sdk/pull/1024

This PR adds the deactivate option on the device resource.

Given a device UUID, it deactivates the device with this UUID. Also, asks for confirmation and shows a warning regarding billing when deactivating a device the same way as the

```
$ balena device deactivate 6xxxxe
?
Warning! Deactivating a device will charge a fee equivalent to the
normal monthly cost for the device (e.g. $1 for an essentials device);
the device will not be charged again until it comes online.

Are you sure you want to deactivate device 6xxxxe ? (y/N)
```


lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,104 @@
...
+					is_active: false,
+				},
+				options: {
+					$filter: {
+						uuid: {
+							$startswith: uuid,
+						},
+					},
+				},
+			} as any);
```
thgreasi, line 98:
```suggestion
			await balena.pine.patch<BalenaSdk.Device>({
				resource: 'device',
				body: {
					is_active: false,
				},
				options: {
					$filter: {
						uuid: {
							$startswith: uuid,
						},
					},
				},
			});
```

lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,104 @@
...
+		try {
+			await balena.pine.patch({
+				resource: 'device',
+				body: {
+					is_active: false,
+				},
+				options: {
+					$filter: {
+						uuid: {
+							$startswith: uuid,
```
thgreasi, line 94: How about doing a `balena.models.device.get(uuid, { $select: 'id' })` right before, so that
* the user gets an error if the device doesn't exist or is ambiguous
* we can use the retrieved `id` directly in the `patch`, instead of a `$startswith` `$filter`

lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,104 @@
...
+		const deactivationWarning = `
+
+Warning! Deactivating a device will charge a fee equivalent to the
+normal monthly cost for the device (e.g. $1 for an essentials device);
+the device will not be charged again until it comes online.
+
+Are you sure you want to deactivate device ${uuid} ?`;
+
+		// Confirm
+		await patterns.confirm(options.yes, deactivationWarning);
```
pdcastro, line 82: `!!options.yes` - To satisfy typing if `options.yes` is declared as possibly undefined as per other comment.

lib/commands/device/deactivate.ts
mbalamat: Also, @thgreasi is it ok that I am using a pine query here or we should move the `deactivation` to the SDK?

thgreasi: My concern is just that $filter $startswith on a PATCH doesn't surface ambiguity issues. I mean, such a patch could apply on more than one devices unintentionally, if it just happens that they share a common uuid prefix and the user passes just that in the CLI. So avoiding $startswith in this case, makes sure that the patch will only apply on a single device, as the command suggests. The cost of the extra GET shouldn't not be a concern, since we also need to surface 404s, which otherwise we would miss.

@mbalamat adding `device.deactivate` to the SDK sounds like a good idea.

srlowe: IMO proper approach would be to add `models.device.disable()` (or something functionally equivalent) to the SDK.

lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,104 @@
...
+
+		const balena = getBalenaSdk();
+		const patterns = await import('../../utils/patterns');
+
+		const uuid = params.uuid;
+		const deactivationWarning = `
+
+Warning! Deactivating a device will charge a fee equivalent to the
+normal monthly cost for the device (e.g. $1 for an essentials device);
+the device will not be charged again until it comes online.
```
pdcastro, line 77: Grammatically, deactivation will not charge a fee... Balena will charge a fee, because of deactivation. 😀   Perhaps 'incur' instead of 'charge':

```
Warning! Deactivating a device will incur a fee equivalent to the
normal monthly cost for the device (e.g. $1 for an essentials device);
the device will not be charged again until it comes online.
```

Also:
* Re "until it comes online" - are devices charged based on being online/offline, or based or being active/inactive? An active device may be online or offline, and is charged regardless, no? Similarly, an inactive device may be online in the sense of connected to internet (and perhaps still communicating with the user's own servers), and yet not be charged. Is this right? I may be wrong about this, don't assume I know better!
* Re "the device will not be charged" - to be picky, users are charged, not devices (we don't ask devices to pay us money, we ask their owners).

If you also think that these notes are correct, then perhaps:

```
Warning! Deactivating a device will incur a fee equivalent to the
normal monthly cost for the device (e.g. $1 for an essentials device);
device fees will not be charged again until the device is reactivated.
```

lib/commands/device/deactivate.ts
pdcastro: > So avoiding $startswith in this case, makes sure that the patch will only apply on a single device, as the command suggests.

Oh yeah. In fact, it would be nice to fetch the full UUID and print the full UUID on the confirmation prompt, e.g.:

```
$ balena device deactivate 59c195a
...
Are you sure you want to deactivate device 59c195afe15206862efc380873e4bdc8 ? (y/N)
```
If the query for the short UUID returns more than one full UUID (ambiguity), the command should aborted, which also covers the `--yes` case.

lib/commands/device/deactivate.ts
pdcastro: > until the device is reactivated

By the way, how can the device be reactivated? Do we need another CLI command for this? `balena device activate`? If so, it would make sense for this PR to add both commands (could be in separate commits).

lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+import { flags } from '@oclif/command';
+import type { IArg } from '@oclif/parser/lib/args';
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+
+interface FlagsDef {
+	yes: boolean;
```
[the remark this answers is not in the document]
srlowe: `yes: boolean` is correct I believe, because the absence of a bool flag is interpreted as a `false` value when parsed, it won't ever be undefined.

lib/commands/device/deactivate.ts
srlowe: see other comment

lib/commands/device/deactivate.ts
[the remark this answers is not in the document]
pdcastro: Oh ok! 👍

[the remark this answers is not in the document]
pdcastro: Except that's not what I am observing in a test:
```
	public async run() {
		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
			DeviceDeactivateCmd,
		);

		console.error(`yes=${options.yes}`);
		return;
...
```
```
$ ./bin/balena-dev device deactivate foo
yes=undefined
```

lib/commands/device/deactivate.ts
mbalamat: @pdcastro yeah we don't have this activate command, but the `until it comes online.` is correct I think, because if it comes online it will get charged.

> An active device may be online or offline, and is charged regardless, no? Similarly, an inactive device may be online in the sense of connected to internet (and perhaps still communicating with the user's own servers), and yet not be charged. Is this right?


this I think is incorrect,

Though these are billing stuff with which I am not familiar, maybe @thgreasi can give us some input here. Also whatever we decide on this message, we should change the corresponding one in the balena-ui in order to match. That's the place I initially took it from.

pdcastro: I see! 👍  So a device is reactivated by coming back online, and therefore there is no need for a reactivate command. In this case, do we need to check that the device is offline (well, not online) before deactivating it? What happens if the user deactivates a device that is currently online?

mbalamat: @pdcastro the logic you describe I think I've seen it somewhere, while I was searching around to do this feature, so I can try and reproduce on balena-staging to see what errors come out of this. I think the approach should be to have as simple logic as possible, for example, if the API does the necessary checks and just returns a nice error then we shouldn't check the same conditions here in the CLI.

lib/commands/device/deactivate.ts
mbalamat: @pdcastro @thgreasi thanks for the suggestions, this https://github.com/balena-io/balena-cli/pull/2112/commits/795d48b4ed921a20f0444c48ff9b8924c781f97d#diff-38726520c30296d8fa2b5e20c4e181ade050b5e9eec92d853b81ddec073fb171R88 will throw a BalenaAmbiguousDevice error if there is any ambiguity, which I am catching. So, the code, as it is right now works only with one device id, and deactivates it, no ambiguity issues.

lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,111 @@
...
+				},
+				options: {
+					$filter: {
+						id,
+					},
+				},
+			});
+		} catch (err) {
+			if (err instanceof BalenaAmbiguousDevice) {
+				console.info(
```
pdcastro, line 102: `console.error` instead of `console.info`, so the message goes to `stderr` instead of `stdout`. Alternatively:
```
		const logger = await DeviceDeactivateCmd.getLogger();
		logger.logError(`...`);
```

lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,111 @@
...
+					`Error: Multiple devices found that their uuid starts with ${uuid}`,
```
pdcastro, line 103: `that their` doesn't sound very good... Alternatives:
* `Error: Multiple devices found whose UUID starts with ${uuid}`,
* `Error: Multiple devices found with UUID starting with ${uuid}`,
* `Error: Multiple devices found of UUID starting with ${uuid}`,

pdcastro: Also, you could have all CLI commands take advantage of this error handling and error message rewriting by doing it in `lib/errors.ts` instead of doing it in this file. Instead of catching the error here, let it bubble up to `errors.ts` (simply don't catch it here) and add a similar handling as for `BalenaExpiredToken` in the following lines:
https://github.com/balena-io/balena-cli/blob/v12.29.0/lib/errors.ts#L140-L142

It works because `BalenaAmbiguousDevice` defines an error code:
https://github.com/balena-io-modules/balena-errors/blob/v4.7.1/lib/errors.ts#L339

You would also add `BalenaAmbiguousDevice` to the following regex array:
https://github.com/balena-io/balena-cli/blob/v12.29.0/lib/errors.ts#L145-L146

The global error handling in `errors.ts` already sets `process.exitCode` to 1 by default:
https://github.com/balena-io/balena-cli/blob/v12.29.0/lib/errors.ts#L181-L184

lib/commands/device/deactivate.ts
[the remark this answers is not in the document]
srlowe: @pdcastro Hmm, the Typescript checking does not predict this - I wonder why?  We should investigate, since all the other commands use non-optional booleans too.  It doesn't seem to affect functionality though thankfully.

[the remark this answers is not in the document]
srlowe: I think the framework should be presenting this as a normal bool (true = flag present, false = flag not present).  I don't think it makes sense to have 3 possible values.  I'll try to check how oclif is supposed to work in these cases.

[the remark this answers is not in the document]
pdcastro: Maybe the `yes` flag needs to specify `default: false` around here:
https://github.com/balena-io/balena-cli/blob/v12.29.0/lib/utils/common-flags.ts#L60-L63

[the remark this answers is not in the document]
srlowe: Yeah I did some tests, and that limits it to true/false. So perhaps we should standardise on that?   We could miss it, but there are no bad side effects if we do (and perhaps we could subclass that flags.boolean, with the default set). I don't know why it doesn't default to false by default (I've asked a question on oclif repo).

lib/errors.ts
```diff
@@ -140,10 +140,14 @@ const messages: {
 	BalenaExpiredToken: () => stripIndent`
 		Looks like the session token has expired.
 		Try logging in again with the "balena login" command.`,
+
+	BalenaAmbiguousDevice: () => stripIndent`
+		Error: Multiple devices found whose UUID starts with the provided UUID.`,
```
pdcastro, line 145: Re _"the provided UUID"_ - as a global error handler in `errors.ts`, we don't know where exactly this error was raised. The error may have come from a dependency in `node_modules`. Also, some CLI commands allow multiple device UUIDs to be typed:
```
$ balena env add EDITOR vim --device 7cf02a6,d6f1433
```
... and the user would not know which of the many UUIDs had a problem if we print "the provided UUID". The default error message for `BalenaAmbiguousDevice` actually includes the offending UUID:

> https://github.com/balena-io-modules/balena-errors/blob/v4.7.1/lib/errors.ts#L336
> `Device is ambiguous: ${device}`

So I'm thinking we might be better off not translating the error message. This would mean deleting the lines above, but keeping the entry in `EXPECTED_ERROR_REGEXES`. I think the user would then get an error message similar to:

```
$ balena device deactivate a782d09
BalenaAmbiguousDevice: Device is ambiguous: a782d09

$ balena env add EDITOR vim --device 7cf02a6,a782d09
BalenaAmbiguousDevice: Device is ambiguous: a782d09
```
It not the nicest error message, but at least it points to the offending UUID.

lib/commands/device/deactivate.ts
```diff
@@ -0,0 +1,86 @@
...
+Warning! Deactivating a device will charge a fee equivalent to the
+normal monthly cost for the device (e.g. $1 for an essentials device);
+the device will not be charged again until it comes online.
+
+Are you sure you want to deactivate device ${uuid} ?`;
+
+		// Confirm
+		await patterns.confirm(!!options.yes, deactivationWarning);
+		// Deactivate
+		await balena.models.device.deactivate(uuid);
```
pdcastro, line 84: Looks like this depends on: balena-io/balena-sdk/pull/1024
So it will also require bumping `balena-sdk` in the `package.json` file.

lib/commands/device/deactivate.ts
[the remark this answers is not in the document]
srlowe: @mbalamat I'd prefer if we use the `default:false` approach, as it keeps casting out of the logic, and better typing in the flags.  Could you please amend the common flags file?

lib/commands/device/deactivate.ts
pdcastro: Given that there is no such thing as an explicit "reactivate" action (devices are automatically reactivated when they come back online, as I've learned!), I retract my suggestion of using the word "reactivated" on the message. The amended suggestion would be:
```
Warning! Deactivating a device will incur a fee equivalent to the normal
monthly cost for the device (e.g. $1 for an Essentials device); device
fees will not be charged again until the device comes back online.
```

npm-shrinkwrap.json
```diff
@@ -2735,6 +2777,17 @@
         "ndjson": "^1.5.0",
         "semver": "^7.3.2",
         "tslib": "^2.0.1"
+      },
+      "dependencies": {
+        "balena-errors": {
+          "version": "4.7.1",
```
thgreasi, line 2783: This also bumped balena-errors, so let's also bump it in the package.json so we use the same version across the whole CLI.

lib/commands/device/deactivate.ts
thgreasi: Wow big discussing on this warning :)
My bits: the warning in this PR is identical with the one in the UI and which was originally suggested by Alison.
I'm up to further improving it, but it might already be good enough and has the bonus of us being consistent with the UI.
See: https://github.com/balena-io/balena-ui/blob/d5cc98107194dea52ddb0a16b687ac7028dda4ad/src/locales/en/translation.json#L488

If we insist on mentioning that the device has to be offline, then it's probably better to mention that in the action's description like: `Deactivate an offline device.`

lib/errors.ts
```diff
@@ -140,10 +140,14 @@ const messages: {
...
 };
 
 const EXPECTED_ERROR_REGEXES = [
 	/^BalenaAmbiguousApplication/, // balena-sdk
+	/^BalenaAmbiguousDevice/, // balena-sdk
```
thgreasi, line 150: Bumping balena-errors and running `npm dedupe` might allow you to drop this.

thgreasi (commented):
Other than the balena-errors bump and the extra regex, LGTM.

lib/errors.ts
pdcastro: `EXPECTED_ERROR_REGEXES` contains regular expressions for a selected list of errors that should not be reported to Sentry.io (in addition to errors that are subclasses of the CLI's own `ExpectedError` class). I don't see how bumping `balena-errors` would remove the need of listing `BalenaAmbiguousDevice` in this array.

lib/utils/common-flags.ts
```diff
@@ -60,6 +60,7 @@ export const verbose: IBooleanFlag<boolean> = flags.boolean({
 export const yes: IBooleanFlag<boolean> = flags.boolean({
 	char: 'y',
 	description: 'answer "yes" to all questions (non interactive use)',
+	default: false,
```
pdcastro, line 63: Might as well add `default: false` to the other boolean flags in this file (`verbose`, `force`).

package.json
```diff
@@ -197,11 +197,11 @@
     "JSONStream": "^1.0.3",
     "balena-config-json": "^4.1.0",
     "balena-device-init": "^5.0.2",
-    "balena-errors": "^4.4.1",
+    "balena-errors": "^4.7.1",
     "balena-image-manager": "^7.0.3",
     "balena-preload": "^10.3.1",
     "balena-release": "^3.0.0",
-    "balena-sdk": "^15.6.0",
+    "balena-sdk": "^15.20.0",
```
pdcastro, line 204: The commit of title `npm-shrinkwrap: Bump balena-sdk to v15.20.0` is missing the following line in its body as explained in the [Contributing](https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#semantic-versioning-commit-messages-and-the-changelog) document.
```
Update balena-sdk from 15.6.0 to 15.20.0
```
Btw, also as explained in the Contributing document, we should squash some commits before merging, as some commits are undoing what is done in earlier commits in this same PR. (I reckon that you are using many commits to make reviews easier.)

npm-shrinkwrap.json
```diff
@@ -1673,9 +1673,9 @@
       "integrity": "sha512-RaE0B+14ToE4l6UqdarKPnXwVDuigfFv+5j9Dze/Nqr23yyuqdNvzcZi3xB+3Agvi5R4EOgAksfv3lXX4vBt9w=="
     },
     "@types/memoizee": {
-      "version": "0.4.4",
-      "resolved": "https://registry.npmjs.org/@types/memoizee/-/memoizee-0.4.4.tgz",
-      "integrity": "sha512-c9+1g6+6vEqcw5UuM0RbfQV0mssmZcoG9+hNC5ptDCsv4G+XJW1Z4pE13wV5zbc9e0+YrDydALBTiD3nWG1a3g=="
+      "version": "0.4.5",
```
pdcastro, line 1676: Builds are failing with the error:
```
** npm-shrinkwrap.json was not deduplicated or not fully committed - FAIL **
** Please run 'npm ci', followed by 'npm dedupe' **
Error: "test-shrinkwrap": Error: /tmp/build/b3c2877b/versioned-source/automation/test-lock-deduplicated.sh failed with exit code 1:
"/tmp/build/b3c2877b/versioned-source/automation/test-lock-deduplicated.sh" [undefined]
```

What I usually do to fix such errors (and always when a switch to a new branch on my computer, where changes to `package.json` are involved) is run:
```
rm -rf node_modules && npm i
```

`npm ci` is not helpful when changes are made to `package.json`, because `npm ci` does not modify the shrinkwrap file as needed. Running `npm install` without first deleting the `node_modules` folder will often (not always) also work correctly, as `npm i` is somewhat influenced by the existing contents of the `node_modules` folder. Combining `npm i` with `dedupe` improves the outcome:

```
npm i && npm dedupe && npm i
```

An easy mistake is to forget to run `npm i` _after_ running `npm dedupe` -- iirc, this one is guaranteed to fail.

package.json
mbalamat: @pdcastro yep as soon as it's ok I'll squash everything, keeping only 2 commits.  The module bump and the code changes.

pdcastro: @mbalamat, re this commit:
```
packages: Bump balena-sdk and balena-errors

Update balena-sdk from 15.6.0 to 15.20.0
Update balena-errors from 4.4.1 to 4.7.1

Change-type: minor
Signed-off-by: Marios Balamatsias <mbalamatsias@gmail.com>
```
Somehow, versionbot thinks it is reason for a major CLI version bump, and proposes to release CLI v13:
* On the PR page, click the Checks tab, then navigate to Versionbot3 > VersionBot/generateversion
* Click on 13.0.0, which links to: https://github.com/balena-io/balena-cli/commit/b9fdcd8b767408160f88e7fa0838fd0e8021c255

I am not sure exactly why, but the Contributing document has some notes about specifying the desired version bump level between patch / minor / major. Also, the document clarifies that only packages listed in repo.yml should have a line in the commit body; `balena-sdk` is listed in `repo.yml`, while `balena-errors` is not. Try deleting the line regarding `balena-errors`, so the commit looks like:
```
packages: Bump balena-sdk and balena-errors

Update balena-sdk from 15.6.0 to 15.20.0

Change-type: minor
```

pdcastro (changes requested):

pdcastro (approved):

Issue #1828, opened 2020-05-15
Livepush process didn't exit after connection hangup


StefKors:
Running live-push with a NUC on balena-on-balena devenv causes problems with connection hangups. 

# Expected Behavior
Livepush workflow for balena on balena development

# Actual Behavior
Device disconnect while developing the balena-api on a bob.
See logs: https://gist.github.com/StefKors/41f6f59d48c4a5fd7934387732eb1343

# Specifications
- **HOST OS VERSION** balenaOS 2.48.0+rev3
- **SUPERVISOR VERSION** 11.4.3
- **balena CLI version:** e.g. 1.2.3 (output of the `"balena version -a"` command)
- **Operating system version:**  macOS 10.15.4
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** npm
- **If npm install, Node.js and npm version:** 12.8.4

# Additional References
FD: https://www.flowdock.com/app/rulemotion/resin-starters/threads/nQMxNkQq9aRIzfMg7cMuuXdKC00

pdcastro:
Further discussion on a [another thread](https://www.flowdock.com/app/rulemotion/r-beginners/threads/-6VzEum82yOienwJxGhypOiZ6C-):

> The "Connection to device lost" message is printed on this line:
> https://github.com/balena-io/balena-cli/blob/v12.25.2/lib/utils/device/logs.ts#L48
> 
> When it happens, the promise returned by `displayDeviceLogs` would be resolved. `displayDeviceLogs` is called either here:  
> https://github.com/balena-io/balena-cli/blob/v12.25.2/lib/utils/device/deploy.ts#L267
> or here:  
> https://github.com/balena-io/balena-cli/blob/v12.25.2/lib/utils/device/deploy.ts#L288
> 
> Assuming the `--nolive` option is not used, it would be the former. In this case, an array of promises is awaited, including the promise returned by `livepush.init()`:
> 
> * https://github.com/balena-io/balena-cli/blob/v12.25.2/lib/utils/device/deploy.ts#L261
> * https://github.com/balena-io/balena-cli/blob/v12.25.2/lib/utils/device/live.ts#L85

PR #2113, opened 2020-12-02
Livepush, logs: Automatically reconnect on 'Connection to device lost'

pdcastro:
This PR started with #1828, then extended to include automatic reconnections to the device on `balena logs <ip-address>` and `balena push <ip-address>`, and then to cover CTRL-C handling on all platforms (it was specially buggy on Windows) for the `push <myApp>`, `push <myApp> --detached`, `push <ip-address>` and `logs` commands. I got into fixing CTRL-C because I was using it to test the issues with CLI process getting stuck on "connection lost".

This PR has not fixed CTRL-C handling for the `build` and `deploy` commands, which have a different code base. CTRL-C is broken there too... But it was too much for a single PR. 😀 

Resolves: #1828
Change-type: minor

### DIFF
--- a/lib/commands/logs.ts
+++ b/lib/commands/logs.ts
@@ -23,6 +23,7 @@ import { LogMessage } from 'balena-sdk';
 import { IArg } from '@oclif/parser/lib/args';
 
 interface FlagsDef {
+	'max-retry'?: number;
 	tail?: boolean;
 	service?: string[];
 	system?: boolean;
@@ -32,6 +33,8 @@ interface FlagsDef {
 interface ArgsDef {
 	device: string;
 }
+
+const MAX_RETRY = 1000;
 
 export default class LogsCmd extends Command {
 	public static description = stripIndent`
@@ -75,6 +78,11 @@ export default class LogsCmd extends Command {
 	public static usage = 'logs <device>';
 
 	public static flags: flags.Input<FlagsDef> = {
+		'max-retry': flags.integer({
+			description: stripIndent`
+				Maximum number of reconnection attempts on "connection lost" errors
+				(use 0 to disable auto reconnection).`,
+		}),
 		tail: flags.boolean({
 			default: false,
 			description: 'continuously stream output',
@@ -105,7 +113,7 @@ export default class LogsCmd extends Command {
 
 		const balena = getBalenaSdk();
 		const { serviceIdToName } = await import('../utils/cloud');
-		const { displayDeviceLogs, displayLogObject } = await import(
+		const { connectAndDisplayDeviceLogs, displayLogObject } = await import(
 			'../utils/device/logs'
 		);
 		const { validateIPAddress, validateDotLocalUrl } = await import(
@@ -153,13 +161,13 @@ export default class LogsCmd extends Command {
 			}
 
 			logger.logDebug('Streaming logs');
-			const logStream = await deviceApi.getLogStream();
-			await displayDeviceLogs(
-				logStream,
+			await connectAndDisplayDeviceLogs({
+				deviceApi,
 				logger,
-				options.system || false,
-				options.service,
-			);
+				system: options.system || false,
+				filterServices: options.service,
+				maxAttempts: 1 + (options['max-retry'] ?? MAX_RETRY),
+			});
 		} else {
 			// Logs from cloud
 			await Command.checkLoggedIn();
--- a/lib/errors.ts
+++ b/lib/errors.ts
@@ -21,7 +21,10 @@ import { TypedError } from 'typed-error';
 import { getChalk, stripIndent } from './utils/lazy';
 import { getHelp } from './utils/messages';
 
-export class ExpectedError extends TypedError {}
+export class ExpectedError extends TypedError {
+	public code?: string;
+	public exitCode?: number;
+}
 
 export class NotLoggedInError extends ExpectedError {}
 
@@ -38,6 +41,8 @@ export class NoPortsDefinedError extends ExpectedError {
 		super('No ports have been provided.');
 	}
 }
+
+export class SIGINTError extends ExpectedError {}
 
 /**
  * instanceOf is a more reliable implementation of the plain `instanceof`
@@ -177,7 +182,7 @@ async function sentryCaptureException(error: Error) {
 		await Sentry.close(1000);
 	} catch (e) {
 		if (process.env.DEBUG) {
-			console.error('Timeout reporting error to sentry.io');
+			console.error('[debug] Timeout reporting error to sentry.io');
 		}
 	}
 }
@@ -212,8 +217,9 @@ export async function handleError(error: Error) {
 		if (!process.env.BALENARC_NO_SENTRY) {
 			await sentryCaptureException(error);
 		}
-
-		// Unhandled/unexpected error: ensure that the process terminates.
+	}
+	if (error instanceof SIGINTError || !isExpectedError) {
+		// SIGINT or unexpected error: ensure that the process terminates.
 		// The exit error code was set above through `process.exitCode`.
 		process.exit();
 	}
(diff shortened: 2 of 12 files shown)


lib/utils/device/deploy.ts
```diff
@@ -210,7 +206,10 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 	if (!opts.nolive) {
 		buildLogs = {};
 	}
-	const buildTasks = await performBuilds(
+
+	const { awaitInterruptibleTask } = await import('../helpers');
+	const buildTasks = await awaitInterruptibleTask<typeof performBuilds>(
```
pdcastro, line 211: This adds CTRL-C support while performing a build on a local device (`balena push <ip-address>`), even before livepush starts watching the filesystem or showing device logs.

lib/utils/device/deploy.ts
```diff
@@ -244,10 +247,11 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 	// Now that we've set the target state, the device will do it's thing
 	// so we can either just display the logs, or start a livepush session
 	// (whilst also display logs)
+	const promises: Array<Promise<void>> = [streamDeviceLogs(api, opts)];
```
pdcastro, line 250: This refactoring (across multiple commits) achieves 3 things:
- Reduces code duplication - there were 2 copies of `getLoStream`, `displayDeviceLogs`, `outputDeferredMessages`..., now there is only one copy.
- Simplifies the logic (isolates some code in a separate function, easier to read). 
- Adds `livepush.close()` to a `finally` block to allow the CLI process to exit if the connection to the device is lost the requested number of connection retries has been reached.

patches/all/exit-hook+1.1.1.patch
```diff
@@ -0,0 +1,17 @@
...
+@@ -14,9 +14,9 @@ function exit(exit, signal) {
+ 		el();
+ 	});
+ 
+-	if (exit === true) {
+-		process.exit(128 + signal);
+-	}
++	// if (exit === true) {
++	// 	process.exit(128 + signal);
++	// }
```
pdcastro, line 14: This patch removes the need for calling `process.removeAllListeners('SIGINT');`, which was deleted in this PR. The obvious problem with calling `process.removeAllListeners('SIGINT');` is that it does not allow multiple SIGINT handlers to be installed at the same time (including by CLI dependencies), and selectively removed. It was a poor design decision by the author of `exit-hook` to call `process.exit` -- it assumes that `exit-hook` is the only dependency that installs signal handlers, that the application will never need more than one handler for SIGINT, and that the application does not require graceful exit with centralised error handling (CLI's `lib/errors.ts`). We cannot get rid of `exit-hook` just yet because it is used by some dependencies. Hence patching.

lib/commands/logs.ts
```diff
@@ -75,6 +78,11 @@ export default class LogsCmd extends Command {
 	public static usage = 'logs <device>';
 
 	public static flags: flags.Input<FlagsDef> = {
+		'max-retry': flags.integer({
+			description: stripIndent`
+				Maximum number of reconnection attempts on "connection lost" errors
+				(use 0 to disable auto reconnection).`,
```
pdcastro, line 84: Sample command output, simulating connection lost with the `tcpkill` utility on the device:
```
$ ./bin/balena logs 192.168.1.46 --max-retry 2
[Logs]    [12/8/2020, 12:37:11 AM] [main] Plain Dockerfile 47 (11602) howdy (5)
[Logs]    [12/8/2020, 12:37:21 AM] [main] Plain Dockerfile 47 (11603) howdy (5)
[Logs]    [12/8/2020, 12:37:31 AM] [main] Plain Dockerfile 47 (11604) howdy (5)
[Warn]    Connection to device lost
Retrying "Streaming logs" after 1.0s (1 of 2) due to: DeviceConnectionLostError: Connection to device lost
[Warn]    Connection to device lost
Retrying "Streaming logs" after 2.0s (2 of 2) due to: DeviceConnectionLostError: Connection to device lost
[Logs]    [12/8/2020, 12:37:41 AM] [main] Plain Dockerfile 47 (11605) howdy (5)
[Logs]    [12/8/2020, 12:37:51 AM] [main] Plain Dockerfile 47 (11606) howdy (5)
[Warn]    Connection to device lost
Max retry count (2) exceeded while attempting to reconnect to the device
```
... and then the process exited with error code 1, as expected.

lib/errors.ts
```diff
@@ -207,9 +212,13 @@ export async function handleError(error: Error) {
...
-
-		// Unhandled/unexpected error: ensure that the process terminates.
+	}
+	if (error instanceof SIGINTError || !isExpectedError) {
+		// SIGINT or unexpected error: ensure that the process terminates.
 		// The exit error code was set above through `process.exitCode`.
+		if (process.env.DEBUG) {
+			console.error('[debug] Forceful process exit');
+		}
 		process.exit();
```
pdcastro, line 222: Calling `process.exit()` here in the event of SIGINT is necessary (arguably) when the `awaitInterruptibleTask` function is used (see `helpers.ts` in this PR). Without `process.exit()` here, the CLI process would stay running until the "other" (non-sigint) promise in the `awaitInterruptibleTask` function is resolved, which substantially defeats the purpose of SIGINT. (Without `process.exit()`, the process would exit gracefully after that other promise resolved, but it could take a while if the promise is for a long-running build...)
The important / nice thing though, is that `process.exit()` is only called at this very late stage, after the promise rejection bubbled up all the way to centralized error handling in `errors.ts`, and even got reported to Sentry.io if it was an "unexpected error" that happened after SIGINT.

lib/utils/remote-build.ts
```diff
@@ -110,68 +109,76 @@ async function getBuilderEndpoint(
...
+	try {
+		if (build.opts.headless) {
+			await handleHeadlessBuildStream(stream);
+		} else {
+			await handleRemoteBuildStream(build, stream);
+		}
+	} finally {
+		process.removeListener('SIGINT', sigintHandler);
+		globalLogger.outputDeferredMessages();
+		await cancellationPromise;
```
pdcastro, line 143: By the way, during testing I came across issue #2121, but it was pre-existing and the fix lies in the balena-builder repo I believe.

lib/utils/device/live.ts
```diff
@@ -219,24 +219,17 @@ export class LivepushManager {
...
-
-			process.exit(0);
-		});
+	/** Delete intermediate build containers from the device */
+	public async cleanup() {
+		this.logger.logLivepush('Cleaning up device...');
+		await Promise.all(
+			_.map(this.containers, (container) =>
+				container.livepush.cleanupIntermediateContainers(),
+			),
```
pdcastro, line 230: There is also a bug fix here: Previously, the `_.map` callback was returning undefined instead of returning the cleanup promise, so the promises were not really awaited for. (The fix was to remove the curly brackets.)

lib/utils/device/deploy.ts
```diff
@@ -220,6 +219,10 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 		buildLogs,
 	);
 
+	globalLogger.outputDeferredMessages();
+	// Print a newline to clearly separate build time and runtime
+	console.log();
```
pdcastro, line 224: Not new code, just moved around to avoid code duplication.

lib/errors.ts
```diff
@@ -21,7 +21,10 @@ import { TypedError } from 'typed-error';
 import { getChalk, stripIndent } from './utils/lazy';
 import { getHelp } from './utils/messages';
 
-export class ExpectedError extends TypedError {}
+export class ExpectedError extends TypedError {
+	public code?: string;
+	public exitCode?: number;
+}
```
srlowe, line 27: I guess this is for error code support?  Shouldn't we add the new properties to `TypedError`?

srlowe: Ok, I think we should do that at some point, but we can switch to that later I guess.

srlowe (approved):

pdcastro:
@balena-ci rebase

PR #2116, opened 2020-12-07
Remove remaining v12 switches

srlowe:
Change-type: patch

### DIFF
--- a/lib/commands/apps.ts
+++ b/lib/commands/apps.ts
@@ -19,7 +19,6 @@ import { flags } from '@oclif/command';
 import Command from '../command';
 import * as cf from '../utils/common-flags';
 import { getBalenaSdk, getVisuals, stripIndent } from '../utils/lazy';
-import { isV12 } from '../utils/version';
 
 interface ExtendedApplication extends ApplicationWithDeviceType {
 	device_count?: number;
@@ -49,9 +48,7 @@ export default class AppsCmd extends Command {
 		verbose: flags.boolean({
 			default: false,
 			char: 'v',
-			description: isV12()
-				? 'No-op since release v12.0.0'
-				: 'add extra columns in the tabular output (SLUG)',
+			description: 'No-op since release v12.0.0',
 		}),
 	};
 
@@ -88,7 +85,7 @@ export default class AppsCmd extends Command {
 			getVisuals().table.horizontal(applications, [
 				'id',
 				'app_name',
-				options.verbose || isV12() ? 'slug' : '',
+				options.verbose || 'slug',
 				'device_type',
 				'online_devices',
 				'device_count',
--- a/lib/commands/envs.ts
+++ b/lib/commands/envs.ts
@@ -22,11 +22,9 @@ import Command from '../command';
 import { ExpectedError } from '../errors';
 import * as cf from '../utils/common-flags';
 import { getBalenaSdk, getVisuals, stripIndent } from '../utils/lazy';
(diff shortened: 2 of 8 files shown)


doc/cli.markdown
```diff
@@ -982,10 +982,6 @@ Examples:
 
 ### Options
 
-#### --all
-
-No-op since balena CLI v12.0.0.
-
```
srlowe, line 988: This should not have been in the docs, since it was a hidden field (capitanodoc does not currently respect the `hidden` property for flags)

lib/commands/envs.ts
```diff
@@ -128,50 +89,23 @@ export default class EnvsCmd extends Command {
...
 		'envs ' + new CommandHelp({ args: EnvsCmd.args }).defaultUsage()
 	).trim();
 
 	public static flags: flags.Input<FlagsDef> = {
-		...(isV12()
-			? {
-					all: flags.boolean({
-						description: stripIndent`
-				No-op since balena CLI v12.0.0.`,
-						hidden: true,
```
pdcastro, line 165: The flag is hidden, but should not yet be deleted from source code? If a CLI v12 user currently provides `--all` on the command line (because they had been doing it in their scripts since v11), it used to work before this PR, and should still work after this PR... (Otherwise, it's a breaking change.) If my reasoning is correct, you could use the newly added `isV13()` to conditionally add the hidden flag.

Btw, something else I remember we want for `isV13()` is to remove the `--gitignore` flag for the push / build / deploy commands... (Not saying you should do it in this PR, unless you want to...) :-)

lib/utils/version.ts
```diff
@@ -22,13 +22,13 @@ export function isVersionGTE(v: string): boolean {
...
 
-let v12: boolean;
+let v13: boolean;
 
-export function isV12(): boolean {
-	if (v12 === undefined) {
+export function isV13(): boolean {
+	if (v13 === undefined) {
 		// This is the `Change-type: major` PR that will produce v12.0.0.
 		// Enable the v12 feature switches and run all v12 tests.
```
pdcastro, line 30: The comment still says v12 :-)  Suggestion: Delete the comment lines, and add some version neutral docs for the function:
```
/** Feature switch for the next major version of the CLI */
export function isV13(): boolean {
    ...
```

pdcastro (changes requested):

lib/commands/envs.ts
srlowe: Yes, you are right.  I'll change to v13 switch.

srlowe: Ah no, I remember now.   Prior to this PR we had this line:
```
		options.all = options.all || isV12();
```
Since v12, `all` has been forced on, and the CLI flag value is ignored.  So, this PR shouldn't change behaviour.

srlowe: Except... Now adding the `-all` flag would cause an `unexpected argument` error.  Ok, so yeah I think we need to leave the flag in until v13, but it's safe to remove all the logic.

pdcastro (approved):

lib/commands/apps.ts
```diff
@@ -88,7 +85,7 @@ export default class AppsCmd extends Command {
 			getVisuals().table.horizontal(applications, [
 				'id',
 				'app_name',
-				options.verbose || isV12() ? 'slug' : '',
+				options.verbose || 'slug',
```
pdcastro, line 88: Oops, we've introduced this bug. Fixed in #2133

PR #2117, opened 2020-12-07
errors: Add expected errors for device deactivation

mbalamat:


lib/errors.ts
```diff
@@ -149,6 +149,8 @@ const EXPECTED_ERROR_REGEXES = [
 	/^BalenaDeviceNotFound/, // balena-sdk
 	/^BalenaExpiredToken/, // balena-sdk
 	/^BalenaInvalidDeviceType/, // balena-sdk
+	/Cannot deactivate devices of Starter applications.$/, // balena-sdk
```
pdcastro, line 152: I guess you meant for the final dot `.` to match the error message:
```
BalenaRequestError: Request error: Cannot deactivate devices of Starter applications.
```
But `.` has a special meaning in the regex, which matches `.` and any other character, so you might have wanted to escape it as `\.`. But really what I would like to see is more generalisation that avoided us having to change the CLI for small changes to the error message. Other than the error codes discussed in [flowdock](https://www.flowdock.com/app/rulemotion/i-cli/threads/bcx7YUa-8fuhJ2lX55sd8sNuRCX), we could at least use case insensitive matching (`/i` flag) and avoid including the application type like `Starter`. I'd suggest:
```
/Cannot deactivate devices/i,
```
If that much matches, changes are we don't want to report it to Sentry.io.

Re comment `// balena-sdk`, perhaps it should be `// balena-api`, because the error message cannot be found in the `balena-sdk`.

lib/errors.ts
```diff
@@ -149,6 +149,8 @@ const EXPECTED_ERROR_REGEXES = [
 	/^BalenaDeviceNotFound/, // balena-sdk
 	/^BalenaExpiredToken/, // balena-sdk
 	/^BalenaInvalidDeviceType/, // balena-sdk
+	/Cannot deactivate devices/i, // balena-sdk
+	/Devices must be offline in order to be deactivated.$/, // balena-sdk
```
pdcastro, line 153: Case insensitive matching for this one as well?

mbalamat: yeah yeah let me check that one, manually testing it now

mbalamat: ok works fine

lib/errors.ts
```diff
@@ -149,6 +149,8 @@ const EXPECTED_ERROR_REGEXES = [
 	/^BalenaDeviceNotFound/, // balena-sdk
 	/^BalenaExpiredToken/, // balena-sdk
 	/^BalenaInvalidDeviceType/, // balena-sdk
+	/Cannot deactivate devices/i, // balena-api
+	/Devices must be offline in order to be deactivated\.$/i, // balena-api
```
pdcastro, line 153: These still don't cover the case of answering 'No' to the warning prompt, right?

mbalamat: @pdcastro yeah it shows the `Additional information...` message. Let me see if I can find it quickly

pdcastro: I guess (didn't test) the fix is to replace `Error` with `ExpectedError` on this line:
https://github.com/balena-io/balena-cli/blob/v12.30.1/lib/utils/patterns.ts#L162

mbalamat: @pdcastro it comes from https://github.com/balena-io/balena-cli/blob/529b98552cf7abbecc9dbb58fa32734805767251/lib/utils/patterns.ts#L142-L168 and I tried to do `await patterns.confirm(options.yes, deactivationWarning, '', true);` but same behaviour, prints `Additional information may...` I'll dig into this tomorrow. My guess is that this also happens elsewhere.

srlowe:
https://github.com/oclif/cli-ux

lib/commands/device/deactivate.ts
```diff
@@ -71,15 +71,16 @@ export default class DeviceDeactivateCmd extends Command {
...
-
 Warning! Deactivating a device will charge a fee equivalent to the
 normal monthly cost for the device (e.g. $1 for an essentials device);
 the device will not be charged again until it comes online.
+`;
 
-Are you sure you want to deactivate device ${uuid} ?`;
+		const warning = `Are you sure you want to deactivate device ${uuid} ?`;
 
+		console.error(deactivationWarning);
```
pdcastro, line 81: 👍

pdcastro (approved):

Issue #2120, opened 2020-12-08
app create: Blank space somehow inserted in app name


pdcastro:
```
$ balena app create TeSt --type raspberrypi3
Application created: gh_paulo_castro/test (raspberrypi3, id 1771727)

$ balena app TeSt
== TE ST
ID:          1771727
DEVICE TYPE: raspberrypi3
SLUG:        gh_paulo_castro/test

$ balena app test
== TEST
ID:          9073
DEVICE TYPE: raspberry-pi2
SLUG:        page/test
COMMIT:      20e567738ed28b591ec43083bfc670056507237f
```

# Expected Behavior

`balena app TeSt` should have printed the app name as `TeSt` (like the name was provided to `balena app create`)

# Actual Behavior

`balena app TeSt` printed the app name as `TE ST`, which is wrong in two aspects:
* There is a strange blank space in the middle of the name.
* The name is printed in all uppercase, instead of being printed with the actual app name case (`TeSt`).

srlowe:
@pdcastro Turns out this is a display artifact with table.vertical, used in `balena app` (not related to `balena app create`).  I suppose changing this would be a breaking change?

pdcastro:
@srlowe, modifying the behavior of `balena app` to display the app name with the correct capitalisation (and without alien 👽  blank spaces) would not be a breaking change in my opinion (it would be 'patch' - bug fix 🐛 ). However, if the fix was to modify the `resin-cli-visuals` package to remove the `table.vertical` feature of camel case conversion (`myApp` -> `MY APP`) for all users of `resin-cli-visuals`, this would be indeed be a breaking change for `resin-cli-visuals`. Thoughts / alternatives:
* Add an option to `resin-cli-visuals` (if one does not already exist) to disable the camel case conversion, and update `balena app` to use that option.
* Change `balena app` so it does not use `resin-cli-visuals`, but maintain the same table-like output format (not difficult to achieve with simple `console.log` statements I think...)

srlowe:
@pdcastro Third option - perhaps there's something appropriate to use in oclif ux?

pdcastro:
oclif ux - yes, however if the overall output format of `balena app` was different, then it would indeed be a breaking change (as users will be parsing the output).
Talking of parsing the output, adding a `--json (-j)` option would be nice, like `balena devices --json` already has -- but this is a different issue, 😀   and does not avoid the breaking change when `--json` is not used.

srlowe:
[...]

PR #2125, opened 2020-12-09
Fix app name output in app command

srlowe:
Change-type: patch
Resolves: #2120


srlowe:
Also updated the app command info for orgs.

pdcastro (approved):

Issue #2119, opened 2020-12-08
Add orgs support to CLI commands


pdcastro:
> first of all creating apps for specific orgs
> app create is arguably the most important command to implement

References:
* https://www.flowdock.com/app/rulemotion/p-organizations/threads/wlA9GRY1O1_wBJfz2YvXPi_E2CR
* https://www.flowdock.com/app/rulemotion/p-organizations/threads/FxhJoEuUe5wcuDBD8-se64ZoxLS

pdcastro:
Some key points from the linked threads:

* `myOrg/myApp` == `app.slug`
  * Even before orgs, I could refer to an app as `gh_paulo_castro/test`. With orgs, this is still valid, and `gh_paulo_castro` now refers to an org.
* Every app belongs to exactly one org (there is no such thing as an org-less app).
* At the moment there is a "default org" named after the username, but we are told we should not rely on this "default org" behaviour.
* The proposal is for the `balena app create` command to have the `--organization` flag, but no other command should have this flag. All other commands should use the app slug in the format 'myorg/myapp'.
* Slugs and orgs handles are always lowercase. App names are mixed case. An app named `myApp` (mixed case) would have a slug like `myorg/myapp` (all lowercase). This is why we decided against the proposal of allowing `balena app create myorg/myApp`, as the `myorg/myApp` part would look like a slug with mixed case, which isn't allowed.

For `balena app create`, if the `--organization` flag is not specified:
* Load orgs user is associated with.
* If more than one org, throw an error.
* If only one org, then use that handle.

cc: @srlowe, @mbalamat. By the way, I have also created this bug issue related to `app create`:
#2120 [app create: Blank space somehow inserted in app name](https://github.com/balena-io/balena-cli/issues/2120)

phil-d-wilson:
@pdcastro @srlowe thanks for progressing this guys! The life of a HHR is tougher right now, since we have LOTS of apps in the balenaLabs org - so commands like `balena devices` are currently un-usable.

pdcastro:
@phil-d-wilson, I thought that `balena devices` had already been updated in the CLI v12.35.0 or later. Could you elaborate on the issue / error you are facing? E.g. sample output or steps to reproduce. Thanks! cc: @srlowe

pdcastro:
[...]

Issue #1808, opened 2020-05-07
balena device rm --yes should not need a TTY


jeff-at-dwelo:
Given a file with one uuid per line, the following should work:
```
$ cat easy_delete.txt | xargs -I '{}' -L1 balena device rm '{}' --yes
Prompts can not be meaningfully rendered in non-TTY environments

Error: Prompts can not be meaningfully rendered in non-TTY environments
```

---
- **balena CLI version:** 11.32.15
- **Operating system version:** Debian 10 buster
- **32/64 bit OS and processor:** 64
- **Install method:** npm
- **If npm install, Node.js and npm version:** Node v12.16.3, npm 6.14.4

---

pdcastro:
Connects-to: #1089 [Implement scriptable (non-interactive) mode](https://github.com/balena-io/balena-cli/issues/1089)

srlowe:
I wasn't able to reproduce this, it could have been fixed in recent change.

pdcastro:
@srlowe, are you able to reproduce it by checking out CLI release/tag v11.32.15? (Or using a standalone zip file from the releases page.) If it can be reproduced with v11.32.15 but not with the latest release, then you could say it is "proof" that it was fixed. :-)

PR #2127, opened 2020-12-09
Add organizations support to app create command

srlowe:
Change-type: minor
Connects-to: #2119

### DIFF
--- a/lib/commands/app/create.ts
+++ b/lib/commands/app/create.ts
@@ -20,9 +20,10 @@ import Command from '../../command';
 import { ExpectedError } from '../../errors';
 import * as cf from '../../utils/common-flags';
 import { getBalenaSdk, stripIndent } from '../../utils/lazy';
-import type * as BalenaSDK from 'balena-sdk';
+import type { Application } from 'balena-sdk';
 
 interface FlagsDef {
+	organization?: string;
 	type?: string; // application device type
 	help: void;
 }
@@ -37,16 +38,26 @@ export default class AppCreateCmd extends Command {
 
 		Create a new balena application.
 
-		You can specify the application device type with the \`--type\` option.
-		Otherwise, an interactive dropdown will be shown for you to select from.
+		You can specify the organization the application should belong to using
+		the \`--organization\` option. The organization's handle, not its name,
+		should be provided. Organization handles can be listed with the
+		\`balena orgs\` command.
 
-		You can see a list of supported device types with:
+		The application's default device type is specified with the \`--type\` option.
+		The \`balena devices supported\` command can be used to list the available
+		device types.
 
-		$ balena devices supported
-`;
+		Interactive dropdowns will be shown for selection if no device type or
+		organization is specified and there are multiple options to choose from.
+		If there is a single option to choose from, it will be chosen automatically.
+		This interactive behavior can be disabled by explicitly specifying a device
+		type and organization.
+	`;
+
 	public static examples = [
 		'$ balena app create MyApp',
-		'$ balena app create MyApp --type raspberry-pi',
+		'$ balena app create MyApp --organization mmyorg',
+		'$ balena app create MyApp -o myorg --type raspberry-pi',
 	];
 
 	public static args = [
@@ -60,6 +71,11 @@ export default class AppCreateCmd extends Command {
 	public static usage = 'app create <name>';
 
 	public static flags: flags.Input<FlagsDef> = {
+		organization: flags.string({
+			char: 'o',
+			description:
+				'handle of the organization the application should belong to',
+		}),
 		type: flags.string({
 			char: 't',
 			description:
@@ -75,30 +91,62 @@ export default class AppCreateCmd extends Command {
 			AppCreateCmd,
 		);
 
-		const balena = getBalenaSdk();
-
-		// Create application
+		// Ascertain device type
 		const deviceType =
 			options.type ||
 			(await (await import('../../utils/patterns')).selectDeviceType());
-		let application: BalenaSDK.Application;
+
+		// Ascertain organization
+		const organization =
+			options.organization?.toLowerCase() || (await this.getOrganization());
+
+		// Create application
+		let application: Application;
 		try {
-			application = await balena.models.application.create({
+			application = await getBalenaSdk().models.application.create({
 				name: params.name,
 				deviceType,
-				organization: (await balena.auth.whoami())!,
+				organization,
 			});
 		} catch (err) {
-			// BalenaRequestError: Request error: Unique key constraint violated
 			if ((err.message || '').toLowerCase().includes('unique')) {
+				// BalenaRequestError: Request error: "organization" and "app_name" must be unique.
 				throw new ExpectedError(
-					`Error: application "${params.name}" already exists`,
+					`Error: application "${params.name}" already exists in organization "${organization}".`,
+				);
+			} else if ((err.message || '').toLowerCase().includes('unauthorized')) {
+				// BalenaRequestError: Request error: Unauthorized
+				throw new ExpectedError(
+					`Error: You are not authorized to create applications in organization "${organization}".`,
 				);
 			}
+
 			throw err;
 		}
-		console.info(
-			`Application created: ${application.slug} (${deviceType}, id ${application.id})`,
+
+		// Output
+		const { isV13 } = await import('../../utils/version');
+		console.log(
+			isV13()
+				? `Application created: slug "${application.slug}", device type "${deviceType}"`
+				: `Application created: ${application.slug} (${deviceType}, id ${application.id})`,
 		);
+	}
+
+	async getOrganization() {
+		const { getOwnOrganizations } = await import('../../utils/sdk');
+		const organizations = await getOwnOrganizations(getBalenaSdk());
+
+		if (organizations.length === 0) {
+			// User is not a member of any organizations (should not happen).
+			throw new Error('This account is not a member of any organizations');
+		} else if (organizations.length === 1) {
+			// User is a member of only one organization - use this.
+			return organizations[0].handle;
+		} else {
+			// User is a member of multiple organizations -
+			const { selectOrganization } = await import('../../utils/patterns');
+			return selectOrganization(organizations);
+		}
 	}
 }
--- a/lib/errors.ts
+++ b/lib/errors.ts
@@ -189,6 +189,7 @@ const EXPECTED_ERROR_REGEXES = [
 	/^BalenaInvalidDeviceType/, // balena-sdk
 	/Cannot deactivate devices/i, // balena-api
 	/Devices must be offline in order to be deactivated\.$/i, // balena-api
+	/^BalenaOrganizationNotFound/, // balena-sdk
 	/Request error: Unauthorized$/, // balena-sdk
 	/^Missing \d+ required arg/, // oclif parser: RequiredArgsError
 	/Missing required flag/, // oclif parser: RequiredFlagError
--- a/lib/utils/patterns.ts
+++ b/lib/utils/patterns.ts
@@ -26,6 +26,7 @@ import { getBalenaSdk, getVisuals, stripIndent, getCliForm } from './lazy';
 import validation = require('./validation');
 import { delay } from './helpers';
 import { isV13 } from './version';
+import { Organization } from 'balena-sdk';
 
 export function authenticate(options: {}): Promise<void> {
 	const balena = getBalenaSdk();
@@ -210,54 +211,18 @@ export function selectApplication(
 		});
 }
 
-export function selectOrCreateApplication() {
-	const balena = getBalenaSdk();
-	return balena.models.application
-		.hasAny()
-		.then((hasAnyApplications) => {
-			if (!hasAnyApplications) {
-				// Just to make TS happy
-				return Promise.resolve(undefined);
-			}
-
-			return (balena.models.application.getAll({
-				$expand: {
-					is_for__device_type: {
-						$select: 'slug',
-					},
-				},
-			}) as Promise<ApplicationWithDeviceType[]>).then((applications) => {
-				const appOptions: Array<{ name: string; value: string | null }> = _.map(
-					applications,
-					(application) => ({
-						name: `${application.app_name} (${application.is_for__device_type[0].slug})`,
-						value: application.app_name,
-					}),
-				);
-
-				appOptions.unshift({
-					name: 'Create a new application',
-					value: null,
-				});
-
-				return getCliForm().ask({
-					message: 'Select an application',
-					type: 'list',
-					choices: appOptions,
-				});
-			});
-		})
-		.then((application) => {
-			if (application) {
-				return application;
-			}
-
-			return getCliForm().ask({
-				message: 'Choose a Name for your new application',
-				type: 'input',
-				validate: validation.validateApplicationName,
-			});
-		});
+export async function selectOrganization(organizations?: Organization[]) {
+	// Use either provided orgs (if e.g. already loaded) or load from cloud
+	organizations =
+		organizations || (await getBalenaSdk().models.organization.getAll());
+	return getCliForm().ask({
+		message: 'Select an organization',
+		type: 'list',
+		choices: organizations.map((org) => ({
+			name: `${org.name} (${org.handle})`,
+			value: org.handle,
+		})),
+	});
 }
 
 export async function awaitDevice(uuid: string) {
--- a/lib/utils/sdk.ts
+++ b/lib/utils/sdk.ts
@@ -32,10 +32,6 @@ export async function getApplication(
 	nameOrSlugOrId: string | number,
 	options?: PineOptions<Application>,
 ): Promise<Application> {
-	// TODO: Consider whether it would be useful to generally include interactive selection of application here,
-	//       when nameOrSlugOrId not provided.
-	//       e.g. nameOrSlugOrId || (await (await import('../../utils/patterns')).selectApplication()),
-	//       See commands/device/init.ts ~ln100 for example
 	const { looksLikeInteger } = await import('./validation');
 	if (looksLikeInteger(nameOrSlugOrId as string)) {
 		try {
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -345,17 +345,26 @@ application name or org/name slug
 
 Create a new balena application.
 
-You can specify the application device type with the `--type` option.
-Otherwise, an interactive dropdown will be shown for you to select from.
+You can specify the organization the application should belong to using
+the `--organization` option. The organization's handle, not its name,
+should be provided. Organization handles can be listed with the
+`balena orgs` command.
 
-You can see a list of supported device types with:
+The application's default device type is specified with the `--type` option.
+The `balena devices supported` command can be used to list the available
+device types.
 
-$ balena devices supported
+Interactive dropdowns will be shown for selection if no device type or
+organization is specified and there are multiple options to choose from.
+If there is a single option to choose from, it will be chosen automatically.
+This interactive behavior can be disabled by explicitly specifying a device
+type and organization.
 
 Examples:
 
 	$ balena app create MyApp
-	$ balena app create MyApp --type raspberry-pi
+	$ balena app create MyApp --organization mmyorg
+	$ balena app create MyApp -o myorg --type raspberry-pi
 
 ### Arguments
 
@@ -364,6 +373,10 @@ Examples:
 application name
 
 ### Options
+
+#### -o, --organization ORGANIZATION
+
+handle of the organization the application should belong to
 
 #### -t, --type TYPE
 


lib/utils/patterns.ts
```diff
@@ -204,54 +205,18 @@ export function selectApplication(
 		});
 }
 
-export function selectOrCreateApplication() {
```
srlowe, line 207: This was unused.

lib/utils/sdk.ts
```diff
@@ -27,10 +32,6 @@ export async function getApplication(
 	nameOrSlugOrId: string | number,
 	options?: PineOptions<Application>,
 ): Promise<Application> {
-	// TODO: Consider whether it would be useful to generally include interactive selection of application here,
-	//       when nameOrSlugOrId not provided.
-	//       e.g. nameOrSlugOrId || (await (await import('../../utils/patterns')).selectApplication()),
-	//       See commands/device/init.ts ~ln100 for example
```
srlowe, line 33: I removed this comment (It may have been me that added it), as I think this is not the right layer for UI converns.  This module is just for working with the SDK, and the methods here might be good candidates for incorporation into the SDK.

lib/commands/app/create.ts
```diff
@@ -75,30 +86,51 @@ export default class AppCreateCmd extends Command {
...
+		if (organizations.length === 0) {
+			// User is not a member of any organizations (should not happen).
+			throw new Error('This account is not a member of any organizations');
+		} else if (organizations.length === 1) {
+			// User is a member of only one organization - use this.
+			return organizations[0].handle;
+		} else {
+			// User is a member of multiple organizations -
+			const { selectOrganization } = await import('../../utils/patterns');
+			return selectOrganization(organizations);
```
pdcastro, line 133: This uses `getCliForm` from `'resin-cli-form'`, which I believe does not do auto TTY detection - and even if it does, there is still issue #1378. And these other issues:

* #1001
* #1043 
* #1089
* #1112
* #1378 
* #1808
* ...

And even attached JF support threads:
* https://jel.ly.fish/1418fb56-c0b7-4861-8ab1-75584e15f367
* https://jel.ly.fish/a675f6e4-6dcf-4de9-a566-a6e52da23104
* ...

I know there is the _"framework level"_ argument and the _"there is already this other prompt"_ argument and the _"it's beyond this PR"_ argument, but these arguments simply result in the problem being compounded by adding more prompts. I suggest a different approach: add a boolean flag like `--noninteractive` or `--notty` to `app create` that results in very simple behavior:

* If `--notty` is given, throw an error instead of raising an interactive prompt, e.g.:
```
throw new ExpectedError("This account has multiple organizations. Please specify an organization with the `-o` option")
```
* Otherwise, raise an interactive prompt.

lib/utils/sdk.ts
```diff
@@ -47,3 +48,27 @@ export async function getApplication(
...
+
+/**
+ * Wraps the sdk organization.getAll method,
+ * restricting to those orgs user is a member of
+ */
+export async function getOwnOrganizations(
+	sdk: BalenaSDK,
+): Promise<Organization[]> {
+	return await sdk.models.organization.getAll({
+		$filter: {
```
pdcastro, line 60: We may or may not want some additional filtering here, because I got the following `BalenaRequestError` while testing the PR:
```
$ ./bin/balena whoami
== ACCOUNT INFORMATION
USERNAME: gh_paulo_castro
EMAIL:    pefcastro@gmail.com
URL:      balena-cloud.com

$ ./bin/balena app create test33 --type raspberrypi3
? Select an organization (Use arrow keys)
❯ gh_paulo_castro’s Organization
  g_paulo_castro1’s Organization
  Page’s Organization
  unicorn’s Organization

? Select an organization g_paulo_castro1’s Organization
BalenaRequestError: Request error: Unauthorized
```
I think what happened is that user `gh_paulo_castro` (the user logged in to the CLI) had been added to another user's org, `g_paulo_castro1’s Organization` that belongs to `g_paulo_castro1`, with the "member" role that the web UI describes as "ready-only access to the org", thus lacking authorization to create apps in that org.

I think that the nicest UX for `balena app create` would be to filter out orgs where the app cannot be created. The second best (also acceptable) would be to issue an error, which actually means doing nothing because the code already does that, and apparently treating the error as 'expected', which is good.

lib/commands/app/create.ts
```diff
@@ -37,16 +38,22 @@ export default class AppCreateCmd extends Command {
...
 
 		$ balena devices supported
-`;
+	`;
+
 	public static examples = [
 		'$ balena app create MyApp',
-		'$ balena app create MyApp --type raspberry-pi',
+		'$ balena app create MyApp --organization MyOrg',
+		'$ balena app create MyApp -o MyOrg --type raspberry-pi',
```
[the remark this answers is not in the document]
pdcastro: Possibly related: https://github.com/balena-io/balena-ui/issues/4349

If that issue means that no 2 orgs (even from different users) can have the same fancy name, then it might be OK to identify orgs by fancy name only.

What is the acceptable character set of org fancy names? We should probably test with non ASCII characters, say an org named `οργάνωση`.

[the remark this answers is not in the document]
pdcastro: If we were to accept org handles in addition to fancy names (which might be a nice thing for users' scripts), we'd have to have a command to list the org handles. Say, `balena orgs`.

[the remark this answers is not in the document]
srlowe: @pdcastro Heres
 - Org handles are unique, but mixed case.  I'll ask about the uniqueness constraints.
 - Yeah, I have org access on prod support account, had to add to staging (didn't have a team_member account there)
 - Will ask about value char sets
 - Org handles === fancy names as far as I understand.

lib/utils/sdk.ts
srlowe: @thgreasi Could you add your opinion here please?

srlowe: @pdcastro Right, I didn't know about that.   I suppose one point of view is that it's better to keep all associated orgs on the list, and then throw a permissions error if they try to use a read only one (maybe a more friendly error).  At least that way they will know why it doesn't work, rather than an org juts missing from the list, which might cause more support queries.

srlowe: @pdcastro If you have time, could you try adding this console log, and posting the data for `g_paulo_castro1’s Organization`
```
	async getOrganization() {
		const { getOwnOrganizations } = await import('../../utils/sdk');
		const organizations = await getOwnOrganizations(getBalenaSdk());

		console.log(organizations);
```

srlowe: @pdcastro Please take a look at the revised version I just pushed, which handles that error.  I think I prefer this solution as it keeps everything transparent.

pdcastro: > could you try adding this console log, and posting the data for `g_paulo_castro1’s Organization`?

@srlowe 
```
  {
    id: 30460,
    created_at: '2018-10-18T20:38:20.946Z',
    name: 'g_paulo_castro1’s Organization',
    handle: 'g_paulo_castro1',
    company_name: 'Balena',
    website: null,
    industry: null,
    billing_account_code: null,
    __metadata: { uri: '/resin/organization(@id)?@id=30460' }
  }
```

lib/commands/app/create.ts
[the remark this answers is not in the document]
pdcastro: @srlowe @thgreasi, some potentially troubling finding re org name charset, conversion of fancy org names to handles, and this PR's design decision of identifying orgs by fancy name rather than handle:

I used the web dashboard, logged in as `gh_paulo_castro`, to create Orgs with the following fancy names:
* `ascii-org`
* `οργάνωση `

Then I used the CLI (this PR) to create apps in those orgs. Results:
```
$ ./bin/balena-dev app create test35 --type raspberrypi3 -o ascii-org
BalenaOrganizationNotFound: Organization not found: ascii-org

$ ./bin/balena-dev app create test34 --type raspberrypi3 -o οργάνωση
BalenaOrganizationNotFound: Organization not found: οργάνωση
```

Adding some log output (as suggested by Scott, thanks!), check the org handles that the web dashboard (or SDK or API?) created for those orgs:

```
[
  {
    id: 172338,
    created_at: '2020-12-10T14:00:35.505Z',
    name: 'ascii-org',
    handle: 'ascii_org',
    company_name: null,
    website: null,
    industry: null,
    billing_account_code: null,
    __metadata: { uri: '/resin/organization(@id)?@id=172338' }
  },
  {
    id: 172337,
    created_at: '2020-12-10T13:57:58.151Z',
    name: 'οργάνωση',
    handle: '_',
    company_name: null,
    website: null,
    industry: null,
    billing_account_code: null,
    __metadata: { uri: '/resin/organization(@id)?@id=172337' }
  },
```

Specially the `οργάνωση` org, which resulted in a handle of `_` ...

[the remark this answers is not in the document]
srlowe: Right, looks like we need to clarify this before we continue.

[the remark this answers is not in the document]
pdcastro: Further discussion: https://www.flowdock.com/app/rulemotion/p-organizations/threads/c1PZ2OS_rNmIRI3UJqKbzv306k-

lib/commands/app/create.ts
```diff
@@ -75,30 +86,57 @@ export default class AppCreateCmd extends Command {
...
+
+		// Create application
+		let application: Application;
 		try {
-			application = await balena.models.application.create({
+			application = await getBalenaSdk().models.application.create({
 				name: params.name,
 				deviceType,
-				organization: (await balena.auth.whoami())!,
+				organization,
```
pdcastro, line 103: Following what Thodoris mentiones in Flowdock:
> https://www.flowdock.com/app/rulemotion/p-organizations/threads/c1PZ2OS_rNmIRI3UJqKbzv306k-
> I think the issue there is that the SDK is only checking the handle

And checking the SDK documentation for `application.create`:
> https://www.balena.io/docs/reference/sdk/node-sdk/#balena.models.application.create
> options.organization - handle (string) or id (number) of the organization that the application will belong to or null

Now I understand that this property is necessarily a handle, not a fancy name. In this case, given:

```
const organization = options.organization || (await this.getOrganization());
```

It follows that `options.organization` should be a handle, not a name. So we to document this, and provide a way of users querying the handles `balena orgs`, and/or add code to fetch orgs by name rather than handle, then extract the handle...

srlowe: I'd favour sticking to handles for everything (rather than names).  It seems like the orgs team are ok with exposing handles to the user.

srlowe: > It follows that options.organization should be a handle, not a name. So we would have to document this, and provide a way of users querying the handles, e.g. balena orgs,

:+1:

srlowe:
I've clarified that --organization expects the handle in latest.  Also the orgs selector now lists the orgs in format `name (handle)`.

srlowe:
I've clarified that --organization expects the handle in latest. Also the orgs selector now lists the orgs in format name (handle).

lib/commands/app/create.ts
```diff
@@ -60,6 +68,11 @@ export default class AppCreateCmd extends Command {
 	public static usage = 'app create <name>';
 
 	public static flags: flags.Input<FlagsDef> = {
+		organization: flags.string({
+			char: 'o',
+			description:
+				'the handle of organization the application should belong to',
```
pdcastro, line 74: `of organization` → `of the organization`

lib/commands/app/create.ts
```diff
@@ -37,16 +38,23 @@ export default class AppCreateCmd extends Command {
 
 		Create a new balena application.
 
-		You can specify the application device type with the \`--type\` option.
-		Otherwise, an interactive dropdown will be shown for you to select from.
+		You can specify the organization the application should belong to using
+		the \`--organization\` option with the organization handle, and the
+		application device type with the \`--type\` option.
```
pdcastro, line 43: Suggestion:
```
You can specify the organization the application should belong to using
the \`--organization\` option. The organization's handle, not its name,
should be provided. Organization handles can be listed with the
\`balena orgs\` command.

The application's default device type is specified with the \`--type\` option.
The \`balena devices supported\` command can be used to list the available
device types.

If no device type or organization is specified, interactive dropdowns
will be shown for selection (if there are multiple options to choose from).
```
This suggestion amends the following paragraphs as well.

lib/commands/app/create.ts
```diff
@@ -75,30 +88,57 @@ export default class AppCreateCmd extends Command {
...
 				name: params.name,
 				deviceType,
-				organization: (await balena.auth.whoami())!,
+				organization,
 			});
 		} catch (err) {
-			// BalenaRequestError: Request error: Unique key constraint violated
 			if ((err.message || '').toLowerCase().includes('unique')) {
+				// BalenaRequestError: Request error: "organization" and "app_name" must be unique.
+				const slug = `${organization.toLowerCase()}/${params.name.toLowerCase()}`;
```
pdcastro, line 110: NIT:
```
const slug = `${organization}/${params.name}`.toLowerCase();
```

lib/commands/app/create.ts
```diff
@@ -33,20 +34,23 @@ interface ArgsDef {
 
 export default class AppCreateCmd extends Command {
 	public static description = stripIndent`
-		Create an application.
+		You can specify the organization the application should belong to using
```
pdcastro, line 37: First line got accidentally deleted I think... (check how it looks in `cli.markdown`)

lib/commands/app/create.ts
srlowe: (I also removed the initial unneeded 'the', matching other flag)

lib/commands/app/create.ts
```diff
@@ -37,16 +38,23 @@ export default class AppCreateCmd extends Command {
 
 		Create a new balena application.
 
-		You can specify the application device type with the \`--type\` option.
-		Otherwise, an interactive dropdown will be shown for you to select from.
+		You can specify the organization the application should belong to using
+		the \`--organization\` option. The organization's handle, not its name,
+		should be provided. Organization handles can be listed with the
+		\`balena orgs\` command.
```
pdcastro, line 44: Now the only thing holding me back from approving the PR is clarification and testing of whether it works with openbalena.

srlowe: It seems like it should work.  Since there is only one org in open balena, admin (if I understand correctly).  It should automatically select that org.  I think we just need a node in the help about --organization being not applicable to open balena.  Is there an openBalena instance set up that I can test against?

pdcastro: > Is there an openBalena instance set up that I can test against?

I am not aware of one. I followed the openBalena [getting started guide](https://www.balena.io/open/docs/getting-started/) to setup openBalena on a VirtualBox VM, using `openbalena.local` as the domain name, such that:

* server: VirtualBox VM (on my macOS laptop), domain configured as `openbalena.local`
* "local machine" where I run the CLI: my macOS laptop, outside the VirtualBox VM, with domain name entries added to the `/etc/local` file.

Adding the following entries to `/etc/hosts` file of my macOS laptop (host OS) gets around the need for CNAME records:

`/etc/hosts`
```
192.168.1.8     api.openbalena.local
192.168.1.8     builder.openbalena.local
192.168.1.8     vpn.openbalena.local
192.168.1.8     s3.openbalena.local
192.168.1.8     registry.openbalena.local
```

Also to make life easier, I've created a `~/openbalena` folder on the "local machine" (macOS host OS, where I run the CLI), and added the `ca.crt` file there and also the following `balenarc.yml` CLI config file there:

`~/openbalena/balenarc.yml`
```
dataDirectory: '/Users/paulo/.openbalena'
balenaUrl: 'openbalena.local'
```

As a result, on the macOS host OS (where I run the CLI), if I `cd openbalena`, the CLI automatically "switches" to contacting the openBalena server, including switching the login token. (The login token for balenaCloud is stored in `~/.balena` by default, and because of my special configuration above, for openBalena it is stored in `~/.openbalena`.)

pdcastro: > I think we just need a node in the help about --organization being not applicable to open balena.

Yep.

lib/commands/app/create.ts
```diff
@@ -75,30 +88,57 @@ export default class AppCreateCmd extends Command {
...
+
+		// Create application
+		let application: Application;
 		try {
-			application = await balena.models.application.create({
+			application = await getBalenaSdk().models.application.create({
 				name: params.name,
 				deviceType,
-				organization: (await balena.auth.whoami())!,
+				organization,
```
dfunckt, line 105:
```suggestion
				organization.toLowerCase(),
```

lib/utils/sdk.ts
```diff
@@ -47,3 +48,27 @@ export async function getApplication(
...
+			organization_membership: {
+				$any: {
+					$alias: 'orm',
+					$expr: {
+						orm: {
+							user: await sdk.auth.getUserId(),
```
thgreasi, line 66: It might also makes sense to also do an in memory filter based on whether you are an org administrator or not, since just being an org member doesn't give you the rights to create apps.
I'm saying in memory, since `orgnaization_membership_role` is not in oB.

dfunckt: No, let's not do that. The CLI must not make these assumptions -- what if we want to also enable a new role that allows app creation in the future? Let's not mix things up -- if the user's role doesn't permit app creation, the request will fail.

srlowe: Right yeah.  We had a discussion about this, and came to the conclusion that it's better to list all the organizations that you're associated with, and then fail with a "You don't have the rights to add an application to this organization" type message, rather than hide those organizations here, since this gives full transparency about what the problem is.  Otherwise users might be confused about why some orgs are not showing.

thgreasi: Makes sense. On the other hand people that so far were creating apps normally, might no more be able to do so if they are also members of other orgs and they will now have to select one of the orgs that they are a member of.
So this was just to make this
* non workflow-breaking (in case the CLI is used in CIs)
* also offer a slightly better experience for cloud users (so that you don't get orgs that you are just a member or).

I was thinking something like `organization_membership_role[0].name !== 'member'` or something.

lib/commands/app/create.ts
```diff
@@ -75,30 +88,58 @@ export default class AppCreateCmd extends Command {
...
-			// BalenaRequestError: Request error: Unique key constraint violated
 			if ((err.message || '').toLowerCase().includes('unique')) {
+				// BalenaRequestError: Request error: "organization" and "app_name" must be unique.
+				const slug = `${organization}/${params.name}`.toLowerCase();
+				throw new ExpectedError(`Error: application "${slug}" already exists`);
+			} else if ((err.message || '').toLowerCase().includes('unauthorized')) {
+				// BalenaRequestError: Request error: Unauthorized
 				throw new ExpectedError(
-					`Error: application "${params.name}" already exists`,
+					`Error: You are not authorized to add applications to organization "${organization}".`,
```
dfunckt, line 116: Maybe worth staying consistent in terms? I don't know if what I suggest is proper English though :)

```suggestion
					`Error: You are not authorized to create applications in organization "${organization}".`,
```

lib/utils/sdk.ts
thgreasi: @srlowe if you do add a filter on organization_member_role it would stop working on oB since we don't have that there.
More info in my other comment (I didn't see this before)
See: https://github.com/balena-io/balena-cli/pull/2127#pullrequestreview-551444748

lib/commands/app/create.ts
dfunckt: I'd rather we don't single out ob. The best way to clarify that would be to say "if no organization is provided and you are a member of a single org, this command assumes you want to create the app in that org. Otherwise the argument must be specified".

lib/commands/app/create.ts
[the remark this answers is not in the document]
srlowe: I think discussion is resolved now?

lib/commands/app/create.ts
srlowe: I've added a message along those lines now.

lib/utils/sdk.ts
pdcastro: > non workflow-breaking (in case the CLI is used in CIs)

Sounds like: Before this PR, a user script around `balena app create` would succeed and create the app in the default org. After this PR, the same script would break because an interactive prompt would be raised for selection between the default org (the only useful selection) and other orgs that would necessarily fail when selected. **In which case, it could be said that this PR, as it is, is both a breaking change and one that adds noise to the workflow.**

We could however counter-argue with:
* "Legit users" (arbitrarily defined as users who are not balena staff testing our system and/or the CLI) are unlikely to script `balena app create`, as they don't create that many apps to warrant scripting. And even for the few legit users who might be doing this, this PR (as it is) would only affect them if their user account was also a member of other orgs. It's unlikely that many legit users would be affected.
* What was referred to above as "adding noise to the workflow" can also be described as adding visibility / to the user's account membership in the platform, which is generally a good thing.
* Users who wish to avoid the interactive prompt can, and should, use the `--organization` flag.

I am not siding with either option (whether or not to filter the orgs), just expanding on what they mean.

lib/commands/app/create.ts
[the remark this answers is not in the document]
pdcastro: Yes, the resolution being that the `--organization` should specify an org handle, rather than its name.

lib/commands/app/create.ts
```diff
@@ -37,16 +38,25 @@ export default class AppCreateCmd extends Command {
...
 
-		You can see a list of supported device types with:
+		The application's default device type is specified with the \`--type\` option.
+		The \`balena devices supported\` command can be used to list the available
+		device types.
+
+		If no device type or organization is specified, interactive dropdowns
+		will be shown for selection (if there are multiple options to choose from).
+		If you are associated with only one organization, then that will be used by default,
+		and there is no need to specify it explicitly.
```
pdcastro, line 53: Would the following suggestion "read a bit easier"?
```
Interactive dropdowns will be shown for selection if no device type or
organization is specified and there are multiple options to choose from.
If there is a single option to choose from, it will be chosen automatically.
This interactive behavior can be disabled by explicitly specifying a device
type and organization.
```

This suggestion purposely starts the paragraph with _"Interactive dropdowns will be shown for selection"_ and includes an explicit hint of how to disable it, so that it sounds alarming and useful to anyone scripting the CLI and skimming through the help output.

Otherwise, nit: One of the original lines is longer than 80 characters.

lib/errors.ts
```diff
@@ -159,6 +159,7 @@ const EXPECTED_ERROR_REGEXES = [
 	/^BalenaInvalidDeviceType/, // balena-sdk
 	/Cannot deactivate devices/i, // balena-api
 	/Devices must be offline in order to be deactivated\.$/i, // balena-api
+	/^BalenaOrganizationNotFound/, // balena-sdk
```
pdcastro, line 162: 👍

lib/utils/sdk.ts
srlowe: > Sounds like: Before this PR, a user script around balena app create would succeed and create the app in the default org.

Not always.  The CLI was making a bad assumption before, by passing the username as the org handle, which does not work in all cases for the default org.

lib/commands/app/create.ts
srlowe: (This text has been updated again re Paulo's suggestion)

pdcastro (approved):

PR #2128, opened 2020-12-09
Add organizations support to app rename command

srlowe:
Change-type: minor
Connects-to: #2119

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -165,7 +165,7 @@ Users are encouraged to regularly update the balena CLI to the latest version.
 	- [app &#60;nameorslug&#62;](#app-nameorslug)
 	- [app create &#60;name&#62;](#app-create-name)
 	- [app purge &#60;name&#62;](#app-purge-name)
-	- [app rename &#60;name&#62; [newname]](#app-rename-name-newname)
+	- [app rename &#60;nameorslug&#62; [newname]](#app-rename-nameorslug-newname)
 	- [app restart &#60;name&#62;](#app-restart-name)
 	- [app rm &#60;name&#62;](#app-rm-name)
 
@@ -386,7 +386,7 @@ application name or numeric ID
 
 ### Options
 
-## app rename &#60;name&#62; [newName]
+## app rename &#60;nameOrSlug&#62; [newName]
 
 Rename an application.
 
@@ -397,12 +397,13 @@ Examples:
 
 	$ balena app rename OldName
 	$ balena app rename OldName NewName
+	$ balena app rename myorg/oldname NewName
 
 ### Arguments
 
-#### NAME
+#### NAMEORSLUG
 
-application name or numeric ID
+application name or org/name slug
 
 #### NEWNAME
 
--- a/lib/commands/app/rename.ts
+++ b/lib/commands/app/rename.ts
@@ -20,14 +20,15 @@ import type { IArg } from '@oclif/parser/lib/args';
 import Command from '../../command';
 import * as cf from '../../utils/common-flags';
 import { getBalenaSdk, stripIndent, getCliForm } from '../../utils/lazy';
-import type { Application, ApplicationType, BalenaSDK } from 'balena-sdk';
+import { lowercaseIfSlug } from '../../utils/normalization';
+import type { ApplicationType } from 'balena-sdk';
 
 interface FlagsDef {
 	help: void;
 }
 
 interface ArgsDef {
-	name: string;
+	nameOrSlug: string;
 	newName?: string;
 }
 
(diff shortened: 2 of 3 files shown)


doc/cli.markdown
```diff
@@ -396,12 +396,13 @@ Examples:
...
 	$ balena app rename OldName
 	$ balena app rename OldName NewName
+	$ balena app rename myorg/oldname NewName
 
 ### Arguments
 
-#### NAME
+#### NAMEORSLUG
 
-application name or numeric ID
```
srlowe, line 404: (still supported, but not mentioned)

lib/commands/app/rename.ts
```diff
@@ -39,16 +39,19 @@ export default class AppRenameCmd extends Command {
...
+		'$ balena app rename myorg/oldname NewName',
 	];
 
 	public static args: Array<IArg<any>> = [
 		{
-			name: 'name',
-			description: 'application name or numeric ID',
+			name: 'nameOrSlug',
+			description: 'application name or org/name slug',
+			parse: lowercaseIfSlug,
```
srlowe, line 54: Converting input to lowercase if slug (on Akis' suggestion), since slugs are lowercase.

lib/commands/app/rename.ts
```diff
@@ -109,28 +96,20 @@ export default class AppRenameCmd extends Command {
...
-	async renameApplication(balena: BalenaSDK, id: number, newName: string) {
-		return balena.pine.patch<Application>({
-			resource: 'application',
-			id,
-			body: {
-				app_name: newName,
-			},
-		});
+		// Output result
+		console.log(`Application ${params.nameOrSlug} renamed to ${newName}`);
```
srlowe, line 113: Command refactored to use new application.rename sdk method.

lib/utils/normalization.ts
```diff
@@ -16,19 +16,8 @@
  */
 
 import type { BalenaSDK } from 'balena-sdk';
-import _ = require('lodash');
 import { ExpectedError } from '../errors';
 
-export function normalizeUuidProp(
```
srlowe, line 22: Not used.

lib/commands/app/rename.ts
```diff
@@ -68,38 +71,22 @@ export default class AppRenameCmd extends Command {
...
+		// Disambiguate target application (if nameOrSlug is a number, it could either be an ID or a numerical name)
+		const { getApplication } = await import('../../utils/sdk');
+		const application = await getApplication(balena, params.nameOrSlug);
 
-		// Check app supports renaming
-		const appType = (app.application_type as ApplicationType[])?.[0];
-		if (appType.is_legacy) {
+		if (!application) {
 			throw new ExpectedError(
-				`Application ${params.name} is of 'legacy' type, and cannot be renamed.`,
```
pdcastro, line 98: Legacy applications are still supported by balena as far as I am aware. I think we need this logic. (I reckon this PR proposes to delete the logic altogether.)

srlowe: If I remember correctly from a conversation with Thodoris, this is handled by the SDK (I'll check).

srlowe: @thgreasi I think you mentioned that the SDK application.rename command handles legacy app checking (though not  sure I remember correctly).  Do you know if this is the case?

srlowe: @pdcastro Ok, I've put this check back.  If it turns out to be redundant we can remove it.

lib/commands/app/rename.ts
```diff
@@ -39,16 +40,19 @@ export default class AppRenameCmd extends Command {
...
 		'$ balena app rename OldName NewName',
+		'$ balena app rename myorg/oldname NewName',
 	];
 
 	public static args: Array<IArg<any>> = [
 		{
-			name: 'name',
-			description: 'application name or numeric ID',
+			name: 'nameOrSlug',
+			description: 'application name or org/name slug',
```
pdcastro, line 54: Approving this PR, but I think we may want to change this code again as part of PR #2131, re documentation of the meaning of "slug" (even the meaning of "org") and a recommendation of slugs over app names or IDs.

To share more data points, given an org named `οργάνωση` that contains an app named `test33`:
```
$ ./bin/balena app rename οργάνωση/test33 new-test-33
BalenaApplicationNotFound: Application not found: οργάνωση/test33
```

Turns out that the org has handle `_`, and the following succeeds:
```
$ ./bin/balena app rename _/test33 new-test-33
Application _/test33 renamed to new-test-33
```

Further discussion: https://www.flowdock.com/app/rulemotion/i-cli/threads/k9NEA-BfE0IbigEuQL85f0z3mZv

pdcastro (approved):

lib/commands/app/rename.ts
thgreasi: I don't see any special check in the SDK, but I'm pretty sure the API blocks it.

PR #2131, opened 2020-12-10
Update various commands to support organizations

srlowe:
Minor changed to the following commands to support organizations (all manually tested):

app purge TESTED
app restart TESTED
app rm TESTED
config generate TESTED
device init TESTED
device move TESTED
device register TESTED 
devices TESTED
env add TESTED
os configure TESTED
tag rm TESTED
tag set TESTED
apps TESTED
envs TESTED
join TESTED
preload TESTED
support TESTED
tags TESTED

Change-type: minor
Connects-to: #2119

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -164,10 +164,10 @@ Users are encouraged to regularly update the balena CLI to the latest version.
 	- [apps](#apps)
 	- [app &#60;nameorslug&#62;](#app-nameorslug)
 	- [app create &#60;name&#62;](#app-create-name)
-	- [app purge &#60;name&#62;](#app-purge-name)
-	- [app rename &#60;nameorslug&#62; [newname]](#app-rename-nameorslug-newname)
-	- [app restart &#60;name&#62;](#app-restart-name)
-	- [app rm &#60;name&#62;](#app-rm-name)
+	- [app purge &#60;application&#62;](#app-purge-application)
+	- [app rename &#60;application&#62; [newname]](#app-rename-application-newname)
+	- [app restart &#60;application&#62;](#app-restart-application)
+	- [app rm &#60;application&#62;](#app-rm-application)
 
 - Authentication
 
@@ -312,7 +312,7 @@ the API key name
 list all your balena applications.
 
 For detailed information on a particular application,
-use `balena app <name> instead`.
+use `balena app <application>` instead.
 
 Examples:
 
@@ -328,6 +328,19 @@ No-op since release v12.0.0
 
 Display detailed information about a single balena application.
 
+Applications may be specified by app name, slug, or numeric ID. App slugs
+are the recommended option, as they are unique and unambiguous. Slugs
+can be listed with the `balena apps` command. Note that slugs may change
+if the application is renamed.
+App names are not unique and may result in "Application is ambiguous" errors
+at any time (even if it "used to work in the past"), for example if the name
+clashes with a newly created public application, or with apps from other balena
+accounts that you may have been invited to as a member. For this reason, app
+names are especially discouraged in scripts (e.g. CI environments).
+Numeric app IDs are deprecated because they consist of an implementation detail
+of the balena backend. We intend to remove support for numeric IDs at some point
+in the future.
+
 Examples:
 
 	$ balena app MyApp
@@ -335,9 +348,9 @@ Examples:
 
 ### Arguments
 
-#### NAMEORSLUG
+#### APPLICATION
 
-application name or org/name slug
+application name, slug (preferred), or numeric ID (deprecated)
 
 ### Options
 
@@ -382,29 +395,56 @@ handle of the organization the application should belong to
 
 application device type (Check available types with `balena devices supported`)
 
-## app purge &#60;name&#62;
+## app purge &#60;application&#62;
 
 Purge data from all devices belonging to an application.
 This will clear the application's /data directory.
 
+Applications may be specified by app name, slug, or numeric ID. App slugs
+are the recommended option, as they are unique and unambiguous. Slugs
+can be listed with the `balena apps` command. Note that slugs may change
+if the application is renamed.
+App names are not unique and may result in "Application is ambiguous" errors
+at any time (even if it "used to work in the past"), for example if the name
+clashes with a newly created public application, or with apps from other balena
+accounts that you may have been invited to as a member. For this reason, app
+names are especially discouraged in scripts (e.g. CI environments).
+Numeric app IDs are deprecated because they consist of an implementation detail
+of the balena backend. We intend to remove support for numeric IDs at some point
+in the future.
+
 Examples:
 
 	$ balena app purge MyApp
+	$ balena app purge myorg/myapp
 
 ### Arguments
 
-#### NAME
+#### APPLICATION
 
-application name or numeric ID
+application name, slug (preferred), or numeric ID (deprecated)
 
 ### Options
 
-## app rename &#60;nameOrSlug&#62; [newName]
+## app rename &#60;application&#62; [newName]
 
 Rename an application.
 
 Note, if the `newName` parameter is omitted, it will be
 prompted for interactively.
+
+Applications may be specified by app name, slug, or numeric ID. App slugs
+are the recommended option, as they are unique and unambiguous. Slugs
+can be listed with the `balena apps` command. Note that slugs may change
+if the application is renamed.
+App names are not unique and may result in "Application is ambiguous" errors
+at any time (even if it "used to work in the past"), for example if the name
+clashes with a newly created public application, or with apps from other balena
+accounts that you may have been invited to as a member. For this reason, app
+names are especially discouraged in scripts (e.g. CI environments).
+Numeric app IDs are deprecated because they consist of an implementation detail
+of the balena backend. We intend to remove support for numeric IDs at some point
+in the future.
 
 Examples:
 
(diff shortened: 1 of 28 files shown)


doc/cli.markdown
```diff
@@ -164,10 +164,10 @@ Users are encouraged to regularly update the balena CLI to the latest version.
 	- [apps](#apps)
 	- [app &#60;nameorslug&#62;](#app-nameorslug)
 	- [app create &#60;name&#62;](#app-create-name)
-	- [app purge &#60;name&#62;](#app-purge-name)
+	- [app purge &#60;nameorslug&#62;](#app-purge-nameorslug)
 	- [app rename &#60;name&#62; [newname]](#app-rename-name-newname)
```
pdcastro, line 168: As part of this PR, I think we should also further update `app rename`, as per comment https://github.com/balena-io/balena-cli/pull/2128#discussion_r540316336

doc/cli.markdown
```diff
@@ -369,20 +369,21 @@ application name
...
 	$ balena app purge MyApp
+	$ balena app purge myorg/myapp
 
 ### Arguments
 
-#### NAME
+#### NAMEORSLUG
 
-application name or numeric ID
+application name or org/name slug
```
pdcastro, line 386: I've raised some matters / questions in flowdock related to documentation of slugs vs IDs:
https://www.flowdock.com/app/rulemotion/i-cli/threads/k9NEA-BfE0IbigEuQL85f0z3mZv

lib/commands/devices/index.ts
```diff
@@ -44,29 +44,25 @@ export default class DevicesCmd extends Command {
...
 		You can filter the devices by application by using the \`--application\` option.
 
-		The --json option is recommended when scripting the output of this command,
-		because field names are less likely to change in JSON format and because it
-		better represents data types like arrays, empty strings and null values.
-		The 'jq' utility may be helpful for querying JSON fields in shell scripts
-		(https://stedolan.github.io/jq/manual/).
+		${applicationIdInfo.split('\n').join('\n\t\t')}
+
+		${jsonInfo.split('\n').join('\n\t\t')}
```
pdcastro, line 49: 👍

lib/commands/devices/index.ts
```diff
@@ -85,6 +81,7 @@ export default class DevicesCmd extends Command {
 		let devices;
 
 		if (options.application != null) {
+			const { tryAsInteger } = await import('../../utils/validation');
 			devices = (await balena.models.device.getAllByApplication(
 				tryAsInteger(options.application),
```
pdcastro, line 86: And here... I'll stop pointing out: It looks like all occurrences of `tryAsInteger` in the CLI codebase, when applied to application names, could / should be replaced with `getApplication` in `sdk.ts` so that app names that look like a number could be used.

lib/commands/preload.ts
```diff
@@ -83,9 +88,11 @@ export default class PreloadCmd extends Command {
 	public static usage = 'preload <image>';
 
 	public static flags: flags.Input<FlagsDef> = {
+		// TODO: Replace with application/a in #v13?
 		app: flags.string({
-			description: 'name of the application to preload',
+			description: 'name of org/name slug of the application to preload',
```
pdcastro, line 93: `name of` → `name or`
`name or slug of the application to preload`
Actually, couldn't this be as follows? -
```
		app: cf.application,
```

tests/commands/device/devices.spec.ts
```diff
@@ -28,7 +28,7 @@ USAGE
   $ balena devices
 
 OPTIONS
-  -a, --application <application>  application name
+  -a, --application <application>  application name, slug (preferred), numeric ID (deprecated)
```
pdcastro, line 31: Tests are failing because some lines exceed a certain number of characters:

https://ci.product-os.io/builds/12766#L5fd4a12f:3501
```
      -  "-a, --application <application>  application name, slug (preferred), numeric"
      -  "ID (deprecated)"
      +  "-a, --application <application>  application name, slug (preferred), numeric ID (deprecated)"
```
I wonder if this `HELP_RESPONSE` test is a bit... overkill. Maybe we could have such a test for one CLI command for the help mechanism in general... Just a thought.

lib/utils/messages.ts
```diff
@@ -39,7 +39,7 @@ export const balenaAsciiArt = `\
 `;
 
 export const registrySecretsHelp = `\
-REGISTRY SECRETS  
+REGISTRY SECRETS
```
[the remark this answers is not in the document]
srlowe: Agh, right.  I wonder if we can add the spaces more explicitly somehow, to stop this happening.

lib/utils/common-args.ts
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { lowercaseIfSlug } from './normalization';
+
+export const applicationRequired = {
+	name: 'application',
+	description: 'application name, slug (preferred), numeric ID (deprecated)',
```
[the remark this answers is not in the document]
srlowe: oops yes, that `or` got removed unintentionally (though I think we should keep the oxford comma), e.g.:
`application name, slug (preferred), or numeric ID (deprecated)`,

lib/utils/common-flags.ts
```diff
@@ -19,14 +19,17 @@ import { flags } from '@oclif/command';
 
 import type { IBooleanFlag } from '@oclif/parser/lib/flags';
 import { stripIndent } from './lazy';
+import { lowercaseIfSlug } from './normalization';
 
 export const application = flags.string({
 	char: 'a',
-	description: 'application name',
+	description: 'application name, slug (preferred), numeric ID (deprecated)',
```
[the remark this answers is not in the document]
srlowe: (same as above)

lib/commands/join.ts
```diff
@@ -72,8 +76,9 @@ export default class JoinCmd extends Command {
 
 	public static flags: flags.Input<FlagsDef> = {
 		application: {
-			description: 'the name of the application the device should join',
 			...cf.application,
+			description:
+				'the name or org/name slug of the application the device should join',
```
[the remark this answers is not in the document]
srlowe: Right. Actually join does accept an ID - this custom message was left because it mentioned join, but perhaps the standard message should be used.   Changing.

lib/commands/preload.ts
srlowe: Yeah, let's standardise

tests/commands/device/devices.spec.ts
srlowe: I was getting this failure seemingly intermittently, it seemed to go away.  Agree, that this is overkill - will remove.

lib/commands/device/move.ts
```diff
@@ -107,14 +113,18 @@ export default class DeviceMoveCmd extends Command {
 		}
 
 		// Get destination application
-		const application =
+		const applicationIdentifier =
 			options.application ||
 			(await this.interactivelySelectApplication(balena, devices));
```
pdcastro, line 118: Probably for another PR... I've noticed that `interactivelySelectApplication()` returns the app name instead of the app slug or the application object -- i.e. the CLI code is doing what this PR advises users not to do: selecting apps by name rather than slug. I think it's possible that users would select the right app via the dropdown menus, only for the CLI to then error with "Application is ambiguous". 🤦  The best would probably be for the application object to be returned, then the caller decides how to use it.

pdcastro (changes requested):

lib/utils/messages.ts
```diff
@@ -94,7 +98,7 @@ option if compatibility is required. This option is mutually exclusive with
 --multi-dockerignore (-m) and will be removed in the CLI's next major version
 release (v13).
 
-Default .dockerignore patterns  
+Default .dockerignore patterns
```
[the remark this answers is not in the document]
srlowe: Agh, I hate meaningful whitespace in languages

lib/commands/device/move.ts
srlowe: I'll look at this in this PR.

lib/commands/device/move.ts
```diff
@@ -152,7 +163,7 @@ export default class DeviceMoveCmd extends Command {
 
 		const patterns = await import('../../utils/patterns');
 		try {
-			const application = await patterns.selectApplication(
+			const applicationId = await patterns.selectApplication(
```
pdcastro, line 166: This is actually the application object, not ID, so the original variable name (`application`) was actually better...

pdcastro (commented):
Nice work 👍 
Seeing as you have chosen to fix the occurrences of `tryAsInteger` for applications in this PR, 😀   I gather that only 3 more occurrences remain in the CLI repo:
* `tryAsInteger(options.application)` in `tags.ts`
* `tryAsInteger(options.application)` in `rm.ts`
* `tryAsInteger(options.application)` in `set.ts`

srlowe:
@pdcastro Yeah perhaps lets leave those 3 other occurrences for another PR, since we are already getting queries in the beginners flow about org usage in the CLI.

lib/commands/device/move.ts
srlowe: Fixing

pdcastro (approved):

lib/commands/device/move.ts
srlowe: Fixed.  Must have mixed it up during refactor.

pdcastro (approved):

PR #2133, opened 2020-12-10
apps: Fix "column.match is not a function" when --verbose is used

pdcastro:
```
$ balena apps --verbose
column.match is not a function

Additional information may be available with the `--debug` flag.

For further help or support, visit:
https://www.balena.io/docs/reference/balena-cli/#support-faq-and-troubleshooting
```

Happened as a result of the `v12` feature switch removal in #2116 -
https://github.com/balena-io/balena-cli/pull/2116/files#diff-706fbb5e5469ab6b5209855b65a3aec467bf9f0bf253b4051737d98604d94570R88

Change-type: patch

### DIFF
--- a/lib/commands/apps.ts
+++ b/lib/commands/apps.ts
@@ -56,7 +56,7 @@ export default class AppsCmd extends Command {
 	public static primary = true;
 
 	public async run() {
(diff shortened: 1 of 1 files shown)


pdcastro:
@balena-ci I self-certify!

PR #2138, opened 2020-12-16
push: Add --release-tag flag

mbalamat:
If the --release-tag flag is set when using the push command then the provided tag is going to be used as a tag/release_version on the produced release.
eg `balena push MyApp --release-tag v1.0`
will result in
Screenshot 2020-12-16 at 4 56 07 PM

### DIFF
--- a/lib/commands/push.ts
+++ b/lib/commands/push.ts
@@ -22,6 +22,7 @@ import { getBalenaSdk, stripIndent } from '../utils/lazy';
 import { dockerignoreHelp, registrySecretsHelp } from '../utils/messages';
 import type { BalenaSDK, Application, Organization } from 'balena-sdk';
 import { ExpectedError, instanceOf } from '../errors';
+import type { RegistrySecrets } from 'resin-multibuild';
 
 enum BuildTarget {
 	Cloud,
@@ -46,6 +47,7 @@ interface FlagsDef {
 	'convert-eol'?: boolean;
 	'noconvert-eol'?: boolean;
 	'multi-dockerignore'?: boolean;
+	'release-tag'?: string[];
 	help: void;
 }
 
@@ -92,6 +94,7 @@ export default class PushCmd extends Command {
 		'$ balena push myApp',
 		'$ balena push myApp --source <source directory>',
 		'$ balena push myApp -s <source directory>',
+		'$ balena push myApp --release-tag key1 "" key2 "value2 with spaces"',
 		'',
 		'$ balena push 10.0.0.1',
 		'$ balena push 10.0.0.1 --source <source directory>',
@@ -224,6 +227,15 @@ export default class PushCmd extends Command {
 			char: 'g',
 			exclusive: ['multi-dockerignore'],
 		}),
+		'release-tag': flags.string({
+			description: stripIndent`
+				Set release tags if the push to a cloud application is successful. Multiple
+				arguments may be provided, alternating tag keys and values (see examples).
+				Hint: Empty values may be specified with "" (bash, cmd.exe) or '""' (PowerShell).
+			`,
+			multiple: true,
+			exclusive: ['detached'],
+		}),
 		help: cf.help,
 	};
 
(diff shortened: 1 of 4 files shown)


lib/commands/push.ts
```diff
@@ -301,9 +310,43 @@ export default class PushCmd extends Command {
 					opts,
 				};
 				await remote.startRemoteBuild(args);
+				// Maybe ^ this should return the id of the release?
+				// instead of getting the latest from below
+				if (options['tag']) {
+					const latestRelease = await sdk.models.release.getLatestByApplication(
```
pdcastro, line 316: Like you hinted in the code comment above, I would also feel more comfortable if `startRemoteBuild` returned the release ID. The worry is that `release .getLatestByApplication` might return the wrong release. If the could builder is busy, it could mean minutes (?) between the time when the builder creates a new release, and the time when the CLI executes this line. Consider also, for example, two end users working on the same application, and running `balena push` more or less at the same time, and the jobs falling on different cloud builders (a fast native ARM builder, and a slow emulated build). This could mean that the build that started first could end _after_ another build submitted shortly after the first.

pdcastro: And I see that you had already considered this in Flowdock too :-)
> https://www.flowdock.com/app/rulemotion/i-cli/threads/XCyAW9_JBlef1rQJdGi21gEpPbX
> The issue that I see with this approach, is that many users try to push in the same App so multiple releases are created so I am afraid that we might end up tagging the wrong release.

mbalamat: We are in agreement then, I'll make the `startRemoteBuild` to return an id.

lib/commands/push.ts
```diff
@@ -224,6 +225,15 @@ export default class PushCmd extends Command {
 			char: 'g',
 			exclusive: ['multi-dockerignore'],
 		}),
+		tag: flags.string({
+			description: stripIndent`
+				When performing a push to an application, use --tag in order set the
+				release version of the produced release with a string of your choice.
+				e.g:
+					--tag v1.0
```
pdcastro, line 233: I think that this example could be added to the examples section of the documentation (`public static examples` array), rather than here. Also, this description mentions "release version", which we have learned is a different feature for another PR. Suggestion:
```
tag: flags.string({
	description: stripIndent`
		Create a new release tag with the given key[:value] pair. The value is optional.
		May be specified multiple multiple times, in which case it must be specified
		at the end of the command line.`
```
See other comment re `multiple: true`. The part regarding _"in which case it must be specified at the end of the command line"_ needs testing / confirmation. (Ideally it would not be needed.)

lib/utils/remote-build.ts
```diff
@@ -142,6 +144,7 @@ export async function startRemoteBuild(build: RemoteBuild): Promise<void> {
 		globalLogger.outputDeferredMessages();
 		await cancellationPromise;
 	}
+	return build.releaseId;
```
pdcastro, line 147: `handleRemoteBuildStream` captures the release ID, but `handleHeadlessBuildStream` currently does not. The latter needs to be updated to capture the release ID. Looks like just a matter of adding a `build` argument, then:
```
build.releaseId = message.releaseId;
```
(Please test it.) :-)

lib/commands/push.ts
```diff
@@ -416,4 +482,17 @@ export default class PushCmd extends Command {
...
 		return selected.extra;
 	}
+
+	checkInvalidOptions(
+		invalidOptions: string[],
+		options: FlagsDef,
+		errorMessage: string,
+	) {
+		invalidOptions.forEach((opt) => {
+			// @ts-ignore : Not sure why typescript wont let me do this?
```
pdcastro, line 492: I believe this `@ts-ignore` line can be deleted, if the following type info is added to the relevant places:
```
				const localOnlyOptions: Array<keyof FlagsDef> = [
					'nolive',
					'service',
					'system',
					'env',
				];

				const remoteOnlyOptions: Array<keyof FlagsDef> = [
					'release-tag-key',
					'release-tag-value',
				];

	checkInvalidOptions(
		invalidOptions: Array<keyof FlagsDef>,
		...
```

lib/commands/push.ts
```diff
@@ -259,17 +289,35 @@ export default class PushCmd extends Command {
 		const buildTarget = await this.getBuildTarget(appOrDevice);
 		switch (buildTarget) {
 			case BuildTarget.Cloud:
+				const _ = await import('lodash');
```
pdcastro, line 292: A fair amount of code is being added to this `case BuildTarget.Cloud` "branch". What about moving both branches of the `switch` statement to separate functions? So it might read something like:
```
		switch (buildTarget) {
			case BuildTarget.Cloud:
				await this.pushToCloud(...);
				break;
			case BuildTarget.Device:
				await this.pushToDevice(...);
				break;
			default:
				...
```
Smaller functions with well defined purposes are usually easier to make sense of, and test (say we would one day want to write a test for it...). And the parent function would be easier to read as well.

lib/commands/push.ts
```diff
@@ -300,10 +348,28 @@ export default class PushCmd extends Command {
...
+				if (releaseId) {
+					// Above we have checked that releaseTagKeys and releaseTagValues are of the same size
+					await Promise.all(
+						(_.zip(releaseTagKeys, releaseTagValues) as Array<
+							[string, string]
+						>).map(async ([key, value]) => {
+							await sdk.models.release.tags.set(releaseId, key, value);
+						}),
+					);
+				}
```
pdcastro, line 361: I think we could add an `else` block here that prints a warning that the tags have not been set because the release ID could not be obtained. For example:
```
	if (releaseId) {
		...
	} else if (releaseTagKeys.length) {
		throw new Error(stripIndent`
			The remote build has started or concluded, but a release ID could not be parsed
			out of the builder's output. As a result, the release tags have not been set.`);
	}
```
Above I've used `Error` instead of `ExpectedError`. Not sure, but if the releaseId could not be parsed, this may actually be an _unexpected_ error worth reporting via Sentry.io.

pdcastro:
The Windows build got stuck: https://ci.balena-dev.com/builds/664625
Running for 2h 38min already. From past experience, it won't unstuck itself. I am clicking the red X button at the top-right corner of that page in order to kill the build. (You can see that page, right? Just checking you have access.)

lib/commands/push.ts
```diff
@@ -224,6 +225,18 @@ export default class PushCmd extends Command {
...
+		'release-tag': flags.string({
+			description: stripIndent`
+				When performing a push to an application, use --release-tag in order set the
+				release tags and their values of your choice.
+
+				e.g:
+					balena push MyApp --release-tag key1 value1 key2 "value2 with spaces" key3 "" key4
+
+				Note that in order for the tags to be applied the push operation should be successful.
+			`,
```
pdcastro, line 237: I think the sample command line fits better in the Examples section of the documentation:
https://github.com/balena-io/balena-cli/blob/v12.35.2/lib/commands/push.ts#L91-L103

`balena push` is an outlier in terms of the length of the documentation of each flag. The documentation of most flags of most other commands is a single line. We can have longer documentation where it is needed, no problem, but we should still aim at being concise. Suggestion:
```
	'release-tag': flags.string({
		description: stripIndent`
			Set release tags if the push to a cloud application is successful. Multiple
			arguments may be provided, alternating tag keys and values (see examples).`,
		multiple: true,
	}),
```

lib/commands/push.ts
```diff
@@ -344,6 +305,143 @@ export default class PushCmd extends Command {
...
+			'is only valid when pushing to a local mode device',
+		);
+
+		const releaseTags = options['release-tag'] ?? [];
+		const releaseTagKeys = releaseTags.filter((_v, i) => i % 2 === 0);
+		const releaseTagValues = releaseTags.filter((_v, i) => i % 2 === 1);
+
+		releaseTagKeys.forEach((key) => {
+			if (key === '') {
+				throw new ExpectedError(`Error: Keys can't be empty "".`);
```
pdcastro, line 340:
```
throw new ExpectedError(`Error: --release-tag keys cannot be empty`);
```
(to clarify which keys this error message refers to, i.e. release tag keys)

lib/commands/push.ts
```diff
@@ -344,6 +305,143 @@ export default class PushCmd extends Command {
...
+			}
+			if (/\s/.test(key)) {
+				throw new ExpectedError(`Error: Keys can't have whitespaces in them.`);
```
pdcastro, line 343:
```
throw new ExpectedError(`Error: --release-tag keys cannot contain whitespaces`);
```
(to clarify which keys this error message refers to, i.e. release tag keys)

lib/commands/push.ts
```diff
@@ -344,6 +305,143 @@ export default class PushCmd extends Command {
...
+				(_.zip(releaseTagKeys, releaseTagValues) as Array<
+					[string, string]
+				>).map(async ([key, value]) => {
+					await sdk.models.release.tags.set(releaseId, key, value);
+				}),
+			);
+		} else if (releaseTagKeys.length > 0) {
+			throw new Error(stripIndent`
+				The remote build has started or concluded, but a release ID could not be parsed
+				out of the builder's output. As a result, the release tags have not been set.`);
```
pdcastro, line 390: I had suggested this error message before we decided to disallow using `--detached` in combination with `--release-tags`. In this case, the build is necessarily "concluded" (except for bugs, incorrect error handling...) and the message could be simplified a bit:
```
			throw new Error(stripIndent`
				A release ID could not be parsed out of the builder's output.
				As a result, the release tags have not been set.`);
```

lib/errors.ts
```diff
@@ -180,6 +180,7 @@ const messages: {
 // related issue https://github.com/balena-io/balena-sdk/issues/1025
 // related issue https://github.com/balena-io/balena-cli/issues/2126
 const EXPECTED_ERROR_REGEXES = [
+	/^--detached= cannot also be provided when using --release-tag=$/, // push.ts
```
srlowe, line 183: I'd suggest changing to something like:
`\ cannot also be provided when using -\`
As then we will handle all cases where oclif `exlude` rules are broken.

pdcastro (approved):

mbalamat:
@balena-ci I self-certify!

Issue #1508, opened 2019-11-10
balena build / deploy --build selects base image of wrong architecture (multiarch image manifest)


MatthewCroughan:
`balena build` and `balena deploy --build` may fail with "exec format error" or related mismatched architecture errors when the base image in a Dockerfile FROM line is a multiarch base image, where a single `name:tag` reference includes multiple images of different architectures. Example: `FROM ubuntu` where [ubuntu](https://hub.docker.com/_/ubuntu) is an image name that refers to multiple images of different architectures (ARM, ARM 64, 386, x86-64, PowerPC 64 LE, IBM Z, etc).

`balena build` and `balena deploy --build` have never had support for multiarch base images. Balena's original solution (that predates Docker's multiarch solution) is [Dockerfile templates](https://www.balena.io/docs/learn/develop/dockerfile/#dockerfile-templates) which work well with [balenalib base images](https://www.balena.io/docs/reference/base-images/base-images/). While balenalib has alternatives to `FROM ubuntu`, e.g. `FROM balenalib/raspberrypi3-ubuntu`, there are no alternatives to other base images like `FROM nginx` or `FROM telegraph`, so the CLI needs to implement support for such multiarch base images.

**Workaround**
A workaround is to append the sha256 hash to image name on the FROM line, as detailed in https://github.com/balena-io/balena-cli/issues/1508#issuecomment-801223357.

***Edited from Matthew's original report:***

> https://forums.balena.io/t/balena-cloud-cli-builder-is-still-not-able-to-interpret-image-manifests/43203/3
> I've already typed a lot here, and the references are included.

pdcastro:
Related to issue [cloud builder should understand and use the docker hub manifest files](https://github.com/balena-io/resin-builder/issues/626) (private repo)

MatthewCroughan:
@pdcastro The builder within BalenaOS does not understand `--platform` whereas the Balena cloud builder and `balena build` *do* understand this argument.

The result is that I cannot work with both the Local and Cloud methods in mind, I can only pick and choose one. This is a bit of a show-stopper for the project I'm working on. 

https://github.com/DynamicDevices/ming/issues/6

pdcastro:
Connects to #1408 [balena push selects incorrect image (multiarch images, Docker manifest)](https://github.com/balena-io/balena-cli/issues/1408)

MatthewCroughan:
@richbayliss This is the issue I referenced tonight at IoT Liverpool.

[...]

Issue #1408, opened 2019-08-27
balena push selects incorrect image (multiarch images, Docker manifest)


pdcastro:
[As reported in the forums](https://forums.balena.io/t/use-pre-build-multiarch-image-from-docker-hub/26110) (with thanks!), running `balena push` with a multiarch image hosted in Docker Hub, with a docker-compose file such as the following, results in "exec format error":

```
services:
  lora-connect:
    image: arribada/lora-connect
    ports:
    - "8070:8070"
```

The problem may not actually be in balena-cli, but rather the cloud builders, but this issue is being created to give more visibility to the issue and keep track of its resolution.

pdcastro:
Link to a related issue in a private GitHub repo (access to balena users only):
https://github.com/balena-io/resin-builder/issues/626

krasi-georgiev:
@pdcastro sorry for the noise, but this is a blocker for a project that is suppose to go in production soon so wondering if there has been any development towards fixing this bug. Some update would be greatly appreciated.

pdcastro:
@krasi-georgiev, I'm afraid there hasn't been activity on this issue as far as I am aware. If helpful, note that balena supports Dockerfile templates to allow a single Dockerfile to pull from base images of different architectures:
https://www.balena.io/docs/learn/develop/dockerfile/#dockerfile-templates

That's how `balena push` has "traditionally" dealt with multiple architectures. I understand that multiarch image support based on manifest files is a relatively new feature introduced by Docker. We intend to add support for it, but at the moment I don't have an ETA to share.

krasi-georgiev:
Thanks. 

> That's how balena push has "traditionally" dealt with multiple architectures.

I might be wrong, but I think this feature has been in docker since one of the early versions so it is not new. Until recently though it wasn't very easy to push multi arch images and this changed with one of the recent versions which simplifies that process.

MatthewCroughan:
https://github.com/balena-io/balena-cli/issues/1508

krasi-georgiev:
it would be useful to get some automated updates from the internal discussions and decisions as well

krasi-georgiev:
Has there been any progress with this.

I just tried it again with 12.20.0 and I still get 

```
.
.
[Success]          Successfully uploaded images
[Error]            Some services failed to build:
[Error]              Service: lora-gps-server
[Error]                Error: no matching manifest for unknown in the manifest list entries
[Info]             Built on arm01
[Error]            Not deploying release.
```

[...]

PR #2145, opened 2020-12-21
build, deploy: Add support for multi-architecture base images

pdcastro:
Resolves: #1408
Resolves: #1508
Change-type: minor

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -243,7 +243,7 @@ export async function buildProject(opts: {
 	projectPath: string;
 	projectName: string;
 	composition: Composition;
-	arch: string;
+	arch: string; // --arch option or application's architecture
 	deviceType: string;
 	emulated: boolean;
 	buildOpts: import('./docker').BuildOpts;
@@ -265,6 +265,7 @@ export async function buildProject(opts: {
 	);
 	const renderer = await startRenderer({ imageDescriptors, ...opts });
 	try {
+		await checkDockerPlatformCompatibility(opts);
 		await checkBuildSecretsRequirements(opts.docker, opts.projectPath);
 
 		const needsQemu = await installQemuIfNeeded({ ...opts, imageDescriptors });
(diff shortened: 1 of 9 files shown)


lib/utils/qemu.ts
```diff
@@ -84,47 +86,50 @@ export const getQemuPath = function (arch: string) {
...
+				.pipe(extract)
+				.on('error', reject)
+				.on('finish', function () {
+					fs.chmodSync(qemuPath, '755');
+					resolve();
+				});
+		});
+	} catch (err) {
+		try {
+			await fs.promises.unlink(qemuPath);
```
pdcastro, line 127: The `fs.createWriteStream(qemuPath)` statement towards the beginning of the function creates a zero-length file on disk. On successful download, the file is expanded to the size of the download. On errors, before this PR the file was left behind and then the CLI would assume the file no longer needed to be downloaded (a bug). I've added this `unlink` call to delete "corrupt" files on errors.

lib/utils/qemu.ts
```diff
@@ -148,18 +155,24 @@ export async function installQemuIfNeeded(
...
 		return false;
 	}
 	const { promises: fs } = await import('fs');
 	const qemuPath = await getQemuPath(arch);
 	try {
+		const stats = await fs.stat(qemuPath);
+		// Earlier versions of the CLI with broken error handling would leave
+		// behind files with size 0. If such a file is found, delete it.
+		if (stats.size === 0) {
+			await fs.unlink(qemuPath);
```
pdcastro, line 169: This is part of a bug fix whereby the CLI would leave behind zero-length files if the user specified an invalid architecture on the command line for the build/deploy commands, e.g. `--arch arm` instead of `--arch armv7hf`. Examples:
```
# ls -l ~/.balena/bin/
total 5476
-rw-r--r-- 1 root root       0 Dec 17 23:57 qemu-execve-386-v4.0.0+balena2
-rwxr-xr-x 1 root root 5606984 Dec 17 23:30 qemu-execve-armv7hf-v4.0.0+balena2
-rw-r--r-- 1 root root       0 Dec 17 23:57 qemu-execve-i386-v4.0.0+balena2
-rw-r--r-- 1 root root       0 Dec 17 23:57 qemu-execve-x86-v4.0.0+balena2
```

package.json
```diff
@@ -256,7 +256,7 @@
     "resin-compose-parse": "^2.1.2",
     "resin-doodles": "^0.1.1",
     "resin-image-fs": "^5.0.9",
-    "resin-multibuild": "^4.7.2",
+    "resin-multibuild": "^4.10.0",
```
pdcastro, line 259: This one line is the "core feature implementation" really... The rest of the PR is bug fixes (qemu download) and printing hints to the user about the `--pull` and `--emulated` options that become particularly relevant in conjunction with multiarch images.

lib/utils/qemu.ts
```diff
@@ -84,47 +86,50 @@ export const getQemuPath = function (arch: string) {
...
+	const installStream = fs.createWriteStream(qemuPath);
+	try {
+		extract.on('entry', function (header, stream, next) {
+			stream.on('end', next);
+			if (header.name.includes(`qemu-${qemuArch}-static`)) {
+				stream.pipe(installStream);
+			} else {
+				stream.resume();
+			}
+		});
```
srlowe, line 111: Does having this part under `try` have any effect?  What could throw?

lib/utils/qemu.ts
```diff
@@ -175,30 +188,27 @@ export async function installQemuIfNeeded(
  * - https://stackoverflow.com/questions/55388725/run-linux-arm-container-via-qemu-binfmt-misc-on-docker-lcow
  *
  * @param docker Dockerode instance
+ * @param emulated The --emulated command-line option
+ * @param logger Logger instance
```
srlowe, line 192: Should we delete the jsdoc params lists?

pdcastro: Why, because they are not useful? The `docker` and `logger` descriptions are not useful indeed, but the `emulated` description actually clarifies that it is the `--emulated` command line option.

lib/utils/qemu.ts
pdcastro: If unnecessary, It is also certainly harmless. I wasn't sure about one specific situation: suppose the execution flow "stopped" at the `await` clause below (`await new Promise(...)`). While "stopped" there, suppose `header.name.includes()` above threw because name is undefined (assume it is possible, though it may not be). In this case, would this `try` block still catch the error? Considering that the error is thrown by a callback of `extract.on`. My gut feeling is that the error would not be caught, but I wasn't sure.

In summary, I think you are right that the `extract.on` block could be moved outside the `try` block, but also harmless to be there. If other changes are needed to the PR I can move this block outside the try-catch, but if this was the only change, I'll just let it be. :-)

lib/utils/qemu.ts
srlowe: I thought I remember you suggesting we stop adding them, since we are using TS.  I don't mind though.

lib/utils/qemu.ts
srlowe: > While "stopped" there, suppose header.name.includes() above threw because name is undefined (assume it is possible, though it may not be). In this case, would this try block still catch the error? Considering that the error is thrown by a callback of extract.on. My gut feeling is that the error would not be caught, but I wasn't sure.

Yeah I think that's outside of the context of the try at that point (since the throw is happening in a callback).  I agree it's harmless, but wanted to check that there wasn't a bad assumption.

lib/utils/qemu.ts
pdcastro: Related to this, what I remember is that some param tags inherited from typed javascript files (.js) include the type hinting that becomes redundant when converted to typescript. Example:

`@param {boolean} emulated The --emulated command-line option`
_vs._
`@param emulated The --emulated command-line option`

So the type hinting should be deleted when converted to Typescript, especially as it introduces the possibility of conflict with the Typescript type specification. But the parameter descriptions may be useful, if they add some value ("Logger instance" does not add value, granted!)

ab77:
@pdcastro @srlowe is fix this blocked by anything? I am also affected by this, via https://github.com/balena-io/balena-cli/issues/2174 cc @klutchell

pdcastro:
> pdcastro srlowe is fix this blocked by anything? I am also affected by this, via #2174 cc klutchell

I've come across complications... Started an FD thread and pinged you guys here:
https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc

lib/utils/qemu.ts
pdcastro: The commit that this comment referred to was moved to PR #2207.

lib/utils/qemu.ts
pdcastro: The commit that this comment referred to was moved to PR #2207.

lib/utils/qemu.ts
pdcastro: The commit that this comment referred to was moved to PR #2207.

lib/utils/docker.ts
```diff
@@ -164,12 +164,85 @@ export function generateBuildOpts(options: {
...
-		dockerVersion.Engine && dockerVersion.Engine.match(/balena|balaena/)
+		// dockerVersion.Engine should be 'balena-engine' for the current
+		// version of balenaEngine, but at one point it was spelt 'balaena':
+		// https://github.com/balena-os/balena-engine/pull/32/files
+		(dockerVersion.Engine && dockerVersion.Engine.match(/balena|balaena/))
 	);
 }
+
+/** Detect whether the docker daemon is Docker Desktop (Windows or Mac) */
+export async function isDockerDesktop(
```
pdcastro, line 179: This function was extracted / refactored from `lib/utils/qemu.ts`.

lib/utils/compose_ts.ts
```diff
@@ -907,6 +908,114 @@ export function printGitignoreWarn(
...
+			If you experience that error, please take any one of the following actions:
+			* Upgrade Docker Engine to version 20.10.4 or later. If you are using
+			  Docker Desktop for Mac or Windows, upgrade it to version 3.2.1 or later.
+			* Downgrade Docker Engine to version 19.X.X. If you are using Docker Desktop
+			  for Mac or Windows, downgrade it to version 2.X.X.
+			* Downgrade the balena CLI to v12.40.3 or earlier. This would however cause
+			  support for multi-architecture base images to be lost.
+			* Manually run the 'docker pull' command for all base images listed in your
+			  Dockerfile(s) prior to executing the 'balena build' or 'balena deploy'
+			  commands, and then do not use the balena CLI's '--pull' flag.
```
pdcastro, line 977: Just learned that this advice is insufficient / incorrect with Docker v20.10.4 and v20.10.5. This PR will require further changes. :-/
https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc

Issue #1714, opened 2020-04-13
Improve `balena scan` error message re insufficient privileges and "sudo"


pdcastro:
As per product feedback ([gsheet](https://docs.google.com/spreadsheets/d/1V6URepXBhqU7WhFU1FwUrn-LdQi95ZF4je49dlM2YWQ/))

Instead of the current message:
```
PS C:\Users\paulo> balena scan
You don't have sufficient privileges to run this operation.
Run a new Command Prompt as administrator and try running this command again.

If this is not the case, and you're trying to burn an SDCard, check that the write lock is not set.
```

* Implement dynamic privilege escalation. On Windows, a pop-up would be raised asking the user to click Yes (or similar) to authorize running the command with administrator permissions. I believe the CLI already does it for other commands: https://github.com/balena-io/balena-cli/blob/v11.30.17/lib/utils/sudo.ts#L73

* Alternatively, perhaps the CLI could re-execute itself using the `runas` shell command:
https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc771525(v=ws.11)

* If that's not easily / reasonably possible, then reword the message:

```
PS C:\Users\paulo> balena scan
This command needs to be run as an administrator. Open a new command prompt
using the start menu "Run As Administrator" option. For additional guidance, see:
https://www.howtogeek.com/194041/how-to-open-the-command-prompt-as-administrator-in-windows-8.1/
```

Perhaps avoid the howtogeek link by choosing a widely available method and explaining how to do it in very few sentences (1-2-3 steps?) on the CLI output itself. Be sure that this explanation is not printed in red-on-bue! See #1712

pdcastro:
Connects-to: #542

Sukhoimaster:
I fix this by adding balena `/bin` directory in the `secure_path` option in `/etc/sudoers`

Using : Linux 16.04.1-Ubuntu  x86_64 x86_64 x86_64 GNU/Linux
Kernel : 4.15.0-106-generic 
NPM : 6.14.4
Nodes : v10.20.0

After complete via NPM installation as per [https://github.com/balena-io/balena-cli/blob/master/INSTALL.md#npm-installation](url)
- Locate balena directory by run in terminal `which balena` 

> the path output will be added to the `secure_path` option in `/etc/sudoers`

>  As in my case `/home/sonix/.nvm/versions/node/v10.20.0/bin/` 

[...]

Issue #542, opened 2017-06-07
`resin local` discovery requires running as `sudo`


hedss:
As part of the discovery of local devices, we've needed to ensure that the implementation of MDNS we use listens/sends on every available NIC. Whilst this works fine when there's no other implementation of MDNS/DNS-SD running, it requires `resin-cli` to be run using `sudo` when there *are* other instances (as `224.0.0.251:5353` needs to be shared). This is not ideal.

A solution to this is to actually hook into any underlying services that are already carrying out these duties:

* Linux - `avahi-daemon` should it have been installed
* OSX - `mDNSResponder` (the Bonjour implementation)
* Windows - `mDNSResponder`, should it have been installed

A fallback for Linux/Windows is to rely on the current JS implementation.

A spec. needs writing for this, along with an implementation.

hedss:
@flesler @afitzek Hi both, you've been potentially earmarked as chaps that could potentially pick this up and run with it. Would either of you like to take this on? ;-)

cc @shaunmulligan @alexandrosm

flesler:
Hi @hedss, I'm currently working on the tableau integration and after that on the persistent api keys which should take some weeks. I could take it but will be on the queue for weeks

hedss:
@flesler Might be better than nothing, dude. I guess it depends on how urgent Shaun and Alex think this is.

pimterry:
@hedss If this needs work sooner, I have a space in my planning once I've got the analytics & signup inbox work sorted, which is getting very very close to done already now. Start of next week at the latest.

Otoh I don't know much about MDNS, but I'm very happy to look into it, and the above all sounds very achievable.

hedss:
@pimterry If you think you could, then that would be wonderful, thank you! I'll hopefully be able to answer any relevant questions about it, too.

pimterry:
An update on this: it's been bumped to second on my list by https://github.com/resin-io/resin-cli/issues/549, but it will be coming straight after that.

srlowe:
@pdcastro I think we should close either this or #1714 as duplicate, since they are the same issue.

PR #2148, opened 2020-12-30
Refactor out command internal scandevices

srlowe:
This removes the internal-only command `internal scandevices`, which was only used by `leave`/`join`.   `Leave` and `join` have also been refactored to not require elevated privileges when scanning for devices (which from my tests appears unnecessary).


lib/commands/leave.ts
```diff
@@ -42,7 +42,7 @@ export default class LeaveCmd extends Command {
 
 		If you don't specify a device hostname or IP, this command will automatically
 		scan the local network for balenaOS devices and prompt you to select one
-		from an interactive picker. This usually requires root privileges.
```
pdcastro, line 45: We need to understand this better -- whether or not root privileges are required. I gather that `balena join` and `balena leave` use the same implementation as `balena scan`:
```
const { discover } = await import('balena-sync');
await discover.discoverLocalBalenaOsDevices();
```
```
const { forms } = await import('balena-sync');
hostnameOrIp = await forms.selectLocalBalenaOsDevice();
```
Where `selectLocalBalenaOsDevice` calls `discoverLocalBalenaOsDevices`:
https://github.com/balena-io-modules/balena-sync/blob/v11.0.2/lib/forms.coffee#L24

If it was the case that we could simply drop the requirement for sudo, it would be great as there are significant issues with `sudo balena scan`. The most recent is #2149 that basically describes how `sudo balena scan` fails on Linux because `balena` is not in the path for `sudo`, and the user himself offers to update the docs to add the requirement of editing the PATH-for-sudo with a `visudo` command that would then be an additional installation requirement of the CLI. It would be much nicer if we simply didn't need sudo...

A preliminary digging reveals #542 and #1714. By the way, re your question there:

> pdcastro I think we should close either this or #1714 as duplicate, since they are the same issue.

The two issues are closely related but I think not duplicates. #542 is about removing the need for superuser privileges (so the command runs as an unprivileged user). #1714 is about implementing dynamic privilege escalation, so the CLI dynamically raises a privilege popup (GUI) or password prompt on the command line. In both issues, `sudo` would be avoided, but by very different means. :-)

Quoting Hedss in #542:
> As part of the discovery of local devices, we've needed to ensure that the implementation of MDNS we use listens/sends on every available NIC. Whilst our fix to MDNS (https://github.com/balena-io-modules/multicast-dns/commit/a15c63464eb43e8925b187ed5cb9de6892e8aacc) works fine when there's no other implementation of MDNS/DNS-SD running, it requires `resin-cli` to be run using `sudo` when there *are* other instances (as `224.0.0.251:5353` needs to be shared). This is not ideal.

I haven't understood it to the point of knowing whether simply removing the sudo requirement would cause `balena scan` to fail frequently or rarely. If the situation was detectable at runtime, we could try to escalate privileges only when really required.

srlowe: @pdcastro I don't have time at the moment to do a full investigation of `balena scan`, so will move this to draft for now.

pdcastro: @srlowe, we can get this PR merged (it's great to remove the `scandevices` internal command 👍 ) almost as it is (or even as it is...). How about just rewording the following:

Original: `This requires root privileges.`
Current PR: (deleted)
Proposed: `This may require administrator/root privileges.`

Then the investigation re removing sudo could be left for #542, #1714, #2149.

srlowe: @pdcastro Ok sure, sounds good.  I'd like to get the scan command investigated too, but I just have too much stuff open at the moment.

pdcastro (approved):

PR #2151, opened 2021-01-08
Handle 'push' edge case with application access

srlowe:
// * Temporary fix *
// When doing `$expand organization` on a public app (when not the owner)
// a partial document is returned without any organization data.
// So e.g. `balena push balena-sound` will break this commands logic.
// The balena cloud builder api will soon change to accept application slugs
// instead of `owner, app`, after which we will not need to do the expand


lib/commands/push.ts
```diff
@@ -374,6 +374,21 @@ export default class PushCmd extends Command {
...
+		// a partial document is returned without any organization data.
+		// So e.g. `balena push balena-sound` will break this commands logic.
+		// The balena cloud builder api will soon change to accept application slugs
+		// instead of `owner, app`, after which we will not need to do the expand
+		//
+		// Users should not be pushing to public apps anyway, so for now catch this situation:
+		if (!application.organization[0]) {
+			throw new ExpectedError(stripIndent`
+			You do no have rights to push to the application ${application.app_name}.
+			If you are trying to deploy a public app, please make sure that you have created the application first.
```
pdcastro, line 388: * Is this because only the owner of a public app can push to it? If so, perhaps it would be informative to users to state it:
```

	If you are trying to deploy a public app, please make sure that you are
	the application owner.
```

* NIT: Line exceeds 80 characters (no line break).

lib/commands/push.ts
```diff
@@ -374,6 +374,21 @@ export default class PushCmd extends Command {
...
+		// When doing `$expand organization` on a public app (when not the owner)
+		// a partial document is returned without any organization data.
+		// So e.g. `balena push balena-sound` will break this commands logic.
+		// The balena cloud builder api will soon change to accept application slugs
+		// instead of `owner, app`, after which we will not need to do the expand
+		//
+		// Users should not be pushing to public apps anyway, so for now catch this situation:
+		if (!application.organization[0]) {
+			throw new ExpectedError(stripIndent`
+			You do no have rights to push to the application ${application.app_name}.
```
pdcastro, line 387: * Would it be better that this was `application.slug` instead of `application.app_name`? (I did not test, needs testing...) The slug, assuming it contains org / user account details, might help clarify to the user that the push was attempted to someone else's application.
* no → not
* Is the word "permissions" more common than "rights"? -  
`You do not have permissions to push to application ${application.app_name}.`

srlowe: > Would it be better that this was application.slug instead of application.app_name

In these cases the partial document you get back from the API doesn't contain the slug.  (Though may be a case of changing the query).

> no → not
:-1:

> Is the word "permissions" more common than "rights"? -
Yeah probably :+1:

lib/commands/push.ts
srlowe: > Is this because only the owner of a public app can push to it

That's my understanding yes.   

> If you are trying to deploy a public app, please make sure that you are the application owner.
	
I'm not sure.  This bug was discovered by a user that was trying to deploy a public app, but misunderstood the process.  Would that message help them?  They might read it as "you need to be the owner of the public app to deploy it".  Whereas of course we are talking about two "apps" (the public app, and the users app they will deploy it too).

pdcastro: I see, I was misunderstanding it too. :-) Yes, the original version is better. Perhaps with a different minor twist: "your own application" instead of "the application" -
```
If you are trying to deploy a public app, please make sure that you have created
your own application first.
```

lib/commands/push.ts
pdcastro: > [...] the partial document you get back from the API doesn't contain the slug. Though may be a case of changing the query.

OK. If a small query change could reveal the slug, I think it would be worth it. Otherwise, something like (for example) _"You do not have permissions to push to application balena-sound"_ could be misleading to a user who had a `balena-sound` application in some of their orgs. Although, if this error message was "impossible" in this situation, then it would not be misleading...

pdcastro (commented):
Note (just in case): This PR is a Draft, therefore I cannot approve it.

PR #2154, opened 2021-01-13
Add release-tag on deploy command

mbalamat:
This is in order to re-use the same logic from the push command on the deploy command.

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -44,6 +44,60 @@ import type { DeviceInfo } from './device/api';
 import { getBalenaSdk, getChalk, stripIndent } from './lazy';
 import Logger = require('./logger');
 
+/**
+ * Given an array representing the raw `--release-tag` flag of the deploy and
+ * push commands, parse it into separate arrays of release tag keys and values.
+ * The returned keys and values arrays are guaranteed to be of the same length.
+ */
+export function parseReleaseTagKeysAndValues(
+	releaseTags: string[],
+): { releaseTagKeys: string[]; releaseTagValues: string[] } {
+	if (releaseTags.length === 0) {
+		return { releaseTagKeys: [], releaseTagValues: [] };
+	}
+
+	const releaseTagKeys = releaseTags.filter((_v, i) => i % 2 === 0);
+	const releaseTagValues = releaseTags.filter((_v, i) => i % 2 === 1);
+
+	releaseTagKeys.forEach((key: string) => {
+		if (key === '') {
+			throw new ExpectedError(`Error: --release-tag keys cannot be empty`);
+		}
+		if (/\s/.test(key)) {
+			throw new ExpectedError(
+				`Error: --release-tag keys cannot contain whitespaces`,
+			);
+		}
+	});
+	if (releaseTagKeys.length !== releaseTagValues.length) {
+		releaseTagValues.push('');
+	}
+	return { releaseTagKeys, releaseTagValues };
+}
+
+/**
+ * Use the balena SDK `models.release.tags.set()` method to set release tags
+ * for the given release ID. The releaseTagKeys and releaseTagValues arrays
+ * must be of the same length; their items map 1-to-1 to form key-value pairs.
+ */
+export async function applyReleaseTagKeysAndValues(
+	sdk: BalenaSDK,
+	releaseId: number,
+	releaseTagKeys: string[],
+	releaseTagValues: string[],
+) {
+	if (releaseTagKeys.length === 0) {
+		return;
+	}
+	await Promise.all(
+		(_.zip(releaseTagKeys, releaseTagValues) as Array<[string, string]>).map(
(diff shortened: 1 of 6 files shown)


lib/common.ts
```diff
@@ -0,0 +1,48 @@
+import { BalenaSDK } from 'balena-sdk';
```
pdcastro, line 1: * Regarding the name of this file, `lib/common.ts`, it sounds very generic, while the contents are quite specific to parsing a flag (`--release-tag`) that applies to two CLI commands only. Instead of creating a new file, what about using one of the following existing files?
  * `lib/utils/common-flags.ts` (though currently it contains the flags themselves, rather than functions to parse them)
  * `lib/utils/compose_ts.ts` - this file contains `composeCliFlags` at the end, and the file is already imported by both `push.ts` (line 279 in master, inside the `run()` method) and `deploy.ts`. (The code in `compose_ts.ts` is mainly used by the build and deploy commands, and `--release-tag`.)

* If a new file is created, we usually add the Apache licence comment at the top.

lib/common.ts
```diff
@@ -0,0 +1,48 @@
...
+export async function applyReleaseTagKeysAndValues(
+	sdk: BalenaSDK,
+	releaseId: number,
+	releaseTagKeys: string[],
+	releaseTagValues: string[],
+) {
+	if (releaseTagKeys.length === 0) {
+		return;
+	}
+	// Above we have checked that releaseTagKeys and releaseTagValues are of the same size
```
pdcastro, line 39: This comment could be moved to a function documentation block (`/** ... */`), to specify a contract that the caller must abide to when calling this function (pompous way of saying 😁  that it is up to the caller to ensure that `releaseTagKeys` and `releaseTagValues` arguments are of the same length).

lib/common.ts
```diff
@@ -0,0 +1,48 @@
+import { BalenaSDK } from 'balena-sdk';
+import { ExpectedError } from './errors';
+
+// TODO fix this any
+export function getReleaseTagKeysAndValues(options: any) {
```
pdcastro, line 5: This function only uses `options` to access `options['release-tag']`. So what about changing its signature to be:
```
export function parseReleaseTagKeysAndValues(releaseTags: string[]) {
```
So the caller does:
```
const { releaseTagKeys, releaseTagValues } = parseReleaseTagKeysAndValues(
    options['release-tag'] ?? [],
);
```
This way, you don't need to worry about typing `options` in a way that covers both the `push` and `deploy` commands. I also renaming it to `parse` instead of `get`.

lib/utils/compose_ts.ts
```diff
@@ -44,6 +44,64 @@ import type { DeviceInfo } from './device/api';
...
+		}
+	});
+	if (releaseTagKeys.length !== releaseTagValues.length) {
+		releaseTagValues.push('');
+	}
+	return { releaseTagKeys, releaseTagValues };
+}
+
+/**
+ * The parseReleaseTagKeysAndValues function returns
```
pdcastro, line 79: This description documents more the `parseReleaseTagKeysAndValues` function than the `applyReleaseTagKeysAndValues` function... Building on my comments for that function, here's a suggested description for `applyReleaseTagKeysAndValues`:
```
/**
 * Use the balena SDK `models.release.tags.set()` method to set release tags
 * for the given release ID. The releaseTagKeys and releaseTagValues arrays
 * must be of the same length; their items map 1-to-1 to form key-value pairs.
 */
export async function applyReleaseTagKeysAndValues(
	sdk: BalenaSDK,
	releaseId: number,
	releaseTagKeys: string[],
	releaseTagValues: string[],
) {
```

pdcastro (commented):
Almost there 👍  I also have a comment regarding the first line of the commit, which is the only line that will be copied to the Changelog file (and to the all-hands call document!) -
```
Re-use release-tag option logic on deploy command
```
End users may find it clearer if we simply state that this CLI release adds the `--release-tag` flag to the `balena deploy` command. The commit that did it for the push command read:
```
push: Add --release-tag flag [Marios Balamatsias]
```
So this commit might as well be:
```
deploy: Add --release-tag flag
```
Alternatively:
```
Add --release-tag flag to the "deploy" command
```

lib/utils/compose_ts.ts
```diff
@@ -44,6 +44,64 @@ import type { DeviceInfo } from './device/api';
 import { getBalenaSdk, getChalk, stripIndent } from './lazy';
 import Logger = require('./logger');
 
+/**
+ * The parseReleaseTagKeysAndValues function returns
+ * releaseTagKeys and releaseTagValues given that the
+ * parameter
+ * @param {string[]} releaseTags
+ * @returns { Object } { releaseTagKeys, releaseTagValues }
```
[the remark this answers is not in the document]
mbalamat: Good points, thanks @pdcastro !

pdcastro (approved):

PR #2156, opened 2021-01-15
ci node 12 14

chenrui333:
- ci: add node12 and node14 to travis
- ci: add node 12 and node 14 to appveyor

---

relates to https://github.com/Homebrew/homebrew-core/pull/63674

PR #2157, opened 2021-01-15
feat(ci): add github action flow

chenrui333:


.github/workflows/ci.yml
```diff
@@ -2,8 +2,6 @@ name: ci
 
 on:
   push:
-    branches:
-      - master
```
pdcastro: What would this achieve? (My knowledge of GitHub actions is limited.) I found this documentation:
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration
If I was to guess... The intention was to ensure that Travis / AppVeyor tests were triggered for external contribution pull requests like #2156. Also, I cannot even find the file `.github/workflows/ci.yml` in the repo: https://github.com/balena-io/balena-cli/tree/master/.github
But this PR clearly amends an existing file rather than add a new file, which makes me properly confused. :-)

pdcastro:
I've converted this PR to draft as I don't understand what it would achieve. It may be related to PR #2156, where I added a [comment](https://github.com/balena-io/balena-cli/pull/2156#issuecomment-760979153) clarifying that the balena-cli repo no longer uses Travis and Appveyor, instead using a custom balena CI setup. PR #2158 deleted old Travis and Appveyor configuration files to avoid future confusion.

Issue #1450, opened 2019-09-25
Release builds for armv7hf and arm64


siredmar:
It would be nice to have the balena-cli builds ready to use for armv7hf and arm64 in the releases download section.
Would that be much work to integrate those builds for arm in your CI chain?
That would eliminate the need for us (or many other?) to build balena cli from source for our target archs.

pdcastro:
Thanks for the suggestion. I'd like to understand a bit more about the use case, i.e. the need for running balena-cli on an ARM architecture. Usually balena-cli is executed on development machines (laptop / desktop computers) or cloud CI environments (AWS, Travis...) which typically have x86 processors, rather than for example on balena devices which often have ARM processors. For x86 platforms, balena-cli offers pre-compiled standalone zip packages (and Mac / Windows native installers) that avoid the need to build balena-cli from source.

In relation to the `balena build` and `balena deploy --build` commands, when building images for the ARM architecture, the `--host` and `--port` options can be used to select a remote Docker or balenaEngine daemon which may be running on a balena device. For example, I would run the following on my x86 Mac laptop:
```
balena build project/folder -a test-rpi -h 192.168.1.10 -p 2375
```
So balena-cli runs on x86 and balenaEngine runs on ARM.

For other balena commands, it may or may not make sense to use the [balena API](https://www.balena.io/docs/reference/api/overview/) directly.

pdcastro:
Connects to #845 (CLI does not install on arm64)

pdcastro:
Note: There is a fresh need / use case for this issue's feature request in the "balena IDE" project where a development environment (IDE, including the CLI) is installed as a Docker container on the device itself.

PR #2159, opened 2021-01-15
dockerfiles: initial commit of balenalib dockerfiles

klutchell:
Add dockerfiles for alpine and debian images, based on upstream
balenalib/arch-distro-node images.


klutchell:
@pdcastro can I get your thoughts on this to get the ball rolling? All of our planned CI stuff will eventually fit around these dockerfiles I think.

docker/DOCKER.md
```diff
@@ -0,0 +1,72 @@
...
+    --pull
+```
+
+## Usage
+
+### basic
+
+```bash
+# print usage
+docker run --rm -it balenalib/amd64-debian-balenacli:latest --help
```
pdcastro, line 37: These examples all use `docker run` and don't include a command that requires logging in as a separate previous step. I understand that `docker run` creates a new container every time, which means that a token saved by `balena login` is not preserved for subsequent commands like `balena apps`. It also means that a docker daemon is started every time, which is verbose and slows things down. I know you already knew all of this and these are just examples, but I'm thinking that some inexperienced users would not know any better and would think that that's how it is. What about adding an example that runs a single container in the background (single `docker run` command, single docker daemon), then uses `docker exec` to run additional processes? For example:

```
$ docker run --detach --privileged --network host --name cli --rm -it balenalib/amd64-debian-balenacli /bin/bash

$ docker exec -it cli balena version -a
balena-cli version "12.38.1"
Node.js version "12.19.1"

$ docker exec -it cli balena login --token abc...

$ docker exec -it cli balena whoami
== ACCOUNT INFORMATION
USERNAME: ...
EMAIL:    ...
URL:      balena-cloud.com

$ docker exec -it cli balena apps
ID      APP NAME         SLUG                            DEVICE TYPE     ONLINE DEVICES DEVICE COUNT
1491721 test-nuc         gh_paulo_castro/test-nuc        intel-nuc       0              1
...

$ docker exec -it cli balena app test-nuc
== test-nuc
ID:          149...
DEVICE TYPE: intel-nuc
SLUG:        gh_.../test-nuc
COMMIT:      ce9...
```

klutchell: Hey, I agree with all your README feedback, it was meant as a placeholder/reference until we have official images and documentation, but there's a good chance external users will notice this so I'll give it a rework with your suggestions.

The examples I have here are trash, it was mostly so I could remember which flags are required for which type of action, but as you said these obviously aren't real use cases and would never work in stand-alone. I'll come up with some proper ones and get in the mindset of using this as a basis for official docs.

klutchell:
@balena-ci rebase

docker/DOCKER.md
```diff
@@ -0,0 +1,167 @@
...
+- `amd64`
+- `i386`
+
+## Basic Usage
+
+Here's a small example of running a single, detached container
+in the background and using `docker exec` to run balena CLI commands.
+
+```
+$ docker run --detach --privileged --network host --name cli --rm -it balenalib/amd64-debian-balenacli /bin/bash
```
pdcastro, line 19: `balenalib/amd64-debian-balenacli`

iirc, we had discussed ([flowdock thread](https://www.flowdock.com/app/rulemotion/resin-projects/threads/enZ9VZOK1lFNUhq5N2AfTssafFb)) that this PR could be merged before the images were actually available in Docker Hub. I may have said that I was OK with this, but I am having second thoughts. :-) What was the benefit of merging this PR before (well before?) transformers / something else was available to actually use the Dockerfiles and upload the images to Docker Hub?  What I am specifically concerned with is that users will see this documentation and Dockerfile in the CLI repo, will try to use it, and then will create CLI issues ("it doesn't work!"), will contact support, and will be disappointed... 

The PR could be left around unmerged, until the time comes. Or, if helpful, we could merge the Dockerfiles without this DOCKER.md file. Or we could replace the current contents of DOCKER.md with:

```
Note: The files in this folder are experimental and a work in progress, and not
ready for end users.
```
But if it's experimental and work in progress, then why merge it...
Ah! Note that transformers may also be triggered (if we implement it so) when commits are pushed to a branch like this one. Well, at least build tests are triggered by merely pushing commits, so we could probably find a way of making sure that whatever we needed was triggered to.

Anyway, perhaps it's just a case of me not seeing (or having already forgotten) "the reason why". :-)

klutchell: @pdcastro you probably aren't forgetting our reason to merge these early, cause I don't think I provided one!

I personally wanted these Dockerfiles put somewhere more "stable" than the playground. As part of my fear of losing track of good bits of code, I like to know everything is stashed in the correct repository with eyes on it so it doesn't get lost or forgotten. That doesn't require by any means that this needs to be merged right now!

I also don't know how long it will be until transformers are ready to pick up this kind of project (I likely don't have the bandwidth to be much help getting them there), and I'd hate to look at these files next year and forget how the heck they work!

We could totally leave this PR as a draft, or merged without the DOCKER.md, or with a disclaimer at the top. Leaving it as a draft is probably easiest?

pdcastro: Sorry for the delay in replying. This excellent bit of code will not be lost if left here as a draft PR... But also, someone at some point will have to pick up the effort of using transformers to build the CLI images and upload them to Dockerhub, otherwise the excellent bits of code will just stay here for a long time. :-)

Hey, what about modifying `DOCKER.md` to include instructions of how users could manually build the images on their own machines / devices? Using Docker on the laptops, or balenaEngine on balenaOS devices. E.g., including `docker build` and `balena-engine build` command lines in addition to `docker run` command lines, and removing references to images that don't yet exist, like `balenalib/amd64-debian-balenacli`. Then this could be merged now and start being of some use to end users. The following bit of the installation instructions could then be amended as well:

> https://github.com/balena-io/balena-cli/blob/master/INSTALL-ADVANCED.md#npm-installation
> _This [sample Dockerfile](https://gist.github.com/pdcastro/5d4d96652181e7da685a32caf629dd44) shows the CLI installation steps on an Ubuntu 18.04 base image._

Instead of that sample Dockerfile, the advanced installation instructions would refer to these Dockerfiles. Perhaps a new section could be added to `INSTALL-ADVANCED.md`, with a short introduction to the concept of running the CLI through docker containers, and a link to DOCKER.md for the command lines and examples.

By the way, here's a half baked idea / question. I know for a fact that some users insist on running the CLI on devices (e.g. because they find it easier to write a shell script around some CLI commands, instead of programming with the Node SDK or making direct use of balena's HTTP API). But an issue for them is that we don't have standalone installers for ARM, which would allow the CLI to be used on the Raspberry Pi. Sometimes they get around it by installing the CLI with npm, but it starts getting messy as it needs the CLI dependencies for compilation of native modules etc. With these Dockerfiles, there'd be the obvious solution of users copying-and-pasting the Dockerfile contents to their application's Dockerfile, or the other way around. But - and here's the half baked question - could they add the CLI as a service to their microservices / multicontainer application? E.g. adding the CLI Dockerfile as a service in the `docker-compose.yml` of their application. In this case, could they do the equivalent of `docker exec -it` inside an app container, to exec the CLI in the other container? Perhaps using DBUS or some other simpler / clever way of doing it? Actually, if the balenaEngine socket is exposed in the container, they could perhaps include a docker client in the app container and literally run `docker exec -it` - maybe we could document some shell `alias` command or shell function to make it user friendly to run `balena` (the CLI) across containers. Could input and output files be placed in a shared volume, for CLI commands that operate on files? ... Anyway, just a thought. :-)

klutchell: I can definitely adjust the readme and documentation to include instructions of how users could manually build the images on their own machines / devices! If you'll recall, this readme initially had **Build** instructions near the top since that was the only way to get an image until we pipelined everything. I can move it back to the top and elaborate a bit more, since I use this as a template for a few projects as it is!

Regarding half-baked, I definitely think there's something here as far as running a balena-cli service in existing apps, I had the same idea. The only thing I haven't solved for yet is the best way to communicate between the containers as you mentioned. We have a similar problem in testbot/exclave where we want to execute commands in a parallel testbot container (ideally without an API or REST interface).

You may be on to something with the docker client though. If we could provide instructions or a pre-built client that would allow executing commands in another container as long as the host balena socket is exposed, that would be applicable to many uses cases, not just the cli.

Maybe two components with pre-packaged files and examples?
1. `balena-exec` unpack this binary your app (preinstalled in balenalib base images) to execute commands in other services
2. `balena-alias` with a balena-cli service (depends on balena-exec) use this alias to run `balena` commands as if the cli were installed (using shared volumes, etc)

pdcastro: > I definitely think there's something here as far as running a balena-cli service in existing apps, I had the same idea.

There is an additional consideration though, before we went ahead with something like that. Currently, we don't advise users to run the CLI on devices, in fact we occasionally advise them not to, e.g.:
https://github.com/balena-io/balena-cli/issues/1450#issuecomment-535024573

The CLI uses Amplitude.com for command usage tracking and Sentry.io for error reporting with the assumption that users run the CLI on their dev workstations, which is a low-volume scenario. Even so, we occasionally hit the Sentry.io rate limit and monthly limit. If users were to deploy the CLI in fleets of thousands of devices, such usage tracking and error reporting would be degraded. So I think there would need to be a production discussion and code and docs adjustments before we did that. Technically, we could try some sort of auto detection to disable Amplitude/Sentry, and/or we could document flags or env vars to disable those features. (Requests to Amplitude use the balena API as a proxy, so work could be done there too.)

> If you'll recall, this readme initially had Build instructions near the top since that was the only way to get an image until we pipelined everything. I can move it back to the top and elaborate a bit more, since I use this as a template for a few projects as it is!

Yes it would be good, but now with the view that the build step is required rather than optional, because "official" images don't exist in `balenalib` or Docker Hub (so no references to `balenalib/amd64-debian-balenacli`. The point above about running the CLI in automated environments may still apply to some extent though. Why would users want to build CLI images and run the CLI using Docker? If it's to run the CLI interactively in their dev workstation then fine, but I suspect it would most often be related to running in CI environments or on devices. That product discussion may be unavoidable... Do we want to make it easier (or even necessary) to disable Amplitude and Sentry.io?...

INSTALL-ADVANCED.md
```diff
@@ -148,3 +148,15 @@ To use a remote Docker Engine (daemon) or balenaEngine, specify the remote machi
 port number with the `--dockerHost` and `--dockerPort` command-line options. For more details,
 check `balena help build` or the [online
 reference](https://www.balena.io/docs/reference/cli/#cli-command-reference).
+
+## Docker Installation
+
+If you're working with Linux and the NPM Installation method doesn't suit your needs
+you may want to try deploying the CLI with Docker.
+
+Official images are not available yet but check out [this guide](docker/DOCKER.md) on how
```
pdcastro, line 157: So just to confirm, the first step for the user would be to `git clone` the CLI repo, right?

klutchell: I suppose so, or downloading the source tarball or just copying the Dockerfiles? What do we normally suggest in these DIY cases?

docker/DOCKER.md
pdcastro: JF brainstorm topic (restricted access): https://jel.ly.fish/faccdba2-becc-40be-8f74-f34b9703c5d5

klutchell:
@balena-ci rebase

docker/DOCKER.md
```diff
@@ -0,0 +1,173 @@
...
+modes in order to allow only the required functionality, and not
+elevate permissions unless required.
+
+### scan
+
+- <https://www.balena.io/docs/reference/balena-cli/#scan>
+
+```bash
+# balena scan requires the host network and NET_ADMIN
+docker run --rm -it --cap-add NET_ADMIN --network host \
```
pdcastro, line 85: In my testing, this was successful on Linux (Debian v10) with "regular Docker" but not on macOS with Docker Desktop. I understand that Docker Desktop uses a virtual machine behind the scenes, and perhaps this is the reason why. Did you get to test with Docker Desktop on macOS or Windows? We may want to document that this is not expected to work with Docker Desktop on macOS or Windows, if you think this sounds right.

Also, on my testing on Linux (Debian v10 and balenaOS), I didn't need `--cap-add NET_ADMIN` for `scan` to work.

Hades32: `NET_ADMIN` isn't needed because it's run as root in the container (which we can't avoid if we want to support Docker in docker)

On Windows, the same behavior is expected. The VM is using a network interface with a separate IP range. It would be great though if we could support this (e.g. by detecting the outer IP range) as this also breaks `balena scan` in WSL2.

klutchell:
@balena-ci retest

klutchell:
@balena-ci retest

.resinci.yml
```diff
@@ -17,4 +17,66 @@ npm:
         - "14"
 
 docker:
-  publish: false
+  builds:
+    - path: .
+      dockerfile: ./docker/alpine/Dockerfile
+      docker_repo: balenalib/amd64-alpine-balenacli
```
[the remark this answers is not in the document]
klutchell: @pdcastro it's not a lack base images, but a failure to `npm install ffi-napi` on aarch64 musl (alpine).

https://github.com/node-ffi-napi/node-ffi-napi/issues/126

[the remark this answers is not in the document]
pdcastro: Thanks for clarifying it -- I thought it would be something like that.
No chance you would have missed those images! 😁

INSTALL-ADVANCED.md
```diff
@@ -148,3 +148,16 @@ To use a remote Docker Engine (daemon) or balenaEngine, specify the remote machi
...
+If you're working with Linux and the NPM Installation method doesn't suit your needs
+you may want to try deploying the CLI with Docker.
+
+Official images are available and can be used either as a standalone service or as a
+base image for your project that leverages the CLI. Check out [this guide](docker/DOCKER.md) on how
+to pick an image and get started!
+
+Note that there are some additional considerations when running the CLI via Docker so pay
+close attention to the [Basic usage](docker/DOCKER.md#basic-usage) and
+[Advanced Usage](docker/DOCKER.md#advanced-usage) sections for examples of different CLI commands.
```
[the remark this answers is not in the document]
klutchell: Good explanation, I'm not sure why I had the Linux restriction either. I think in my mind Windows and MacOS users were advised to use the installer but it's really not a restriction.

docker/DOCKER.md
```diff
@@ -0,0 +1,207 @@
+# Docker Images for balena CLI
```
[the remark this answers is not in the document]
klutchell: > There were past discussions about styling balena CLI as balenaCLI

I vaguely recall these discussions which is why I avoided using `balenaCLI` but making sure to include `the balena CLI` is another good consideration.

docker/DOCKER.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Features Overview
+
+These CLI images are based on the popular [Balena base images](https://www.balena.io/docs/reference/base-images/base-images/)
+so they include many of the features you see there.
+
+- Multiple Architectures:
+    - `rpi`
+    - `armv7hf`
+    - `aarch64` (debian only)
```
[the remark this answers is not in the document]
klutchell: unfortunately it did not, the issue is with one of our dependencies as I linked above

docker/DOCKER.md
```diff
@@ -0,0 +1,207 @@
...
+    - `debian`
+    - `alpine`
+- [cross-build](https://www.balena.io/docs/reference/base-images/base-images/#building-arm-containers-on-x86-machines) functionality for building ARM containers on x86.
+- Helpful package installer script called `install_packages` inspired by [minideb](https://github.com/bitnami/minideb#why-use-minideb).
+
+## Image Names
+
+`balenalib/<arch>-<distro>-balenacli:<cli_ver>`
+
+- `<arch>` is the architecture and is mandatory. If using Dockerfile.templates, you can replace this with `%%BALENA_ARCH%%`.
```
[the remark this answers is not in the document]
klutchell: Agreed, should we update the base image docs as well since that's where I copied these lines from?

https://github.com/balena-io/docs/blob/master/pages/reference/base-images/base-images.md#image-names

docker/DOCKER.md
```diff
@@ -0,0 +1,207 @@
...
+    - `i386`
+- Multiple Distributions
+    - `debian`
+    - `alpine`
+- [cross-build](https://www.balena.io/docs/reference/base-images/base-images/#building-arm-containers-on-x86-machines) functionality for building ARM containers on x86.
+- Helpful package installer script called `install_packages` inspired by [minideb](https://github.com/bitnami/minideb#why-use-minideb).
+
+## Image Names
+
+`balenalib/<arch>-<distro>-balenacli:<cli_ver>`
```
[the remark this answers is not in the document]
klutchell: Pinning is always the way to go and I'm usually very strict about that. I'm surprised I forgot to do that here but I guess there's been a lot of churn.

I like the automation idea, I'll make a note to look into that!

docker/DOCKER.md
```diff
@@ -0,0 +1,207 @@
...
+- `<distro>`: alpine - smaller footprint than debian
+- `<cli_ver>`: omitted - the latest available CLI version will be used
+
+## Basic Usage
+
+Here's a small example of running a single, detached container
+in the background and using `docker exec` to run balena CLI commands.
+
+```
+$ docker run --detach --privileged --network host --name cli --rm -it balenalib/amd64-debian-balenacli /bin/bash
```
pdcastro, line 59: Should we also add `--pull` to this list?

docker/DOCKER.md
```diff
@@ -0,0 +1,207 @@
...
+
+`balenalib/<arch>-<distro>-balenacli:<cli_ver>`
+
+- `<arch>` is the architecture and is mandatory. If using Dockerfile.templates, you can replace this with `%%BALENA_ARCH%%`.
+For a list of available device names and architectures, see the [Device types](https://www.balena.io/docs/reference/base-images/devicetypes/).
+- `<distro>` is the Linux distribution and is mandatory. Currently there are 2 distributions, namely `debian` and `alpine`.
+
+## Image Tags
+
+In the tags, all of the fields are optional, and if they are left out, they will default to their `latest` pointer.
```
pdcastro, line 35: How about adding the following note here:
```
To ensure that the latest available CLI image is used, it is also necessary
to add the `--pull` flag to `docker run` commands. Otherwise, Docker will
reuse a locally cached image, without checking for available updates.
```
The CLI itself notifies users when new versions are available, but not less than 24 hours since installation, and not when executed as the `root` user. The last check timestamp is stored in `~/.config/configstore/update-notifier-balena-cli.json`. With these Docker images, I think users would never get notified of available updates. I think the CLI update notification mechanism will need improvements (in some other PR) to work better with these Docker images (and any installation where the CLI is executed as the root user).

.dockerignore
```diff
@@ -0,0 +1,42 @@
...
+.nvmrc
+.nyc_output
+.vscode
+coverage
+lib-cov
+logs
+pids
+
+# OS cache files
+**/.DS_Store
```
pdcastro, line 23: I have added a commit to review these patterns. A difference between `.gitignore` and `.dockerignore` is that a pattern like `.DS_Store` applies to subdirectories as well in `.gitignore`, but only to the project root directory in `.dockerignore`. So I rewrote some such `.dockerignore` patterns as `**/.DS_Store`.
I have also added comments at the top of the the files to highlight other differences between them.

docker/DOCKER.md
```diff
@@ -0,0 +1,214 @@
...
+- `<distro>`: alpine - smaller footprint than debian
+- `<cli_ver>`: omitted - the latest available CLI version will be used
+
+## Basic Usage
+
+Here's a small example of running a single, detached container
+in the background and using `docker exec` to run balena CLI commands.
+
+```
+$ docker run --pull --detach --privileged --network host --name cli --rm -it balenalib/amd64-debian-balenacli /bin/bash
```
pdcastro, line 63: I had suggested adding `--pull` here, but it may have to be `--pull always`. Without argument, I get an error if the image does not yet exist locally:
```
$ docker rmi balenalib/amd64-debian-balenacli:1021344afddd361752852d5b4475a6042c767b66

$ docker run --pull --privileged --network host --name cli --rm -it balenalib/amd64-debian-balenacli:1021344afddd361752852d5b4475a6042c767b66
docker: Error response from daemon: No such image: balenalib/amd64-debian-balenacli:1021344afddd361752852d5b4475a6042c767b66.
See 'docker run --help'.
```
```
$ docker run --help
...
      --pull string    Pull image before running ("always"|"missing"|"never") (default "missing")
```
```
$ docker rmi balenalib/amd64-debian-balenacli:1021344afddd361752852d5b4475a6042c767b66

$ docker run --pull always --privileged --network host --name cli --rm -it balenalib/amd64-debian-balenacli:1021344afddd361752852d5b4475a6042c767b66
(success)
```
I am guessing that `--pull` without arguments is taken to mean `--pull never`.

docker/DOCKER.md
```diff
@@ -0,0 +1,214 @@
...
+- `<distro>` is the Linux distribution and is mandatory. Currently there are 2 distributions, namely `debian` and `alpine`.
+
+## Image Tags
+
+In the tags, all of the fields are optional, and if they are left out, they will default to their `latest` pointer.
+
+- `<cli_ver>` is the version of the balena CLI, for example, `12.40.2`, it can also be substituted for `latest`.
+
+To ensure that the latest available CLI image is used, it is also necessary
+to add the `--pull` flag to `docker run` commands. Otherwise, Docker will
```
pdcastro, line 40: `--pull always` (see other comment)

docker/DOCKER.md
```diff
@@ -0,0 +1,214 @@
...
+`balenalib/armv7hf-alpine-balenacli`
+
+- `<arch>`: armv7hf - suitable for running on a Raspberry Pi 3 for example
+- `<distro>`: alpine - smaller footprint than debian
+- `<cli_ver>`: omitted - the latest available CLI version will be used
+
+## Basic Usage
+
+Here's a small example of running a single, detached container
+in the background and using `docker exec` to run balena CLI commands.
```
pdcastro, line 60: NIT, to help novice users understand it the interplay between `docker run` and `docker exec` when running the CLI in this manner.

-------------
\## Basic Usage

The recommended basic usage consists of running a single, detached container
named "cli" in the background, with the `docker run` command:

```
$ docker run --name cli --detach --pull always --privileged --network host --rm -it balenalib/amd64-debian-balenacli /bin/bash
...
Status: Image is up to date for balenalib/amd64-debian-balenacli
3fd0fd2c86049c5c6f6dad665aef0e6a0ffc83003d73a3522924f844270dbde4

$ docker ps -a
CONTAINER ID   IMAGE                              NAMES
3fd0fd2c8604   balenalib/amd64-debian-balenacli   cli

$ docker logs cli
...
INFO[2021-03-10T21:14:48.826556111Z] Daemon has completed initialization
INFO[2021-03-10T21:14:48.838769027Z] API listen on /var/run/docker.sock
```

Then, use `docker exec` to run balena CLI commands in that container, as follows.

```
< docker exec command examples go here >
```

When the background container is no longer required, it can be stopped with:
```
$ docker stop cli
cli
```
-------------

pdcastro: Then again, I am having second thoughts about the usefulness of this "basic usage" with `docker run` + `docker exec`, mainly because we are not giving examples of how to share the filesystem between "host OS" and the CLI containers, and for sure users will be trying to run commands like `balena push`, especially if they take "basic usage" to mean "essential usage" (`balena push` is one of the most essential commands).

We would also have to add a note about the fact that `balena login` web authentication (that the CLI highlights as the recommended option) won't work at all. Which is OK given the container limitations, but a note would be needed.

How bad would it be to bind mount the user's homedir to `/root` in the container? Just tested and got `nvm` errors (script running from `.profile` / `.bash_profile` / `.bashrc` files). Probably not the best idea, but perhaps we could document bind mounting a specific folder for data sharing between host OS and container.

I seem to remember that `balena preload` does not work if the source image is loaded through a bind mount (it had to be copied to the container's internal filesystem). If so, we would have to document that too, otherwise users will be wasting time attempting and will bring the matter to support...

docker/DOCKER.md
klutchell: Looks like the default without arguments is actually `missing`, which also isn't what we want. I'll change it to `always`.

pdcastro: They document `default "missing"`, but the observed behaviour is that:
* No flag: behaves like "missing" (as documented)
* `--pull` provided without argument (which is undocumented usage): behaves like "never"

docker/DOCKER.md
klutchell: > We would also have to add a note about the fact that balena login web authentication (that the CLI highlights as the recommended option) won't work at all

Good point, this will never work even by copying the URL as the callback auth hook won't be able to reach the CLI. I'll add a note.

> How bad would it be to bind mount the user's homedir to /root in the container?

I wouldn't suggest doing this, since the container runs as root and your host OS user would end up losing access to some important files. We can show an example of a named volume to persist cache and credentials if it helps?

> I seem to remember that balena preload does not work if the source image is located in a bind mount (it had to be copied to the container's internal filesystem).

Preload does work in a bind mount, and I have an example farther down, but the absolute path to the image file must be the same from the perspective of the docker host and the CLI. So it only matters if using the host docker socket, and can be managed by mounting the path to the image on the host to an identical path in the container (eg. `-v $PWD:$PWD -w $PWD`).

pdcastro: Stepping back and thinking from the point of view of the end user... As an end user, I would like the DOCKER.md guide to show me how I am supposed to use the CLI's docker images for, say, at least the 15 "primary commands" listed with `balena help` (as opposed to the "additional commands" listed with `balena help -v`) -

```
$ balena help
...
PRIMARY COMMANDS
  login                                  login to balena
  push <applicationOrDevice>             start a build on the remote balenaCloud build servers, or a local mode device
  logs <device>                          show device logs
  ssh <applicationOrDevice> [service]    SSH into the host or application container of a device
  apps                                   list all applications
  app <nameOrSlug>                       display information about a single application
  devices                                list all devices
  device <uuid>                          show info about a single device
  tunnel <deviceOrApplication>           tunnel local ports to your balenaOS device
  preload <image>                        preload an app on a disk image (or Edison zip archive)
  build [source]                         build a project locally
  deploy <appName> [image]               deploy a single image or a multicontainer project to a balena application
  join [deviceIpOrHostname]              move a local device to an application on another balena server
  leave [deviceIpOrHostname]             remove a local device from its balena application
  scan                                   scan for balenaOS devices on your local network
```

It doesn't matter so much if we are using `docker run` + `docker exec`, or just `docker run`. I think that using `docker exec` was supposed to be a performance optimization (so that the docker daemon doesn't get started again and again) and a way of persisting the `login` information across multiple CLI commands. This is good, as long as the 15 primary commands can be used this way, and we have examples.

DOCKER.md is currently split between Basic Usage and Advanced Usage. I think it doesn't matter if we have one or two such sections and which section uses `docker exec`. The sections could be swapped, e.g. Basic Usage could use `docker run` only, and Advanced Usage could use `docker exec`, or one of the 2 sections could be deleted or shortened if they are redundant (meaning: no benefit over the alternative usage). Whichever section is titled Basic Usage, I think it should cover all 15 primary commands, with some "usability requirements" if possible:

- We should document that users have the option of using a host Docker Engine or a Docker Engine inside the CLI container, and how to switch between them. A specific note about the `deploy` command would be nice, as it deploys a pre-built image through the chosen Docker Engine.
- If possible, the project data for push/build/deploy should be outside the container, which implies a bind mount. An example should be provided of bind mounting.
- We should document the restrictions on `balena login` (that web auth doesn't work).
- We should document your explanation re preload that makes a lot of sense:
> Preload does work in a bind mount, but the absolute path to the image file must be the same from the perspective of the docker host and the CLI. So it only matters if using the host docker socket, and can be managed by mounting the path to the image on the host to an identical path in the container (eg. -v $PWD:$PWD -w $PWD).

Much of the above is already done / available, it's just a matter of moving things around so it's easier for users.

klutchell: My latest push just added some much better notes to each of the example commands, but I'm definitely missing some of the top 15 like `push`, `logs, `apps`, `devices`, `device`, `tunnel` (this one is tricky and requires some explanation), `join`, `leave`.

pdcastro: Some of them are trivial because it's just API queries -- apps, devices, device... Others have forms that take the IP address of a local device and will also probably work without problems: logs, join, leave. The trickier ones that require bind mounting are push/build and `deploy --build`. The `deploy` command without the `--build` flag will load a pre-build image using some Docker Engine. The `tunnel` command may not work at all and that's fine, as long as we add a note so that users won't spend time trying it and then come to support and say "it's broken!"

pdcastro: Sounds good! Avoiding duplication is good! 😁 
All the "purely API" commands could be combined as there's nothing special about them, like bind mounting or docker sockets... `apps | app | devices | device`

klutchell: I'm dropping the current exec "basic" usage example as the container definition needs to change for many commands. Instead, I'm providing two examples for each command, one with a named volume for credentials and one with a shell session and multiple commands.

eg.
### login

- <https://www.balena.io/docs/reference/balena-cli/#login>

The `balena login` command cannot be used with web authorization and a browser
when running in a container. Instead it must be used with `--token` or `--credentials`.

Notice that in this example we've used a named volume `balena_cache` to store credentials
for future runs of the image. This is optional but avoids having to run the login
command again for each of the other commands.

```bash
$ docker volume create balena_cache

# example 1: login with credentials
$ docker run --pull always --rm -it \
    -v balena_cache:/root/.balena \
    balenalib/amd64-debian-balenacli login --credentials --email johndoe@gmail.com --password secret

# example 2: login with token
$ docker run --pull always --rm -it \
    -v balena_cache:/root/.balena \
    balenalib/amd64-debian-balenacli login --token "..."
```

### app | apps

- <https://www.balena.io/docs/reference/balena-cli/#app-nameorslug>
- <https://www.balena.io/docs/reference/balena-cli/#apps>

```bash
# example 1: use existing credentials from login example
$ docker run --pull always --rm -it \
    -v balena_cache:/root/.balena \
    balenalib/amd64-debian-balenacli app myApp

# example 2: start a shell session and execute multiple commands
$ docker run --pull always --rm -it \
    balenalib/amd64-debian-balenacli bash

> balena login --credentials --email johndoe@gmail.com --password secret
> balena app myApp
> exit
```

pdcastro: It's looking good! 👍 
What do you think about running `dockerd` conditionally on an env var that was defined through  
`docker run --env DOCKERD=1`  
instead of checking for a bind mount of `docker.sock`? The reasoning being that most (?) commands don't use docker, like `login` and `apps`, and yet the Docker Engine starts verbosely every time. It would make some `docker run` command lines even longer than they already are, but any user who regularly wanted to use the CLI this way would for sure write a 2-line script or `alias` shell command to save typing.

Then for commands that do use docker like `build` and `preload`, the "2 examples" could include an example with a bind mount of `docker.sock` and no env var, and another example with `--env DOCKERD=1`.
Just a thought!

klutchell: It's a good thought, I'm on it!

.dockerignore
klutchell: I just noticed that I force pushed over these changes! Do you still have the .dockerignore and .gitignore locally?

klutchell: Nevermind, I found the orphaned commit and reapplied it!

docker/DOCKER.md
klutchell: Docker 19.09 is the first version to include the `--pull` option with `docker run` so we may want to recommend a manual pull command rather than a flag that doesn't exist for most versions in use today.

.dockerignore
pdcastro: I had myself removed my commit to test whether it was the cause of the build failures. It wasn't -- the build was failing with or without that commit, and I had not understood why. Let's see if the build passes now...

docker/docker-init.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+# install private ssh key if one is provided
+if [ -n "${SSH_PRIVATE_KEY}" ]
+then
+    echo "${SSH_PRIVATE_KEY}" | tr -d '\r' | ssh-add -
+fi
+
+# try to determine if an executable was provided or just args
+if [ "${1}" = "balena" ] || [ -x "${1}" ] || "${1}" -v >/dev/null 2>&1
```
klutchell, line 23: This isn't working as intended, I'm retesting it now with

```bash
# try to determine if an executable was provided or just balena args
if ! balena "${1}" --help >/dev/null 2>&1
then
    exec "$@"
else
    exec balena "$@"
fi
```

pdcastro: > if ! balena "${1}" --help >/dev/null 2>&1

hmm a disadvantage of this approach is that every CLI command results in invoking the CLI twice, which causes additional "unnecessary" delay and lag in CLI usage. I think the `--help` form also tracks usage with Amplitude.com, and sometimes this can add variable delays. Oh, at least you could disable this tracking with:

```
if ! BALENARC_NO_ANALYTICS=1 balena "${1}" --help >/dev/null 2>&1
```

But even disabling analytics, I am not fan of this approach of executing the CLI twice, because it is slow.

docker/DOCKER.md
pdcastro: Awesome job with the latest DOCKER.md documentation, Kyle! 💯  🥇 
It's looking great! 👌

docker/docker-init.sh
klutchell: Okay, let me think about it some more.

docker/DOCKER.md
```diff
@@ -0,0 +1,400 @@
...
+This will bind mount your current working directory into the container at the same absolute path.
+
+This bind mount is required so the CLI has access to your app sources.
+
+```bash
+$ docker run --pull always --rm -it -v "balena_data:/root/.balena" \
+    -v "$PWD:$PWD" -w "$PWD" \
+    balenalib/amd64-debian-balenacli bash
+
+> balena login --credentials --email "johndoe@gmail.com" --password "secret"
```
klutchell, line 137: I find the login command in every example to be really excessive. Do you think we can drop it and when users realize they aren't logged in they will scroll up to the corresponding login section?

docker/DOCKER.md
klutchell: Thank you! :blush:

docker/DOCKER.md
pdcastro: Yeah, I agree, no need to repeat the `balena login` line in several places. 👍

klutchell: The catch so far has been some balena commands are also valid executables in the system path, so just saying "is this cmd an executable file in PATH" will fall over with things like `ssh` and some others.

I'm trying to be too clever by supporting both args and commands in the CMD field, but I'm determined to make it work in a predictable way!

I was pretty close when I didn't check `[ -x "$(which "${1}") ]` and only checked `[ -x "${1} ]` but then it fails when providing `bash` instead of `/bin/bash`.

Is there a file in the source code where we could quickly grep whether something is a real balena command?

klutchell: Related, I'm going to limit the cases where the ssh-agent starts to only when the SSH_PRIVATE_KEY is provided and the agent is not already mapped from the host.

docker/docker-init.sh
klutchell: The catch so far has been some balena commands are also valid executables in the system path, so just saying "is this cmd an executable file in PATH" will fall over with things like `ssh` and some others.

I'm trying to be too clever by supporting both args and commands in the CMD field, but I'm determined to make it work in a predictable way!

I was pretty close when I didn't check `[ -x "$(which "${1}")" ]` and only checked `[ -x "${1}" ]` but then it fails when providing `bash` instead of `/bin/bash`. I might go back to this as it worked the best.

Is there a file in the source code where we could quickly grep whether something is a real balena command?

klutchell: Related, I'm going to limit the cases where the ssh-agent starts to only when the SSH_PRIVATE_KEY is provided and the agent is not already mapped from the host.

docker/DOCKER.md
pdcastro: Right. Docker v20 is already out, but because of the `preload` AUFS issue we still ask people to downgrade to Docker v18. I believe that removing `--pull always` from the `docker run` command lines will result in people running old versions of the CLI though. There is that note, currently under the Image Tags section:

> To ensure that the latest available CLI image is used, it is also necessary to add the `--pull always` flag to `docker run` commands. Otherwise, Docker will reuse a locally cached image, without checking for available updates.

How about we move that note from the Image Tags section to the Usage section, and reword it a bit, for example as follows:

> We recommend adding the `--pull always` flag to `docker run` commands, so that Docker checks for available updates of the balena CLI image (new image layers will only be downloaded if a new version is available). This is needed even if the `:latest` image tag is used. Note however that `--pull always` is only available with Docker v19 or later. If using an older version of Docker, run a separate `docker pull balenalib/<arch>-<distro>-balenacli` command.

And then we keep the `--pull always` flag in the `docker run` usage examples. If someone is using Docker v18 and `docker run --pull always` fail, they'll easily notice it and hopefully they'll find that note.

docker/docker-init.sh
pdcastro: > Is there a file in the source code where we could quickly grep whether something is a real balena command?

Yes, we do actually: `oclif.manifest.json`
```
{ ...
  "commands": {
    "apps": {
    ...
    "build": {
    ...
    "deploy": {
    ...
```
It's not committed to the git repo, but produced by `npm run build:fast`. We could extract the list of command names during the image build (in the Dockerfile), to even reduce JSON parsing time at runtime (it's a large file, 123KB, as it contains all the flags as well...)

klutchell: That `oclif.manifest.json` file is handy, but this is all getting a bit messy. What do you think about these three options:

### 1. Try to determine if command is executable or args

```bash
if [ "${1}" = "balena" ] || [ -x "${1}" ]
then
    exec "$@"
else
    exec balena "$@"
fi
```

- shell can be started with `/bin/bash`
- the default CMD could be `/bin/bash` if we assume a container shell is the common use case?
- the default CMD could be `--help` if we assume a single balena command is the common use case?
- failing to provide the full path to an exec will treat it as an arg (eg. `bash` => `balena bash`)

### 2. Assume command is executable

```bash
exec "$@"
```

- shell can be started with `bash`
- one-liner docker run commands require `balena ` prefix
- the default CMD could be `bash` if we assume a container shell is the common use case?
- the default CMD could be `balena --help` if we assume a single balena command is the common use case?

### 3. Assume command is args

```bash
exec "balena $@"
```

- **shell cannot be started without skipping dockerd and ssh-agent init steps via custom entrypoint**
- one-liner docker run commands do not require `balena ` prefix
- the default CMD could be `--help` if we assume a single balena command is the common use case?

docker/DOCKER.md
klutchell: My version of Docker 19.03 fails as well, the command didn't release until 19.09. What about putting the `docker pull balenalib/amd64-debian-balenacli` before every run command?

pdcastro: Shame. Adding the `docker pull` command in every example would be a bit noisy and repetitive. In this case, how about we simply move that note from the Image Tags section to a new section (so it gets more prominence), just before the Usage section, along the lines of:

```
## Keeping the CLI image up to date

Please note that using the `:latest` tag is not enough to keep the image up to date,
because Docker will reuse a locally cached image. To update the image to the latest
version, run:

` ` `bash
$ docker pull balenalib/<arch>-<distro>-balenacli
` ` `

Replacing `<arch>` and `<distro>` with the image architecture and distribution as
described earlier.

If you are using Docker v19.09 or later, you can also add the `--pull always` flag to
`docker run` commands, so that Docker automatically checks for available updates
(new image layers will only be downloaded if a new version is available).
```

.dockerignore
klutchell: Still seeing the build failures, but I can't figure out from the logs what's wrong. They build fine locally?

docker/docker-init.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
pdcastro, line 1: `bash` instead of `sh` in order to use `[[ =~ ]]` string matching later in the file.

docker/debian/Dockerfile
```diff
@@ -31,10 +31,12 @@ RUN install_packages avahi-daemon ca-certificates docker.io jq openssh-client
 
 COPY docker/docker-init.sh init.sh
 
-RUN chmod +x ./init.sh
+RUN CLI_CMDS=$(jq -r '.commands | keys | map(.[0:index(":")]) | unique | join("\\ ")' < oclif.manifest.json); \
+    sed -ie "s/CLI_CMDS=\"help\"/CLI_CMDS=\"help\\ ${CLI_CMDS}\"/" init.sh && \
+    chmod +x init.sh
```
pdcastro, line 36: Yes, the expressions are a bit ugly. :-) But only 2 lines of that, and it seems to work.

docker/docker-init.sh
pdcastro: I've added a commit that attempts to deliver "option zero", the one you described as getting a bit messy. :-)  Do you think it works? I am happy for you to modify it!

pdcastro: It's like option 1, but better discriminates when the argument is something like `bash` or `whoami`, treating the latter as a CLI command but not the former, giving priority to CLI commands when there is a clash.

.dockerignore
pdcastro: Yes, the build works on my laptop as well. By the way, As of Docker Desktop v3.2.1 (macOS), Docker Engine v20.10.5, the default builder is buildkit, and the output of `docker build` is quite different than it used to be. So now my machine is quite different from the CI. I don't particularly like that it is different, but I get to see what many CLI end users will be experiencing as well.

I don't know what's wrong, but we can merge the PR until we figure it out... If in doubt, perhaps you could create a separate PR where you push the last known-to-succeed commit, if you still have it - or something really basic, like a single line Dockerfile with just a FROM instruction, which might succeed or fail... Then you could ping the CI team (Giovanni...) and ask - "what's changed in the CI? This build used to pass!"

docker/alpine/Dockerfile
```diff
@@ -31,8 +31,8 @@ RUN install_packages avahi bash ca-certificates docker jq openssh
 
 COPY docker/docker-init.sh init.sh
 
-RUN CLI_CMDS=$(jq -r '.commands | keys | map(.[0:index(":")]) | unique | join("\\ ")' < oclif.manifest.json); \
-    sed -ie "s/CLI_CMDS=\"help\"/CLI_CMDS=\"help\\ ${CLI_CMDS}\"/" init.sh && \
+# write CLI commands to a file to reference on init
```
klutchell, line 34: While this sed command works, it's nice to have the list of CLI commands in a separate file so we don't have to be concerned with breaking the sed matching when updating the init script in the future. Just a preference, your method worked well too!

docker/docker-init.sh
```diff
@@ -15,12 +15,10 @@ then
     echo "${SSH_PRIVATE_KEY}" | tr -d '\r' | ssh-add -
 fi
 
-# space-separated list of balena CLI commands (filled in through `sed`
-# in a Dockerfile RUN instruction)
-CLI_CMDS="help"
-
-# try to determine if command is a balena CLI command
-if [[ " ${CLI_CMDS} " =~ " ${1} " ]]
```
klutchell: Shellcheck wasn't happy with this line. It seemed to work anyway when I tested it though?
https://github.com/koalaman/shellcheck/wiki/SC2076

docker/docker-init.sh
```diff
@@ -15,12 +15,10 @@ then
...
+# treat the provided command as a balena CLI arg...
+# 1. if the first word is present in cli_cmds matching word boundaries
+# 2. OR if the first character is a hyphen (eg. -h or --help)
```
klutchell, line 20: Added support to detect global flags if the first character is a hyphen. Shouldn't be required but nice to have I think.

.dockerignore
klutchell: I pushed a change to remove the base image pinning, which is one of the changes I made early in the week.

docker/docker-init.sh
pdcastro: > SC2076: _it'll match literally rather than as a regex_

That's how I meant it to work, nothing wrong with it. Removing the quotes as per their advice would introduce a bug.

docker/alpine/Dockerfile
pdcastro: I had also considered it, but it means the extra file loading and grep'ing at runtime, versus sed at build time. I had opted to shift the cost to build time. It's very minor though, either way is fine.

.dockerignore
klutchell: Found the issue, some of the images don't exist with that date tag. I'll move the base image definitions to `.resinci.yml` so it can be unique per arch/distro and make it easier to automate updating in the future.

docker/docker-init.sh
klutchell: That's what I thought, odd behaviour from shellcheck.

docker/alpine/Dockerfile
klutchell: Good call, my change was missing the `help` command anyway so I just reverted to what you added.

.dockerignore
klutchell: The latest v12 available for i386-debian-node is 12.16.3, but the rest are at various releases of 12.19.1.
Do you think we should align them all on 12.16.3 (if possible) or whatever the latest v12 is for the respective platforms?

pdcastro: It is already the case that a single CLI release / version uses different versions of Node.js bundled in the different installers: GUI installers for Windows and macOS, and the standalone zip packages. I don't readily see the benefit of having all Docker images of a CLI release use the same version of Node.js (if was even possible), but also it's not a big deal either way. So it's really up to you... If it was me making the choice, I would probably have each Docker image use the latest available base image image version for that arch-distro combination.

pdcastro: Having each Docker image use the latest available base image for that arch-distro combination is also probably slightly easier to automate, as it avoids having to find the latest common version -- which may not even exist for all images. And if one of the base images was lagging behind the other for any reason -- e.g. one image in Node v10 and the other images in Node v14 -- we would not want to penalise all other images because of that one. So I am actually starting to see a reason to have each image use the latest available base image for its arch-distro combination.

klutchell: Agreed, that's what I was thinking as well. Nice to see the builds working now!

docker/DOCKER.md
```diff
@@ -0,0 +1,410 @@
...
+## Volumes
+
+Volumes can be used to persist data between instances of the CLI container, or to share
+files between the host and the container.
+In most cases these are optional, but some examples will highlight when volumes are required.
+
+- `-v "balena_data:/root/.balena"`: persist balena credentials and downloads between instances
+- `-v "docker_data:/var/lib/docker"`: persist cache between instances when using Docker-in-Docker (requires `-e "DOCKERD=1"`)
+- `-v "$PWD:$PWD" -w "$PWD"`: bind mount your current working directory in the container to share app sources or balenaOS image files
+- `-v "${SSH_AUTH_SOCK}:/ssh-agent"`: bind mount your host ssh-agent socket with preloaded SSH keys
```
pdcastro, line 62: This worked for me on Windows, but not on macOS. Maybe related to my system's permissions.
```
$ echo "${SSH_AUTH_SOCK}:/ssh-agent"
/private/tmp/com.apple.launchd.CJQ4XK0H8e/Listeners:/ssh-agent
```
Even when testing with `sudo` and `--privileged`, it would fail (same as when not using `sudo` or `--privileged`):
```
$ sudo docker run -it --privileged -v "${SSH_AUTH_SOCK}:/ssh-agent" -v balena_data:/root/.balena balenalib/amd64-alpine-balenacli:d40af7fdcff1943ffe488c3067400dbb57f8645d /bin/bash

bash-5.0# ssh-add -l
Error connecting to agent: Connection refused

bash-5.0# env
SSH_AUTH_SOCK=/ssh-agent
...

bash-5.0# ls -la /ssh-agent
srw-rw-rw-    1 root     root             0 Feb 15 01:58 /ssh-agent
```
But on the "host OS" (my laptop):

```
$ env | grep SSH
SSH_AUTH_SOCK=/private/tmp/com.apple.launchd.CJQ4XK0H8e/Listeners

$ ls -la /private/tmp/com.apple.launchd.CJQ4XK0H8e/Listeners
srw-rw-rw-  1 paulo  wheel  0 15 Feb 01:58 /private/tmp/com.apple.launchd.CJQ4XK0H8e/Listeners

$ ssh-add -l
2048 SHA256:0XZ70WNycHiGhmVsGjqK2ajiuUvRPG1xoUaI/J/s2Uw /Users/paulo/.ssh/id_rsa (RSA)
```

We don't need to change DOCKER.md, unless you have suggestions.

pdcastro (commented):
I think the only thing left now (before merging) is to squash some of the 7 commits, only because the commits become bullet points in the CHANGELOG which is currently the main source of "what's new" in the CLI for end users. I leave it up to you to decide how many. The main concern is that it should make sense from end user's point of view.

(Side note: for the balena-cli repo, I think it would be good to have the ability of decoupling commits from the CHANGELOG. It's a discussion to have in the balenaci flow I think.)

I am also leaving a couple more review comments, but it's minor -- I would have approved the PR if it wasn't for the commit / CHANGELOG issue.

.dockerignore
```diff
@@ -0,0 +1,43 @@
...
+oclif.manifest.json
+package-lock.json
+resinrc.yml
+tmp
+
+# Docker
+docker-compose*.yml
+.dockerignore
+**/Dockerfile
+**/DOCKER.md
```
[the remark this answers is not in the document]
klutchell: I don't mind putting them all back in, but I found it really annoying that the entire image would rebuild when I was making minor changes to documentation. Because these files get included in the context, and are copied in the first build stage, adding a character to DOCKER.md invalidated my entire image cache and would rebuild the CLI from source :cry: 

I can remove them from the ignore file now though since we are nearly done testing!

docker/DOCKER.md
klutchell: I don't have a Mac to test on, but is it possible that `/private/tmp/` isn't shared with Docker Desktop by default?

.dockerignore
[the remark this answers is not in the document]
pdcastro: Oh right. Eventually we could replace `COPY . .` with "a list of the folders and files that matter" and maybe multiple COPY instructions at strategic positions in the Dockerfiles. A list of files and folders that matter exists in `package.json`, `files` section.

pdcastro (approved):

PR #2161, opened 2021-01-15
Handle 'push' edge case with application access

srlowe:
@pdcastro , this includes the changes we discussed on the other PR, but something strange happened with the git rebase, so I created a new PR.

// * Temporary fix *
// When doing `$expand organization` on a public app (when not the owner)
// a partial document is returned without any organization data.
// So e.g. `balena push balena-sound` will break this commands logic.
// The balena cloud builder api will soon change to accept application slugs
// instead of `owner, app`, after which we will not need to do the expand

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/commands/push.ts
+++ b/lib/commands/push.ts
@@ -359,10 +359,26 @@ export default class PushCmd extends Command {
 					$select: ['handle'],
 				},
 			},
-			$select: ['app_name'],
+			$select: ['app_name', 'slug'],
 		})) as Application & {
 			organization: [Organization];
 		};
+
+		// * Temporary fix *
+		// When doing `$expand organization` on a public app (when not the owner)
+		// a partial document is returned without any organization data.
+		// So e.g. `balena push balena-sound` will break this commands logic.
+		// The balena cloud builder api will soon change to accept application slugs
+		// instead of `owner, app`, after which we will not need to do the expand
+		//
+		// Users should not be pushing to public apps anyway, so for now catch this situation:
+		if (!application.organization[0]) {
+			throw new ExpectedError(stripIndent`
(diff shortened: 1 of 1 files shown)


mbalamat (approved):

lib/commands/push.ts
```diff
@@ -359,11 +359,27 @@ export default class PushCmd extends Command {
...
+		// When doing `$expand organization` on a public app (when not the owner)
+		// a partial document is returned without any organization data.
+		// So e.g. `balena push balena-sound` will break this commands logic.
+		// The balena cloud builder api will soon change to accept application slugs
+		// instead of `owner, app`, after which we will not need to do the expand
+		//
+		// Users should not be pushing to public apps anyway, so for now catch this situation:
+		if (!application.organization[0]) {
+			throw new ExpectedError(stripIndent`
+				You do not have permissions to push to application ${application.slug}.
```
pdcastro, line 377: Does `${application.slug}` work then? (Instead of `application.app_name`.) In the previous PR (#2151), you had explained:
> https://github.com/balena-io/balena-cli/pull/2151#discussion_r555148382
> In these cases the partial document you get back from the API doesn't contain the slug. (Though may be a case of changing the query).

pdcastro (changes requested):
"Requesting changes" just for clarification of a question (temporarily preventing merge just in case...)

lib/commands/push.ts
srlowe: Yeah, I just had to add `slug` to the selected fields of the query.

pdcastro (approved):

Issue #2042, opened 2020-09-22
tunnel command uses unencrypted connection to backend server


pdcastro:
The `balena tunnel` command establishes a TCP connection to a CONNECT HTTP proxy server at `vpn.balena-cloud.com:3128` with 'basic' HTTP authentication. This connection is not currently encrypted, making it subject to network sniffing - a security issue. (The connection is encrypted between the balenaCloud backend and the device, but not between the CLI and the balenaCloud backend.) The connection should be encrypted end to end.

This issue was originally kindly reported by a user in this forum thread:
* https://forums.balena.io/t/balena-tunnel-connections-are-not-encrypted/191826/
* Internal thread: https://jel.ly.fish/9ee8724d-50d6-40a8-8ccf-3ee7911cf6be

Proposed solutions:
* Enable (and require) TLS encryption on the existing TCP connection.
* Alternatively, use 'ssh' to the device in combination with `/usr/bin/nc`, as suggested in the forum thread linked above.
* Internal discussion: https://www.flowdock.com/app/rulemotion/resin-tech/threads/nOheKZaBwyMwC2PmzfPw0NxYEMl

Related `open-balena-vpn` issue: https://github.com/balena-io/open-balena-vpn/issues/149

mcamou:
Hi, do you have a rough timeline when this might be fixed? Right now we're using the `socat`/`ssh` hack but it's not working reliably especially when we need to e.g. transfer binary files via a REST interface. I'm really not comfortable with my Balena token flying around the Internet unencrypted.

pdcastro:
Thanks for chasing progress on this issue, @mcamou. 👍   I have re-raised it with the backend team. I gather that a decision was made to enable TLS on port `vpn.balena-cloud.com:3128`, but no one in team had picked it up -- that is, until today when @zwhitchcox bravely took it on him 👏    to implement it soon (at least on the backend, then we can amend the CLI to use it).

Thank you again for raising this issue.

pdcastro:
@mcamou, starting with CLI version 12.38.5, the `balena tunnel` commands connects to a new balenaCloud endpoint:
* New: `tunnel.balena-cloud.com`, port 443, using TLS (CLI v12.38.5 and later)
* Old: `vpn.balena-cloud.com`, port 3128, no encryption (CLI v12.38.4 and earlier)

[...]

PR #2163, opened 2021-01-19
tls: Use TLS for tunnel connection

richbayliss:
Switch to using the exposed tunnelUrl and TLS for making
tunnels to the device, to improve security.

Connects-to: #2042 
Change-type: patch

### DIFF
--- a/lib/utils/tunnel.ts
+++ b/lib/utils/tunnel.ts
@@ -15,9 +15,13 @@ limitations under the License.
 */
 import type { BalenaSDK } from 'balena-sdk';
 import { Socket } from 'net';
+import * as tls from 'tls';
 import { TypedError } from 'typed-error';
+import { ExpectedError } from '../errors';
 
 const PROXY_CONNECT_TIMEOUT_MS = 10000;
+
+class TunnelServerNotTrustedError extends ExpectedError {}
 
 class UnableToConnectError extends TypedError {
 	public status: string;
@@ -42,17 +46,17 @@ export const tunnelConnectionToDevice = (
 	sdk: BalenaSDK,
 ) => {
 	return Promise.all([
-		sdk.settings.get('vpnUrl'),
+		sdk.settings.get('tunnelUrl'),
 		sdk.auth.whoami(),
 		sdk.auth.getToken(),
-	]).then(([vpnUrl, whoami, token]) => {
+	]).then(([tunnelUrl, whoami, token]) => {
 		const auth = {
 			user: whoami || 'root',
 			password: token,
 		};
 
 		return (client: Socket): Promise<void> =>
-			openPortThroughProxy(vpnUrl, 3128, auth, uuid, port)
+			openPortThroughProxy(tunnelUrl, 443, auth, uuid, port)
 				.then((remote) => {
 					client.pipe(remote);
 					remote.pipe(client);
@@ -96,30 +100,41 @@ const openPortThroughProxy = (
 	}
 
 	return new Promise<Socket>((resolve, reject) => {
-		const proxyTunnel = new Socket();
-		proxyTunnel.on('error', reject);
-		proxyTunnel.connect(proxyPort, proxyServer, () => {
-			const proxyConnectionHandler = (data: Buffer) => {
-				proxyTunnel.removeListener('data', proxyConnectionHandler);
-				const [httpStatus] = data.toString('utf8').split('\r\n');
-				const [, httpStatusCode, ...httpMessage] = httpStatus.split(' ');
-
-				if (parseInt(httpStatusCode, 10) === 200) {
-					proxyTunnel.setTimeout(0);
-					resolve(proxyTunnel);
-				} else {
+		const proxyTunnel = tls.connect(
+			proxyPort,
+			proxyServer,
+			{ servername: proxyServer }, // send the hostname in the SNI field
+			() => {
+				if (!proxyTunnel.authorized) {
+					console.error('Unable to authorize the tunnel server');
 					reject(
-						new UnableToConnectError(httpStatusCode, httpMessage.join(' ')),
+						new TunnelServerNotTrustedError(proxyTunnel.authorizationError),
 					);
+					return;
 				}
-			};
 
-			proxyTunnel.on('timeout', () => {
-				reject(new RemoteSocketNotListening(devicePort));
-			});
-			proxyTunnel.on('data', proxyConnectionHandler);
-			proxyTunnel.setTimeout(PROXY_CONNECT_TIMEOUT_MS);
-			proxyTunnel.write(httpHeaders.join('\r\n').concat('\r\n\r\n'));
-		});
+				proxyTunnel.once('data', (data: Buffer) => {
+					const [httpStatus] = data.toString('utf8').split('\r\n');
+					const [, httpStatusCode, ...httpMessage] = httpStatus.split(' ');
+
+					if (parseInt(httpStatusCode, 10) === 200) {
+						proxyTunnel.setTimeout(0);
+						resolve(proxyTunnel);
+					} else {
+						reject(
+							new UnableToConnectError(httpStatusCode, httpMessage.join(' ')),
+						);
+					}
+				});
+
+				proxyTunnel.on('timeout', () => {
+					reject(new RemoteSocketNotListening(devicePort));
+				});
+
+				proxyTunnel.setTimeout(PROXY_CONNECT_TIMEOUT_MS);
+				proxyTunnel.write(httpHeaders.join('\r\n').concat('\r\n\r\n'));
+			},
+		);
+		proxyTunnel.on('error', reject);
 	});
 };
--- a/package.json
+++ b/package.json
@@ -203,7 +203,7 @@
     "balena-release": "^3.0.0",
     "balena-sdk": "^15.20.0",
     "balena-semver": "^2.3.0",
-    "balena-settings-client": "^4.0.5",
+    "balena-settings-client": "^4.0.6",
     "balena-settings-storage": "^7.0.0",
     "balena-sync": "^11.0.2",
     "bluebird": "^3.7.2",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2760,9 +2760,9 @@
       }
     },
     "balena-settings-client": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/balena-settings-client/-/balena-settings-client-4.0.5.tgz",
-      "integrity": "sha512-w1SWIQYViMP51PYnPvbwgGavipkBv8wbRj1ISjPYZ5M45oEVRcktDfix8c3xOlWl+vWqW8aA4L8BjhqnxhAvRQ==",
+      "version": "4.0.6",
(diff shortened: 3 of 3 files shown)


lib/utils/tunnel.ts
```diff
@@ -96,11 +97,14 @@ const openPortThroughProxy = (
...
-			const proxyConnectionHandler = (data: Buffer) => {
-				proxyTunnel.removeListener('data', proxyConnectionHandler);
+		const proxyTunnel = tls.connect(proxyPort, proxyServer, {}, () => {
+			if (!proxyTunnel.authorized) {
+				reject(proxyTunnel.authorizationError);
+				return;
+			}
+
+			const proxyDataHandler = (data: Buffer) => {
+				proxyTunnel.removeListener('data', proxyDataHandler);
```
pdcastro, line 107: Does this `removeListener` call do the same as using `proxyTunnel.once(...)` instead of `proxyTunnel.on(...)`?

lib/utils/tunnel.ts
```diff
@@ -114,12 +118,13 @@ const openPortThroughProxy = (
 				}
 			};
 
+			proxyTunnel.on('data', proxyDataHandler);
 			proxyTunnel.on('timeout', () => {
```
pdcastro, line 122: I know it was existing code, but it "feels uncomfortable" that the timeout handler is added inside the callback that gets called after a connection is established or throws an error - e.g. after it has timed out. Though I guess event handlers are still called even if added _after_ the event was produced. Would it be clearer to move both the `'timeout'` and `'data'` handlers outside the callback, alongside the `'error'` handler, so that the event handlers are added before the events are produced?

.versionbot/CHANGELOG.yml
```diff
@@ -1,3 +1,14 @@
+- commits:
```
pdcastro, line 1: These extra files may be just a matter of re-fetching the master branch and rebasing, no?

lib/utils/tunnel.ts
```diff
@@ -96,11 +97,14 @@ const openPortThroughProxy = (
 	}
 
 	return new Promise<Socket>((resolve, reject) => {
-		const proxyTunnel = new Socket();
-		proxyTunnel.on('error', reject);
-		proxyTunnel.connect(proxyPort, proxyServer, () => {
-			const proxyConnectionHandler = (data: Buffer) => {
-				proxyTunnel.removeListener('data', proxyConnectionHandler);
+		const proxyTunnel = tls.connect(proxyPort, proxyServer, {}, () => {
+			if (!proxyTunnel.authorized) {
```
pdcastro, line 101: Is this test necessary? The documentation says that the `rejectUnauthorized` property defaults to `true`:
> https://nodejs.org/docs/latest-v10.x/api/tls.html#tls_tls_connect_options_callback
> `rejectUnauthorized` <boolean> If not `false`, the server certificate is verified against the list of supplied CAs. An `'error'` event is emitted if verification fails; `err.code` contains the OpenSSL error code. **Default:** `true`.

The documentation mentions a "list of supplied CAs", but it also applies to the ``tlsSocket.authorized`` property:

> https://nodejs.org/docs/latest-v10.x/api/tls.html#tls_tlssocket_authorized
> Returns true if the peer certificate was signed by one of the CAs specified when creating the tls.TLSSocket instance, otherwise false.

> https://nodejs.org/docs/latest-v10.x/api/tls.html#tls_event_secureconnection
> The tlsSocket.authorized property is a boolean indicating whether the client has been verified by one of the supplied Certificate Authorities for the server. If tlsSocket.authorized is false, then socket.authorizationError is set to describe how authorization failed. Note that depending on the settings of the TLS server, unauthorized connections may still be accepted.

I assume that in the absence of an explicit list of CAs, it uses Node's default / system default / environment variables.

lib/utils/tunnel.ts
richbayliss: @pdcastro actually yes, it should, probably a cleaner way to do it also 👍🏻

lib/utils/tunnel.ts
richbayliss: It's done this way because we need different timeouts based on the stage of the connection. It takes time for the connect proxy to establish the link and so the timeout is higher before the initial data is sent, then we set it to a normal value.

.versionbot/CHANGELOG.yml
richbayliss: I pulled clean into a fresh laptop... idk

lib/utils/tunnel.ts
richbayliss: ...from what I read, I think it is.

> The 'secureConnect' event is emitted after the handshaking process for a new connection has successfully completed. The listener callback will be called regardless of whether or not the server's certificate has been authorized. It is the client's responsibility to check the tlsSocket.authorized property to determine if the server certificate was signed by one of the specified CAs. If tlsSocket.authorized === false, then the error can be found by examining the tlsSocket.authorizationError property. If ALPN was used, the tlsSocket.alpnProtocol property can be checked to determine the negotiated protocol.

lib/utils/tunnel.ts
pdcastro: OK. The `setTimeout()` calls would stay where they are, just moving the `on('timeout', ...)` handler setup. But if it works, it works. :-)

lib/utils/tunnel.ts
```diff
@@ -15,10 +15,13 @@ limitations under the License.
 */
 import type { BalenaSDK } from 'balena-sdk';
 import { Socket } from 'net';
+import * as tls from 'tls';
 import { TypedError } from 'typed-error';
 
 const PROXY_CONNECT_TIMEOUT_MS = 10000;
 
+class TunnelServerNotTrustedError extends TypedError {}
```
pdcastro, line 23: As coded, `new TunnelServerNotTrustedError(...)` will cause the CLI to report this error to Sentry.io. I assume that such errors do not represent CLI issues that need fixing, rather represent issues in the user's local network -- which would be reasonable for users to bring to support attention, at which point support would investigate the usual Deep Packet Inspection, SSL interception issues in the local network. If you agree with this reasoning, then the way to avoid the error being reported to Sentry.io would be to extend `ExpectedError`:
```
import { ExpectedError } from '../errors';

class TunnelServerNotTrustedError extends ExpectedError {}
```

richbayliss: What happens to the UX, the same?

pdcastro: Here's how the global error handler treats the error depending on whether or not it is "expected":
https://github.com/balena-io/balena-cli/blob/v12.38.3/lib/errors.ts#L247-L262

richbayliss: OK, so UX is the same.

I leave it to you @pdcastro to make that call then; personally I would rather see if people are hitting issues, and at what frequency vs good connections, but this is your ship dude 😁

pdcastro: It depends on what `TunnelServerNotTrustedError` means. My view is that an error reported to Sentry.io is something that we need to fix (and after fixed, it stops being reported). Two extremes are:
* If the user types a command-line option that doesn't exist, e.g. `--nono`, this definitely should not be reported to Sentry.io. It's just noise and counts towards Sentry.io's monthly report limit and max API rate limit that then prevents "legitimate" errors from being reported.
* Properly unexpected and unhandled errors (excluding the catch-all global handler) such as "TypeError" of "startsWith does not exist on undefined" should definitely be reported and it should definitely be fixed - typically a logic error in source code, and often one that blocks users / renders the command unusable.

If `TunnelServerNotTrustedError` means a problem in the user's local network (SSL interception), then it's not normally something actionable by balena, and becomes noise in Sentry.io -- noise that in the past resulted in CLI pull requests to replace `Error` or `TypedError` with `ExpectedError`, so that it stops being reported. However, if we had the view that `TunnelServerNotTrustedError` signified something actionable such as, say, the server certificate has expired and we need to update it, then we'd want it reported in Sentry.io.  But dev ops takes care of replacing certificates when they expire, especially at the ELB/NLB level.

`TunnelServerNotTrustedError` feels like a grey area. We will be testing that `balena tunnel` works before the PR is merged, and on balance I don't think it should be reported to Sentry.io. If it's for me to decide, it's an `ExpectedError`. :-)  (I also take into account the fact that there are plenty of bugs and longstanding issues reported by users themselves, let alone Sentry.io, which are known and in the queue to be fixed, so we don't need additional reports of things that most likely aren't even actionable...)

package.json
```diff
@@ -203,7 +203,7 @@
     "balena-release": "^3.0.0",
     "balena-sdk": "^15.20.0",
     "balena-semver": "^2.3.0",
-    "balena-settings-client": "^4.0.5",
+    "balena-settings-client": "^4.0.6",
```
pdcastro, line 206: Good, but the PR is missing the respective changes to `npm-shrinkwrap.json`. Running `npm install` results in the following changes on my laptop:
```
diff --git a/npm-shrinkwrap.json b/npm-shrinkwrap.json
index b970547c..e29661ac 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2760,9 +2760,9 @@
       }
     },
     "balena-settings-client": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/balena-settings-client/-/balena-settings-client-4.0.5.tgz",
-      "integrity": "sha512-w1SWIQYViMP51PYnPvbwgGavipkBv8wbRj1ISjPYZ5M45oEVRcktDfix8c3xOlWl+vWqW8aA4L8BjhqnxhAvRQ==",
+      "version": "4.0.6",
+      "resolved": "https://registry.npmjs.org/balena-settings-client/-/balena-settings-client-4.0.6.tgz",
+      "integrity": "sha512-bB14Zvg1N6t7XXPJqZs48SajgTuk2WTMm2AnxcOfoIQ2d/Lh0RsEGxD9toF2v+WhF2Ip4u7ko5tKlCr2kFddXA==",
       "requires": {
         "@resin.io/types-hidepath": "1.0.1",
         "@resin.io/types-home-or-tmp": "3.0.0",
```

richbayliss: wasnt sure it was needed, added.

pdcastro:
> requested a review from pdcastro

Looks good 👍  (I cannot approve draft PRs though, as versionbot dismisses such reviews.)

pdcastro:
> marked this pull request as ready for review

OK, 😀   now I cannot approve it because approval means merge / release, and `tunnel.balena-cloud.com:443` is not yet available in production, and I think openBalena also has not yet released `tunnel.my-openbalena-domain.com:443`. When it's available everywhere, we can merge this PR. Sounds right?

richbayliss:
> > marked this pull request as ready for review
> 
> OK,    now I cannot approve it because approval means merge / release, and `tunnel.balena-cloud.com:443` is not yet available in production, and I think openBalena also has not yet released `tunnel.my-openbalena-domain.com:443`. When it's available everywhere, we can merge this PR. Sounds right?

Yep @pdcastro 👍🏻 I'll put the VB draft label on to keep it from merging.

pdcastro (approved):

pdcastro:
> I'll put the VB draft label on to keep it from merging

Versionbot still dislikes approval :-)

pdcastro:
@balena-ci rebase

pdcastro (approved):

Issue #2165, opened 2021-01-21
livepush: CLI exits unexpectedly (no error) with Node.js v14


pdcastro:
As originally reported in a [flowdock thread](https://www.flowdock.com/app/rulemotion/resin-starters/threads/k2KO6r6D4ep5CyKC_Sf-DaU01Po) (private):

> livepushing my BoB:
```
[Debug]   Found project types:
[Debug]       admin: External image
[Debug]       builder: External image
[Debug]       db: External image
[Debug]       haproxy: External image
[Debug]       s3: External image
[Debug]       redis: External image
[Debug]       balena-mdns-publisher: External image
[Debug]   Probing remote daemon for cache images
[Debug]   Using 586 on-device images for cache...
[Debug]   Starting builds...

~/balena/balena-on-balena master* 1m 5s
```
> basically my push just craps out with no error/log. there's also nothing supervisor-side. this is the latest version 12.38.3 and the latest supervisor 12.3.0. i've also commented out nearly all of my services, as you can tell from the logs above

This issue only applies to the advanced `npm` installation of balena CLI, where the system installation of Node.js is used. It does not apply to the standalone zip package (all platforms) or the GUI installers (Windows and Mac), as these options currently use a built-in Node.js v12.

Possibly related to #1847 or #1878 (maybe not).

**Known workarounds:**
* Use the standalone zip package (all platforms)
* Use the GUI installer (Windows or macOS)
* If using the `npm` installation, downgrade the Node.js to v10 or v12, instead of v14.

LucianBuzzo:
@pdcastro Do you recommend switching to node v12 for a better `balena push local` experience?

LucianBuzzo:
I'm developing on an intel NUC (Os: balenaOS 2.68.1+rev1, super 12.3.0) and my main workstation is windows 10 running WSL 1.
I was having a lot of issues with livepush aborting without an error message or segfaulting. I wasn't able to ctrl+c the build and restart it without it failing and the only way to get it working again would be to disable then re-enable local mode. 
I switched from node 14 to node 12 and it is now working flawlessly 👍

StefKors:
@pdcastro is this the same as https://github.com/balena-io/balena-supervisor/issues/1568 ?

pdcastro:
> Do you recommend switching to node v12 for a better `balena push local` experience?

[...]

Issue #1990, opened 2020-08-11
`os configure` crashes with "Illegal instruction: 4" (npm install with Node v14)


pdcastro:
```
$ balena version -a
balena-cli version "12.14.4"
Node.js version "14.7.0"

$ which balena
/Users/paulo/.nvm/versions/node/v14.7.0/bin/balena

$ balena os configure my.img -a myApp
? Network Connection ethernet

#
# Fatal error in , line 0
# Check failed: result.second.
#
#
#
#FailureMessage Object: 0x7ffeefbfd6d0
 1: 0x100111772 node::NodePlatform::GetStackTracePrinter()::$_3::__invoke() [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
 2: 0x100fe4552 V8_Fatal(char const*, ...) [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
 3: 0x1004c17b4 v8::internal::GlobalBackingStoreRegistry::Register(std::__1::shared_ptr<v8::internal::BackingStore>) [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
 4: 0x1001f8336 v8::ArrayBuffer::GetBackingStore() [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
 5: 0x10007fa3b node::Buffer::Data(v8::Local<v8::Value>) [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
 6: 0x1000ad9a0 node::fs::Read(v8::FunctionCallbackInfo<v8::Value> const&) [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
 7: 0x1002572b8 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
 8: 0x10025684c v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/Users/paulo/.nvm/versions/node/v14.7.0/bin/node]
[...]

PR #2166, opened 2021-01-21
Update supported Node.js version range in package.json (<13.0.0)

pdcastro:
Connects-to: #2079, #1990, #2165

Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
   "author": "Juan Cruz Viotti <juan@balena.io>",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.20.0"
+    "node": ">=10.20.0 <13.0.0"
   },
   "husky": {
     "hooks": {


package.json
```diff
@@ -84,7 +84,7 @@
   "author": "Juan Cruz Viotti <juan@balena.io>",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.20.0"
+    "node": ">=10.20.0 <13.0.0"
```
pdcastro, line 87: With this change, when using Node.js v14, `npm install` still succeeds but then when running the CLI the result is:
```
$ ./bin/balena version -a
------------------------------------------------------------------------------
Warning: Node version "v14.15.4" does not match required versions ">=10.20.0 <13.0.0".
This may cause unexpected behavior. To upgrade Node, visit:
https://nodejs.org/en/download/
------------------------------------------------------------------------------
balena-cli version "12.38.3"
Node.js version "14.15.4"
```

As question was whether to use `<13.0.0` or `<14.0.0`. We know the CLI works with Node v12, we know it does not work with Node v14, but we are not sure about Node v13. I think that a good principle would be _"it works unless we know it doesn't"_ (as this encourages people to test and report bugs), by which we should use `<14.0.0`. However, because this change will be disruptive to many users who've been using Node.js v14 for a while (we've allowed it for several months), it would be a shame if many users went through the trouble of downgrading to Node.js v13, only for us later finding that v13 also doesn't work, and disrupt users again.

srlowe (approved):

package.json
pdcastro: Also of potential interest:
* https://github.com/Homebrew/homebrew-core/pull/63674#discussion_r558329656
* https://aur.archlinux.org/cgit/aur.git/commit/?h=balena-cli&id=fc21a59bec910c9a92c0d172c645d092fe643bea

PR #2168, opened 2021-01-21
Make `os configure` and `local flash` work with Node 14

mbalamat:
This PR adds Node14 support for the `os configure` and `local flash` commands. Also, when push is used it doesn't segfault. Local push with Node14 still needs investigation.


mbalamat:
@balena-ci retest!

mbalamat:
@balena-ci test!

mbalamat:
@balena-ci test

stathismor:
@balena-ci test

package.json
```diff
@@ -221,7 +222,7 @@
     "docker-toolbelt": "^3.3.8",
     "dockerode": "^2.5.8",
     "ejs": "^3.1.3",
-    "etcher-sdk": "^2.0.20",
+    "etcher-sdk": "^5.2.2",
```
pdcastro, line 225: Looks like we now have v6.2.0 available:
https://github.com/balena-io-modules/etcher-sdk/blob/master/CHANGELOG.md

mbalamat: @pdcastro sure, I'll add it as soon as we merge the balena-device-init pr. I'll add this ^ and replace the npm pre-releases.

pdcastro (commented):
I see there are 7 commits, each of which would become a bullet point in the CHANGELOG file that end users look at for "what's new" in the CLI:
* Replace old resin-image-fs with newer balena-image-fs
* package.json: Remove resin-image-fs package
* package: Install dependencies that work with node14
* Remove resin-image-fs typings
* Fix etcher-sdk related types
* tests/utils/eol-conversion: Remove ext2fs sample binary
* tests/test-data/pkg: Add new expected warnings darwin/linux/windows

There is always a conflict between granular commits that may make sense to CLI developers, and commits that make sense to end users. From an end user's point of view, they'd be better served by a single commit (squashing all the above) that read something like:
* Add support for Node.js v14

By the way, so far in the balena-cli repo, we haven't made much use of `Changelog-entry:` footer, but I know it exists and it allows specifying an alternative message for the Changelog file. I don't know if it also allows omitting some commits from the Changelog file. I also don't remember which repo implements the commit parsing, so we could learn exact what's possible and available. The following two repos come to mind:
* https://github.com/balena-io/versionist
* https://github.com/balena-io-modules/resin-commit-lint/

lib/commands/local/flash.ts
```diff
@@ -16,6 +16,7 @@
  */
 
 import { flags } from '@oclif/command';
+import { BlockDevice } from 'etcher-sdk/build/source-destination';
```
srlowe, line 19: Maybe 
```suggestion
import type { BlockDevice } from 'etcher-sdk/build/source-destination';
```

tests/test-data/pkg/expected-warnings-darwin.txt
```diff
@@ -24,6 +24,42 @@
   The file must be distributed with executable as %2.
   %1: node_modules/open/xdg-open
   %2: path-to-executable/xdg-open
+> Warning Cannot include addon %1 into executable.
+  The addon must be distributed with executable as %2.
+  %1: node_modules/@ronomon/direct-io/binding.node
+  %2: path-to-executable/binding.node
+> Warning Cannot include file %1 into executable.
+  The file must be distributed with executable as %2.
+  %1: node_modules/etcher-sdk/node_modules/drivelist/build/Release/drivelist.node
```
pdcastro, line 33: Because of these changes to expected warnings, I decided to test the draft standalone zip package downloaded from here:
https://github.com/balena-io/balena-cli/releases/tag/untagged-29cb5354571d6faf986b

Which actually still bundles Node.js v10:

```
$ ./balena-cli-12.41.0-draft/balena version -a
balena-cli version "12.41.0"
Node.js version "10.21.0"
```

I tried the `local flash` command (because it exercises the Etcher SDK) and it resulted in an error:
```
$ sudo DEBUG=1 ./balena-cli-12.41.0-draft/balena local flash balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip
Password:
[debug] original argv0="./balena-cli-12.41.0-draft/balena" argv=[/Users/paulo/Downloads/balena-cli-12.41.0-draft/balena,/snapshot/volume/bin/balena,local,flash,balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip] length=5
[debug] new argv=[/Users/paulo/Downloads/balena-cli-12.41.0-draft/balena,/snapshot/volume/bin/balena,local:flash,balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip] length=4
? Select a drive /dev/disk2 (4.1 GB) - TOSHIBA USB Flash Memory Media
? This will erase the selected drive. Are you sure? Yes
EBADF: bad file descriptor, write
Segmentation fault: 11
```

Same error on Linux and macOS. The error does not happen with the master branch, standalone zip package version 12.40.3, so it's something introduced by this PR.

I also tested building the standalone zip package on my laptop using Node.js v12 and Node.js v14, and got the same `EBADF: bad file descriptor, write` error (as above with Node.js v10) -

```
$ ./build-bin/balena version -a
balena-cli version "12.40.3"
Node.js version "12.18.1"

$ sudo DEBUG=1 ./build-bin/balena local flash ~/Downloads/balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip
Password:
[debug] original argv0="./build-bin/balena" argv=[/Users/paulo/balena/balena-io/balena-cli/build-bin/balena,/snapshot/balena-cli/bin/balena,local,flash,/Users/paulo/Downloads/balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip] length=5
[debug] new argv=[/Users/paulo/balena/balena-io/balena-cli/build-bin/balena,/snapshot/balena-cli/bin/balena,local:flash,/Users/paulo/Downloads/balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip] length=4
? Select a drive /dev/disk2 (4.1 GB) - TOSHIBA USB Flash Memory Media
? This will erase the selected drive. Are you sure? Yes
EBADF: bad file descriptor, write

$ nvm use 14
Now using node v14.16.0 (npm v6.14.11)

$ rm -rf node_modules && npm i && npm run build:fast && npm run build:standalone

$ ./build-bin/balena version -a
balena-cli version "12.40.3"
Node.js version "14.4.0"

$ sudo DEBUG=1 ./build-bin/balena local flash ~/Downloads/balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip
[debug] original argv0="./build-bin/balena" argv=[/Users/paulo/balena/balena-io/balena-cli/build-bin/balena,/snapshot/balena-cli/bin/balena,local,flash,/Users/paulo/Downloads/balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip] length=5
[debug] new argv=[/Users/paulo/balena/balena-io/balena-cli/build-bin/balena,/snapshot/balena-cli/bin/balena,local:flash,/Users/paulo/Downloads/balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip] length=4
? Select a drive /dev/disk2 (4.1 GB) - TOSHIBA USB Flash Memory Media
? This will erase the selected drive. Are you sure? Yes
EBADF: bad file descriptor, write
```

So it's not related to the version of Node.js.

I have noticed that the CLI's `node_modules` folder now includes 2 installations of `drivelist`:

```
$ find node_modules -type d -name drivelist
node_modules/drivelist
node_modules/etcher-sdk/node_modules/drivelist
```

The `npm-shrinkwrap.json` file clarifies it's because `etcher-image-write` and `resin-device-operations` require `drivelist` v8:

```
    "etcher-image-write": {
      "version": "9.1.6",
      "resolved": "https://registry.npmjs.org/etcher-image-write/-/etcher-image-write-9.1.6.tgz",
      "requires": {
        "drivelist": "^8.0.10",
        ...
```
```
    "resin-device-operations": {
      "version": "1.7.0",
      "resolved": "https://registry.npmjs.org/resin-device-operations/-/resin-device-operations-1.7.0.tgz",
      "requires": {
        "drivelist": "^8.0.10",
        ...
```
```
    "drivelist": {
      "version": "8.0.10",
      "resolved": "https://registry.npmjs.org/drivelist/-/drivelist-8.0.10.tgz",
```

... while `etcher-sdk` requires `drivelist` v9:

```
    "etcher-sdk": {
      "version": "6.2.0",
      "resolved": "https://registry.npmjs.org/etcher-sdk/-/etcher-sdk-6.2.0.tgz",
      "requires": {
        "drivelist": "^9.2.4",
        ...
      },
      "dependencies": {
        ...,
        "drivelist": {
          "version": "9.2.4",
          "resolved": "https://registry.npmjs.org/drivelist/-/drivelist-9.2.4.tgz",
```

This may or may not have something to do with the `EBADF: bad file descriptor, write` error.

If `etcher-image-write` and `resin-device-operations` could be updated to use `drivelist` v9 like `etcher-sdk`, then we would not need to installations of it.

I thought it might be case of having to add some files to the `pkg.scripts` or `pkg.assets` sections of the CLI's `package.json`:
https://github.com/balena-io/balena-cli/blob/v12.40.3/package.json#L26-L46

But then I realised that I also get the error with the npm installation - a straight `./bin/balena` execution:

```
$ sudo ./bin/balena local flash ~/Downloads/balena-cloud-test-rpi-raspberrypi3-2.69.1+rev1-dev-v12.3.5.img.zip
? Select a drive /dev/disk2 (4.1 GB) - TOSHIBA USB Flash Memory Media
? This will erase the selected drive. Are you sure? Yes
EBADF: bad file descriptor, write
```

So it may actually not have anything to do with the standalone zip package. If it's not the Node.js version and not the standalone zip package, then it may actually be a source code bug somewhere...

pdcastro (changes requested):

package.json
```diff
@@ -84,7 +84,7 @@
   "author": "Juan Cruz Viotti <juan@balena.io>",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.20.0 <13.0.0"
+    "node": ">=10.20.0 <15.0.0"
```
pdcastro, line 87: My thinking may be flawed, but on this line in `package.json`, I think that we should only "prevent" (warn against) usage of versions that we know not to work. So I'd recommend:
```
    "node": ">=10.20.0"
```
At the same time, this PR could amend the [INSTALL-ADVANCED.md](https://github.com/balena-io/balena-cli/blob/master/INSTALL-ADVANCED.md#npm-installation):

**From:** `Node.js version 10 (min 10.20.0) or 12 (version 14 is not yet fully supported)`
**To:** `Node.js version 10 (min 10.20.0), 12 or 14 (other versions may also work, but are not well tested)`

The GUI installers and the standalone zip package bundle their own Node.js (currently v10 or v12), so only users who choose the _not-recommended_ advanced installation option -- npm install -- are affected by this change in `package.json`, and even then the documentation warns them that only versions 10, 12 and 14 are tested. So if they choose to use Node.js v16, v25... and find bugs and report them, they'd effectively be testing Node.js for us, and saving us the trouble. :-)

tests/test-data/pkg/expected-warnings-darwin.txt
mbalamat: @pdcastro it just needed 3 extra arguments https://github.com/balena-io-modules/etcher-sdk/blob/cac65b6864619a349000209cf284a9bb110d3068/examples/multi-destination.ts#L119-L121 .

package.json
mbalamat: I like this idea. @srlowe what do you think?

pdcastro (changes requested):
In my testing I have found that, so far, this PR does not resolve issue #2165. I have tested the following `balena push` commands with Node.js v10, v12 and v14, pushing the BoB / devenv repo to a NUC device in local mode. The command succeeds with Node.js v10 and v12, but fails with Node.js v14, as reported in that issue and also as reported by Matthew in this FD thread:
https://www.flowdock.com/app/rulemotion/resin-starters/threads/k2KO6r6D4ep5CyKC_Sf-DaU01Po

```
$ nvm use 10
Now using node v10.24.0 (npm v6.14.11)

$ rm -rf node_modules && npm i && npm run build
$ ./bin/balena push 192.168.1.46 -m -s ~/balena/balena-io/balena-on-balena -R ~/balena/balena-io/balena-on-balena/secrets.yml --debug
...
[Logs]    [3/14/2021, 12:47:05 AM] [wifi-connect] default via 192.168.1.1 dev wlp58s0 proto dhcp metric 600
[Logs]    [3/14/2021, 12:47:05 AM] [wifi-connect] Skipping WiFi Connect\n
^C[Warn]    Connection to device lost
[Live]    Cleaning up device...
[Debug]   Cleaning up done.
Task aborted on SIGINT signal

SIGINTError: Task aborted on SIGINT signal
    at process.sigintHandler (/Users/paulo/balena/balena-io/balena-cli/build/utils/helpers.js:315:20)
    at Object.onceWrapper (events.js:286:20)
    at process.emit (events.js:203:15)
    at process.EventEmitter.emit (domain.js:448:20)
```

Above we see correct behaviour with Node v10: the CLI was running continuously and displaying logs, until I hit CTRL-C to abort it. The `SIGINTError` is displayed because of `--debug`, but it is correct and expected behaviour.

```
$ nvm use 12
Now using node v12.21.0 (npm v6.14.11)
$ rm -rf node_modules && npm i && npm run build
$ ./bin/balena push 192.168.1.46 -m -s ~/balena/balena-io/balena-on-balena -R ~/balena/balena-io/balena-on-balena/secrets.yml --debug
...
[Logs]    [3/14/2021, 00:52:25] [monitor] oauth2-proxy[8507]: [2021/03/14 00:52:25] [logger.go:490] invalid configuration:
[Logs]    [3/14/2021, 00:52:25] [monitor] oauth2-proxy[8507]:   missing setting: cookie-secret
^C[Warn]    Connection to device lost
[Live]    Cleaning up device...
[Debug]   Cleaning up done.
Task aborted on SIGINT signal

SIGINTError: Task aborted on SIGINT signal
    at process.sigintHandler (/Users/paulo/balena/balena-io/balena-cli/build/utils/helpers.js:315:20)
    at Object.onceWrapper (events.js:421:26)
    at process.emit (events.js:326:22)
    at process.EventEmitter.emit (domain.js:483:12)
```

Above we see correct behavior with Node v12, just as described above for Node.js v10. The CLI correctly exited when I hit CTRL-C, and `SIGINTError` was correctly printed because of the `--debug` flag.

```
$ nvm use 14
Now using node v14.16.0 (npm v6.14.11)
$ rm -rf node_modules && npm i && npm run build
$ ./bin/balena push 192.168.1.46 -m -s ~/balena/balena-io/balena-on-balena -R ~/balena/balena-io/balena-on-balena/secrets.yml --debug

... prints hundreds of lines, then
[Build]   [netdata] Successfully built 84f611fd0d7e
[Build]   [netdata] Successfully tagged local_image_netdata:latest
```
Above, with Node.js v14, the CLI exited suddenly and unexpectedly and soon after the container builds completed, without me hitting CTRL-C. Interestingly, it exited with process exit code 0, which normally indicates success, as if no error had happened, which may be a debugging clue as well: it is as if the event loop had concluded that there was nothing else to do, and the process exited "gracefully" -- although incorrectly insofar as it does not behave as we want it to. The CLI should not have exited, it should have kept running and displaying logs. This "graceful exit" observation is just speculation, I don't really know the reason why the CLI decided to exit.

You will notice that I added a commit that bumps `net-keepalive`. It added it because I thought it might fix the issue, because `net-keepalive` is used by `balena push` and the changelog (or recent commits) of their dependencies `ffi-napi` and `ref-napi` mention improved support for NAPI v6 and added CI tests for Node.js 14. Unfortunately, that commit I added did not seem to make any difference re issue described above. `balena push` exits unexpectedly with Node.js v14 regardless of my commit. Still, I think that updating `net-keepalive` is desirable -- a step in the right direction, even if it hasn't fixed the problem.

Given that `balena push` is one of the most important CLI commands and it still does not work with Node.js v14, I think we couldn't merge this PR while it claims to make the CLI work with Node.js v14 generally. However, if you like, you could split this PR in two, having a separate PR that adds support for Node.js v14 to the `os configure` and `local flash` commands only. But we'd still have to generally recommend users to use Node.js v10 or v12.

NIT about the 2 commits (before I added mine):
* The commit that updates the install docs says `Change-type: minor`
* The commit that does everything else says `Change-type: patch`

Wouldn't that be the other way around? :-) It doesn't really matter though, because either way the CLI would have a minor version bump overall.

Hades32:
Another hint at debugging this:
* it works for me with v14.15.4 on Linux (WSL2 in Win10)
* there were issues for @LucianBuzzo on WSL1

Maybe the culprit is some underlying system library...

package.json
```diff
@@ -84,7 +84,7 @@
   "author": "Juan Cruz Viotti <juan@balena.io>",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.20.0 <15.0.0"
+    "node": ">=10.20.0"
```
pdcastro, line 87: As this PR is now for `os configure` and `local flash` only, I think this commit (that updates `package.json` `node` versions, and `INSTALL-ADVANCED.md`) should be dropped from this PR and added to the other PR that resolves `balena push` with Node.js v14.
This is because, after this PR is merged, I understand we will still have to ask users to use Node.js v10 or v12 only (like the balena team, using `balena push` with the devenv / BoB).

pdcastro: Btw, my commit re `net-keepalive` would also fit better the PR for `balena push` than the PR for `os configure` and `local flash`, but it should also be harmless to have it here. It's up to you.

mbalamat: I agree with the first point. I'd say keep the `net-keepalive` bump. Also, we could probably edit the issue about the segfault on push. Nothing segfaults now.

pdcastro: > we could probably edit the issue about the segfault on push. Nothing segfaults now.

Issue #2200, right? Were you able to reproduce it? So that we could demonstrate that it crashed before this PR, and not after this PR. I wasn't able to reproduce it myself and I'm thinking I may have jumped to conclusions when I assumed that issue #2200 was related to Node.js v14.

pdcastro: By the way, the first commit still claims `Resolves: #2079`, but I think it should not. There are users watching issue #2079, which I think was also linked in the forums and in homebrew, as the issue that makes the balena CLI fully support Node.js v14. I think we need to leave that issue open until `balena push` is resolved in another PR.

mbalamat:
@balena-ci retest

pdcastro:
The only thing preventing me from approving this PR at the moment is the `Resolves: #2079` footer in one of the commits. I think that footer line should be removed.

package.json
```diff
@@ -84,7 +84,7 @@
   "author": "Juan Cruz Viotti <juan@balena.io>",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.20.0 <13.0.0",
+    "node": ">=10.20.0 <15.0.0",
```
pdcastro, line 87: I understand that, after this PR is merged, we still don't support Node.js v14 generally (e.g. `balena push` with devenv / BoB works with Node.js v10 and v12, but not v14). Therefore, this PR should not be amending this line. No?

mbalamat: Basically let change that and see.

pdcastro: If the Node.js v14 checks fail because of the warnings printed in connection with this `engines.node` restriction, we could either ignore it, or maybe hardcode exceptions for example here:
https://github.com/balena-io/balena-cli/blob/v12.42.0/tests/helpers.ts#L46-L52

lib/commands/local/flash.ts
```diff
@@ -75,6 +75,21 @@ export default class LocalFlashCmd extends Command {
 			LocalFlashCmd,
 		);
 
+		if (process.platform === 'linux') {
+			const { promisify } = await import('util');
+			const { exec } = await import('child_process');
+			const execAsync = promisify(exec);
+			const info = await execAsync('cat /proc/version');
```
pdcastro, line 82: I don't know if it would be possible for this statement to throw an error in some Linux distribution... Perhaps it would be a bit more robust to have something along the lines of:
```ts
		let distroVersion = '';
		try {
			const info = await execAsync('cat /proc/version');
			distroVersion = info.stdout.toLowerCase();
		} catch {
			// pass
		}
		if (distroVersion.includes('microsoft')) {
			throw new ExpectedError(stripIndent`
				This command is known not to work on WSL. Please use a CLI release
				for Windows (not WSL), or balenaEtcher.`);
		}
```
Above I've also replaced `process.exit(0)` with `throw new ExpectedError`, which allows "graceful exit" through `errors.ts`. This also changes the process exit code from 0 to 1, which I think is good too.

pdcastro: > graceful exit

`process.exit()` causes immediate process termination, preventing network requests from completing (e.g. HTTP to Amplitude or Sentry) and I believe preventing buffered filesystem I/O from being flushed to disk.

pdcastro: https://nodejs.org/docs/latest-v10.x/api/process.html#process_process_exit_code
>Calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.
> In most situations, it is not actually necessary to call process.exit() explicitly. The Node.js process will exit on its own if there is no additional work pending in the event loop. The process.exitCode property can be set to tell the process which exit code to use when the process exits gracefully.

lib/commands/os/configure.ts
```diff
@@ -279,37 +280,23 @@ export default class OsConfigureCmd extends Command {
...
-			the Windows Subsystem for Linux in order to run on Windows. It was tested with
-			the Ubuntu 18.04 distribution from the Microsoft Store. With WSL, a balena CLI
-			release for Linux (rather than Windows) should be installed: for example, the
-			standalone zip package for Linux. (It is possible to have both a Windows CLI
-			release and a Linux CLI release installed simultaneously.) For more information
-			on WSL and the balena CLI installation options, please check:
-			- https://docs.microsoft.com/en-us/windows/wsl/about
-			- https://github.com/balena-io/balena-cli/blob/master/INSTALL.md
-		`);
-	}
```
pdcastro, line 312: Nice 👍  This PR could also delete a similar note in the `INSTALL-WINDOWS.md` file:
https://github.com/balena-io/balena-cli/blob/v12.43.0/INSTALL-WINDOWS.md#balena-os-configure

pdcastro:
@mbalamat, I've added a commit and tests are now passing. Try rebasing on master -- it's possible there'll be some small conflicts.

pdcastro:
@mbalamat, after you get the tests to pass, I'll do some testing of my own (e.g. running `os configure` on Windows) and hopefully we will get this PR merged. Notes:
* I expect you'll want to squash most of the 10 commits. I understand you keep adding commits to make review easier, right? But this doesn't match balena's "approval = merge" workflow by which a non-draft PR should be ready to merge as is (with all the commits). You can have a non-squashed backup branch on your local machine for your own benefit (e.g. ability to revert to a known good state), but I suggest keeping the commits in their final form even if you have to force-push changes to previous commits (e.g. through interactive rebase). As a reviewer, I prefer that changes are force-pushed to previous commits, so that I can review the PR in its final form -- otherwise, I don't know what comments to make regarding commit messages / changelog / patch/minor, and I cannot approve the PR because approval means that all the commits are merged straight away. (The only time when I find force pushing less nice is after a reviewer has approved the PR, which can happen if the reviewer approves it with failed tests or before the build completes.)

* I think this comment hasn't been addressed: https://github.com/balena-io/balena-cli/pull/2168/files#r601638358

mbalamat:
@pdcastro yes I'll squash everything and ping you when ready. I just don't squash until the last minute because it's easier to solve merge conflicts as it is.

mbalamat:
@pdcastro it's ok now, feel free to test it.

pdcastro (approved):
Awesome! 🎉

PR #2170, opened 2021-01-22
tunnel: Add note re openBalena version compatibility

pdcastro:
Update `balena help tunnel` following the release of openBalena v3.1.2:
https://github.com/balena-io/open-balena/blob/master/CHANGELOG.md#v312

Change-type: patch

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -1830,17 +1830,25 @@ bypass global proxy configuration for the ssh connection
 
 ## tunnel &#60;deviceOrApplication&#62;
 
-Use this command to open local ports which tunnel to listening ports on your balenaOS device.
+Use this command to open local TCP ports that tunnel to listening sockets in a
+balenaOS device.
 
-For example, you could open port 8080 on your local machine to connect to your managed balenaOS
-device running a web server listening on port 3000.
+For example, this command could be used to expose the ssh server of a balenaOS
+device (port number 22222) on the local machine, or to expose a web server
+running on the device. The port numbers do not have be the same between the
+device and the local machine, and multiple ports may be tunneled in a single
+command line.
 
 Port mappings are specified in the format: <remotePort>[:[localIP:]localPort]
-localIP defaults to 'localhost', and localPort defaults to the specified remotePort value.
+localIP defaults to 'localhost', and localPort defaults to the specified
+remotePort value.
 
-You can tunnel multiple ports at any given time.
+Note: the -p (--port) flag must be provided at the end of the command line,
+as per examples.
 
-Note: Port mappings must come after the deviceOrApplication parameter, as per examples.
+In the case of openBalena, the tunnel command in CLI v12.38.5 or later requires
+openBalena v3.1.2 or later. Older CLI versions work with older openBalena
+versions.
 
 Examples:
 
--- a/lib/commands/tunnel.ts
+++ b/lib/commands/tunnel.ts
@@ -39,17 +39,25 @@ export default class TunnelCmd extends Command {
 	public static description = stripIndent`
 		Tunnel local ports to your balenaOS device.
 
-		Use this command to open local ports which tunnel to listening ports on your balenaOS device.
+		Use this command to open local TCP ports that tunnel to listening sockets in a
+		balenaOS device.
 
-		For example, you could open port 8080 on your local machine to connect to your managed balenaOS
-		device running a web server listening on port 3000.
+		For example, this command could be used to expose the ssh server of a balenaOS
+		device (port number 22222) on the local machine, or to expose a web server
+		running on the device. The port numbers do not have be the same between the
+		device and the local machine, and multiple ports may be tunneled in a single
+		command line.
 
 		Port mappings are specified in the format: <remotePort>[:[localIP:]localPort]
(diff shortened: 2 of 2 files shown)


lib/commands/tunnel.ts
```diff
@@ -39,17 +39,25 @@ export default class TunnelCmd extends Command {
 	public static description = stripIndent`
 		Tunnel local ports to your balenaOS device.
 
-		Use this command to open local ports which tunnel to listening ports on your balenaOS device.
+		Use this command to open local TCP ports that tunnel to listening sockets in a
+		balenaOS device.
```
pdcastro, line 43: Some of the changes were for the lines to fit in 80 characters. Some changes were minor, for improved clarity (hopefully).

lib/commands/tunnel.ts
```diff
@@ -39,17 +39,25 @@ export default class TunnelCmd extends Command {
...
+		remotePort value.
 
-		You can tunnel multiple ports at any given time.
+		Note: the -p (--port) flag must be provided at the end of the command line,
+		as per examples.
 
-		Note: Port mappings must come after the deviceOrApplication parameter, as per examples.
+		In the case of openBalena, the tunnel command in CLI v12.38.5 or later requires
+		openBalena v3.1.2 or later. Older CLI versions work with older openBalena
+		versions.
```
pdcastro, line 60: This is the key doc addition that motivated the PR.

lib/commands/tunnel.ts
```diff
@@ -66,7 +74,7 @@ export default class TunnelCmd extends Command {
 		'$ balena tunnel myApp -p 22222:0.0.0.0:222',
 		'',
 		'# multiple port tunnels can be specified at any one time',
-		'$ balena tunnel myApp -p 8080:3000 -p 8081:9000',
+		'$ balena tunnel myApp -p 8080:3000 8081:9000',
```
pdcastro, line 77: The second `-p` is optional. I've removed it because it kind of clarifies the reason why the `-p` option must be provided at the end of the command line. (Which is really an oclif design bug, if you ask me...)

dfunckt (commented):
It looks great but I'd argue that we don't really need to include this sort info in command help.

pdcastro:
> [...] but I'd argue that we don't really need to include this sort info in command help.

I think it helps towards reducing friction for users, avoiding the need for users to create CLI bug reports (which aren't really bugs), support interactions and async pings that take hours to resolve (or days if the user creates a thread in a Saturday morning...)

dfunckt:
I understand and agree with the premise -- I'm only not sure command help is the right place for this sort of info. Up to you though, I can't think of a better alternative for now, may be worth discussing it separately.

pdcastro:
We can/should add the compatibility note in other places as well, I think especially the openBalena Getting Started Guide:
https://www.balena.io/open/docs/getting-started/

The `balena help tunnel` info might be most useful to users who already have a working (old) openBalena installation, who then decide to upgrade the CLI. Some time later, they try `balena tunnel ...` and it fails. They'd thing, _"what am I doing wrong?",_ then they run `balena help tunnel`, see the note regarding version compatibility, and realise - _"oh right, I've recently updated the CLI, so that's why it's failing..."_

pdcastro:
@balena-ci rebase

doc/cli.markdown
```diff
@@ -1857,7 +1865,7 @@ Examples:
 	$ balena tunnel myApp -p 22222:0.0.0.0:222
 	
 	# multiple port tunnels can be specified at any one time
-	$ balena tunnel myApp -p 8080:3000 -p 8081:9000
+	$ balena tunnel myApp -p 8080:3000 8081:9000
```
srlowe, line 1868: Are both these formats supported? `-p 8080:3000 -p 8081:9000` / `-p 8080:3000 8081:9000` 

I wonder if the latter might be confusing for people who have seen the comma seperated convention for we have adopted for multi arguments?

lib/commands/tunnel.ts
srlowe: Ah.  I wonder if this spaced separated convention might be confusing for people who have seen the comma separated convention we have adopted for multi arguments?

pdcastro: Most recently we used the space separated syntax for the `--release-tags` option in the `push` and `deploy` commands:
```
$ balena help push
EXAMPLES
...
  $ balena push myApp --release-tag key1 "" key2 "value2 with spaces"

$ balena help deploy
EXAMPLES
...
  $ balena deploy myApp myApp/myImage --release-tag key1 "" key2 "value2 with spaces"
```
If it's controversial, I will revert the example.

pdcastro: I have already reverted the change made in this PR to the usage example of the `-p` option.

pdcastro:
PR converted back to draft pending confirmation that the advice is actually correct, as I seem to have tested that only older CLI releases work the latest openBalena release (the opposite of what the PR proposed to advise) -
https://www.flowdock.com/app/rulemotion/resin-tech/threads/ZY_iGlst9bPMr337xR7ay6BlurH

srlowe (approved):

PR #2173, opened 2021-01-29
build/deploy: Update QEMU to speed up emulated builds 

klutchell:
QEMU v5 has quite a few improvements over v4, and the speed
difference when emulating arm is quite noticible.

We tested this with, and without, our single-core limitation
patch and have not been able to reproduce the stability
issues we were seeing in v4 so the patch was removed in
this release.

Change-type: patch
Connects-to: balena-io/balena-io#2340

---

### DIFF
--- a/lib/utils/qemu.ts
+++ b/lib/utils/qemu.ts
@@ -19,7 +19,7 @@ import type * as Dockerode from 'dockerode';
 import { getBalenaSdk, stripIndent } from './lazy';
 import Logger = require('./logger');
 
-export const QEMU_VERSION = 'v4.0.0+balena2';
+export const QEMU_VERSION = 'v5.2.0+balena1';
 export const QEMU_BIN_NAME = 'qemu-execve';
 
 export function qemuPathInContext(context: string) {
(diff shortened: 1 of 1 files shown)


pdcastro (commented):
Great job @klutchell !! 💯 

Just being picky about the Changelog, which end users will be looking at. Only the first line of the commit gets copied to the Changelog, so at moment it would look like:

* qemu: update to v5.2.0+balena1

But many CLI users may not know or remember what `qemu` is about, and won't know which CLI commands are affected or the benefit of this CLI release. I suggest rewording the first line of the commit to:

`build/deploy: Update QEMU to speed up emulated builds`

So the Changelog reads:

* build/deploy: Update QEMU to speed up emulated builds

Then CLI users of `balena build -e` and `balena deploy --build -e` will realise that their builds are affected and should run faster. And in case of build problems, users may be able to associate it with this CLI release, to help with any investigations.

klutchell:
@pdcastro Good to know! I didn't know that the changelog used the first line of the commit for this repo, I'm used to adding `Changelog-entry:` to the bottom of my commit template. I guess this is different for each repo? Does balena lean one way or the other regarding how the changelog picks things up?

pdcastro:
> @pdcastro Good to know! I didn't know that the changelog used the first line of the commit for this repo, I'm used to adding `Changelog-entry:` to the bottom of my commit template. I guess this is different for each repo? Does balena lean one way or the other regarding how the changelog picks things up?

This repo should be treated the same as any other, and the default (in the absence of `Changelog-entry:`) is to copy the first line only. I had not noticed - my bad - that your previous commit had a `Changelog-entry` footer:

```
Changelog-entry: update qemu to v5.2.0 with balena patches [klutchell]
```

But the same review comment would have applied. :-)

pdcastro (approved):

PR #2175, opened 2021-02-04
Modify push to pass app slug to builder

srlowe:
As described in this comment: https://github.com/balena-io/balena-cli/issues/2119#issuecomment-769894278
 
Change-type: patch


pdcastro (approved):
Nice 👍

PR #2176, opened 2021-02-04
Add command `device local-mode`

srlowe:
Switching local mode on and off is currently not convenient as it requires navigating the web dashboard.  This has become more of an issue recently because livepush seems to require this to get out of some (so far undiagnosed) broken states.

This implementation mirrors the recently adopted conventions for `device public-url`.  The linked issue contains alternative suggestions, but these are very old and not practical with oclif's limitations (they also discuss public-url, which has now been implemented differently).

Change-type: minor
Resolves: #1304

### DIFF
--- a/lib/commands/device/local-mode.ts
+++ b/lib/commands/device/local-mode.ts
@@ -1,0 +1,114 @@
+/**
+ * @license
+ * Copyright 2021 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { flags } from '@oclif/command';
+import type { IArg } from '@oclif/parser/lib/args';
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+import { tryAsInteger } from '../../utils/validation';
+
+interface FlagsDef {
+	enable: boolean;
+	disable: boolean;
+	status: boolean;
+	help?: void;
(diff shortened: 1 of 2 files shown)


lib/utils/device/deploy.ts
```diff
@@ -140,9 +141,11 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 		globalLogger.logDebug('Checking we can access device');
 		await api.ping();
 	} catch (e) {
-		throw new ExpectedError(
-			`Could not communicate with local mode device at address ${opts.deviceHost}`,
-		);
+		throw new ExpectedError(stripIndent`
+			Could not communicate with local mode device at address ${opts.deviceHost}.
```
pdcastro, line 145: In this sentence, `local mode device` implies the device is known to be in local mode, which is then readily cast into doubt by the subsequent sentence. :-) Anyway, I think it would be nice to be specific that the CLI was not able to contact the supervisor on the device:
```
	Could not communicate with device supervisor at address ${opts.deviceHost}.
	Device may not have local mode enabled. Check with:
	  balena device local-mode <device-uuid>
```
We could even mention the port number. A few lines above (line 137), we could do:
```
const port = 48484;
const api = new DeviceAPI(globalLogger, opts.deviceHost, port);
```
And then:
```
	Could not communicate with device supervisor at address ${opts.deviceHost}:${port}.
	Device may not have local mode enabled. Check with:
	  balena device local-mode <device-uuid>
```

lib/commands/device/local-mode.ts
```diff
@@ -0,0 +1,114 @@
...
+
+interface FlagsDef {
+	enable: boolean;
+	disable: boolean;
+	status: boolean;
+	help?: void;
+}
+
+interface ArgsDef {
+	uuid: string;
```
pdcastro, line 33: Would it be `string | number` ? Because of `parse: (dev) => tryAsInteger(dev)`
Where `params.uuid` is used in the code, I think that currently `tsc` is type checking against `string` only, but it could be `number` as well, so potential bugs are not being caught by `tsc`.

srlowe: I seem to recall there was some reason we didn't do this, but I can't remember exactly, will check.

srlowe: Ah it was flags where that caused a bit of an issue - ok in args.

pdcastro (approved):

pdcastro:
@balena-ci rebase

PR #2186, opened 2021-02-11
Update macOS installer to avoid Apple's warning pop-up

danthegoodman1:
Update macOS installer to avoid Apple's warning pop-up

Call notarization hook in `build-bin.ts`

Change-type: patch

Connects-to: #1479

### DIFF
--- a/automation/build-bin.ts
+++ b/automation/build-bin.ts
@@ -28,6 +28,8 @@ import * as path from 'path';
 import * as rimraf from 'rimraf';
 import * as semver from 'semver';
 import * as util from 'util';
+import * as klaw from 'klaw';
+import { Stats } from 'fs';
 
 import { stripIndent } from '../lib/utils/lazy';
 import {
@@ -38,6 +40,7 @@ import {
 	StdOutTap,
 	whichSpawn,
 } from './utils';
+
 export const packageJSON = loadPackageJson();
 export const version = 'v' + packageJSON.version;
 const arch = process.arch;
@@ -296,6 +299,88 @@ async function zipPkg() {
 	});
 }
 
+async function signFilesForNotarization() {
+	console.log('Deleting unneeded zip files...');
+	await new Promise((resolve, reject) => {
+		klaw('node_modules/')
+			.on('data', (item: { path: string; stats: Stats }) => {
+				if (!item.stats.isFile()) {
+					return;
+				}
+				if (
+					path.basename(item.path).endsWith('.zip') &&
+					path.dirname(item.path).includes('test')
+				) {
+					console.log('Removing zip', item.path);
+					fs.unlinkSync(item.path);
+				}
+			})
+			.on('end', resolve)
+			.on('error', reject);
+	});
+	// Sign all .node files first
+	console.log('Signing .node files...');
+	await new Promise((resolve, reject) => {
+		klaw('node_modules/')
+			.on('data', async (item: { path: string; stats: Stats }) => {
+				if (!item.stats.isFile()) {
+					return;
+				}
+				if (path.basename(item.path).endsWith('.node')) {
+					console.log('running command:', 'codesign', [
+						'-d',
+						'-f',
+						'-s',
+						'Developer ID Application: Balena Ltd (66H43P8FRG)',
+						item.path,
+					]);
+					await whichSpawn('codesign', [
+						'-d',
+						'-f',
+						'-s',
+						'Developer ID Application: Balena Ltd (66H43P8FRG)',
+						item.path,
+					]);
+				}
+			})
+			.on('end', resolve)
+			.on('error', reject);
+	});
+	console.log('Signing other binaries...');
+	console.log('running command:', 'codesign', [
+		'-d',
+		'-f',
+		'--options=runtime',
+		'-s',
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/denymount/bin/denymount',
+	]);
+	await whichSpawn('codesign', [
+		'-d',
+		'-f',
+		'--options=runtime',
+		'-s',
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/denymount/bin/denymount',
+	]);
+	console.log('running command:', 'codesign', [
+		'-d',
+		'-f',
+		'--options=runtime',
+		'-s',
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/macmount/bin/macmount',
+	]);
+	await whichSpawn('codesign', [
+		'-d',
+		'-f',
+		'--options=runtime',
+		'-s',
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/macmount/bin/macmount',
+	]);
+}
+
 export async function buildStandaloneZip() {
 	console.log(`Building standalone zip package for CLI ${version}`);
 	try {
@@ -344,6 +429,20 @@ async function signWindowsInstaller() {
 }
 
 /**
+ * Wait for Apple Installer Notarization to continue
+ */
+async function notarizeMacInstaller(): Promise<void> {
+	const appleId = 'accounts+apple@balena.io';
+	const { notarize } = await import('electron-notarize');
+	await notarize({
+		appBundleId: 'io.balena.etcher',
+		appPath: renamedOclifInstallers.darwin,
+		appleId,
+		appleIdPassword: '@keychain:CLI_PASSWORD',
+	});
+}
+
+/**
  * Run the `oclif-dev pack:win` or `pack:macos` command (depending on the value
  * of process.platform) to generate the native installers (which end up under
  * the 'dist' folder). There are some harcoded options such as selecting only
@@ -369,6 +468,10 @@ export async function buildOclifInstaller() {
 			console.log(`rimraf(${dir})`);
 			await Bluebird.fromCallback((cb) => rimraf(dir, cb));
 		}
+		if (process.platform === 'darwin') {
+			console.log('Signing files for notarization...');
+			await signFilesForNotarization();
+		}
 		console.log('=======================================================');
 		console.log(`oclif-dev "${packCmd}" "${packOpts.join('" "')}"`);
 		console.log(`cwd="${process.cwd()}" ROOT="${ROOT}"`);
@@ -381,6 +484,10 @@ export async function buildOclifInstaller() {
 		// (`oclif.macos.sign` section).
 		if (process.platform === 'win32') {
 			await signWindowsInstaller();
+		} else if (process.platform === 'darwin') {
+			console.log('Notarizing package...');
+			await notarizeMacInstaller(); // Notarize
+			console.log('Package notarized.');
 		}
 		console.log(`oclif installer build completed`);
 	}
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -6085,6 +6085,36 @@
         "jake": "^10.6.1"
       }
     },
+    "electron-notarize": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/electron-notarize/-/electron-notarize-1.0.0.tgz",
+      "integrity": "sha512-dsib1IAquMn0onCrNMJ6gtEIZn/azG8hZMCYOuZIMVMUeRMgBYHK1s5TK9P8xAcrAjh/2aN5WYHzgVSWX314og==",
+      "dev": true,
+      "requires": {
+        "debug": "^4.1.1",
+        "fs-extra": "^9.0.1"
+      },
+      "dependencies": {
+        "fs-extra": {
+          "version": "9.1.0",
+          "resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-9.1.0.tgz",
+          "integrity": "sha512-hcg3ZmepS30/7BSFqRvoo3DOMQu7IjqxO5nCDt+zM9XWjb33Wg7ziNT+Qvqbuc3+gWpzO02JubVyk2G4Zvo1OQ==",
+          "dev": true,
+          "requires": {
+            "at-least-node": "^1.0.0",
+            "graceful-fs": "^4.2.0",
+            "jsonfile": "^6.0.1",
+            "universalify": "^2.0.0"
+          }
+        },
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -221,6 +221,7 @@
     "docker-toolbelt": "^3.3.8",
     "dockerode": "^2.5.8",
     "ejs": "^3.1.3",
+    "electron-notarize": "^1.0.0",
```
pdcastro, line 224: This was added to the `dependencies` section of `package.json`, but I think it should be added to the `devDependencies` section instead (because it's not needed when CLI end users run `npm i -g --production balena-cli`).

automation/build-bin.ts
```diff
@@ -382,6 +383,7 @@ export async function buildOclifInstaller() {
 		if (process.platform === 'win32') {
 			await signWindowsInstaller();
 		}
+		await afterSignHook.sign(path.join(ROOT, 'dist', renamedOclifInstallers.darwin)); // File to notarize
```
pdcastro, line 386: I think the path is incorrect here, because `renamedOclifInstallers.darwin` already includes ROOT/dist.

pdcastro (commented):
* The build is failing because the commit is missing the `Change-type` footer. See:
https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#semantic-versioning-commit-messages-and-the-changelog

* It would also be good to add a `Connects-to: #1479` footer. (For now I suggest `Connects-to` instead of `Resolves` because it seems that this PR is not yet addressing notarization of the standalone zip package, only the GUI installer.)

* The first line of the commit message automatically becomes a bullet point in the Changelog file: https://github.com/balena-io/balena-cli/blob/master/CHANGELOG.md and many CLI end users (including the balena team) will be looking there to figure out "what's new". So I suggest rewording the first line of the commit:  
From: `Add eletrcon-notarize for notarization`
To: `Update macOS installer to avoid Apple's warning pop-up`

* There is a typo in the commit message and the PR title: `eletrcon` → `electron`

automation/afterSignHook.ts
```diff
@@ -0,0 +1,13 @@
+import { notarize } from 'electron-notarize'
+const { ELECTRON_SKIP_NOTARIZATION } = process.env
+
+export async function sign (filePath: string): Promise<void> {
+  const appleId = 'accounts+apple@balena.io'
+
+  await notarize({
+    appBundleId: 'io.balena.etcher',
```
pdcastro, line 8: In `package.json`, there is a `macos.identifier: "io.balena.cli"` value that I believe is currently used by oclif for signing (not notarizing) the installer. I wonder if that value would (eventually) work here.

pdcastro:
@balena-ci retest

danthegoodman1:
@balena-ci restest

pdcastro:
The last retest was requested within a minute of Alexis starting an operating system update on the Mac Mini (13:59 UTC). That's mean :-) https://www.flowdock.com/app/rulemotion/r-etcherprv/threads/eIgmi9pJMnJtvRZlMBDiYMA0OHY

danthegoodman1:
@balena-ci retest

danthegoodman1:
THERE WE GO NOW WE'RE COOKING -- Kept thinking the `Change-type` had to be in the comment, not the commit!

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci retest

nazrhom:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

package.json
```diff
@@ -235,13 +235,15 @@
...
     "js-yaml": "^3.14.0",
+    "JSONStream": "^1.0.3",
     "klaw": "^3.0.0",
     "livepush": "^3.5.0",
     "lodash": "^4.17.20",
     "minimatch": "^3.0.4",
     "mixpanel": "^0.10.3",
     "moment": "^2.27.0",
     "moment-duration-format": "^2.3.2",
+    "net-keepalive": "^1.3.6",
```
pdcastro, line 246: I think `net-keepalive` and `windosu` were listed only in the `optionalDependencies` section, not the `dependencies` section, because they are not available for all platforms:
* `net-keepalive` require a native node module compilation that only succeeds on Linux and macOS, not Windows, and causes the build and installation to fail on Windows. (The CLI dynamically checks the platform before loading it at runtime.)
* `windosu` is only needed on Windows (iirc, the CLI won't try to use it on other platforms).

Were these changes necessary for any specific reason? If not, I suggest reverting them. The `net-keepalive` change is indeed currently preventing the build from succeeding on Windows.

package.json
```diff
@@ -21,7 +21,7 @@
     "oclif.manifest.json"
   ],
   "bin": {
-    "balena": "./bin/balena"
+    "balena": "bin/balena"
```
pdcastro, line 24: Was this change necessary for some specific reason? At least on Linux, the current working directory is sometimes not included in the executable search path, so attempting to execute `myApp` fails where `./myApp` succeeds.

npm-shrinkwrap.json
```diff
@@ -1,8 +1,17290 @@
 {
   "name": "balena-cli",
   "version": "12.40.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
pdcastro, line 4: I believe that this change to the format and version of the shrinkwrap file is causing the build to produce the following warning and error:
```
+ node --version
v10.21.0
+ npm --version
6.14.4
+ npx check-node-version@4 --npm '^6.9.0'
npx: installed 13 in 6.391s
+ run_as npm install
+ su resin -c 'npm install'
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1,
but npm-shrinkwrap.json was generated for lockfileVersion@2. I'll try to do my best with it!
...
npm ERR! code ENOLOCAL
npm ERR! Could not install from "node_modules/etcher-sdk/unbzip2-stream@github:balena-io-modules/unbzip2-stream#942fc218013c14adab01cf693b0500cf6ac83193" as it does not contain a package.json file.
```
I was able to reproduce the build warning and error on my laptop with:
```
$ rm -rf node_modules
$ npm install
$ npm run package
```
It happens with this PR's branch, but does not happen with the master branch. I gather that lockfileVersion 2 is produced by npm version 7: https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

It is described as backwards compatible but perhaps not sufficiently so. :-) The CLI only supports Node v10 and v12, and I suggest using the versions of npm that "ship with" those versions of Node.js -- currently npm v6.14.11:
```
$ nvm use 10
Now using node v10.23.3 (npm v6.14.11)

$ nvm use 12
Now using node v12.20.2 (npm v6.14.11)
```
After switching to npm v6.14.11, the shrinkwrap file may be reverted and fixed with:
```
$ rm -rf node_modules
$ npm install
```

package.json
danthegoodman1: Yeah npm did that for me, how nice.

pdcastro:
The current build error is:
```
npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/balena-cli - You cannot publish over the previously published versions: 12.40.1-notarization-ea41dfb379004e2e5df1a02c4ed41dc27d35df9c.
```
This one happens every now and then, and I never fully understood what the fix would be. Will try retesting.

pdcastro:
@balena-ci retest

danthegoodman1:
@balena-ci retest

package.json
```diff
@@ -235,13 +235,14 @@
...
     "js-yaml": "^3.14.0",
+    "JSONStream": "^1.0.3",
     "klaw": "^3.0.0",
     "livepush": "^3.5.0",
     "lodash": "^4.17.20",
     "minimatch": "^3.0.4",
     "mixpanel": "^0.10.3",
     "moment": "^2.27.0",
-    "moment-duration-format": "^2.3.2",
+    "moment-duration-format": "^2.3.2","net-keepalive": "^1.3.6",
```
pdcastro, line 245: The build is failing on Windows because `net-keepalive` was added here. Was it added by accident?

danthegoodman1: I guess so! I'll remove and commit

automation/build-bin.ts
```diff
@@ -343,6 +343,20 @@ async function signWindowsInstaller() {
 	}
 }
 
+/**
+ * Wait for Apple Installer Notarization to continue
+ */
+async function afterSignHook(): Promise<void> {
```
pdcastro, line 349: No need to make this change straight away, but I think this function's name would be clearer as `notarizeMacInstaller()`.

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

automation/build-bin.ts
```diff
@@ -354,6 +370,17 @@ export async function buildOclifInstaller() {
 	let packOpts = ['-r', ROOT];
 	if (process.platform === 'darwin') {
 		packOS = 'macos';
+		await new Promise((resolve, reject) => {
+			klaw('node_modules/node-unzip-2/testData/')
```
pdcastro, line 374: This will miss most of the zip files:
```
-rw-r--r--  1 0      0        1636 Mar  2 10:19 ./node_modules/unzip-stream/testData/compressed-flags-set/archive.zip
-rw-r--r--  1 0      0        1598 Mar  2 10:19 ./node_modules/unzip-stream/testData/compressed-OSX-Finder/archive.zip
-rw-r--r--  1 0      0        1636 Mar  2 10:19 ./node_modules/unzip-stream/testData/compressed-standard/archive.zip
...
```
To be more "robust", I'd suggest navigating the whole of `node_modules`, look for folders with the word "test" in the name (case insensitive) and delete any zip files under those folders. It should be fairly safe to delete zip files that used in test code.

error-e.json
```diff
@@ -0,0 +1,333 @@
+{
```
pdcastro, line 1: I think this file was committed by accident.

package.json
pdcastro: It looks like you're using `npm` v7 again, and committed these changes again, alongside a huge `npm-shrinkwrap.json` file that will break the build again. Please use `npm` v6.

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

automation/build-bin.ts
```diff
@@ -354,6 +370,39 @@ export async function buildOclifInstaller() {
...
+		// Sign all .node files first
+		await new Promise((resolve, reject) => {
+			klaw('node_modules/')
+				.on('data', async (item: { path: string; stats: Stats }) => {
+					if (!item.stats.isFile()) {
+						return;
+					}
+					if (path.basename(item.path).endsWith('.node')) {
+						await whichSpawn(
+							`productsign --sign "Developer ID Installer: Rulemotion Ltd (66H43P8FRG)" ${item.path}`,
```
pdcastro, line 399: This should be something like:
```js
await whichSpawn(
    'productsign',
    [
        '--sign',
        'Developer ID Installer: Rulemotion Ltd (66H43P8FRG)',
        item.path,
    ],
);
```

pdcastro: This is what makes it shell agnostic (independent of the OS shell being cmd.exe, bash, PowerShell, zsh, etc.) If it took a single long string as argument, it would need to use a shell to parse it. By taking a Javascript array of arguments, it does not need to get any shell involved.

automation/build-bin.ts
```diff
@@ -354,6 +370,41 @@ export async function buildOclifInstaller() {
...
+		});
+		// Sign all .node files first
+		await new Promise((resolve, reject) => {
+			klaw('node_modules/')
+				.on('data', async (item: { path: string; stats: Stats }) => {
+					if (!item.stats.isFile()) {
+						return;
+					}
+					if (path.basename(item.path).endsWith('.node')) {
+						await whichSpawn('productsign', [
```
pdcastro, line 398: Oh, can't use `await` inside an event handler like `on('data')`. The result is `UnhandledPromiseRejectionWarning` if the promise is rejected - https://ci.balena-dev.com/builds/803380
```
Usage: productsign [options] --sign <identity> <input-product-path> <output-product-path>

(node:65442) UnhandledPromiseRejectionWarning: Error: productsign failed with exit code 1:
"/usr/bin/productsign" [--sign,Developer ID Installer: Rulemotion Ltd (66H43P8FRG),/private/var/lib/concourse/worker/volumes/live/797930ff-065b-4eac-71a4-afa1b720e87e/volume/node_modules/fsevents/fsevents.node]
    at Object.whichSpawn (/private/var/lib/concourse/worker/volumes/live/797930ff-065b-4eac-71a4-afa1b720e87e/volume/automation/utils.ts:236:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:65442) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:65442) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

pdcastro: So there are 2 separate errors there:
* `productsign` exits with error code 1 and the only debugging output for that is:  
`Usage: productsign [options] --sign <identity> <input-product-path> <output-product-path>`  
I.e., a malformed command line. The way to further debug this is to run `productsign` manually on a fly shell, and/or to add a `console.error(...)` message printing the full command line before attempting to execute it.

* `UnhandledPromiseRejectionWarning` - this is thrown because the promise rejection is not handled in the `on('data')` event handler. The promise is rejected because `productsign` returns a non-zero error code.

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci retest

danthegoodman1:
@balena-ci rebase

pdcastro (approved):

Issue #1908, opened 2020-07-01
Allow setting `release_version` during a deploy/push operation


xginn8:
We now support setting a unique `release_version` property per release. We should allow that to be set during a `deploy`/`push` operation (and any others I may have missed), perhaps behind a flag?

related to https://github.com/balena-io/balena-io/issues/2219 and https://github.com/balena-io/balena-api/issues/2655

pdcastro:
Connects to:
* #892 Add 'balena push' flag to specify a release tag
* #1356 Successful balena deploy reports release commit rather than release id

pdcastro:
PR #2193 was closed without merging because (as I understand it) the following discussion concluded that the release version should be specified through of a contract file:
https://www.flowdock.com/app/rulemotion/resin-tech/threads/sxmoCgQkKE9-OPuJtf2tHnX--rf

The contract file being:
https://www.balena.io/docs/learn/deploy/deploy-with-balena-button/#balenayml-configuration-file

In the case of `balena push`, I believe the contract file is parsed by the balenaCloud builder. In the case of `balena deploy`, I believe it depends on issue #2183 [`balena deploy` should parse, validate, and create a contract as part of the release it creates](https://github.com/balena-io/balena-cli/issues/2183) being implemented.

pdcastro:
Connects to JF Product Improvement: [Draft releases and release versioning](https://jel.ly.fish/8ea1c390-9a85-402d-978c-4d31dcb0d235) (restricted access)

pdcastro:
By the way, at the API data model level, I hear that the `release.release_version` field is now deprecated in favor of `release.semver`. The balenaCloud web dashboard will present the value of `release.semver`.
Ref.: [FD thread](https://www.flowdock.com/app/rulemotion/public-s-community/threads/Rr7rfak-9UuJTTemEco2Klg_yaO) (restricted access)

pdcastro:
Closing this issue in favour of #2183 with regard to `balena deploy`. For `balena push` (cloud, not local mode), this feature is already implemented by the balenaCloud builder. The builder will set either the `release.semver` field or the `release.release_version` field of the release object (API model) (see [balena-builder PR 892](https://github.com/balena-io/balena-builder/pull/892) (restricted access)), depending on whether the `version` field in the [balena.yml contract file](https://www.balena.io/docs/learn/deploy/deploy-with-balena-button/#balenayml-configuration-file) looks like a [semver](https://semver.org/) by regex matching.

PR #2193, opened 2021-02-22
deploy: support passing in `release_version` as an arg

xginn8:
Connects-to: #1908
Depends-on: https://github.com/balena-io-modules/balena-release/pull/16
Change-type: minor
Signed-off-by: Matthew McGinn <matthew@balena.io>

### DIFF
--- a/lib/commands/deploy.ts
+++ b/lib/commands/deploy.ts
@@ -54,6 +54,7 @@ interface FlagsDef extends ComposeCliFlags, DockerCliFlags {
 	build: boolean;
 	nologupload: boolean;
 	'release-tag'?: string[];
+	'release-version'?: string;
 	help: void;
 }
 
@@ -94,6 +95,7 @@ ${dockerignoreHelp}
 		'$ balena deploy myApp',
 		'$ balena deploy myApp --build --source myBuildDir/',
 		'$ balena deploy myApp myApp/myImage',
+		'$ balena deploy myApp myApp/myImage --release-version "v1.1.1"',
 		'$ balena deploy myApp myApp/myImage --release-tag key1 "" key2 "value2 with spaces"',
 	];
 
@@ -124,6 +126,11 @@ ${dockerignoreHelp}
 		nologupload: flags.boolean({
 			description:
 				"don't upload build logs to the dashboard with image (if building)",
+		}),
+		'release-version': flags.string({
+			description: stripIndent`
+				Set release version to uniquely identify this release within your app.
+			`,
 		}),
 		'release-tag': flags.string({
 			description: stripIndent`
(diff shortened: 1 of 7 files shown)


lib/commands/deploy.ts
```diff
@@ -125,6 +127,11 @@ ${dockerignoreHelp}
 			description:
 				"don't upload build logs to the dashboard with image (if building)",
 		}),
+		'release-version': flags.string({
+			description: stripIndent`
+				Set release version to uniquely identify this release within your app.
```
pdcastro, line 132:
```suggestion
				Set a release version to uniquely identify this release within your app.
```

Also, kindly remind me: does the version have to be valid semver? Or any arbitrary string will do? If the former, we could add input validation in the CLI.

pdcastro (commented):
Nice 👍 
* The commit `Change-type` footer should be `minor` -- new flag, new feature.
* I suggest rewording the commit's first line to the following, which I think users will find  easier to understand in the Changelog file:  
***From:*** `deploy: support passing in release_version as an arg`  
***To:*** `deploy: Add --release-version flag to help identify a release in the app`

pdcastro:
Question: It doesn't need to be done in this PR, but the `--release-version` flag would be applicable to `balena push myApp` as well, right? I've linked this PR to #1908 with _"connects-to"_ rather than _"resolves"_ because that issue asks for the flag to be added to `deploy` and `push`.

pdcastro:
Ongoing discussion on whether we really want to have a `--release-version` flag:
https://www.flowdock.com/app/rulemotion/resin-tech/threads/sxmoCgQkKE9-OPuJtf2tHnX--rf

pdcastro:
Some context on why this PR was closed:

> (restricted access) https://www.flowdock.com/app/rulemotion/resin-tech/threads/sxmoCgQkKE9-OPuJtf2tHnX--rf
> [...] balena.yml being the source of truth for any versioning
> this will allow us to have versionbot control that value, and get us one step closer to deploying directly to balenaCloud
part of this spec: (restricted access) [balenaCloud-native CD from GitHub](https://docs.google.com/document/d/1UY38o-RCK-JgSKczbeai7JSWqGEAr9RDpO7Hmw1F8Kg/edit#heading=h.gsl21lx6udla)

> (restricted access) https://www.flowdock.com/app/rulemotion/resin-frontend/threads/y6ip5IRO-SUUNJZJoOLcn8SeNJp
> If the release version was just a CLI flag, it would not be version controlled and could contradict `balena.yml`
> In the past, there was also discussion that a `--release-version` flag could update or create a balena.yml file for the user, but there were objections on the grounds that running balena push and balena deploy should not result in "uncommitted changes" to a git repo

PR #2194, opened 2021-02-23
docs: emphasize that push emulation is not required in most cases

klutchell:
Docs clarification as a result of this forums post:
https://forums.balena.io/t/problems-with-balena-push/247082/38

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -2583,8 +2583,9 @@ Source directory to be sent to balenaCloud or balenaOS device
 
 #### -e, --emulated
 
-Don't use native ARM servers; force QEMU ARM emulation on Intel x86-64
-servers during the image build (balenaCloud).
+Don't use the faster, native balenaCloud ARM builders; force slower QEMU ARM
+emulation on Intel x86-64 builders. This flag is sometimes used to investigate
+suspected issues with the balenaCloud backend.
 
 #### --dockerfile DOCKERFILE
 
--- a/lib/commands/push.ts
+++ b/lib/commands/push.ts
@@ -135,8 +135,9 @@ export default class PushCmd extends Command {
 		}),
 		emulated: flags.boolean({
 			description: stripIndent`
-				Don't use native ARM servers; force QEMU ARM emulation on Intel x86-64
-				servers during the image build (balenaCloud).`,
+				Don't use the faster, native balenaCloud ARM builders; force slower QEMU ARM
+				emulation on Intel x86-64 builders. This flag is sometimes used to investigate
+				suspected issues with the balenaCloud backend.`,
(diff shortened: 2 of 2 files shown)


doc/cli.markdown
```diff
@@ -2584,7 +2584,8 @@ Source directory to be sent to balenaCloud or balenaOS device
 #### -e, --emulated
 
 Don't use native ARM servers; force QEMU ARM emulation on Intel x86-64
-servers during the image build (balenaCloud).
+servers during the image build (balenaCloud). This should only be used in
+special circumstances as the dedicated ARM builders are much faster.
```
pdcastro, line 2588: * The advice that a native ARM server/builder is faster is useful, but I think that the advice that this flag "should only be used in special circumstances" is less useful because there is not clarification what such circumstances might be. How about:
```
	Don't use the faster, native ARM servers; force slower QEMU ARM emulation on
	Intel x86-64 servers (balenaCloud). This flag is sometimes used to investigate
	suspected issues with the balenaCloud builders.
``` 
Alternatively, you could have the view that, if this flag is only useful for balena engineers while debugging the balena backend, then it should not even be visible to end users. A flag can have the `hidden: true` attribute so that it can still be used, but will not be listed with `balena help push`. However, in practice this would mean that balena engineers would be less likely to know or remember about this flag when they needed it. And also, if the flag is useful for balena engineers to debug the backend, it could also be useful for CLI end users to debug the backend before reaching out to balena's support.

* The `cli.markdown` file is auto generated and the build is expected to fail with an "uncommitted changes" error.
Make these changes to `lib/commands/push.ts` instead, run `npm run build` and then commit the changes to both `push.ts` and `cli.markdown`. More details in the Contributing file: https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#editing-documentation-files-readme-install-reference-website

pdcastro: Hmm. Reading the user's comment in the forum thread:
> _I misinterpreted the description in the documentation and thought that pushing an application on ARM bases was impossible on AMD64._

Perhaps the user interpreted the word "server" in "native ARM servers" to mean their workstation (which is Intel), not realising the documentation meant the balenaCloud servers. The proposed changes so far were also mixing the words "server" and "builder". Maybe it would be a bit clearer still as:
```
	Don't use the faster, native balenaCloud ARM builders; force slower QEMU ARM
	emulation on Intel x86-64 builders. This flag is sometimes used to investigate
	suspected issues with the balenaCloud backend.
```

klutchell: I actually had the same thought when I saw the forum post, I think saying "balenaCloud ARM builders" in all cases is more clear.

pdcastro (approved):

PR #2197, opened 2021-02-23
docs: Fix missing markdown docs for device `deactivate` and `local-mode`

pdcastro:
Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2199, opened 2021-02-28
Completion

khancyr:
Change-type: patch

This PR propose to add ZSH completion to Balena-cli. The completion is as basic as the bash one but that is a start.
It also put the bash-completion file into the same directory for easier maintenance.
The doc has been adjusted and a simple example is provided to enable the completion.


completion/zsh/_balena
```diff
@@ -0,0 +1,81 @@
...
+  local commands; commands=(
+            app apps build config deploy device devices env envs help key \
+            keys local login logout logs note os preload quickstart settings \
+            scan ssh util version whoami
+  )
+
+  # Sub-completions
+  app_cmds=( create restart rm )
+  config_cmds=( generate inject read reconfigure write )
+  device_cmds=( identify init move public-url reboot register rename rm \
```
pdcastro, line 18: The contents of file are mainly from the previous `balena-completion.bash`, right? Totally not your fault, but that file was quite outdated, missing new commands. Developers forget to update it. As it happens, now that we are fully using oclif, when `npm run build` is executed the following large-ish JSON file is automatically generated: `oclif.manifest.json` (It's not committed to the git repo, it's listed in `.gitignore`, but it's included in the CLI installers so you should be able to find it in the folder where the CLI is installed - except for the zip package...)
```
$ jq . < oclif.manifest.json
{
  "version": "12.40.0",
  "commands": {
    "apps": {
      "id": "apps",
      "description": "List all applications.\n\nlist all your balena applications.\n\nFor detailed information on a particular application,\nuse `balena app <application>` instead.",
      "usage": "apps",
      "pluginName": "balena-cli",
      "pluginType": "core",
      "aliases": [],
      "examples": [
        "$ balena apps"
      ],
      "flags": {
        "help": {
          "name": "help",
          "type": "boolean",
          "char": "h",
          "description": "show CLI help",
          "allowNo": false
        },
        "verbose": {
          "name": "verbose",
          "type": "boolean",
          "char": "v",
          "description": "No-op since release v12.0.0",
          "allowNo": false
        }
      },
      "args": []
    },
    "build": {
      "id": "build",
      "description": "Build a project locally.\n\n ...
      ...
```
The great thing about this JSON file is that it is auto generated in every CLI release (every time we run `npm run build`), and thus always up to date with all commands, subcommands and flags. I would like to tempt you to come up with a solution that auto generates this `_balena` file, having `oclif.manifest.json` as input. Do you know a bit of Javascript? The solution would be preferably written in Typescript or plain Javascript (Node.js). Say, this file:

`./completion/generate-completion.js`
* Input: `oclif.manifest.json`
* Output: `_balena`

Alternatively, `_balena` could dynamically load `oclif.manifest.json`, or some other auto generated file, but in this case, an extract of what matters should probably be cached somehow and not reloaded every time one hits the tab character on the command line, because `oclif.manifest.json` is quite chunky - 124KB in compact form.

Writing the solution in shell script may not be a good idea because of Windows compatibility: We would have to hook the code to run when `npm run build` is executed, which on Windows currently happens on `cmd.exe` (e.g. in balena CI prior to creation of the CLI's GUI installer for Windows). Ironically, we also want to have bash or zsh autocompletion on Windows, because end users may be using MSYS, Cygwin, WSL etc -- but that's after the CLI was built and installed.

Talking of Windows, files starting with an underscore, like `_balena`, are traditionally hidden files on Windows -- like files starting with a dot on Linux or macOS. For this reason, I suggest renaming `_balena` to something else. How about `completion.common` or `completion-common`? (Because it is used in common by `completion.bash` and `completion.zsh`.)

pdcastro (changes requested):
Hey @khancyr, thank you very much for your contribution! 👍 
I'm leaving some initial comments.

INSTALL-LINUX.md
```diff
@@ -15,7 +15,7 @@ Selected operating system: **Linux**
    `balena-cli` folder.
 
 3. Add the `balena-cli` folder to the system's `PATH` environment variable. There are several
-   ways of achieving this on Linux: See this [StackOverflow post](https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix). Close and reopen the terminal window
+   ways of achieving this on Linux: See this [StackOverflow post](https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix). The simpliest way is to edit you `.bashrc` or `.zshrc` files. They are on your Home directory but it is a hidden file (CTRL+H on Ubuntu to reveal them). Then put at the end of the file : `export PATH=PATH_TO_BALENA_CLI_DIRECTORY:$PATH`. Where PATH_TO_BALENA_CLI_DIRECTORY is the path to Balena-cli directory. After any change on `PATH`, close and reopen the terminal window
```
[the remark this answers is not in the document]
khancyr: Hello,
That explain why the doc was voluntary detailed for Linux installation. I understand your point and know well those user that need a map to go to the bathroom in a 1 room flat.
I agree that the apt install or a one-line installer would be the best, but normally for linux user giving an example how to set the PATH correctly is quite simple and well understand ! (that what is done on Arduino, java, ROS, ArduPilot, etc.) 
But I respect your choice, no issue !

doc/cli.markdown
```diff
@@ -39,10 +39,20 @@ are supported. Alternative shells include:
   [FAQ](https://github.com/balena-io/balena-cli/blob/master/TROUBLESHOOTING.md) for using the
   balena CLI with WSL and Docker Desktop for Windows.
 
-On **macOS** and **Linux,** the standard terminal window is supported. Optionally, `bash` command
-auto completion may be enabled by copying the
-[balena-completion.bash](https://github.com/balena-io/balena-cli/blob/master/balena-completion.bash)
-file to your system's `bash_completion` directory: check [Docker's command completion
+On **macOS** and **Linux,** the standard terminal window is supported.
+
+##### Optionally
```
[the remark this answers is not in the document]
khancyr: oh, my bad, I tried to follow correctly which files to edit but I should have mixed stuff

completion/zsh/_balena
khancyr: Yep, I just copied the existing completion, adding a full regex from cmdline help completion file isn't a great first commit ^^

If you got a json that explain all cmdline options, I could forge the completion with it and add it as part of the install and update process. I will look at that.

pdcastro: As a coincidence, while working on another PR (#2159), I came across the following `jq` expression that could be used for inspiration:
```bash
# print a list of top-level balena CLI commands
$ jq -r '.commands | keys | map(.[0:index(":")]) | unique | join(" ")' < oclif.manifest.json

api-key app apps build config deploy device devices env envs internal join key keys leave local login logout logs note orgs os preload push scan settings ssh support tag tags tunnel util version whoami
```
`jq` docs: https://stedolan.github.io/jq/manual/

khancyr:
@pdcastro hello,

I have made a new version that have .js file that take as input `oclif.manifest.json` file and generate both bash and zsh completion file.
I am not JS expert so, I will be listing any comments on the code I made.
I don't know yet how to include the file as part of building, any guidance ?

for those that want to give a try : 
`node completion/generate-completion.js`
to generate the completion files. Then, add `source PATH_TO_BALENA_CLI_DIRECTORY/completion/completion.bash` to your .bashrc or `source "PATH_TO_BALENA_CLI_DIRECTORY/completion/completion.zsh"` on you .zshrc

completion/generate-completion.js
```diff
@@ -0,0 +1,95 @@
+const path = require('path');
```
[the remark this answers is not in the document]
pdcastro: > I don't know yet how to include the file as part of building, any guidance ?

I think we can add it to the `build:src` npm script in the `package.json` file. Talking of which, the `completion` folder could also be added to the `lint` npm script in `package.json`.

I have made these two changes and ran `npm run lint`. The result is in this diff:
https://github.com/khancyr/balena-cli/compare/completion...balena-io:completion-lint

Most changes are a result of running `npm run lint`. Feel free to copy any changes from there to this PR.

pdcastro (commented):
Thank you for this contribution @khancyr 👍

completion/generate-completion.js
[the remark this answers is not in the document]
khancyr: Excellent ! Thanks for the correction, I wasn't sure how to do that. I have just cherry-pick your patch

.gitignore
```diff
@@ -34,3 +34,5 @@
 /package-lock.json
 /resinrc.yml
 /tmp/
+/completion/bash/
+/completion/zsh/
```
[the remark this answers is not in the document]
khancyr: That is corrected now. 
For the generation that was an open question as I didn't know how you will to integrate this the in the installation flow, but it seems solved now !

khancyr:
@pdcastro it should be good now ! I can squash the commits if that is better

completion/completion.bash
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+_BALENA_COMPLETION_DIR=$(builtin cd "`dirname "${BASH_SOURCE[0]}"`" > /dev/null && pwd)
+source "$_BALENA_COMPLETION_DIR/bash/balena_comp"
```
pdcastro, line 4: What is the purpose of this `completion.bash` file? (And the `completion.zsh` file.) If `completion.bash` is executed through the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)), thus creating a child process, for example:
```sh
cd balena-cli
chmod +x ./completion/completion.bash
./completion/completion.bash
```
... then I understand that the `source` line will have no practical effect, because it will only apply to the child process that quickly terminates. In other words, it doesn't work. :-) It works if this file is sourced:
```sh
cd balena-cli
source ./completion/completion.bash
```
But in this case, we could directly source `./completion/bash/balena_comp` instead (which is what the README documents, as amended by this PR) -
```sh
cd balena-cli
source ./completion/bash/balena_comp
```
And I think it is somewhat confusing to end users to have these 2 files:
```sh
completion/completion.bash
completion/bash/balena_comp
```
or
```sh
completion/completion.zsh
completion/bash/_balena
```
Unless there are some specific reasons to do it like you did, I would suggest the following:
* Delete the old (existing before this PR) `balena-completion.bash`
* Delete `completion/completion.bash`
* Delete `completion/completion.zsh`
* Rename/move `completion/bash/balena_comp` to `completion/balena-completion.bash`
* Rename/move `completion/zsh/_balena` to `completion/balena-completion.zsh`
* Delete subfolder `completion/bash`
* Delete subfolder `completion/zsh`

To make it even clearer, I would also move the template files to a template subfolder:
* Rename/move `completion/balena_bash_template` to `completion/templates/bash.template`
* Rename/move `completion/balena_zsh_template` to `completion/templates/zsh.template`

So we would end up with the following structure:
```sh
completion/balena-completion.bash
completion/balena-completion.zsh
completion/generate-completion.js
completion/templates/bash.template
completion/templates/zsh.template
```

README.md
```diff
@@ -50,7 +50,7 @@ are supported. Alternative shells include:
 
 On **macOS** and **Linux,** the standard terminal window is supported. Optionally, `bash` command
 auto completion may be enabled by copying the
-[balena-completion.bash](https://github.com/balena-io/balena-cli/blob/master/balena-completion.bash)
+[balena_comp](https://github.com/balena-io/balena-cli/blob/master/completion/bash/balena_comp)
```
pdcastro, line 53: After changes are made to some source code files and to some sections of the README file, including this section, `npm run build` should be run in order to update the `doc/cli.markdown` file, which is the source of the [reference website](https://www.balena.io/docs/reference/balena-cli/). This PR's CI build is currently failing because the automatic changes to `doc/cli.markdown` (made when `npm run build` is run) were not committed:
```
-[balena-completion.bash](https://github.com/balena-io/balena-cli/blob/master/balena-completion.bash)
+[balena_comp](https://github.com/balena-io/balena-cli/blob/master/completion/bash/balena_comp)
 
Error: "catch-uncommitted": Error: npx failed with exit code 1:
"/Users/balena/.nvm/versions/node/v12.20.0/bin/npx" [catch-uncommitted,--catch-no-git,--skip-node-versionbot-changes,--ignore-space-at-eol]
```
So basically, to fix this:
```
npm run build
git add doc/cli.mardown
git commit
```

pdcastro (commented):
> I can squash the commits if that is better

Yes please -- the PR currently has 5 commits, and each would automatically become a bullet point in the [CHANGELOG](https://github.com/balena-io/balena-cli/blob/master/CHANGELOG.md) file. Feel free to squash and force push - thanks! 🙏

completion/completion.bash
khancyr: The completion.bash/.zsh allow to be an interface for completion script. You just need to source them once and further changes on completion will be automatically loaded on update. This also allow to have multiple completion script loaded for user without them to have to manually source them.

but yes for balena-cli, there is only one tools so the interface isn't needed ...
One issue is that zsh is expecting a file starting with _ for completion and we could only add a whole directory to the completion search path

README.md
khancyr: ok done !

completion/completion.bash
khancyr: I have modify accordingly. Only the zsh completion file name was keeped to _balena as this is needed to work

pdcastro:
Hi @khancyr, sorry for the delay in replying. Thanks for implementing the changes, 👍    it's looking good! 💯   Would you mind rebasing it on master, so that I can merge the PR? (I tried the `@balena-ci rebase` command that should have done this for us, but I gather that our CI cannot yet cope with PRs from a forked repo -- something that the CI team is working on.)

On a side note, I have noticed that, when I use the bash completion script, I lose file/path auto completion. For example, if the completion script is **not** loaded, I can auto complete the "Downloads" word on the following line, by hitting the tab key:
```
$ balena push test-rpi -s ~/Down<hit tab here, 'Downloads' auto completes>
```
However when I load the CLI completion script:
```
$ source ./completion/completion.bash
$ balena push test-rpi -s ~/Down<hit tab here, no auto completion :-( >
$ balena preload ~/Down<hit tab here, no auto completion :-( >
```

This limitation already existed before this PR and I am happy to merge it -- the PR makes a great improvement over what we had. I am mentioning this issue just in case you knew a way of easily solving it. :-) But it can be left for a future PR, no worries.

khancyr:
@pdcastro wonderfull. Indeed, as we are trying to catch everthing we don't complete the directories anymore.
I will fix that in another PR that will bring the right completion for each word !

pdcastro (approved):

PR #2204, opened 2021-03-05
Notarization

danthegoodman1:
Update macOS installer to avoid Apple's warning pop-up

Call notarization hook in build-bin.ts

Change-type: patch

Connects-to: https://github.com/balena-io/balena-cli/issues/1479, https://github.com/balena-io/balena-cli/pull/2186


danthegoodman1:
Idk why it created here...

PR #2207, opened 2021-03-06
build, deploy: Fix error handling when QEMU download fails

pdcastro:
Fixes a bug where the CLI would leave behind zero-length QEMU executable files if the user specified an invalid architecture on the command line for the build/deploy commands, e.g. `--arch foo` instead of `--arch armv7hf`, and subsequent executions would no longer re-download QEMU because a QEMU file already existed (despite being corrupt).

Sample broken files before this PR:
```
# ls -l ~/.balena/bin/
total 5476
-rw-r--r-- 1 root root       0 Dec 17 23:57 qemu-execve-arm-v4.0.0+balena2
-rwxr-xr-x 1 root root 5606984 Dec 17 23:30 qemu-execve-armv7hf-v4.0.0+balena2
-rw-r--r-- 1 root root       0 Dec 17 23:57 qemu-execve-i386-v4.0.0+balena2
-rw-r--r-- 1 root root       0 Dec 17 23:57 qemu-execve-x86-v4.0.0+balena2
```

This PR also "standardizes" the cached file names to use the "QEMU arch" instead of the "balena arch", meaning that there can only be 2 variants on disk: `arm` and `aarch64` (as opposed to multiple other archs like `rpi`, `armv7hf`, `i386` etc).

Note: this PR is a spin-off PR #2145 (commit extracted from there, to simplify that PR).

Change-type: patch

### DIFF
--- a/lib/utils/qemu.ts
+++ b/lib/utils/qemu.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017-2020 Balena Ltd.
+ * Copyright 2017-2021 Balena Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
  */
 
 import type * as Dockerode from 'dockerode';
+
+import { ExpectedError } from '../errors';
 import { getBalenaSdk, stripIndent } from './lazy';
 import Logger = require('./logger');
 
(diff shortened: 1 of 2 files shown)


lib/utils/qemu.ts
```diff
@@ -79,64 +82,76 @@ export const getQemuPath = function (arch: string) {
...
-					.on('finish', function () {
-						fs.chmodSync(qemuPath, '755');
-						resolve();
-					});
-			}),
-	);
+		});
+	} catch (err) {
+		try {
+			await fs.promises.unlink(qemuPath);
```
pdcastro, line 133: The `fs.createWriteStream(qemuPath)` call earlier in this function creates a zero-length file on disk. On successful download, the file is expanded to the size of the download. On errors, before this PR the file was left behind and, the next time the CLI was executed, the CLI would assume the file no longer needed to be downloaded (a bug). This `unlink` call deletes "corrupt" files.

pdcastro:
@balena-ci I self-certify!

Self-certifying because the commit in this PR had already been reviewed by @srlowe in PR #2145, e.g. this comment:
https://github.com/balena-io/balena-cli/pull/2145/commits/e09392153867baa785d056abfdb3a97fc920c71a#r547340164

pdcastro:
@balena-ci I self-certify!

PR #2206, opened 2021-03-06
Adds info about fixing "sudo balena" issues to docs

frederikheld:
Resolves: #2149   
Change-type: patch   
Depends-on: none   
See: none   

---
Please check the CONTRIBUTING.md file for relevant information and some
guidance. Keep in mind that the CLI is a cross-platform application that runs
on Windows, macOS and Linux. Tests will be automatically run by balena CI on
all three operating systems, but this will only help if you have added test
code that exercises the modified or added feature code.

Note that each commit message (currently only the first line) will be
automatically copied to the CHANGELOG.md file, so try writing it in a way
that describes the feature or fix for CLI users.

If there isn't a linked issue or if the linked issue doesn't quite match the
PR, please add a PR description to explain its purpose or the features that it
implements. Adding PR comments to blocks of code that aren't self explanatory
usually helps with the review process.

If the PR introduces security considerations or affects the development, build
or release process, please be sure to highlight this in the PR description.

Thank you very much for your contribution!

Issue #2149, opened 2021-01-01
`balena scan` sudo issue


frederikheld:
System is Ubuntu 20.04 (64 Bit)
balena-cli version is 12.36.1

(the issue exists with balena-cli v12.40.2 on Ubuntu 20.10 (64 Bit) as well)

When I run `$ balena scan` as described in the [balena-cli master class](https://www.balena.io/docs/learn/more/masterclasses/cli-masterclass/), I get the message "You need root/admin privileges to run this command". So I run `$ sudo balena scan`, which gives me the following message "sudo: balena: command not found".

This is very odd. What am I doing wrong?

Full terminal output:
```sh
frederik@mycomputer:~$ balena scan
You need root/admin privileges to run this command

frederik@mycomputer:~$ sudo balena scan
[sudo] password for frederik: 
sudo: balena: command not found
frederik@mycomputer:~$ balena version -a
balena-cli version "12.36.1"
Node.js version "12.18.1"
frederik@mycomputer:~$ whereis balena
balena: /opt/balena-cli/balena
```
Installed via zip package as described [here](https://github.com/balena-io/balena-cli/blob/master/INSTALL-LINUX.md). Instead of adding the extracted path to $PATH, I copied it to `/opt/balena-cli` and added `export PATH="$PATH:/opt/balena-cli"
` to `~/.bashrc`. I came to the decision to to do it this way after reading a lot about the history of linux file system and install paths. But I'm not an expert about what side-effects this choice might have. So may this cause the issue?

In an attempt to fix this, I uninstalled `balena-cli` and tried to install it via npm as described [here](https://github.com/balena-io/balena-cli/blob/master/INSTALL-ADVANCED.md), but it failed with a long list of warnings and errors:

```sh
frederik@mycomputer:~$ sudo rm -r /opt/balena-cli
frederik@mycomputer:~$ sudo npm install balena-cli -g --unsafe-perm --production
[...]
npm WARN tar ENOENT: no such file or directory, open '/usr/lib/node_modules/.staging/es-abstract-f6c79366/helpers/maxSafeInteger.js'
npm WARN tar ENOENT: no such file or directory, open '/usr/lib/node_modules/.staging/es-abstract-f6c79366/2015/MinFromTime.js'
npm ERR! code 128
npm ERR! Command failed: git clone --mirror -q git://github.com/balena-io-modules/unbzip2-stream.git /root/.npm/_cacache/tmp/git-clone-74700e65/.git
[...]

PR #2208, opened 2021-03-06
Linux installation instructions: Add sudo configuration section

pdcastro:
Connects-to: #2149
Change-type: patch

This PR was inspired by #2206 and special thanks go to @frederikheld 🙏

### DIFF
--- a/INSTALL-LINUX.md
+++ b/INSTALL-LINUX.md
@@ -11,35 +11,52 @@ Selected operating system: **Linux**
    with "-standalone.zip", for example:  
    `balena-cli-vX.Y.Z-linux-x64-standalone.zip`
 
-2. Extract the zip file contents to any folder you choose. The extracted contents will include a
-   `balena-cli` folder.
+2. Extract the zip file contents to any folder you choose, for example `/home/james`.
+   The extracted contents will include a `balena-cli` folder.
 
-3. Add the `balena-cli` folder to the system's `PATH` environment variable. There are several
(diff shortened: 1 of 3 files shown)


INSTALL-LINUX.md
```diff
@@ -11,35 +11,52 @@ Selected operating system: **Linux**
...
+If your Linux distribution has an `/etc/sudoers` file that defines a `secure_path`
+setting, run `sudo visudo` to edit it and add the balena CLI's installation folder to
+the ***pre-existing*** `secure_path` setting, for example:
+
+```text
+Defaults   secure_path="/home/james/balena-cli:<pre-existing entries go here>"
+```
+
+If an `/etc/sudoers` file does not exist, or if it does not contain a pre-existing
+`secure_path` setting, do not change it.
```
pdcastro, line 44: As documented at:
> https://www.sudo.ws/man/1.8.18/sudoers.man.html
> `secure_path`
> Path used for every command run from `sudo`. If you don't trust the people running `sudo` to have a sane `PATH` environment variable you may want to use this. Another use is if you want to have the “root path” be separate from the “user path”. Users in the group specified by the _exempt_group_ option are not affected by _secure_path._ This option is not set by default.

ab77:
@balena-ci retest

ab77:
@balena-ci rebase

pdcastro:
@balena-ci rebase

srlowe (approved):

PR #2210, opened 2021-03-12
Add macOS uninstall script (sudo /usr/local/lib/balena-cli/bin/uninstall)

pdcastro:
The uninstall script is actually a "harmless side effect" of updating the `@oclif/dev-cli` dependency from v1.22.2 to v1.26.0, which was done with the purpose of facilitating the macOS installer notarization as further discussed in this FD thread (private): https://www.flowdock.com/app/rulemotion/i-cli/threads/Kt78PVEE0OMrH1szWdTn6lc1xgS 

Another benefit of this PR is that `npm run build:installer` should run a bit faster on macOS and Windows, because `npm install --production` was replaced with cloning `node_modules` and running `npm prune --production`.

Change-type: minor
Connects-to: #2186


patches/all/@oclif+dev-cli+1.26.0.patch
```diff
@@ -0,0 +1,250 @@
...
++async function msysExec(cmd, options = {}) {
++    if (process.platform !== 'win32') {
++        return qq.x(cmd, options);
++    }
++    const sh = MSYSSHELLPATH;
++    const args = ['-c', cmd];
++    console.error(`[debug] @oclif/dev-cli msysExec sh="${sh}" args=${JSON.stringify(args)} options=${JSON.stringify(options)}`);
++    return qq.x(sh, args, options);
++}
++exports.msysExec = msysExec;
```
pdcastro, line 250: Some of code in this file was moved here from `qqjs+0.3.11.dev.patch` and `qqjs++execa+0.10.0.dev.patch`, which allowed deleting those other two patch files. I think it is beneficial to concentrate the patches in fewer modules and files, as it reduces the amount of patch errors / merge conflicts when dependencies are updated. This PR also deleted more than 500 lines of code, some of which in patch files.

patches/all/@oclif+dev-cli+1.26.0.patch
```diff
@@ -0,0 +1,250 @@
...
++        if (process.platform === 'win32') {
++            // xcopy is much faster than `qq.cp(source, ws)`
++            await qq.x(`xcopy "${source}" "${ws}\\node_modules" /S /E /B /I /K /Q /Y`);
++        } else {
++            // use the shell's `cp` on macOS in order to preserve extended
++            // file attributes containing `codesign` digital signatures
++            await qq.x(`cp -pR "${source}" "${ws}"`);
+         }
++        console.error(`[debug] @oclif/dev-cli running "npm prune --production" in "${ws}"`);
++        await qq.x('npm prune --production');
```
pdcastro, line 124: This is the main change that helps with the notarization of macOS installers. Previously (code deleted 15 lines above), oclif would run `npm install --production` to produce a production install of the CLI that would then be packaged in the GUI installer. With these changes, we instead clone the `node_modules` folder of a development installation, preserving extended file attributes such as the macOS `codesign` digital signature of `.node` native module files (PR #2186), and then run `npm prune --production` to remove dev packages. This is similar to what we've already been doing for the CLI Docker images (check [this line](https://github.com/balena-io/balena-cli/blob/v12.42.2/docker/debian/Dockerfile#L17)), so it's been already tested a bit, and I found that it is also overall faster than running `npm install --production`, as it avoids the compilation of the native node modules.

pdcastro:
@balena-ci I self certify!

PR #2218, opened 2021-03-17
Public address

chriswiggins:
Change-type: patch

This PR updates the `balena-sdk` to the latest, and with that, displays the public_address information on the command line for a device


lib/commands/device/index.ts
```diff
@@ -169,6 +170,7 @@ export default class DeviceCmd extends Command {
 				'is_online',
 				'ip_address',
 				'mac_address',
+				'public_address',
```
pdcastro, line 173: Nitpick: Maybe move this one line up (between `ip_address` and `mac_address`), so that the private and public IP addresses are displayed next to each other?

pdcastro (changes requested):
Thank you for this contribution! The build is failing early on because the commits (at least one of them, I think) is/are missing a `Change-type:` footer as described in:
https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#semantic-versioning-commit-messages-and-the-changelog

I suggest:

* `Updates to latest balena-sdk` → `Change-type: patch`
* `Display public address on cli for device` → `Change-type: minor`

Also, because each commit message automatically becomes a bullet point in the [CHANGELOG](https://github.com/balena-io/balena-cli/blob/master/CHANGELOG.md) file, which is the source of "what's new" for CLI end users, and because `balena-sdk` supports nested changelogs when a specific commit message format is used (also as described in the contributing doc), I would suggest rewording the commit messages to:

* `Update balena-sdk from 15.20.0 to 15.29.0` → `Change-type: patch`
* `balena device: Display public IP address field` → `Change-type: minor`

chriswiggins:
@pdcastro Thanks for all the advice! I've updated those now. Hopefully this is what you were after

pdcastro:
Thanks @chriswiggins 👍  Almost there, but now the PR has a 3rd commit that reads `Merge branch 'master' into public-address`. Could you drop that commit and use rebase instead of merge? (I.e., rebase the `public-address` branch on the `master` branch, instead of merging the `master` branch on the `public-address` branch.) We follow this practice across all balena repos.

By the way, when the PR gets outdated before review approval, we can also add a `@balena-ci rebase` comment to the PR, which causes versionbot / balena CI to automatically do the rebase for us (as an alternative to doing it locally on the command line).

chriswiggins:
Done :-)

pdcastro (approved):

PR #2224, opened 2021-03-19
fix proxy support bullet point spacing

gantonayde:
component: Make `Proxy Support` sub-bullet point spacing uniform

Change-type: patch
Signed-off-by: Genadi Naydenov genadi@balena.com

Resolves: balena-io/docs#797

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -68,7 +68,6 @@ HTTP(S) proxies can be configured through any of the following methods, in prece
   file](https://www.npmjs.com/package/balena-settings-client#documentation). It may be:
   * A string in URL format, e.g. `proxy: 'http://localhost:8000'`
   * An object in the format:
-
     ```yaml
     proxy:
         protocol: 'http'


gantonayde:
Before and after preview

bullets

doc/cli.markdown
```diff
@@ -68,7 +68,6 @@ HTTP(S) proxies can be configured through any of the following methods, in prece
   file](https://www.npmjs.com/package/balena-settings-client#documentation). It may be:
   * A string in URL format, e.g. `proxy: 'http://localhost:8000'`
   * An object in the format:
-
```
pdcastro, line 71: `cli.markdown` is automatically generated: https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#editing-documentation-files-readme-install-reference-website

This change should go to README.md instead. By the way, this change causes markdownlint (running on my laptop only) to complain about rule MD031:

> https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines
> MD031 - Fenced code blocks should be surrounded by blank lines

That's OK if it actually fixes balena-io/docs/issues/797, but I fear that future linting might cause a regression.

pdcastro (changes requested):

doc/cli.markdown
pdcastro: > This change should go to README.md instead

After which, `npm run build` needs to be executed to update `cli.markdown`.

pdcastro: Oh, I've found more context about the issue here: https://github.com/balena-io/docs/pull/1654#issuecomment-802652571
It doesn't change the previous comments though. :-)

pdcastro: And only now I see that the screenshot was also added to this PR too. :-)

mbalamat: @pdcastro @gantonayde I also think WRT to the commit message that Resolves should be above the Change-type/Signed-off-by block and it should have : in the end so it would be `Resolves: <url>`

pdcastro: That's probably why VersionBot didn't like the commit:
https://github.com/balena-io/balena-cli/pull/2224/checks?check_run_id=2148327228

gantonayde: I used the `Improve this doc` button [here](https://www.balena.io/docs/reference/balena-cli/) which points directly to `cli.markdown` instead of the README.md. (I assume this is another issue?)

@Page- has another fix which is more uniform so the spacing only between these two bullet points might not be such a big deal

pdcastro: > I used the `Improve this doc` button [here](https://www.balena.io/docs/reference/balena-cli/) which points directly to `cli.markdown` instead of the README.md. (I assume this is another issue?)

Yes: https://github.com/balena-io/docs/issues/1619

gantonayde: I agree that fixing the visual markdown glitch by not following the linting rules might not be the best practice. The problem here is that the code block is below a sub-bullet point rather than a bullet point. Let's leave it as it is then.

pdcastro: I'm OK with approving this PR with the lint warning -- after wall, someone took the time to create issue https://github.com/balena-io/docs/issues/797 :-)
In the future, the "proper fix" would be to incorporate markdown linting in the CLI's `package.json` (not just my VS Code plugin), and add rule exceptions for this case.

gantonayde: Let's close this one as I see there is another issue with the section above after the latest fix. 
I will edit the README.md and open a new one.

Issue #2226, opened 2021-03-19
docker-compose.dev.yml causes volumes to become invalid


sgrimm:
# Expected Behavior

Creating a `docker-compose.dev.yml` file with a service definition overrides that service's definition but has no impact on the rest of the project as defined in `docker-compose.yml`.

# Actual Behavior

If `docker-compose.yml` has a `volumes` section with a volume called `chromium-settings` and the local device's IP address is 192.168.1.73, `balena push 192.168.1.73` bombs out with an error:

```
data/volumes/chromium-settings should be object,null
```

Output when I add `--debug`:

```
[Debug]   Using build source directory: . 
[Debug]   Pushing to local device: 192.168.1.73
[Debug]   Checking we can access device
[Debug]   Sending request to http://192.168.1.73:48484/ping
[Debug]   Checking device supervisor version: 12.4.3
[Info]    Starting build on device 192.168.1.73
[Debug]   Loading project...
[Debug]   Resolving project...
[Debug]   docker-compose.yml file found at "."
[Info]    Docker compose dev overlay detected (docker-compose.dev.yml) - merging.
[Debug]   Creating project...
data/volumes/chromium-settings should be object,null

ValidationError: data/volumes/chromium-settings should be object,null
    at Object.normalize (/usr/local/lib/balena-cli/node_modules/resin-compose-parse/build/compose.js:77:19)
    at createProject (/usr/local/lib/balena-cli/build/utils/compose.js:30:33)
    at Object.loadProject (/usr/local/lib/balena-cli/build/utils/compose_ts.js:79:12)
    at async Object.deployToDevice (/usr/local/lib/balena-cli/build/utils/device/deploy.js:97:21)
    at async PushCmd.pushToDevice (/usr/local/lib/balena-cli/build/commands/push.js:94:13)
    at async PushCmd.run (/usr/local/lib/balena-cli/build/commands/push.js:37:17)
    at async PushCmd._run (/usr/local/lib/balena-cli/node_modules/@oclif/command/lib/command.js:43:20)
    at async Config.runCommand (/usr/local/lib/balena-cli/node_modules/@oclif/config/lib/config.js:175:24)
    at async CustomMain.run (/usr/local/lib/balena-cli/node_modules/@oclif/command/lib/main.js:27:9)
    at async CustomMain._run (/usr/local/lib/balena-cli/node_modules/@oclif/command/lib/command.js:43:20)
    at async Promise.all (index 1)
[...]

PR #2227, opened 2021-03-20
push: Fix docker-compose.dev.yml serialization ("should be object,null" error)

pdcastro:
Resolves: #2226 
Change-type: patch

The following `volumes` section is typical in a `docker-compose.yml` file of a balena application:
```
volumes:
  resin-data:
```

Before this PR, when `docker-compose.dev.yml` was merged with `docker-compose.yml` and then serialized, the above would become:

```
volumes:
  resin-data: null
```

And then `balena push` would result in:
```
$ balena push 192.168.1.46
[Info]    Starting build on device 192.168.1.46
[Info]    Docker compose dev overlay detected (docker-compose.dev.yml) - merging.
data/volumes/resin-data should be object,null
```

According to the Yaml spec, `resin-data:` means the same as `resin-data: null`, so that error was unexpected. I thought it might thus be a case of incorrect validation (a case of fixing the `resin-compose-parse` module) but, on investigation, the schema of `docker-compose.yml` already defines `volume` as `"type": ["object", "null"]`, which I believe is correct:

https://github.com/balena-io-modules/resin-compose-parse/blob/v2.1.3/src/schemas/config_schema_v2.1.json#L34
```
    "volume": {
      "id": "#/definitions/volume",
      "type": ["object", "null"],
```

So I don't know if there is something to fix there. In any case, I found that `js-yaml` v4.0.0 introduces a new serialization option `styles: { '!!null': 'empty' }` that causes the serialization to produce:
```
volumes:
  resin-data:
```

Which then avoids the error of issue #2226. So that's what I did in this PR.
cc: @srlowe


automation/update-module.ts
```diff
@@ -58,7 +58,7 @@ const getUpstreams = async () => {
 	const repoYaml = fs.readFileSync(__dirname + '/../repo.yml', 'utf8');
 
 	const yaml = await import('js-yaml');
-	const { upstream } = yaml.safeLoad(repoYaml) as {
+	const { upstream } = yaml.load(repoYaml) as {
```
pdcastro, line 61: `safeLoad` was removed in `js-yaml` v4.0.0 (breaking change), which is updated by this PR.

lib/utils/compose_ts.ts
```diff
@@ -179,13 +179,18 @@ async function mergeDevComposeOverlay(
...
 		try {
-			const yaml = await import('js-yaml');
-			const compose = yaml.load(composeStr);
-			const devOverlay = yaml.load(await fs.readFile(devOverlayPath, 'utf8'));
+			const compose = loadObj(composeStr);
+			const devOverlay = loadObj(await fs.readFile(devOverlayPath, 'utf8'));
 			// We only want to merge the services section
 			compose.services = { ...compose.services, ...devOverlay.services };
-			composeStr = yaml.dump(compose);
+			composeStr = yaml.dump(compose, { styles: { '!!null': 'empty' } });
```
pdcastro, line 193: `{ styles: { '!!null': 'empty' } }` is what fixes issue #2226. This option was introduced in `js-yaml` v4.0.0, which is why this PR updates `js-yaml`.

pdcastro:
@balena-ci I self-certify!

pdcastro:
@srlowe, do you think that the root cause of #2226 is in `resin-compose-parse`? I was going to add a test case to this PR to ensure `resin-data:` instead of `resin-data: null`, but then I thought that perhaps we should make `resin-data: null` work, instead of avoiding it.

PR #2228, opened 2021-03-24
Docs: Update INSTALL-MAC.md - Fix incorrect reference to Windows OS

maggie0002:
Fix incorrect reference to Windows OS.

Hope I have interpreted this correctly? 18.06.1 is years old, is it really still a requirement? What if Balena Engine is installed on the Mac, will it use that instead and then Docker can be kept up to date? Running a version of Docker from 2018 is going to have a lot of impact on other work.

### DIFF
--- a/INSTALL-MAC.md
+++ b/INSTALL-MAC.md
@@ -65,7 +65,7 @@ restriction that Docker must be installed on the local machine (because Docker's
 feature is used). Also, for some device types (such as the Raspberry Pi), the `preload` command
 requires Docker to support the [AUFS storage
 driver](https://docs.docker.com/storage/storagedriver/aufs-driver/). Unfortunately, Docker Desktop
-for Windows dropped support for the AUFS filesystem in Docker CE versions greater than 18.06.1. The
+for Mac dropped support for the AUFS filesystem in Docker CE versions greater than 18.06.1. The
 present workaround is to either:
 
 * Downgrade Docker Desktop to version 18.06.1. Link: [Docker CE for


pdcastro:
Thanks for this fix @maggie0002. 👍  The CI build is failing because the commit needs a `Change-type: patch` footer, as further detailed at: https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#semantic-versioning-commit-messages-and-the-changelog

FYI, the commit message will be automatically copied to the Changelog file when the PR is merged. While editing the commit to add the `Change-type` footer, perhaps at the same time the following part of the commit message could be dropped:

>Hope I have interpreted this correctly? 18.06.1 is years old, is it really still a requirement? What if Balena Engine is installed on the Mac, will it use that instead and then Docker can be kept up to date? Running a version of Docker from 2018 is going to have a lot of impact on other work.

Answering the questions, you're right that Docker v18.06.1 is very old and only getting older. The solution we are gradually deploying is to convert balenaOS images from AUFS to Overlay2. For example, I understand that recent versions of balenaOS for the Raspberry Pi 4 already use Overlay2, and therefore the `balena preload` command can be used with latest version of Docker for Mac or Windows. For older balenaOS images that still used AUFS, a Linux Virtual Machine (e.g. Ubuntu 20.04 on VirtualBox) could be used instead of downgrading Docker. Despite these options, some users still prefer the alternative of downgrading Docker.

> What if Balena Engine is installed on the Mac, will it use that instead and then Docker can be kept up to date?

hmm I have never tried running balenaEngine and Docker at the same time. I suppose it is possible, as long as they don't attempt to listen on the same port numbers. The `balena preload` command has the `--dockerHost` flag to specify an alternative engine -- either balenaEngine or Docker.

After this PR is resolved, I can further update the docs to add the clarification that newer versions of balenaOS (for some device types) no longer require downgrading Docker.

maggie0002:
Thanks for the info. Will have to have a think about the best way to go, probably the Virtual Machine as you mention. 

Will close this PR, no point merging a one word change if you are going to do larger updates.

Thanks again.

Issue #1099, opened 2019-02-13
Preload fails on macOS with error "error initializing graphdriver: driver not supported"


mschm:
- **Cli version: 9.12.2**
- **Standalone install: no**
- **Node version: 11.1.0**
- **Npm version: 6.4.1**
- **Operating system and architecture: macOS 10.14.3**
- **Docker version: 18.09.1 stable (Docker Desktop for Mac)**

---

First off, the preloader is an awesome tool and I really appreciate it!
I have used it successfully in a Linux environment, but under macOS it fails (using the Intel NUC base image https://files.resin.io/resinos/intel-nuc/2.29.2%2Brev2.prod/image/balena.img.zip)

```
$ balena preload balena.img
Building Docker preloader image. [========================] 100%



/ Creating preloader container
\ Starting preloader container
| Reading image information
1: Step 1/7 : FROM docker:17.10.0-ce-dind
 ---> 9769e0f3f9cb
Step 2/7 : RUN apk update && apk add --no-cache python3 parted btrfs-progs docker util-linux sfdisk file coreutils sgdisk
 ---> Using cache
 ---> 8e051386d0ca
Step 3/7 : COPY ./requirements.txt /tmp/
 ---> Using cache
 ---> acbb5071e45c
Step 4/7 : RUN pip3 install -r /tmp/requirements.txt
 ---> Using cache
 ---> bf95343e0281
Step 5/7 : COPY ./src /usr/src/app
 ---> Using cache
 ---> 1a0ed03b49a9
Step 6/7 : WORKDIR /usr/src/app
 ---> Using cache
 ---> 334af6145047
Step 7/7 : CMD ["python3", "/usr/src/app/preload.py"]
 ---> Using cache
 ---> 510a76c6dcde
Successfully built 510a76c6dcde
Successfully tagged balena/balena-preload:latest
Waiting for Docker to start...
Exception in thread background thread for pid 209:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.6/site-packages/sh.py", line 1540, in wrap
    fn(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/sh.py", line 2459, in background_thread
    handle_exit_code(exit_code)
[...]

PR #2230, opened 2021-03-25
Improve installation docs regarding Docker Desktop version requirements

pdcastro:
Connects-to: #2228
Change-type: patch

### DIFF
--- a/INSTALL-MAC.md
+++ b/INSTALL-MAC.md
@@ -62,15 +62,21 @@ command set can also be used to list and manage SSH keys: see `balena help -v`.
 
 Like the `build` and `deploy` commands, the `preload` command requires Docker, with the additional
 restriction that Docker must be installed on the local machine (because Docker's bind mounting
-feature is used). Also, for some device types (such as the Raspberry Pi), the `preload` command
-requires Docker to support the [AUFS storage
+feature is used). Also, preloading balenaOS images for some older device types (like the Raspberry
+Pi 3, but not the Raspberry 4) requires Docker to support the [AUFS storage
 driver](https://docs.docker.com/storage/storagedriver/aufs-driver/). Unfortunately, Docker Desktop
-for Windows dropped support for the AUFS filesystem in Docker CE versions greater than 18.06.1. The
-present workaround is to either:
+for Windows or macOS dropped support for the AUFS filesystem in Docker CE versions greater than
+18.06.1. The present workarounds are to either:
 
+* Install the balena CLI on Linux (e.g. Ubuntu) with a virtual machine like VirtualBox.
+  This works because Docker for Linux still supports AUFS. Hint: if using a virtual machine,
+  copy the image file over, rather than accessing it through "file sharing", to avoid errors.
 * Downgrade Docker Desktop to version 18.06.1. Link: [Docker CE for
   Mac](https://docs.docker.com/docker-for-mac/release-notes/#docker-community-edition-18061-ce-mac73-2018-08-29)
-* Install the balena CLI on a Linux machine (as Docker for Linux still supports AUFS). A Linux
-  Virtual Machine also works, but a Docker container is _not_ recommended.
 
-Long term, we are working on replacing AUFS with overlay2 for the affected device types.
+The [balena CLI Docker images](./docker/DOCKER.md) also support the `balena preload` command,
+but they do not avoid the requirement of using a Linux physical or virtual machine. This is
+because the AUFS storage driver is not present in the Linux kernel used by the newer versions
+of Docker Desktop for Windows or macOS.
+
+We are working on replacing AUFS with overlay2 in balenaOS images of the affected device types.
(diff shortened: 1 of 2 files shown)


pdcastro:
@balena-ci retest

pdcastro:
@balena-ci I self-certify!

maggie0002:
Thanks for the info.

What is it about the Raspberry Pi 3 that means this won't work? I am trying a preload on Orange Pi Zero BalenaOS 2.60.1+rev1 and it appears to go through (at least appears too, I assume I would receive error messages rather than a corrupt build).

Should I be looking for Balena OS version numbers? 64bit vs 32 bit builds?

pdcastro:
> What is it about the Raspberry Pi 3 that means this won't work?

@maggie0002, it is not exactly the balenaOS version, nor 64bit vs 32bit, nor a lack of capability of the device type (RPi 3 vs RPi 4). If you open a Host OS terminal on a device running balenaOS, with the RPi 3 you will see this `aufs` folder:
```
root@9a56d9b:~# ls -l /mnt/data/docker/aufs/
total 468
drwx------ 1071 root root 155648 Mar 26 16:28 diff
drwx------    2 root root 155648 Mar 26 16:28 layers
drwx------ 1071 root root 155648 Mar 26 16:28 mnt

root@9a56d9b:~# ls -l /mnt/data/docker/aufs/layers/
total 4068
-rw-r--r-- 1 root root 1495 Mar 14 00:08 00032c7a32c37bd238449d3de6ada51f6d363192f76b111dd10b1893e995f738
-rw-r--r-- 1 root root 1755 Mar 14 00:08 006f1e72b62e297a62658b8d61e8c077e7a1e4edcf374b76e05a769893c1fded
```

While with the RPi 4 you will see an `overlay2` folder instead:
```
root@b408a70:~# ls -l /mnt/data/docker/overlay2/
total 164
drwx------ 4 root root 4096 Mar 16 13:37 05eab514d7c4be8fb43f9d8149b9ae3178577a26397772bb854f1e780d763d6e
drwx------ 4 root root 4096 Mar 16 13:38 0a0d69cabf943d97016b5c1ce6b7a43acd8e66f9ca8216204b78f68c430fdaad
...
```

The contents of these folders are the storage filesystem of application containers and images listed with `balena-engine images` (equivalent of `docker images`), in either the [aufs](https://docs.docker.com/storage/storagedriver/aufs-driver/) or [overlay2](https://docs.docker.com/storage/storagedriver/overlayfs-driver/) formats (which of course are incompatible).

I understand that the same version of balenaOS will include either the `aufs` folder or the `overlay2` folder, depending on the device type. I believe this is currently a "hardcoded setting" per device type. A device image downloaded from the dashboard, even before preloading, will already contain one folder or the other. The `balena preload` command will then populate one of those folders (inside the balenaOS image) with an application release.

Also, if I recall correctly, balenaOS host OS updates were designed so that the [rootA / rootB partitions](https://www.balena.io/docs/reference/OS/overview/2.x/#image-partition-layout) are also structured as Docker / balenaEngine images in one of those formats. References:

* https://github.com/balena-os/balenahup/issues/301  
_"Assume you have a device that uses AUFS for the container filesystem, and are upgrading to a version of the OS that uses overlay2. Currently, `balenahup` will happily apply this update and reboot; however, the device will panic at boot time because the containers have not been migrated to the overlay2 filesystem."_
* https://github.com/balena-os/balena-engine/issues/205

So I think the difficulty of converting from aufs to overlay2 is managing the data format migration on existing devices, as the `/mnt/data/docker` partition persists across Host OS upgrades, and the `rootA` / `rootB` partitions as well. The balenaOS team has been working on a [overlay2 migration utility](https://github.com/balena-os/balena-engine/pull/168) which I believe is almost complete and will be included in balenaOS 3.0, which is planned to be released in the next few months. I am not an "authoritative source" for this information but it is what I am aware of at the moment.

maggie0002:
Thanks @pdcastro. I have taken enough from that to have concluded that doing all preloads on a Linux VM for all devices is going to be the way to go, just to avoid the uncertainty until OS 3.0.

pdcastro:
> I am trying a preload on Orange Pi Zero BalenaOS 2.60.1+rev1 and it appears to go through

@maggie0002, I have now tested and confirmed that the Orange Pi Zero, like the RPi 4, already uses overlay2 instead of aufs, so you can use `balena preload` with the latest version of Docker Desktop on macOS or Windows, and there is no need to use a Linux VM (for the Orange Pi Zero).

> it appears to go through (at least appears too, I assume I would receive error messages rather than a corrupt build).

`balena preload` would fail and produce an error message: `error initializing graphdriver: driver not supported` if the device type used AUFS and Docker didn't support AUFS. More details on issue #1099.

maggie0002:
Thanks again @pdcastro.

OPI Zero happened to just be the one I was working on, I intend to be building a lot more images too. Using the Linux VM seemed like an easier way to go than checking and tracking them all. It was something on my to do list for a while anyway, and overall I think this improves my workflow.

If interested, I ended up using a GitHub Action: https://github.com/LearnersBlock/learners-block/blob/develop/.github/workflows/deploy-release.yml

PR #2232, opened 2021-03-26
docker: Install official static docker binaries

klutchell:
We saw fewer environmental issues when running the alpine container vs debian, likely due to the differences in the docker versions (20 vs 18).

(private link) https://www.flowdock.com/app/rulemotion/r-beginners/threads/pnN3-UtI9EqD1oeJ2DZUzbFlvs9

This PR will install the same static docker binaries on both platforms.

Change-type: patch
Changelog-entry: docker: Install official static docker binaries

### DIFF
--- a/docker/DOCKER.md
+++ b/docker/DOCKER.md
@@ -397,12 +397,12 @@ export BALENA_DISTRO="debian"
 # provide the architecture where you will be testing the image
 export BALENA_ARCH="amd64"
 
-# optionally register QEMU binfmt if building for other architectures (eg. armv7hf)
-$ docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+# optionally register QEMU binfmt if building for other architectures (eg. rpi)
+docker run --rm --privileged multiarch/qemu-user-static:5.2.0-2 --reset -p yes
 
 # build and tag an image with docker
 docker build . -f docker/${BALENA_DISTRO}/Dockerfile \
-    --build-arg "BUILD_BASE=balenalib/${BALENA_ARCH}-${BALENA_DISTRO}-node:12.19.1-build" \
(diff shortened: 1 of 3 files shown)


klutchell:
@balena-ci retest

klutchell:
This is failing because Docker only provides official static binaries for a limited set of architectures.

Similarly, they only provide recent debian packages for a limited set of architectures.

It got complicated enough that I made this table here:
(private link) https://docs.google.com/spreadsheets/d/1I8mknQK0jEBF-ZHNeU7-1jroiyaGaQXCKsKQ8f0a6oc/edit?usp=sharing

We may have to revisit which platforms are available for these images based on ongoing Docker binary support.

klutchell:
The official Docker-in-Docker images only support the following combinations:
- `alpine` x `amd64` x `docker 19.03.8`
- `alpine` x`arm32v6` x `docker 19.03.8`
- `alpine` x`arm32v7` x `docker 19.03.8`
- `alpine` x`arm64v8` x `docker 19.03.8`

So I think we should stick to those as close as possible, which would mean the following changes:
- switch to static binary installs of `docker 19.03.8` and avoid package repos
- remove `i386` images since docker does not provide binaries
- remove `rpi` x `debian` since the `armhf` and `armel` binaries won't run on `balenalib/rpi-debian` thus far

Which leaves us with the following (messy) supported architectures list:
- `amd64`
- `aarch64 (debian only)` <-- https://github.com/node-ffi-napi/node-ffi-napi/issues/126
- `armv7hf`
- `rpi (alpine only)` <-- GH issue tbd

OR we can simplify and go all in on alpine as Docker has done:
- `amd64`
- `aarch64 (WIP)` <-- https://github.com/node-ffi-napi/node-ffi-napi/issues/126
- `armv7hf`
- `rpi`

Any other ideas or combinations I haven't tried? The Dockerfiles in this PR were trimmed down and used for testing.

Issue #1142, opened 2019-03-15
Add ability to select a docker-compose file


CameronDiver:
Given a directory with more than one docker-compose file, users would like the ability to be able to select one or another of them.

Probably the best way to do this would be a flag to `balena push`.

---

*Please keep in mind that we try to use the issue tracker of this repository for specific bug reports & CLI feature requests. General & troubleshooting questions are encouraged to be posted to the [balena forums](https://forums.balena.io) where the community can both contribute and benefit from the answers.*

*Before submitting this issue please check that this issue is not a duplicate. If there is another issue describing the same problem or feature please add your information to the existing issue's comments.*

pdcastro:
Related to issue #587 _(Let users specify a Dockerfile with `balena build -f`)_

smilykoch:
+1 on this feature

krasi-georgiev:
+1 on this feature

What are the chances of someone starting to work on this?

there is already a flag:`--dockerfile` so something like `--composefile` would be ideal

robsonvn:
+1000 to this feature, I'm currently using bash script on my deploys just because of it.

pdcastro:
@krasi-georgiev and others -- thanks for the feedback and upvoting. 👍
I am currently auditing and sorting all open issues in the CLI repo, and this issue is currently ranking in the top five. :-) Link: [Sorted Backlog](https://github.com/balena-io/balena-cli/milestone/2)
It may go down a bit as I find other high priority issues, but it will still end up in the top league. So, there is hope it will get implemented in the near future.

rjhuijsman:
+1. We use different compose files to specify different environments, and having to use scripts is (1) a headache to develop, maintain, document (2) just lost me a bunch of time to even figure out.

pdcastro:
Implementation note: Docker's documentation for the docker-compose.yml CONTEXT instruction states that the context path is _"interpreted as relative to the location of the Compose file"_ (https://docs.docker.com/compose/compose-file/compose-file-v2/#context). Therefore, allowing the specification of an alternative `docker-compose.yml` file **may** also require changes to CLI dependency modules like [resin-bundle-resolve](https://github.com/balena-io-modules/resin-bundle-resolve), which may currently assume that the context path is relative to the "project root" (?). (It's probably also a good idea to test that the docker-compose tool actually behaves as documented.)

hedss:
[...]

PR #2241, opened 2021-04-02
Rename supported docker-compose overlay file

maggie0002:
This amends a previous commit allowing users to specify a docker-compose.override.yml instead of a docker-compose.dev.yml file, which is merged in local pushes, and allows extra/alternative services to be declared.
Change-type: minor
Signed-off-by:

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -172,7 +172,7 @@ async function mergeDevComposeOverlay(
 	composeStr: string,
 	projectRoot: string,
 ) {
-	const devOverlayFilename = 'docker-compose.dev.yml';
+	const devOverlayFilename = 'docker-compose.override.yml';
 	const devOverlayPath = path.join(projectRoot, devOverlayFilename);
 
 	if (await exists(devOverlayPath)) {


maggie0002:
For further info on the discussion, see https://github.com/balena-io/balena-cli/issues/1142

This request beings the Balena CLI inline with standard Docker practice. Docker will automatically merge a file called docker-compose.override.yml (https://docs.docker.com/compose/extends/#multiple-compose-files). 

Before this change, it would automatically merge docker-compose.dev.yml. This is not a standard Docker practice. It is also a common naming system for compose files, which would result in user inadvertently merging their development files.

srlowe:
Thanks for the PR @maggie0002 , unfortunately I can't approve it, mainly because the merge behaviour of docker's .override file is slightly different to our .dev file (the .dev file only does a shallow merge on the services section - simply adding or replacing services.)

I've read the linked discussion, and understand most of the issues you describe.  Note that this is something that is under active discussion.  We will discuss the possibility of aligning the .dev behaviour with the standard docker behaviour (along with naming) and we hope to be able add the ability to explicitly specify alternative compose files in the future.

PR #2242, opened 2021-04-04
docs: Further clarify Docker requirements for preload

pdcastro:
Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2246, opened 2021-04-07
Direct missing release installs to npm install method

20k-ultra:
Users that do not find their achitecture in the release page might not know that the npm method is their only option to install.

### DIFF
--- a/INSTALL-LINUX.md
+++ b/INSTALL-LINUX.md
@@ -1,8 +1,6 @@
 # balena CLI Installation Instructions for Linux
 
-These instructions are for the recommended installation option. They are suitable for most Linux
-distributions, except notably for **Linux Alpine** or **Busybox**. For these distros, see [advanced
-installation options](./INSTALL-ADVANCED.md).
+These instructions are suitable for most Linux distributions on Intel x86, except notably for **Linux Alpine** or **Busybox**. For these distros or for the ARM architecture, follow the [NPM Installation](./INSTALL-ADVANCED.md#npm-installation) method.
 
 Selected operating system: **Linux**
 


INSTALL-LINUX.md
```diff
@@ -1,8 +1,6 @@
 # balena CLI Installation Instructions for Linux
 
-These instructions are for the recommended installation option. They are suitable for most Linux
-distributions, except notably for **Linux Alpine** or **Busybox**. For these distros, see [advanced
-installation options](./INSTALL-ADVANCED.md).
+These instructions are the recommended installation option. They are suitable for most Linux distributions and architectures, except notably for **Linux Alpine** or **Busybox**. For these distros, see [advanced installation options](./INSTALL-ADVANCED.md). Additionally, if you do not find a suitable release for your machine's architecture, then follow the [NPM Installation](./INSTALL-ADVANCED.md#npm-installation) method.
```
pdcastro, line 3: Would the following be a little less verbose overall? (I care because of regular criticism that CLI installation process and instructions are too complicated.)
```
These instructions are the recommended installation option. They are suitable for most Linux
distributions on Intel x86, except notably for **Linux Alpine** or **Busybox**. For these distros
or for the ARM architecture, follow the [NPM Installation](./INSTALL-ADVANCED.md#npm-installation)
method.
```

20k-ultra: that's a great optimization.

20k-ultra:
@pdcastro I actually changed the beginning to not mention this is the recommended install option.

pdcastro (approved):

PR #2247, opened 2021-04-09
lib/commands/local/configure: Fix local configure when resin-wifi is …

mbalamat:


mbalamat:
@balena-ci I self-certify!

Issue #2185, opened 2021-02-11
When using local mode, any multi-line prints to the console will not be sent back using either balena push or balena logs.


sradevski:

edorgeville:
Original forum thread: https://forums.balena.io/t/balena-push-logs-local-mode-multi-line-prints-are-not-shown/250083

edorgeville:
Update: using the `--debug` flag gives this error for every new log line:
```
[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 138[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 140[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 157[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 172[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 111[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 187[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 231[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 138[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 151[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 155[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token â in JSON at position 9  
[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 181[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 146[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 177[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 177[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 177[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 176[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token { in JSON at position 116[Debug]   Dropping device log due to failed parsing: SyntaxError: Unexpected token m in JSON at position 2  
[...]

PR #2248, opened 2021-04-09
push, logs: Fix parsing of local mode device logs (NDJSON stream)

pdcastro:
Resolves: #2185
Change-type: patch


pdcastro:
@balena-ci I self-certify!

Issue #1003, opened 2018-10-25
Config inject should work directly on images, not just drives


pimterry:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_em7f)

pdcastro:
Even more importantly, if it does not work on images, it should not document that it does! -
```
$ balena help config inject
...
OPTIONS
  -d, --drive <drive>  device filesystem or OS image location
```

pdcastro:
Fixed/implemented in CLI v12.44.9. (Might have been implemented before, but the CLI was also calling `umount` on unmounted image files which would cause errors, and this was fixed in #2251.)

PR #2251, opened 2021-04-13
config inject/read/write: Fix umount errors with OS image files

pdcastro:
Resolves: #1003
Change-type: patch

### DIFF
--- a/lib/commands/config/inject.ts
+++ b/lib/commands/config/inject.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016-2020 Balena Ltd.
+ * Copyright 2016-2021 Balena Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


lib/utils/helpers.ts
```diff
@@ -16,16 +16,11 @@ limitations under the License.
...
 import * as _ from 'lodash';
-import * as os from 'os';
-import type * as ShellEscape from 'shell-escape';
+import { promisify } from 'util';
 
-import type { Device, PineOptions } from 'balena-sdk';
-import { ExpectedError, SIGINTError } from '../errors';
 import { getBalenaSdk, getChalk, getVisuals } from './lazy';
-import { promisify } from 'util';
-import { isSubcommand } from '../preparser';
```
pdcastro, line 28: I've done a cleanup of module-level imports in this file, so that there are fewer side effects (unnecessary cascading imports) of merely importing this file to use perhaps a single function like `safeUmount`.

lib/commands/config/inject.ts
```diff
@@ -76,12 +68,11 @@ export default class ConfigInjectCmd extends Command {
...
 		);
 
-		const { promisify } = await import('util');
-		const umountAsync = promisify((await import('umount')).umount);
+		const { safeUmount } = await import('../../utils/helpers');
 
 		const drive =
 			options.drive || (await getVisuals().drive('Select the device/OS drive'));
-		await umountAsync(drive);
+		await safeUmount(drive);
```
pdcastro, line 75: If `drive` was a path to an image file like `./balena.img` (as opposed to a path to a block device like `/dev/disk2`), `umountAsync('./balena.img')` would raise an error. This PR adds a check on whether `drive` is something currently mounted before attempting to umount it.

pdcastro:
@balena-ci I self-certify!

Issue #2045, opened 2020-09-24
preload: stuck at "Reading image information" spinner, errors not reported


pdcastro:
Originally reported by @chrisys ([internal flowdock thread](https://www.flowdock.com/app/rulemotion/r-product/threads/nSzIWfXQcDabItaQS5ZzSVNTfU9)):

> CLI preload looks like it's working but in the background has failed without informing the user, so you're left looking at a spinner thinking it's doing it's job (there are related issues for this but I think the worst aspect is that it doesn't even fail, it looks from the user pov as if everything is fine).

Steps to reproduce:

1. Using a recent version of Docker Desktop for macOS or Windows (which does not support the AUFS filesystem, e.g. Docker Desktop v2.3.0.5, Engine 19.03.12), attempt to preload an image that requires AUFS (raspberrypi3, intel-nuc). Do **not** enable debug mode (`export DEBUG=`).
2. Observe the CLI get stuck as follows:
```
$ balena preload balena-cloud-test-nuc-intel-nuc-2.50.1+rev1-v11.4.10.img -a test-nuc
Building Docker preloader image. [========================] 100%
| Checking that the image is a writable file
| Finding a free tcp port and getting balena settings
| Checking if the image is an edison zip archive
| Creating preloader container
\ Starting preloader container
/ Fetching application test-nuc
\ Reading image information
```

vasdee:
I am facing this problem now. Are there any workarounds to get through? 

A team member didn't have this problem on a particular version of ubuntu, so my thinking is using a VM that has an older docker engine version might get me across the line

chrisys:
@vasdee if you're on Linux it's my understanding that it should work OK, if you're on Windows/MacOS the options are to downgrade docker, or as you suggested, use a Ubuntu/Linux VM. I believe Docker Desktop 18.06.1 is the latest that supported the operation for Windows/Mac; see here for more: https://github.com/balena-io/balena-cli/issues/1099

vasdee:
@chrisys interesting, this issue has occurred on both of my Linux installs - first was Arch Linux, second is a fresh install of Fedora 33.

As a test, I spun up a Ubuntu 20.04 VM in my Fedora install, it didn't have a problem running the same preload commands.  It looked like the docker versions between the Ubuntu VM and Fedora install were the same too, but I'll need to confirm that.

pdcastro:
[...]

PR #2259, opened 2021-05-03
Fix "Total: undefined" image pull progress report and update dependencies

pdcastro:
Connects-to: #2165 
Connects-to: #2045 

This PR is a by-product of the investigation on Node.js v14 support (#2165 and #2079).

Four commits in the same PR because they were tested together, after bumping `resin-multibuild`, `docker-progress`, `balena-preload` and other modules. While testing, I also fixed a pre-existing bug by which `balena build` would print "Total: undefined" instead of a pull image progress, as shown below:

```
$ balena build ~/balena/other/simple_multicontainer -h 192.168.1.46 -p 2375 -a test-nuc
[Info]    Building for amd64/intel-nuc
[Build]   Building services...
[Build]   foo Preparing...
[Build]   bar Preparing...
[Build]   zip Preparing...
...
[Build]   foo Total: undefined
[Build]   foo Total: undefined
[Build]   foo Total: undefined
...
```

After the fix:

```
...
[Build]   foo Pull progress: 54%
[Build]   foo Pull progress: 56%
[Build]   foo Pull progress: 59%
...
```

This PR also converts some JavaScript code to TypeScript (`docker-js.js` → `docker.ts`).


klutchell:
I can give this a test in the morning if you'd like more coverage.

pdcastro:
@balena-ci retest

lib/utils/docker.ts
```diff
@@ -173,3 +171,120 @@ export async function isBalenaEngine(docker: dockerode): Promise<boolean> {
...
+
+export interface ExtendedDockerOptions extends dockerode.DockerOptions {
+	docker?: string; // socket path, e.g. /var/run/docker.sock
+	dockerHost?: string; // host name or IP address
+	dockerPort?: number; // TCP port number, e.g. 2375
+}
+
+export async function getDocker(
+	options: ExtendedDockerOptions,
+): Promise<dockerode> {
```
pdcastro, line 183: This block of code in `docker.ts` is a conversion of `docker-js.js` (deleted by this PR) from JavaScript to TypeScript.

lib/utils/compose_ts.ts
```diff
@@ -1015,6 +1015,9 @@ export async function makeBuildTasks(
 			infoStr = `build [${task.context}]`;
 		}
 		logger.logDebug(`    ${task.serviceName}: ${infoStr}`);
+		// Workaround for Docker v20.10 + single-arch base images. See:
+		// https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc
+		task.dockerPlatform = 'none';
```
klutchell, line 1020: Just to clarify, looking at `resin-multibuild` I believe this means the behaviour will be to not provide the platform flag for now?

https://github.com/balena-io-modules/balena-multibuild/blob/cda88bc0f434cae3b1594ee15e532d2d455ae510/lib/resolve.ts#L48-L56

pdcastro: That's right. The CLI was stuck with `resin-multibuild` v4.7.2 (could not update it) because newer versions of `resin-multibuild` provided the `platform` flag to the Docker engine, which in turn would cause `balena build` to fail with Docker v20.10 and later (that long thread ™ https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc ).
That issue is not resolved but I needed to update `resin-multibuild` to take advantage of the bump of `docker-progress` (Node.js v14 investigation), and generally because being unable to update `resin-multibuild` is a bad thing for `balena-cli`, as `resin-multibuild` is a fundamental component. So I came up with a way of updating `resin-multibuild` while disabling the `platform` feature. cc: @toochevere @robertgzr @dfunckt @Page-

pdcastro: @klutchell, I see that I had not invited you as reviewer of the corresponding recent `resin-multibuild` PR. Check this comment: https://github.com/balena-io-modules/balena-multibuild/pull/88#discussion_r626907561

lib/utils/compose_ts.ts
```diff
@@ -1493,10 +1494,15 @@ function pullProgressAdapter(outStream: Duplex) {
...
+		id ||= '';
+		status ||= '';
+		const isTotal = id && id.toLowerCase() === 'total';
+		if (status) {
 			status = status.replace(/^Status: /, '');
+		} else if (isTotal && typeof percentage === 'number') {
+			status = `Pull progress: ${percentage}%`;
 		}
-		if (id != null) {
+		if (id && status && !isTotal) {
```
klutchell, line 1505: It took me a while to track down where these values were coming from but it looks like it's from this build log task?
https://github.com/balena-io/balena-cli/blob/4911db640f4979c8561ba1a4b8d103bcf7db4ae2/lib/utils/device/deploy.ts#L485-L489

Is it parsing raw logs from a docker pull command?

package.json
```diff
@@ -219,11 +219,10 @@
...
     "common-tags": "^1.7.2",
     "denymount": "^2.3.0",
-    "docker-modem": "^2.1.4",
-    "docker-progress": "^4.0.1",
+    "docker-modem": "^3.0.0",
+    "docker-progress": "^5.0.0",
     "docker-qemu-transpose": "^1.1.1",
-    "docker-toolbelt": "^3.3.8",
-    "dockerode": "^2.5.8",
+    "dockerode": "^3.3.0",
```
pdcastro, line 225: My testing so far did not reveal any problems with `dockerode` v3. However, the CLI is only partially using it, as a few dependencies still require `dockerode` v2:
```
$ npm list
...
├─┬ balena-sync@11.0.2
│ ├─┬ docker-toolbelt@3.3.8
│ │ ├─┬ dockerode@2.5.8

├─┬ livepush@3.5.0
│ ├─┬ dockerode@2.5.8

├─┬ resin-multibuild@4.11.0
│ ├─┬ dockerode@2.5.8

├─┬ resin-multibuild@4.11.0
│ ├─┬ resin-docker-build@1.1.6
│ │ ├─┬ dockerode@2.5.8
```

`resin-multibuild` still declares a dependency on `dockerode` v2, but it mostly uses a `dockerode` instance injected as an argument in several functions, which after this PR will be `dockerode` v3 for `balena-cli`. The balenaCloud builder will still use `dockerode` v2 as long as that's the version it injects. I have tested this PR quite a bit (and automated tests cover it too). We'll keep an eye on it over time, and gradually move components from v2 to v3.

lib/utils/compose_ts.ts
pdcastro: `utils/device/deploy.ts` is a different code path, used by `balena push <ip-addr>`.
The changes in this commit affect `balena build`.

The `id` and `percentage` fields are assigned in the following function:
https://github.com/balena-io-modules/docker-progress/blob/v5.0.0/lib/index.ts#L332-L348

The source is indeed logs from a `docker pull` command (API call through `dockerode`), but not raw, as it is a second layer of parsing after the `docker-progress` module has already done some parsing/transforming of its own:

* `task.progressHook = pullProgressAdapter(captureStream);`
at https://github.com/balena-io/balena-cli/blob/v12.44.12/lib/utils/compose_ts.ts#L496

* `const progressHook = task.progressHook;`  
`return dockerProgress.pull(imageName, progressHook, opts);`  
at https://github.com/balena-io-modules/balena-multibuild/blob/v4.11.0/lib/external.ts#L68

* `onProgress = reporter.pullProgress(onProgress);`
at https://github.com/balena-io-modules/docker-progress/blob/v5.0.0/lib/index.ts#L428

By the way, `Total: undefined` is nothing introduced (nor fixed) by `docker-progress` v5, this bug was around for a long time, maybe longer than a year.

Before this commit (and for a long time, maybe longer than a year):
```
$ balena build ~/balena/other/simple_multicontainer -h 192.168.1.46 -p 2375 -a test-nuc
[Info]    Building for amd64/intel-nuc
[Build]   Building services...
[Build]   foo Preparing...
[Build]   bar Preparing...
[Build]   zip Preparing...
...
[Build]   foo Total: undefined
[Build]   foo Total: undefined
[Build]   foo Total: undefined
...
```
After this commit:
```
...
[Build]   foo Pull progress: 54%
[Build]   foo Pull progress: 56%
[Build]   foo Pull progress: 59%
...
```

pdcastro:
I have tested this PR enough that I could self certify it, but I added a number of reviewers for awareness of changes like `dockerode` v3, `docker-progress` v5, and the bump to `resin-multibuild` while disabling the "platform feature" with `task.dockerPlatform = 'none'`.
So @klutchell if you're hesitating, hopefully this reassures you. 😀    No pressure though, there's also no hurry with getting this merged!

klutchell:
I don't see any issues, I just want to make sure I understand it fully (mostly the current functionality)!

klutchell (approved):
LGTM!

PR #2265, opened 2021-05-15
devices supported: Use new DeviceType data model as source of truth 

pdcastro:
Connects-to: #2266 
Connects-to: balena-io/balena-sdk/pull/1060
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/pub/threads/AE_IXh9Xs_I7MYOHS98oLb0xaya

### DIFF
--- a/lib/commands/devices/supported.ts
+++ b/lib/commands/devices/supported.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016-2019 Balena Ltd.
+ * Copyright 2016-2021 Balena Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,13 +15,13 @@
  * limitations under the License.
  */
 import { flags } from '@oclif/command';
-import type * as SDK from 'balena-sdk';
 import * as _ from 'lodash';
 import Command from '../../command';
 
 import * as cf from '../../utils/common-flags';
 import { getBalenaSdk, getVisuals, stripIndent } from '../../utils/lazy';
 import { CommandHelp } from '../../utils/oclif-utils';
+import { isV13 } from '../../utils/version';
 
 interface FlagsDef {
 	discontinued: boolean;
@@ -30,17 +30,24 @@ interface FlagsDef {
 	verbose?: boolean;
 }
 
+const deprecatedInfo = isV13()
+	? ''
+	: `
+The --verbose option may add extra columns/fields to the output. Currently
+this includes the "STATE" column which is DEPRECATED and whose values are one
+of 'new', 'released' or 'discontinued'. However, 'discontinued' device types
+are only listed if the '--discontinued' option is also used, and this option
+is also DEPRECATED.
+`
+			.split('\n')
+			.join(`\n\t\t`);
+
 export default class DevicesSupportedCmd extends Command {
 	public static description = stripIndent`
 		List the supported device types (like 'raspberrypi3' or 'intel-nuc').
 
 		List the supported device types (like 'raspberrypi3' or 'intel-nuc').
-
-		The --verbose option adds extra columns/fields to the output, including the
-		"STATE" column whose values are one of 'new', 'released' or 'discontinued'.
-		However, 'discontinued' device types are only listed if the '--discontinued'
-		option is used.
-
+		${deprecatedInfo}
 		The --json option is recommended when scripting the output of this command,
 		because the JSON format is less likely to change and it better represents data
 		types like lists and empty strings (for example, the ALIASES column contains a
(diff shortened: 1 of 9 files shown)


lib/commands/devices/supported.ts
```diff
@@ -69,45 +69,50 @@ export default class DevicesSupportedCmd extends Command {
...
-				} else {
-					// ensure it is always an array (for the benefit of JSON output)
-					d.aliases = [];
-				}
-				return d;
-			},
-		);
+		const [dts, configDTs] = await Promise.all([
+			getBalenaSdk().models.deviceType.getAll({
+				$expand: { is_of__cpu_architecture: { $select: 'slug' } },
```
pdcastro, line 80: @thgreasi, thanks for the following comment on Flowdock:
> https://www.flowdock.com/app/rulemotion/pub/threads/AE_IXh9Xs_I7MYOHS98oLb0xaya
> _Always think of `/v6/device_type` as holding 1000's of records with all DTs of the world. You would probably never want to query the DTs w/o a filter to limit the results, to at least those for which we have a hostApp (and maybe at least one release depending on the use case)._

So it sounds like the query above is incomplete, missing filters. Could you help me figuring out what the filter should be here?

>> Or is it the case that the API's `/device-types/v1/` endpoint is now sharing the same source of truth as the `/v6/device_type` endpoint?
> Kind of. The even though `/device-types/v1/` is getting some of its data from S3, it should not serve anything that's not in `/v6/device_type` (which is populated from the DT contract repo).

I am confused: Should the CLI query `models.deviceType.getAll(... $filter ...)` or not? Was this PR unnecessary? As is, this PR added some 11 device types to the output of `balena devices supported`, which I thought were "good" new entries.

lib/commands/devices/supported.ts
```diff
@@ -69,45 +69,50 @@ export default class DevicesSupportedCmd extends Command {
...
+				(alias) => alias !== dt.slug,
+			);
+			const arch = (dt.is_of__cpu_architecture as any)?.[0]?.slug;
+			deviceTypes.push({
+				slug: dt.slug,
+				aliases: options.json ? aliases : [aliases.join(', ')],
+				arch: arch || 'n/a',
+				// 'BETA' renamed to 'NEW'
+				// https://www.flowdock.com/app/rulemotion/i-cli/threads/1svvyaf8FAZeSdG4dPJc4kHOvJU
+				state: (configDT.state || 'NEW').replace('BETA', 'NEW'),
```
pdcastro, line 105: I found that `models.deviceType.getAll()` was listing some 11 additional device types compared to `models.config.getDeviceTypes()`, so `configDT.state` above would be `undefined`. I assumed, perhaps incorrectly, that those were new device types, hence `configDT.state || 'NEW'`.

lib/commands/devices/supported.ts
thgreasi: @pdcastro as per the FD thread
> You would probably never want to query the DTs w/o a filter to limit the results, to at least those for which we have a hostApp (and maybe at least one release depending on the use case).

This is one of those cases where you care for the DTs that already have a hostApp aka are supported in balenaCloud. Luckily, we had considered this case and have a convenience method for that `models.deviceType.getAllSupported()`
See: https://github.com/balena-io/balena-sdk/blob/a42c9f13f554cebd23f5c0b011b95b043f66e072/lib/models/device-type.ts#L149

thgreasi, line 80:
```suggestion
			getBalenaSdk().models.deviceType.getAllSupported({
				$select: [/*only the props you need*/],
				$expand: { is_of__cpu_architecture: { $select: 'slug' } },
```

pdcastro: Thanks @thgreasi 🙏  I have changed the code to use `getAllSupported()` instead of `getAll()`. Consistently to your explanation, `getAll()` returns additional results, which were the "11 extra device types" I had mentioned before.

> `getBalenaSdk().models.deviceType.getAllSupported({`
> To be clear, the above probably does not include those that are discontinued (which afair is represented with `is_archived: true` in the hostApp. If you also think we need the discontinued as well (do you really?)

Indeed. Here's what I've done: I have marked the `--discontinued` and `--verbose` flags deprecated, and I have already added `isV13()` feature switches that will automatically disable the flags in CLI v13. While `isV13()` is false (while in v12), I am still taking the "discontinued" devices from `models.config.getDeviceTypes()`, but already taking the non-discontinued devices from `models.deviceType.getAllSupported()` -- the new source of truth. The `STATE` column and `state` JSON field will automatically disappear when CLI v13 is released, thanks to the feature switches.

The overall result, not particularly exciting, :-) is that the list of the devices printed before and after this PR is unchanged.

pdcastro:
@balena-ci rebase

pdcastro:
This PR has been awaiting a follow-up review for almost a month after I addressed Thodoris' comments... And now he is on leave. :-) I will self certify. If anyone has further comments, feel free to add them and they can be addressed in another PR.

pdcastro:
@balena-ci I self-certify!

Issue #2183, opened 2021-02-10
`balena deploy` should parse, validate, and create a contract as part of the release it creates


sradevski:
It should essentially do the same as here: https://github.com/balena-io/balena-builder/blob/master/src/app-contract.ts

pdcastro:
There is a related product call topic pending discussion:

> https://jel.ly.fish/view-all-brainstorm-topics/brainstorm-topic-7478d22a-59b8-44a3-a5bf-eaaa247cb396
> 
> Discuss what to do (if anything) about the fact that two `balena.yml` files may exist in a balena app with different purposes and formats.
> ```
> ./balena.yml - contracts / Deploy With Balena
> ./.balena/balena.yml - build secrets and variables
> ```
> Example: the ARG and ENV instructions in a Dockerfile are used for build time variables and runtime environment variables respectively. To set build time variables, one should use `./.balena/balena.yml`, but to set environment variables, one should use `./balena.yml`. This separation seems difficult to justify, especially when both files have the same name. Also:
> * `./.balena/balena.yml` allows specifying service-specific variables, whereas `./balena.yml` only allows "global" (all services) variables, which is an inconsistency.
> * `./.balena/balena.yml` is supported across `balena push`, `balena build`, `balena deploy`, whereas `./balena.yml` is only supported by `balena push` (because it is only the cloud builder that interprets `./balena.yml`, not the CLI).
> 
> Files documented at:
> * https://www.balena.io/docs/learn/deploy/deploy-with-balena-button/
> * https://www.balena.io/docs/learn/deploy/deployment/#build-time-only-secret-file
>
> At least one user was confused by it and brought it to support / forums: https://forums.balena.io/t/git-push-and-buildargs/227059/13
> 
> Would we want to rename one of the files? Merge them? Leave it as it is?

pdcastro:
I didn't investigate it yet, but we may need to implement support for local push as well - `balena push <ip-addr>`. For example, the contract file includes an `applicationEnvironmentVariables` section that should be honoured when pushing to a local device. I suspect this doesn't currently happen.

sradevski:
@pdcastro yes it doesn't happen currently, but we might need support for it in the future, if the envvars are read (which would be really convenient). For now, `balena deploy` should suffice

pdcastro:
Connects to JF Product Improvement: [Draft releases and release versioning](https://jel.ly.fish/8ea1c390-9a85-402d-978c-4d31dcb0d235) (restricted access)

PR #2270, opened 2021-05-19
Add `--draft` option to `balena push`

pipex:
This change will allow to build releases as draft and have them being
set as final at a later stage. This change is part of a larger feature towards
using the builder as part of CI/CD pipelines.

Depends-on: https://github.com/balena-io/balena-builder/pull/868
Change-type: minor

### DIFF
--- a/lib/commands/push.ts
+++ b/lib/commands/push.ts
@@ -58,6 +58,7 @@ interface FlagsDef {
 	'noconvert-eol': boolean;
 	'multi-dockerignore': boolean;
 	'release-tag'?: string[];
+	draft: boolean;
 	help: void;
 }
 
@@ -267,6 +268,14 @@ export default class PushCmd extends Command {
 			multiple: true,
 			exclusive: ['detached'],
 		}),
+		draft: flags.boolean({
+			description: stripIndent`
+				Instruct the builder to create the release as a draft. Draft releases are ignored
+				by the 'track latest' release policy but can be used through release pinning.
+				Draft releases can be marked as final through the API. Releases are created
+				as final by default unless this option is given.`,
+			default: false,
+		}),
 		help: cf.help,
 	};
 
@@ -362,6 +371,7 @@ export default class PushCmd extends Command {
 			registrySecrets,
 			headless: options.detached,
 			convertEol: !options['noconvert-eol'],
+			isDraft: options.draft,
 		};
 		const args = {
 			appSlug: application.slug,
@@ -394,7 +404,7 @@ export default class PushCmd extends Command {
 		registrySecrets: RegistrySecrets,
 	) {
 		// Check for invalid options
-		const remoteOnlyOptions: Array<keyof FlagsDef> = ['release-tag'];
+		const remoteOnlyOptions: Array<keyof FlagsDef> = ['release-tag', 'draft'];
 		this.checkInvalidOptions(
 			remoteOnlyOptions,
 			options,
--- a/lib/utils/remote-build.ts
+++ b/lib/utils/remote-build.ts
@@ -42,6 +42,7 @@ export interface BuildOpts {
 	headless: boolean;
 	convertEol: boolean;
 	multiDockerignore: boolean;
+	isDraft: boolean;
 }
 
 export interface RemoteBuild {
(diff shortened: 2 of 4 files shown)


pipex:
For more context about this feature, see https://www.flowdock.com/app/rulemotion/r-supervisor/threads/JN2-gnspQ-v6WaeWCvm9T8NYDY1

lib/commands/push.ts
```diff
@@ -267,6 +268,14 @@ export default class PushCmd extends Command {
 			multiple: true,
 			exclusive: ['detached'],
 		}),
+		draft: flags.boolean({
+			description: stripIndent`
+				Instruct the builder to create the release as a draft. Draft releases are ignored
+				by the 'track latest' release policy but can be used through release pinning.
+				Draft releases can be marked as final later through the API. All releases as created
```
pdcastro, line 275:
```suggestion
				Draft releases can be later marked as final through the API. Releases are created
```

lib/commands/push.ts
```diff
@@ -373,6 +382,7 @@ export default class PushCmd extends Command {
 			auth: token,
 			baseUrl,
 			nogitignore: !options.gitignore,
+			isdraft: options.draft,
```
pdcastro, line 385: I think `isdraft: options.draft` should be move a few lines above to the `opts` object instead of the `args` object, and also be listed in BuildOpts here:
https://github.com/balena-io/balena-cli/blob/v12.44.16/lib/utils/remote-build.ts#L37-L44
and then also be listed in this query string block:
https://github.com/balena-io/balena-cli/blob/v12.44.16/lib/utils/remote-build.ts#L89-L95

pipex: Done! Thanks for that!

lib/utils/compose_ts.ts
```diff
@@ -1405,6 +1418,18 @@ export function createRunLoop(tick: (...args: any[]) => void) {
 	return runloop;
 }
 
+async function tryGetContractContent(
```
pdcastro, line 1421: @toochevere, your commit looks like a great contribution! 👍 
Does it go (or would it go) as far as addressing issue #2183 [balena deploy should parse, validate, and create a contract as part of the release it creates](https://github.com/balena-io/balena-cli/issues/2183) ?
If your commit does not depend on Felipe's commit, and even more so if it addresses #2183, I would suggest creating a separate PR for it.

pipex: Not sure entirely about #2183, but I think PJ's commit will allow to close #1908, since this will add the version from the contract to the release during deploy

ab77:
@balena-ci retest

ab77:
@balena-ci retest

lib/utils/remote-build.ts
```diff
@@ -42,6 +42,7 @@ export interface BuildOpts {
 	headless: boolean;
 	convertEol: boolean;
 	multiDockerignore: boolean;
+	isdraft: boolean;
```
pdcastro, line 45:
```suggestion
	isDraft: boolean;
```
... to follow the camelCase convention for variable / function / class names, even if the query string argument had to read `isdraft` because perhaps `balena-builder` requires it (does it? Another query string argument reads `dockerfilePath`, with camelCase).

lib/commands/push.ts
```diff
@@ -267,6 +268,14 @@ export default class PushCmd extends Command {
 			multiple: true,
 			exclusive: ['detached'],
 		}),
+		draft: flags.boolean({
+			description: stripIndent`
+				Instruct the builder to create the release as a draft. Draft releases are ignored
+				by the 'track latest' release policy but can be used through release pinning.
+				Draft releases can be marked as final through the API. Releases as created
```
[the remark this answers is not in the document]
pipex: done!

lib/utils/remote-build.ts
pipex: balena-builder requires it like this. @toochevere how do you feel about changing your PR to use camelCase for the `isDraft` query param?

pipex: @pdcastro I did change other instances in this PR though

toochevere:
Moved back to draft because something is broken along the way... the builder is only creating draft releases, never creating non-draft.

toochevere:
> Moved back to draft because something is broken along the way... the builder is only creating draft releases, never creating non-draft.

The problem was on the builder.  It has been corrected.

toochevere:
@balena-ci I self certify!

toochevere (approved):

PR #2272, opened 2021-05-20
doc: Document the image nameing scheme used by deploy

robertgzr:
When using deploy to create a release for a multi-container project one
needs to follow this to avoid triggering a build. Relevant for CI/CD scenarios
with their own build pipeline.

Change-type: patch
Signed-off-by: Robert Günzler <robertg@balena.io>

### DIFF
--- a/lib/commands/deploy.ts
+++ b/lib/commands/deploy.ts
@@ -81,10 +81,12 @@ the image in the balenaCloud build servers.)
 Unless an image is specified, this command will look into the current directory
 (or the one specified by --source) for a docker-compose.yml file.  If one is
 found, this command will deploy each service defined in the compose file,
-building it first if an image for it doesn't exist. If a compose file isn't
-found, the command will look for a Dockerfile[.template] file (or alternative
-Dockerfile specified with the \`-f\` option), and if yet that isn't found, it
-will try to generate one.
+building it first if an image for it doesn't exist. Image names will be looked
+up according to the scheme: \`<projectName>_<serviceName>\`.
+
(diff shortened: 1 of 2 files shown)


lib/commands/deploy.ts
```diff
@@ -81,10 +81,12 @@ the image in the balenaCloud build servers.)
 Unless an image is specified, this command will look into the current directory
 (or the one specified by --source) for a docker-compose.yml file.  If one is
 found, this command will deploy each service defined in the compose file,
-building it first if an image for it doesn't exist. If a compose file isn't
-found, the command will look for a Dockerfile[.template] file (or alternative
-Dockerfile specified with the \`-f\` option), and if yet that isn't found, it
-will try to generate one.
+building it first if an image for it doesn't exist. Image names will be looked
+up according to the scheme: <projectName>_<serviceName>.
```
pdcastro, line 85: The CI tests are failing because this line is not matching between `deploy.ts` and `cli.markdown` -

```
diff --git a/doc/cli.markdown b/doc/cli.markdown
index 4f87ee7a..4ee75d51 100644
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
-up according to the scheme: `<projectName>_<serviceName>`.
+up according to the scheme: <projectName>_<serviceName>.

Error: "catch-uncommitted": Error: npx failed with exit code 1:
```
https://ci.balena-dev.com/builds/1082955

robertgzr: fixed o/

pdcastro (approved):

Issue #1708, opened 2020-04-13
Investigate variable delay after `balena help`.


pdcastro:
`balena help` is currently slow because is loads all command modules at runtime to extract usage information. We could / should instead do this at compile time, like we do for `cli.markdown` documentation. But perhaps differently from `cli.markdown`, the file did not have to be committed to the github repo, instead treated like the `build` folder and uploaded to the npm registry. (Maybe it could be generated in the `build` folder.)

pdcastro:
We could perhaps also remove mixpanel/amplitude events from `balena help`, so there's nothing to "hang" an immediate command exit.

srlowe:
@pdcastro I think this is no longer relevant?, as help seems quite fast now (I think it uses the oclif manifest, which caches this info).

pdcastro:
I think the olicf manifest plus the full CLI conversion to oclif made a good difference, yes! There is still this comment:
> We could perhaps also remove mixpanel/amplitude events from `balena help`, so there's nothing to "hang" an immediate command exit.

But it's also not a big deal. @srlowe, if you don't think that removing the mixpanel/amplitude event makes much of a difference, or if the effort doesn't justify it, feel free to close this issue.

srlowe:
@pdcastro Ah yes, we probably should.  I'll leave it open but change the name then.

srlowe:
@pdcastro I've tested, and mixpanel is not currently being called with `balena` `balena help` or `balena help devices`, so will close.

pdcastro:
> @pdcastro I've tested, and mixpanel is not currently being called with `balena` `balena help` or `balena help devices`, so will close.

Ha. OK. Then it's something else that hangs for half a second or so after the output of `balena help` is printed.

srlowe:
@pdcastro Yeah it seems to be quite intermittent - from no delay, to about half a second as you say.  I'll reopen this as an investigation task.

srlowe:
@pdcastro Is this still happening for you?  I don't get a delay now.

pdcastro:
@srlowe, indeed it is not happening right now. Would it be because some network endpoint is currently returning a response very quickly, whereas a few days ago it was taking longer? In which case, the investigation might be to find out which endpoint, perhaps add an artificial delay for testing purposes, and then decide whether there is a case for `unref`'ing the socket to allow the CLI to exit promptly. I mean something like this line of code:
[...]

PR #2273, opened 2021-05-23
Fix fast-boot module caching with read-only installation folders

pdcastro:
Connects-to: #1708

While comparing the runtime performance of a caxa-based Linux installer ([Flowdock thread](https://www.flowdock.com/app/rulemotion/i-cli/threads/lqsgv7orNqQAinI5hwKYDs66nLw)) with the pkg-based standalone zip package for Linux, I noticed that the latter was consistently taking more than 1 second to run `balena version`, not taking advantage of [fast-boot](https://www.npmjs.com/package/fast-boot2) caching. On investigation, I found it was because of 2 issues:

* The CLI was attempting to create a `.fast-boot.json` file on the parent directory of the `./bin/balena` entry-point script which, in the case of the standalone zip package, was `pkg`'s internal read-only filesystem. This would silently fail.
* Compounding that issue, the CLI was missing a call to `fast-boot2`'s `stop()` function on graceful exit. By default, `fast-boot2` sets a 1-second timeout before writing the module cache changes to disk (in order to batch multiple changes, improving performance). This applies whether or not permissions exist to save the `.fast-boot.json` file to disk, but where permissions exist, it causes a one-off CLI exit delay, and where permission is denied, this causes a delay on every CLI execution. It's mainly noticeable on short-lived commands.

Change-type: patch

### DIFF
--- a/lib/fast-boot.ts
+++ b/lib/fast-boot.ts
@@ -1,0 +1,119 @@
+/**
+ * @license
+ * Copyright 2021 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This module sets up the `fast-boot2` module, including testing whether
+ * we have permissions over the cache file before even attempting to load
+ * fast boot.
+ * DON'T IMPORT BALENA-CLI MODULES HERE, as this module is loaded directly
+ * from `bin/balena`, before the CLI's entrypoint in `lib/app.ts`.
+ */
+
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+// `@types/node` does not know about `options: { bigint?: boolean }`
+type statT = (
+	fPath: string,
+	options: { bigint?: boolean },
+) => fs.Stats | Promise<fs.Stats>;
+
+// async stat does not work with pkg's internal `/snapshot` filesystem
+const stat: statT = process.pkg ? fs.statSync : fs.promises.stat;
+
+let fastBootStarted = false;
+
+export async function start() {
+	if (fastBootStarted) {
+		return;
+	}
+	try {
+		await $start();
+		fastBootStarted = true;
+	} catch (e) {
+		if (process.env.DEBUG) {
+			console.error(`\
+[debug] Unable to start 'fast-boot2':
+[debug] ${(e.message || '').split('\n').join('\n[debug] ')}
+[debug] The CLI should still work, but it will run a bit slower.`);
+		}
+	}
+}
+
+export function stop() {
+	if (fastBootStarted) {
+		require('fast-boot2').stop();
+	}
+	fastBootStarted = false;
+}
+
+async function $start() {
+	const dotBalena = process.platform === 'win32' ? '_balena' : '.balena';
+	// TODO: take into account `~/.balenarc.yml` or `./balenarc.yml`,
+	// without hurting performance at this early loading stage.
+	const dataDir = path.normalize(
+		process.env.BALENARC_DATA_DIRECTORY || path.join(os.homedir(), dotBalena),
+	);
+	// Consider that the CLI may be installed to a folder owned by root
+	// such as `/usr[/local]/lib/balena-cli`, while being executed by
+	// a regular user account.
+	const cacheFile = path.join(dataDir, 'cli-module-cache.json');
+	const root = path.join(__dirname, '..');
+	const [, pJson, pStat, nStat] = await Promise.all([
+		ensureCanWrite(dataDir, cacheFile),
+		import('../package.json'),
+		stat(path.join(root, 'package.json'), { bigint: true }),
+		stat(path.join(root, 'npm-shrinkwrap.json'), { bigint: true }),
+	]);
+	// Include timestamps to account for dev-time changes to node_modules
+	const cacheKiller = `${pJson.version}-${pStat.mtimeMs}-${nStat.mtimeMs}`;
+	require('fast-boot2').start({
+		cacheFile,
+		cacheKiller,
+		cacheScope: root,
+	});
+}
+
+/**
+ * Check that `file` has write permission. If so, return straight away.
+ * Throw an error if:
+ * - `file` exists but does have write permissions.
+ * - `file` does not exist and `dir` exists, but `dir` does not have
+ *    write permissions.
+ * - `file` does not exist and `dir` does not exist, and an attempt
+ *    to create `dir` failed.
+ */
+async function ensureCanWrite(dir: string, file: string) {
+	const { access, mkdir } = fs.promises;
+	try {
+		try {
+			await access(file, fs.constants.W_OK);
+			return;
+		} catch (e) {
+			// OK if file does not exist
+			if (e.code !== 'ENOENT') {
+				throw e;
+			}
+		}
+		// file does not exist; ensure that the directory is writable
+		await mkdir(dir, { recursive: true, mode: 0o755 });
+		await access(dir, fs.constants.W_OK);
+	} catch (e) {
+		throw new Error(`Unable to write file "${file}":\n${e.message}`);
+	}
+}
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
       "node_modules/open/xdg-open",
       "node_modules/windosu/*.bat",
       "node_modules/windosu/*.cmd",
+      "npm-shrinkwrap.json",
       "oclif.manifest.json"
     ]
   },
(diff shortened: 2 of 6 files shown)


lib/fast-boot.ts
```diff
@@ -0,0 +1,110 @@
...
+	const dotBalena = process.platform === 'win32' ? '_balena' : '.balena';
+	// TODO: take into account `~/.balenarc.yml` or `./balenarc.yml`,
+	// without hurting performance at this early loading stage.
+	const dataDir = path.normalize(
+		process.env.BALENARC_DATA_DIRECTORY || path.join(os.homedir(), dotBalena),
+	);
+	// Consider that the CLI may be installed to a folder owned by root
+	// such as `/usr[/local]/lib/balena-cli`, while being executed by
+	// a regular user account.
+	const cacheFile = path.join(dataDir, 'cli-module-cache.json');
```
pdcastro, line 43: The code comment above, and this PR's description, explains the motivation for changing the location of the module cache file:
* Before this PR: `.fast-boot.json` file at the root of the CLI installation folder.
* After this PR: `~/.balena/cli-module-cache.json` (user's homedir) (`_balena` instead of `.balena` on Windows, see [FAQ](https://github.com/balena-io/balena-cli/blob/master/TROUBLESHOOTING.md#how-do-i-make-the-balena-cli-persist-data-in-another-directory)).

Consider not just `/usr[/local]/lib/balena-cli` (used by the caxa-based installer in the works and the GUI installer for macOS), but also `pkg`'s internal read-only snapshot filesystem and alternative user-defined installation folders like `/opt/balena-cli`.

Basically `fast-boot` never worked properly in such installations or with `pkg`, actually adding a small performance hit instead of a small performance benefit, and none ever noticed (or cared to fix?). This PR is doing something about it.

(An alternative "doing something about it" is to simply remove `fast-boot2` as "not worth the trouble", but on balance and considering it's already being used, I thought this PR was preferable.)

On Linux when `sudo` is used, depending on system configuration, the homedir may be `/root/` instead of the regular user's homedir. It means there may be 2 separate module caches files (root and non-root), but that's OK. Trying to unify them would add complexity that I suggest is not worth the benefit.

package.json
```diff
@@ -41,6 +41,7 @@
       "node_modules/open/xdg-open",
       "node_modules/windosu/*.bat",
       "node_modules/windosu/*.cmd",
+      "npm-shrinkwrap.json",
```
pdcastro, line 44: `npm-shrinkwrap.json` is now added to the `pkg` executable so that its timestamp can be used for the `cacheKiller` option of `fast-boot2`. It should be less than 200KB when compressed. (We could be better, use an empty file with a similar timestamp, hardcode exceptions... But probably not worth it.)

pdcastro:
@srlowe, @Page-, have a look at this PR:
* Page, inviting you to review it because you had forked `fast-boot2` from `fast-boot` and made improvements to it, so you know what it is about. :-)
* Scott, I think this PR address ***a component*** of issue #1708 [Investigate variable delay after balena help](https://github.com/balena-io/balena-cli/issues/1708), by chance. With the standalone zip package (`pkg`), a delay of up to 1 second would always be added to the `balena version` command, because of the read-only internal filesystem. With other installers, a similar delay would be added but only when the module cache was modified, which would happen the first time a command was executed or the first that different flags were used.

I think I've tested the PR well enough (Windows admin / non-admin, Linux with and without sudo, docker container, Mac GUI installer).

srlowe (commented):
Well spotted, and nice work.   Looks good to me, but I'll give Page a chance to comment before approving.

lib/fast-boot.ts
```diff
@@ -0,0 +1,110 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This module sets up the `fast-boot2` module, including testing whether
+ * we have permissions over the cache file before even attempting to load
+ * fast boot.
+ * AVOID LOADING OTHER MODULES HERE AS MUCH AS POSSIBLE, AND KEEP ALL
+ * FUNCTIONS SYNCHRONOUS!
```
pdcastro, line 23: I had thought that functions had to be synchronous in this file because the `await` keyword cannot be used in the module top level in `bin/balena`, but I could probably refactor `bin/balena` to have an `async` function (so that await can be used) and then this file could have async functions and I could actually try parallelising the execution of some code with `Promise.all()`. Converting the PR back to draft while I test this.

pdcastro:
@balena-ci retest

pdcastro:
@Page-, @srlowe, PR ready for review again, now nicely async. 😀

pdcastro:
@Page- wrote on [FD](https://www.flowdock.com/app/rulemotion/i-cli/threads/FknMwrLeJ4IiMbqvvcOgZgHbBLa):
> I think you could fix most of the problems and get more benefit if the cache was bundled into the installers/standalone versions
> at least the minimal cache that comes from running eg `balena help`

Bundling an initial cache file in the installers would be a performance optimization, but it would still require the initial cache file to be writable and updated at runtime (unless the cache file was "maximal", but I wouldn't even know how to produce it and it might be so large as to negate the performance improvement afforded by `fast-boot2`). Bundling an initial cache file would not solve the 2 bugs addressed in this PR: 1- that currently the cache file may be located on a non-writable folder (due to permissions or a read-only filesystem), and 2- that the CLI was not calling `require('fast-boot2').stop()` on graceful exit, which would then cause a cache file write timeout of up to 1 second when the module cache was updated.

I had a go at measuring the difference between having an initial cache file (assuming it is the same as "subsequent executions" after deleting the cache file) and not having it, and how it compares with not using `fast-boot2` at all. On my Mac laptop, executing `./bin/balena version` with the changes in this PR:

`./bin/balena help`
* Without `fast-boot2`: ~360ms
* With `fast-boot2`, no initial cache file, first execution: ~400ms
* With `fast-boot2`, no initial cache file, subsequent executions: ~340ms

`./bin/balena version`
* Without `fast-boot2`: ~780ms
* With `fast-boot2`, no initial cache file, first execution: ~800ms
* With `fast-boot2`, no initial cache file, subsequent executions: ~750ms

`./bin/balena whoami`
* Without `fast-boot2`: ~870ms
* With `fast-boot2`, no initial cache file, first execution: ~910ms
* With `fast-boot2`, no initial cache file, subsequent executions: ~830ms

Having an initial cache file would be comparable to the "subsequent executions" without an initial cache file, so it might save save 60ms for `balena help`, but it's on the first execution only. We would still need to copy the initial cache file to the user's homedir, so it's writeable / updatable.

I think the difference was more significant in the past -- [the original PR from Shaun 2.5 years ago](https://github.com/balena-io/balena-cli/pull/1071) reported larger gains:

* `balena whoami` and `balena apps` from ~3.5s to ~1.3s
* `balena help` from ~1.8s to ~1.4s

srlowe:
@pdcastro @Page- Looking at those benchmarks, it seems to me like we would do better just to get rid of fastboot altogether.  Is 20-40ms performance gain really worth the extra complexity? (especially considering the first execution is slower with fastboot)

pdcastro:
> Looking at those benchmarks, it seems to me like we would do better just to get rid of fastboot altogether. Is 20-40ms performance gain really worth the extra complexity? (especially considering the first execution is slower with fastboot)

The same thought occurred to me. The first execution can be ignored because:
* The numbers I shared above apply after the CLI has been "primed" by executing it several times, then using the near-best result. "Repeated first executions" were obtained by repeatedly running:
  *  `rm ~/.balena/cli-module-cache.json && time ./bin/balena help`  
(note that `time` does not include the time it takes to run `rm`)
* The _very first_ execution after the CLI is installed is an outlier -- it can takes several seconds, faster on Linux, a bit worse on macOS, a lot worse on Windows, because:
  * The OS will read the files from disk (possibly spinning magnetic HDD) and cache them in RAM memory, especially the large Node.js executable.
  * Windows will be inspecting the CLI for malware, which takes some 10 or 20 seconds (?) (we plan to have the Windows GUI installer execute the CLI at the end of the installation, so that the malware scanning happens during installation rather than during first execution).

* It is one-off, until the CLI is updated to a new version.
* The difference is only ~50ms which is really negligible on the _very first_ execution.

Comparing with / without `fast-boot2` (considering primed, "subsequent executions" only), an improvement of less than 50ms makes it indeed hard to justify the additional complexity. However, I would say that, at this point in time, we are better off keeping `fast-boot2` because:

* If we remove it, at some point in the future someone could come up with a PR adding `fast-boot` to the CLI, like Shaun did in the past. By already having `fast-boot`, that someone could instead focus on "what else" can be done to improve performance, or what tweaks can be made to the usage of `fast-boot`.
* I believe that the largest cost of adopting `fast-boot` has already been incurred: the dev time spent by Shaun and Page (who forked `fast-boot` as `fast-boot2`) to add `fast-boot` in the first place, then Thodoris, myself and others to cover the edge cases (like this PR). While there may still be additional issues to address in the future, I think it will be only a small fraction of the effort already spent at adopting `fast-boot`.
* Though ~50ms isn't a great improvement, it's still an improvement. It seems that in the past the improvement was much greater than this, and maybe in the future it could be the case again (I guess if `npm-shrinkwrap.json` became significantly suboptimal).
  * By the way, I've just realized that `npm-shrinkwrap.json` was added in CLI v11.4.1, while `fast-boot` was added in v9.10.0. It's possible that this made a difference -- and even improvements to the `npm` tool itself might have made a difference over time.

pdcastro:
> It's possible that this made a difference

Actually what really made a difference must have been Page's effort in a flurry of CLI PRs focusing at startup time (and other improvements) starting at CLI v11.23.0, where he moved import clauses from the top of files to inside functions, which we then kept doing. By loading much fewer modules, by significantly reducing the number of `require()` statements, the benefits of `fast-boot` are reduced (as `fast-boot` caches the resolution of `require()` statements), and the CLI loads faster whether with or without `fast-boot`.

pdcastro:
@balena-ci rebase

srlowe (approved):

PR #2274, opened 2021-05-24
Add balena.yml handling to `balena deploy` release creation

pipex:
This change allows use of a contract and release semver when doing a push, and is part of the larger feature to use the builder as part of a CI/CD pipeline.

Change-type: minor
Depends-on: https://github.com/balena-io-modules/balena-release/pull/19


toochevere:
Tests?

pdcastro:
Build failing with:
```
...
Validating commit: f5e087e3c968196470c0324e5023c0589d4d90a7

WIP:  Add balena.yml handling to `balena deploy` release creation

Found the following errors:

Error: The commit subject must not start with a leading space
```
https://ci.balena-dev.com/builds/1099124

If I was to guess, it's the 2 consecutive blank spaces between the "word" `WIP:` and the word `Add`. I guess it is related to the fact that the semver change type may be _optionally_ specified at the beginning of the subject line, for example:

```
patch: Add balena.yml handling to `balena deploy` release creation
minor: Add balena.yml handling to `balena deploy` release creation
major: Add balena.yml handling to `balena deploy` release creation
```
I think that people usually write `[WIP]` within square brackets instead of `WIP:`. Replacing 2 spaces with 1 may also do the trick, not sure. This is just a guess, however informed. :-)

lib/utils/compose_ts.ts
```diff
@@ -1314,12 +1316,23 @@ export async function deployProject(
 
 	let $release: Release;
 	try {
+		let contractContent: string | undefined;
+		let contract = await tryGetContractContent(`${projectPath}/balena.yaml`);
```
pdcastro, line 1320: * Not a big deal, but is there a spec or requirement or precedent or anticipated use case or such to support the `.yaml` file extension in addition to `.yml`? If not, I suggest accepting `.yml` only. Otherwise, it's just extra logic to maintain over time. Less code, less chance of bugs.
* It might work either way, but the preferred way of dealing with paths in the CLI codebase -- considering we support Windows as well -- is to use `path.join(projectPath, 'balena.yml')`. On Windows, this will insert a backslash instead of forward slash as path separator.

lib/utils/compose_ts.ts
```diff
@@ -1405,6 +1418,18 @@ export function createRunLoop(tick: (...args: any[]) => void) {
 	return runloop;
 }
 
+async function tryGetContractContent(
+	filePath: string,
+): Promise<any | undefined> {
+	try {
+		const fileContentAsString = await fs.readFile(filePath, 'utf8');
+		return jsyaml.load(fileContentAsString) as any;
```
pdcastro, line 1426: Does the builder have any logic to validate contracts, like a schema? If not, at least a basic sanity test to check that this is not a blank string or null / undefined or empty object, and also a test that differentiates between:
* `./balena.yml` and
* `./.balena/balena.yml` - this one representing build secrets and variables and documented at https://www.balena.io/docs/learn/deploy/deployment/#build-time-only-secret-file

Users are likely to swap these files at some point, so it would be good to have at least some minimal validation. We have already discussed the intention of unifying these two `balena.yml` (while somehow preserving backwards compatibility) in two product calls, but the work hasn't been done. You're super welcome to do it if you feel like! Some more details on this brainstorm topic: https://jel.ly.fish/90f5a591-db7e-4a1a-8f72-1a38f203e3f2

lib/utils/compose_ts.ts
```diff
@@ -1314,12 +1316,23 @@ export async function deployProject(
 
 	let $release: Release;
 	try {
+		let contractContent: string | undefined;
+		let contract = await tryGetContractContent(`${projectPath}/balena.yaml`);
+		if (!contract) {
+			contract = await tryGetContractContent(`${projectPath}/balena.yml`);
+		}
+		if (contract) {
```
pdcastro, line 1324: Is contract content allowed without a (valid) contract version? If not, it might make sense to check for a valid contract version. See other comment for contract validation.

pdcastro:
> Tests?

Indeed. We already have some tests for `balena deploy` that could be extended, so we don't have to start from scratch.
https://github.com/balena-io/balena-cli/blob/v12.44.17/tests/commands/deploy.spec.ts
`build` and `push` have more test cases / examples:
https://github.com/balena-io/balena-cli/blob/v12.44.17/tests/commands/build.spec.ts
https://github.com/balena-io/balena-cli/blob/v12.44.17/tests/commands/push.spec.ts

Most of the test cases work by mocking the balena API, balenaEngine and device supervisor, at the HTTP level using [nock](https://www.npmjs.com/package/nock). We then check that the HTTP payload matches what's expected. In the case of this PR, this might mean checking that the create release requests to the API include contracts as expected. If unit tests are sufficient that's all fine, but most of the existing test cases are more like integration / end-to-end tests where HTTP communication is mocked. They are not very easy to digest at first, but they're quite effective. An essential tool to write and debug these tests is a HTTP sniffer / inspector, such as ngrep, [HTTP Toolkit](https://httptoolkit.tech/), Wireshark, etc. The workflow would be:

* Run the CLI command normally (not tests) while recording / viewing the HTTP requests.
* Find the requests that create releases, including the contract content.
* Decide whether it looks good. If not good, fix it. If good, copy and paste the payloads to data files such as:
https://github.com/balena-io/balena-cli/tree/v12.44.17/tests/test-data/api-response
* Then write the tests cases that ensure the CLI produces that output.

Some would way it should be other way around -- write test first, code later -- and that's somewhat doable with unit tests or smaller scopes, which we should perhaps do too. With end-to-end tests at the HTTP level, it is rather difficult to write the tests first.

lib/utils/compose_ts.ts
pipex: All documentation and code I have found for this accepts `balena.yml` only. So I think it makes sense to maintain that here.

https://www.balena.io/docs/learn/deploy/deploy-with-balena-button/#balenayml-configuration-file
https://github.com/balena-io/balena-ui/blob/998d0ffbe8465447c960dc7472c204a4108a0cac/src/utils/balenaConfig.ts#L206-L234

lib/utils/compose_ts.ts
pipex: Looks like the only required field in a contract is `type`, with a  descriptive error message this might be enough to identify and clue the user about the wrong configuration path

https://www.balena.io/docs/learn/deploy/deploy-with-balena-button/#balenayml-configuration-file
https://github.com/product-os/contrato/blob/master/lib/types/definitions.cue

lib/utils/compose_ts.ts
pipex: Also https://github.com/balena-io/balena-ui/blob/998d0ffbe8465447c960dc7472c204a4108a0cac/src/utils/balenaConfig.ts#L116-L156

pipex: Sorry about all the comments, just leaving some context to pick this up later

pdcastro: Great comments 👍

toochevere: Just FYI, the reason I went with both extensions was that is how the builder does it.  So the builder could technically take either.  It seems right to keep it consistent?

lib/utils/compose_ts.ts
toochevere: I believe it is ok to have no version.

lib/utils/compose_ts.ts
pipex: @toochevere I'd opt to remove the `.yaml` extension from the builder since it makes things simpler and consistent all around and sounds like it shouldn't have any dire consequences. I'll ask on FD to see if there is some definite information about `balena.yml`

lib/utils/compose_ts.ts
pipex: I wonder if we should validate semver as well. If an invalid semver is used, the API request will fail, but not sure what is the good practice here.

pipex:
- [x] Add basic contract validation
- [x] Add version on release creation
- [x] Add contract contents when updating the release
- [ ] Update balena-release 
- [ ] Add `--draft` option to deploy
- [ ] Add tests

pdcastro:
Will this PR address issues #2183 and #1908? Are there any contradiction between those issues and this PR?

pipex:
@pdcastro As you pointed out in #1908, `release.semver` replaces `release.release_version` so we could close that as wontfix once this is merged. Regarding #2183, i think this addresses that issue, so the behaviors of `balena push` and `balena deploy`are equivalent with regards to contracts.

tests/commands/deploy.spec.ts
```diff
@@ -18,6 +18,7 @@
 import { expect } from 'chai';
 import { promises as fs } from 'fs';
 import * as _ from 'lodash';
+import * as nock from 'nock';
```
pdcastro, line 21: @toochevere, just a note as I see that this branch/PR is "out of date with the base branch" and test code is being added, while quite a few changes were made to the master branch affecting test code, including bumping the balena-sdk which resulted in new API endpoints being used and required some changes to nock mocked endpoints.
Maybe it won't make any difference to this PR, 🤞  but I suggest rebasing on master sooner than later.

toochevere: thanks for the heads up... will do.

lib/utils/compose_ts.ts
```diff
@@ -1288,6 +1289,9 @@ async function pushServiceImages(
 	);
 }
 
+// TODO: This should be shared between the CLI & the Builder
+const PLAIN_SEMVER_REGEX = /^([0-9]+)\.([0-9]+)\.([0-9]+)$/;
```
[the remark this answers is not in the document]
pdcastro: Connects-to: #2337

PR #2276, opened 2021-05-27
Update advanced installation instructions

pdcastro:
Replace 'python' with 'python3' in `apt-get install` command line, and refactor for clarity.

Motivated by making the CI worker setup simpler ([FD thread](https://www.flowdock.com/app/rulemotion/london_office/threads/AfBrtp-Wd0M2YlXb_1nvOKHO3_K)). Python 2 is way past its end of life date (1st Jan 2020). I gather that Python is required by [node-gyp](https://www.npmjs.com/package/node-gyp) which the CLI uses to compile native Node.js modules, and `node-gyp` is happy with either python 2 or python 3.

Change-type: patch


pdcastro:
@balena-ci retest

PR #2283, opened 2021-06-23
preload: Catch sfdisk errors that result in an endless spinner

klutchell:
See: https://github.com/balena-io-modules/balena-preload/pull/247
Resolves: #1810, #2045
Change-type: patch
Changelog-entry: preload: Catch sfdisk errors that result in an endless spinner

### DIFF
--- a/lib/commands/preload.ts
+++ b/lib/commands/preload.ts
@@ -169,6 +169,14 @@ Can be repeated to add multiple certificates.\
 		try {
 			const fs = await import('fs');
 			await fs.promises.access(params.image);
+			const path = await import('path');
+			if (path.extname(params.image) === '.zip') {
+				console.warn(stripIndent`
+					------------------------------------------------------------------------------
+					Warning: A zip file is only accepted for the Intel Edison device type.
+					------------------------------------------------------------------------------
+					`);
+			}
 		} catch (error) {
 			throw new ExpectedError(
 				`The provided image path does not exist: ${params.image}`,
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3443,9 +3443,9 @@
       }
     },
     "balena-preload": {
-      "version": "10.4.7",
-      "resolved": "https://registry.npmjs.org/balena-preload/-/balena-preload-10.4.7.tgz",
-      "integrity": "sha512-lsE0aVhgm7pPFVm1Rvt12RvZpP4pN8N5mXDEvETh6o16RQLNwJAek9HzBY0yenM/D0CEoFjRkI/QavCiSdgn+A==",
+      "version": "10.4.8",
+      "resolved": "https://registry.npmjs.org/balena-preload/-/balena-preload-10.4.8.tgz",
+      "integrity": "sha512-HENu1BfBW+f9VthtpentxU24inhGI1pJX1Q42Qv+iHyXoBxbd2GzCW88vvWjRbll8j+jM25PzkSABqxFRWYXMA==",
       "requires": {
         "archiver": "^3.1.1",
         "balena-sdk": "^15.3.1",
--- a/package.json
+++ b/package.json
@@ -205,7 +205,7 @@
     "balena-errors": "^4.7.1",
     "balena-image-fs": "^7.0.6",
     "balena-image-manager": "^7.0.3",
-    "balena-preload": "^10.4.7",
+    "balena-preload": "^10.4.8",
     "balena-release": "^3.0.0",
     "balena-sdk": "^15.36.0",
     "balena-semver": "^2.3.0",


pdcastro:
I think this PR is enough to close issue #2045 (it could be reopened if we find additional cases of endless spinners).

As for issue #1810, what could we (easily) do in this PR to be able to close it as well? How about: check whether the input image file name has the `.zip` extension. If so, print a warning message that `.zip` files can only be used with the Intel Edison device type. That's all. The documentation of `preload` already touches on this:


```
$ balena help preload
...
DESCRIPTION
  Preload a balena application release (app images/containers), and optionally
  a balenaOS splash screen, in a previously downloaded '.img' balenaOS image file
  in the local disk (a zip file is only accepted for the Intel Edison device type).
```

klutchell:
> How about: check whether the input image file name has the `.zip` extension. If so, print a warning message that `.zip` files can only be used with the Intel Edison device type. That's all. The documentation of `preload` already touches on this:

So a warning for every zip detected, including Intel Edison because we haven't checked the device type? This warning would be printed by the CLI itself rather than returned by the preload module?

lib/commands/preload.ts
```diff
@@ -169,6 +169,14 @@ Can be repeated to add multiple certificates.\
 		try {
 			const fs = await import('fs');
 			await fs.promises.access(params.image);
+			const path = await import('path');
```
klutchell, line 172: There are other ways to get the file extension without importing path (if desired). LMK

pdcastro:
> So a warning for every zip detected, including Intel Edison because we haven't checked the device type? This warning would be printed by the CLI itself (eg `preload.ts`) rather than returned by the preload module `preload.js`?

Yes and yes, it is what I had in mind. But then you wrote _"because we haven't checked the device type"_ and it led me to trying to remember why this wasn't implemented before. One reason was that the `-a` flag is optional, and when not provided, we don't know the device type and thus cannot tell whether or not the image file should be a zip file. Another reason is that past discussions concluded that the CLI should also extract device type information from the image itself. Between these reasons, we implemented nothing and users kept coming across issues like #1810. Also, often the `-a` will be provided even if not required, and in this case we can at least tell whether or not the device type is Intel Edison (even if we couldn't tell about "compatible architectures" like the RPi family).

I am contributing a commit that improves validation further with what I thought was relatively easy to do. Please review it and feel free to squash it into your validation commit, as there's duplicate functionality.

lib/commands/preload.ts
```diff
@@ -305,6 +304,33 @@ Can be repeated to add multiple certificates.\
...
+			throw new ExpectedError(`Cannot access image file: "${image}"`);
+		}
+		const path = await import('path');
+		const zipMsg =
+			'The image file must be unzipped unless the device type is Intel Edison';
+		const isZip = path.extname(image) === '.zip';
+		if (deviceType) {
+			if (deviceType === 'intel-edison' && !isZip) {
+				throw new ExpectedError(
+					'Error: The Intel Edison device type requires the image to be a zip file',
```
pdcastro, line 321: Oh, I think this validation is broken. Using the web dashboard, I just realised that an `intel-edison` _app_ also allows downloading images for the 'qemu x86 32bit' device type, and in this case, the zip file must be unzipped even though the app type is `intel-edison`.

pdcastro: I've amended my commit to cope with 'qemu x86 32bit' devices (non zip image files) in 'intel-edison' apps. Still needs your review and squashing. :-)

lib/commands/preload.ts
```diff
@@ -305,6 +304,31 @@ Can be repeated to add multiple certificates.\
...
+			await fs.promises.access(image);
+		} catch (error) {
+			throw new ExpectedError(`Cannot access image file: "${image}"`);
+		}
+		const path = await import('path');
+		if (path.extname(image) === '.zip') {
+			const zipMsg =
+				'The image file must be unzipped unless the device type is Intel Edison';
+			if (appDeviceType && appDeviceType !== 'intel-edison') {
+				throw new ExpectedError(`Error: ${zipMsg}`);
```
pdcastro, line 319: So basically the improvement is that, if we opportunistically know for a fact that the image file must not be a zip file, we abort rather than just warn. Further UX improvement might be to unzip the file for the user, but I'd say that this is already way better than infinite spinners or error messages about partition tables (when we can avoid them).

lib/commands/preload.ts
```diff
@@ -164,29 +165,27 @@ Can be repeated to add multiple certificates.\
...
-		} catch (error) {
-			throw new ExpectedError(
-				`The provided image path does not exist: ${params.image}`,
-			);
-		}
+		let appDeviceType: string | undefined;
 
 		// balena-preload currently does not work with numerical app IDs
 		// Load app here, and use app slug from hereon
-		if (options.app && !options.app.includes('/')) {
```
klutchell, line 180: Was this intentionally blocking the org/app syntax?

pdcastro: Yes. Before this PR, the purpose of this code block was to "convert" an app ID in the format `1301645`, or an app name in the format `test-rpi`, to an app slug in the format `gh_paulo_castro/test-rpi`, so as to resolve ambiguities using the logic in the CLI's implementation of `getApplication`: https://github.com/balena-io/balena-cli/blob/v12.44.21/lib/utils/sdk.ts#L30-L50
(Also, not sure if it's still the case, but at one point the `balena-preload` package would not accept numeric app IDs, and this code block would avoid that problem as well.)
The test `!options.app.includes('/')` would check whether the user had already specified a slug in the format `org/appname` on the command line, in which case this code block could be skipped.

In this PR, we are choosing to run this code block regardless of the format of `options.app`, as long as it is defined, in order to obtain `appDeviceType` in addition to addressing the concerns above, when they apply.

lib/commands/preload.ts
pdcastro: Ah, the logic is still broken, isn't it? If I create an app of type QEMU X86 32bit with the web dashboard, then `appDeviceType` will be `qemux86`, but I am allowed to add devices of type `intel-edison` to that app, whose images would be Intel Edison zip files. So, basically, there are no circumstances in which the image file is necessarily a zip file or necessarily not a zip file, based solely on the application's default device type. In other words, the logic cannot rely on `appDeviceType` and there is no point obtaining it.

If you have saved a copy of your commit as it was yesterday, perhaps we should revert to it. Sorry for the mess. :-/

pdcastro (changes requested):

lib/commands/preload.ts
klutchell: `git reflog` knows all!

pdcastro (approved):

klutchell:
Does bulldozer bot only close the first issue linked in `Resolves:`?

pdcastro:
> Does bulldozer bot only close the first issue linked in `Resolves:`?

Looks like that. Actually I thought that it was "GitHub itself" that did the issue closing, rather than bulldozer. It would have to be two lines I think:
```
Resolves: #1810
Resolves: #2045
```

klutchell:
I think on non-balena repos it would be GH doing the closing, and it underlines the `Resolves` string with a tooltip.

PR #2291, opened 2021-07-08
config write: Fix parsing of 'key' argument with numeric components

pdcastro:
Change-type: patch
See: https://forums.balena.io/t/programatically-github-actions-add-udev-rule-to-config-file/337399


pdcastro:
@balena-ci I self-certify!

PR #2295, opened 2021-07-20
Simplify runSpinner api

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -36,7 +36,6 @@ import {
 	ComposeCliFlags,
 	ComposeOpts,
 	ComposeProject,
-	Release,
 	TaggedImage,
 	TarDirectoryOptions,
 } from './compose-types';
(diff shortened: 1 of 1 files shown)


lib/utils/compose_ts.ts
```diff
@@ -1355,15 +1347,12 @@ export async function deployProject(
...
-		runloop = runSpinner(tty, spinner, `${prefix}Saving release...`);
-		release.end_timestamp = new Date();
-		if (release.id != null) {
-			try {
+		await runSpinner(tty, spinner, `${prefix}Saving release...`, async () => {
+			release.end_timestamp = new Date();
+			if (release.id != null) {
 				await releaseMod.updateRelease(pineClient, release.id, release);
-			} finally {
-				runloop.end();
```
Page-, line 1364: This also fixes a bug where if `release.id == null` then we never actually hit this specific `runloop.end` and so would continue spinning forever - I don't know if that ever happens in practice but the callback api makes it far easier to ensure the spinner is always ended and avoids the potential for the issue

pdcastro:
@balena-ci retest

pdcastro:
Retesting because of the following unusual error in the Linux build:
> https://ci.balena-dev.com/teams/main/pipelines/github-events-balena-io/jobs/node-cli/builds/629
> `connection: process error: could not find the exitcode file for the process:`
> `stat /worker-state/depot/1855445c-a1ce-47a9-4b95-e7058c20c3a6/processes/resource/exitcode:`
> `no such file or directory`

pdcastro (approved):

Issue #2302, opened 2021-07-31
Rename 'application' to 'fleet' in command names, flags and output


pdcastro:
# Summary

The objective is to rename user-visible `app` and `application` references to `fleet`:
* Rename `app` / `apps` / `application` command names and flags to `fleet` / `fleets`
* Rename the `-a` single-char flag to `-f`
* Rename such references in the output of commands such as `balena devices` and `balena envs` -- column headers in tabular output or field names in JSON output.

> _Update:_ A full list of renamed commands and flags can be found in this Wiki page:
> https://github.com/balena-io/balena-cli/wiki/Renaming-of-Applications-to-Fleets
> also shortened as: https://git.io/JRuZr

The CLI is currently at major version 12 (v12). The renaming will take place in stages, the first stage in v12 and the last stage a number of weeks later, in v13.

# Details

* Initially, while in v12, backwards compatibility should be maintained. This will be achieved by creating aliases to the existing command names and flags, so that `app(s)` or `application(s)` or `fleet(s)` could be used interchangeably. However, when the user uses `app(s)` or `--application(s)`, a warning message will be printed. For example:

```
---------------------------------------------------------------------------
[Warn] The `apps` command was renamed to `fleets`. Find out more at:
       <link to blog or wiki or docs website>
---------------------------------------------------------------------------
---------------------------------------------------------------------------
[Warn] The `--application` flag was renamed to `--fleet`. Find out more at:
       <link to blog or wiki or docs website>
---------------------------------------------------------------------------
```

* This warning message will not be printed in scripted usage (where the CLI output is parsed by users’ code), if `stderr` is detected as not being a `tty`. This would mean that a user's script could continue using CLI v12 with the `app` or `application` terminology for a long time, but this is also subject to the CLI's overall deprecation policy enforcement feature described next.

* Independently of the app/fleet renaming but coinciding in time, while at v12, a CLI deprecation policy enforcement feature will be implemented as discussed in an arch call (JF brainstorm). This feature means that, 12 months after the release of the next major version (v13), the previous major version (v12) will stop working (the CLI will abort execution for all commands) unless the user uses a `--unsupported` flag. Half way through that period (6 months after the release of the next major), the previous major will print a warning message referring to the deprecation policy and to the `--unsupported` flag (only under interactive use -- `stderr` being a `tty`).

[...]

PR #2304, opened 2021-08-05
Rename applications to fleets (stage 1)

pdcastro:
Change-type: minor
Resolves: #2302


pdcastro:
@balena-ci I self-certify!

PR #2305, opened 2021-08-05
push, build: Improve error handling (identify which service failed)

pdcastro:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/i-cli/threads/i8M4YKcfEW2FOyOGuxQEUmIC7-Z

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -1063,18 +1063,21 @@ async function performResolution(
 					if (!buildTask.buildStream) {
 						continue;
 					}
-					// Consume each task.buildStream in order to trigger the
-					// resolution events that define fields like:
-					//     task.dockerfile, task.dockerfilePath,
-					//     task.projectType, task.resolved
-					// This mimics what is currently done in `resin-builder`.
-					const clonedStream: Pack = await cloneTarStream(
-						buildTask.buildStream,
-					);
-					buildTask.buildStream = clonedStream;
-					if (!buildTask.external && !buildTask.resolved) {
+					let error: Error | undefined;
+					try {
+						// Consume each task.buildStream in order to trigger the
+						// resolution events that define fields like:
+						//     task.dockerfile, task.dockerfilePath,
+						//     task.projectType, task.resolved
+						// This mimics what is currently done in `resin-builder`.
+						buildTask.buildStream = await cloneTarStream(buildTask.buildStream);
+					} catch (e) {
+						error = e;
+					}
+					if (error || (!buildTask.external && !buildTask.resolved)) {
+						const cause = error ? `${error}\n` : '';
 						throw new ExpectedError(
-							`Project type for service "${buildTask.serviceName}" could not be determined. Missing a Dockerfile?`,
+							`${cause}Project type for service "${buildTask.serviceName}" could not be determined. Missing a Dockerfile?`,
 						);
 					}
 				}


pdcastro:
@balena-ci retest

lib/utils/compose_ts.ts
```diff
@@ -1063,15 +1063,17 @@ async function performResolution(
...
-					buildTask.buildStream = clonedStream;
+					try {
+						// Consume each task.buildStream in order to trigger the
+						// resolution events that define fields like:
+						//     task.dockerfile, task.dockerfilePath,
+						//     task.projectType, task.resolved
+						// This mimics what is currently done in `resin-builder`.
+						buildTask.buildStream = await cloneTarStream(buildTask.buildStream);
+					} catch (e) {
+						e.message = `Error resolving service "${buildTask.serviceName}":\n${e.message}`;
```
pranavpeshwe, line 1074: Not sure whether the phrase 'Error resolving service xyz' jels well here. How about 'Failed to complete build task for service xyz'? Borrowed the 'build task' terminology from the message that logically precedes this error message (viz. 'Found build tasks:').
And, I agree, this comment borders nitpicking :) Hence feel free to ignore.

pdcastro: Good call! "Error resolving service" does not mean much to end users, despite technically the error originating from the call to `multiBuild.performResolution`. "Failed to complete build task" is quite generic though, and might also cover any build errors, including errors produced by the Docker engine (or balenaEngine) during the image build. But at this point, we know that the error is specific to the project resolution stage, before the CLI asks Docker or balenaEngine to build the image, so we could try and be a bit more specific.

Project resolution means an attempt to detect the project type. For example, the CLI / builders accept Node.js projects including a `package.json` file but no `Dockerfile` nor `docker-compose.yml` file, in which case a default basic Dockerfile is automatically generated.

Paying more attention to the surrounding code, I've noticed that, a few lines below, there was already an error message that might be a compromise between error specificity and clarity for end users:
```ts
	if (!buildTask.external && !buildTask.resolved) {
		throw new ExpectedError(
			`Project type for service "${buildTask.serviceName}" could not be determined. Missing a Dockerfile?`,
```
I've now pushed a commit that re-uses that error message, instead of introducing a new error message. What do you think?

lib/utils/compose_ts.ts
```diff
@@ -1063,18 +1063,21 @@ async function performResolution(
...
+						// This mimics what is currently done in `resin-builder`.
+						buildTask.buildStream = await cloneTarStream(buildTask.buildStream);
+					} catch (e) {
+						error = e;
+					}
+					if (error || (!buildTask.external && !buildTask.resolved)) {
+						const cause = error ? `${error}\n` : '';
 						throw new ExpectedError(
-							`Project type for service "${buildTask.serviceName}" could not be determined. Missing a Dockerfile?`,
+							`${cause}Project type for service "${buildTask.serviceName}" could not be determined. Missing a Dockerfile?`,
```
pdcastro, line 1080: I've tested this by injecting a dummy `throw new Error('oops')` statement right after the call to `cloneTarStream`. The console output was:
```
$ ./bin/balena build ~/balena/balena-io-examples/balena-idling/ -f test-rpi
[Info]    No "docker-compose.yml" file found at "/Users/paulo/balena/balena-io-examples/balena-idling"
[Info]    Creating default composition with source: "/Users/paulo/balena/balena-io-examples/balena-idling"
[Info]    Building for armv7hf/raspberrypi3
[Build]   Building services...
[Build]   main Preparing...
[Info]    Docker Desktop detected (daemon architecture: "x86_64")
[Info]      Docker itself will determine and enable architecture emulation if required,
[Info]      without balena-cli intervention and regardless of the --emulated option.
[Build]   Built 1 service in 0 seconds
[Error]   Build failed.
Error: oops
Project type for service "main" could not be determined. Missing a Dockerfile?
```

lib/utils/compose_ts.ts
pranavpeshwe: My exposure (unlike yours) to balena-cli is only as a consumer of the cli, so, TBH 'Project type for service xyz could not be determined' would have confused me. Having said that, knowing which service faulted is critical. Thus, if you think this message is technically appropriate then we should go with it 👍🏻

pdcastro: > 'Project type for service xyz could not be determined' would have confused me

Note that's not the only error message, though. It would have 2 lines, the first line being, for example, _"Could not find a Dockerfile for this service"_ if that's what the error was, as in [the FD thread](https://www.flowdock.com/app/rulemotion/i-cli/threads/i8M4YKcfEW2FOyOGuxQEUmIC7-Z) :
```
...
[Build]   Built 3 services in 0 seconds
[Error]   Build failed.
Could not find a Dockerfile for this service
Project type for service <serviceName> could not be determined. Missing a Dockerfile?
```

So the user would know that issue is that a Dockerfile could not be found, and would have the name of the service. The additional "Project type could not be determined" helps support agents and CLI maintainers distinguish between a Docker engine build error and a project resolution error.

pdcastro:
@balena-ci retest

pdcastro:
@balena-ci I self-certify!

Issue #2306, opened 2021-08-10
 "Application is ambiguous" but device UUID is specified


20k-ultra:
In the following snippet I am setting a device variable for the sideshow-bob service. I am explicitly saying a device UUID which is unique across balenaCloud. However, the CLI responds saying that it can't figure out what to do because my account had access to more then 1 fleet (application) named balena-on-balena. After renaming one of those fleets

```
mig@ghost ~$ balena env add API_TOKEN ${balena_api_token} \                                                                                                               
  --service sideshow-bob \
  --device 0bd0bfe
Application is ambiguous: balena-on-balena, service: sideshow-bob
```

**Expected results**
To set service variable for device specified

**Actual results**
CLI says it cannot figure out what device to update because "Application is ambiguous"

**Steps to reproduce**
1. create an application with the same name as an app in another org so your account has access to 2 applications of the same name
2. run above command

pdcastro:
Fixed in CLI v12.45.1. Thanks for reporting it! 👍

PR #2307, opened 2021-08-11
envs, env add: Fix "Application is ambiguous" when using device UUID

pdcastro:
Resolves: #2306
Change-type: patch

Originally reported in [an FD thread](https://www.flowdock.com/app/rulemotion/r-beginners/threads/bavbqlHOYilvn--QG_qc49I2I5M) (restricted access).

### DIFF
--- a/lib/commands/envs.ts
+++ b/lib/commands/envs.ts
@@ -174,11 +174,11 @@ export default class EnvsCmd extends Command {
 				balena,
 				options.device,
 				['uuid'],
-				['app_name'],
+				['slug'],
 			);
 			fullUUID = device.uuid;
 			if (app) {
-				appNameOrSlug = app.app_name;
+				appNameOrSlug = app.slug;
 			}
 		}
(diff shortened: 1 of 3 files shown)


lib/commands/envs.ts
```diff
@@ -174,11 +174,11 @@ export default class EnvsCmd extends Command {
 				balena,
 				options.device,
 				['uuid'],
-				['app_name'],
+				['slug'],
 			);
 			fullUUID = device.uuid;
 			if (app) {
-				appNameOrSlug = app.app_name;
+				appNameOrSlug = app.slug;
```
pdcastro, line 181: It turns out that even listing env vars by device UUID could produce the error:
```
$ ./bin/balena envs -d b9a7aae
BalenaAmbiguousApplication: Application is ambiguous: balena-on-balena
```

pdcastro:
@balena-ci I self-certify!

PR #2309, opened 2021-08-17
Add deprecation policy checker and --unsupported global flag

pdcastro:
Change-type: minor

See: (restricted access) https://jel.ly.fish/ed8d2395-9323-418c-bb67-d11d32a17d00


pdcastro:
@balena-ci I self-certify!

Issue #2318, opened 2021-08-30
Should move away from the device-types/v1 endpoints in favor of the model


thgreasi:
Placeholder issue
See: https://www.flowdock.com/app/rulemotion/pub/threads/AE_IXh9Xs_I7MYOHS98oLb0xaya

thgreasi:
No we can't @myarmolinsky b/c we still use device-types/v1 in `config generate` :(

PR #2324, opened 2021-09-07
preload: Rely on the device type model to get the compatible apps

thgreasi:
On top of #2323, so let's get that in first.

Connects-to: #2318
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/i-cli/threads/s6x4Z_LoH8IG4PC_YeXMC0TP6v-

---

### DIFF
--- a/lib/commands/preload.ts
+++ b/lib/commands/preload.ts
@@ -36,13 +36,7 @@ import { isV13 } from '../utils/version';
 
 import { flags } from '@oclif/command';
 import * as _ from 'lodash';
-import type {
-	Application,
-	BalenaSDK,
-	DeviceTypeJson,
-	PineExpand,
-	Release,
-} from 'balena-sdk';
+import type { Application, BalenaSDK, PineExpand, Release } from 'balena-sdk';
 import type { Preloader } from 'balena-preload';
 
 interface FlagsDef extends DockerConnectionCliFlags {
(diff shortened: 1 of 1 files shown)


pdcastro:
@balena-ci rebase

pdcastro:
Weird -- it looks like `balena-ci rebase` kept one of the two commits, even though it is now an empty commit. It would become a duplicate bullet point in the CLI's Changelog, so deleting the empty commit would be nicer.

lib/commands/preload.ts
```diff
@@ -351,43 +345,47 @@ Can be repeated to add multiple certificates.\
...
+									$any: {
+										$alias: 'ioca',
+										$expr: {
+											ioca: {
+												is_supported_by__device_type: {
+													$any: {
+														$alias: 'isbdt',
+														$expr: {
+															isbdt: {
+																slug: deviceTypeSlug,
```
pdcastro, line 380: Would this work with device type aliases, e.g. where `deviceTypeSlug` was `nuc` instead of `intel-nuc`?
If not or in doubt, earlier the code was already executing:
```ts
    await balena.models.deviceType.get(deviceTypeSlug);
```
So I guess we could use the returned device type object to get its "canonical" slug (assuming that `models.deviceType.get` accepts device type aliases).

pdcastro: Actually I am not sure if `deviceTypeSlug` could be an alias or not. It looks like the slug could come from the `config.json` file found in the OS image being preloaded:
* https://github.com/balena-io/balena-cli/blob/v12.48.5/lib/commands/preload.ts#L531
* https://github.com/balena-io-modules/balena-preload/blob/v10.5.0/lib/preload.js#L481
* https://github.com/balena-io-modules/balena-preload/blob/v10.5.0/src/preload.py#L929
* https://github.com/balena-io-modules/balena-preload/blob/v10.5.0/src/preload.py#L700

I guess someone could manually edit the `config.json` file and replace, say, "raspberrypi3" with "raspberrypi2" when using an old `config.json` file as a template, not realising that "raspberrypi2" is an alias for "raspberry-pi2".

By the way, the `balena-preload` repo also contains an occurrence of the deprecated `models.config.getDeviceTypes()`:
https://github.com/balena-io-modules/balena-preload/blob/v10.5.0/lib/preload.js#L309

cc: @klutchell as maintainer of `balena-preload` -- As "warned" by Alex M during the all hands call, we should remove usage of the API's `/device-types/v1`, which is used by the balena SDK's deprecated `models.config.getDeviceTypes()` method. Unless Thodoris says he will create a PR for `balena-preload` as well. :-)

thgreasi: > Would this work with device type aliases

@pdcastro no it will not, but the previous implementation was also not working with aliases, so there is no change of behavior.

> earlier the code was already executing `await balena.models.deviceType.get(deviceTypeSlug);`

I don't see that in the diff. I only see previously doing `await balena.models.config.getDeviceTypes()`` and then filtering on the slug, which would not allow using aliases, since it's filtering on and then returning only the canonical DT slug.

klutchell: > By the way, the balena-preload repo also contains an occurrence of the deprecated models.config.getDeviceTypes():

Good catch, I already opened an issue for our usage of `deployArtifact` but I will rename it to cover any `/device-types/v1` endpoints.

klutchell: > It looks like the slug could come from the config.json file found in the OS image being preloaded:

This looks correct, we read it from `config.json` on the device and that should never be an alias. If someone used an alias there manually I would expect a number of issues either way.

pdcastro: > earlier the code was already executing `await balena.models.deviceType.get(deviceTypeSlug);`

By "earlier" I meant a few lines of code before the line my comment was added to. Code introduced by this PR, rather than pre-existing code. `await balena.models.deviceType.get(deviceTypeSlug);` was introduced by this PR but the return value is discarded. If we didn't discard the return value, I thought we might take the "canonical" (non alias) slug from the return value, assuming that `models.deviceType.get(deviceTypeSlug)` works with an alias.

pdcastro: > assuming that `models.deviceType.get(deviceTypeSlug)` works with an alias.

Well, it doesn't, I've just tested. `models.deviceType.get('nuc')` throws an error.

thgreasi: No it doesn't since aliases are not yet part of the model.
Atm `models.deviceType.get(deviceTypeSlug)` is there just to throw a DeviceTypeNotFoundError, essentially replacing the `throw new Error(`Device type "${deviceTypeSlug}" not found in API query`);` that we had before.

pdcastro (approved):

PR #2326, opened 2021-09-08
v13 preparations: Add feature switch for removal of '--gitignore' (push, build)

pdcastro:
Preparations for the release of CLI v13. The changes take effect when the version field in `package.json` is eventually updated to v13.0.0 or later (feature switch: avoids the need for a separate release branch).

Change-type: patch


pdcastro:
@balena-ci I self-certify!

Issue #1018, opened 2018-11-08
Deploy command ignores --tag argument


kfrance:
- Using cli version 9.1.0
- Using intel x86 archlinux system that is up to date

The tag argument is ignored when using deploy. This seems to be related to issue #825 where the tag argument is also ignored for the build command.

kfrance:
I tried pushing with projectName but I don't see any tags on the release, or anything else related to the project name I gave it.

thgreasi:
@kfrance can I get some context on what you are trying to achieve and the complete command you are using for that?
The `--projectName` parameter is only used in the deploy command when you are also using the `--build` parameter and it affects the name of the build image.
You can check the name of the resulting image name locally by running `docker images`, which will also have a suffix with the name of the service that this image is for (for single container apps that would be `myprojectname_main`).

pdcastro:
Internal discussion thread: https://www.flowdock.com/app/rulemotion/i-cli/threads/S2M8AnPbIbbH6qKQHUOEb9lYDYb

pdcastro:
Fixed in CLI v12.48.11. The `--tag` flag specifies only the `:tag` portion of the image name. For example:
```
$ balena deploy myFleet --build -s . --projectName dronecontrol --tag v2.2.4
```
Images created (e.g. as listed by `docker images` on the workstation):
```
dronecontrol_service1:v2.2.4
dronecontrol_service2:v2.2.4
dronecontrol_service3:v2.2.4
```
The default if `--tag` is not specified is `:latest`.

Help output:

```
$ balena help deploy
...
OPTIONS
...
-n, --projectName <projectName>  Name prefix for locally built images. This is the 'projectName' portion
                                 in 'projectName_serviceName:tag'. The default is the directory name.

-t, --tag <tag>                  Tag locally built Docker images. This is the 'tag' portion
                                 in 'projectName_serviceName:tag'. The default is 'latest'.
```

Issue #825, opened 2018-03-21
Build command ignores --tag argument


dfunckt:
- **resin-cli version: >=7.0.0**
- **Operating system and architecture: All**

It is currently completely ignored, but it's functionality overlaps with `--projectName`, so we need to either:

a) remove it
b) make it an alias to `--projectName`
c) replace `--projectName` with `--tag`

(c) is the most likely candidate given `--tag` existed in previous versions, but it's less accurate now with multi-container.

pdcastro:
Fixed in CLI v12.48.11. The `--tag` flag specifies only the `:tag` portion of the image name. For example:
```
$ balena build . --projectName dronecontrol --tag v2.2.4
```
Images created (e.g. as listed by `docker images` on the workstation):
```
dronecontrol_service1:v2.2.4
dronecontrol_service2:v2.2.4
dronecontrol_service3:v2.2.4
```
The default if `--tag` is not specified is `:latest`.

Help output:

```
$ balena help build
...
OPTIONS
...
-n, --projectName <projectName>  Name prefix for locally built images. This is the 'projectName' portion
                                 in 'projectName_serviceName:tag'. The default is the directory name.

-t, --tag <tag>                  Tag locally built Docker images. This is the 'tag' portion
                                 in 'projectName_serviceName:tag'. The default is 'latest'.
```

Issue #2327, opened 2021-09-08
balena build does not tag the resulting image when using `--tag`


pipex:
# Expected Behavior

Image tag should be set on resulting image when doing `balena build --tag`

```
my-project $ balena build -f my-fleet -t my-image:v1.1.0
[Build]   Built 1 service in 0 seconds
[Success] Build succeeded!
my-project $ docker images 
REPOSITORY                 TAG              IMAGE ID       CREATED          SIZE
my-project_main           latest           0ab63e543ead   21 minutes ago   221MB
my-image                      v1.1.0           0ab63e543ead   21 minutes ago   221MB
```

# Actual Behavior

Tag does not get set
```
my-project $ balena build -f my-fleet -t my-image:v1.1.0
[Build]   Built 1 service in 0 seconds
[Success] Build succeeded!
my-project $ docker images 
REPOSITORY                 TAG              IMAGE ID       CREATED          SIZE
my-project_main           latest           0ab63e543ead   21 minutes ago   221MB
```
balena push 192.168.0.12
balena deploy myFleet
balena deploy myFleet --build
balena build . -f myFleet
balena build . -A armv7hf -d raspberrypi3
```

```

# Steps to Reproduce the Problem

1. Run `balena build -f <fleet name> -t <some tag>`
2. Check the list of built images with `docker images`

# Specifications

- **balena CLI version:** 
balena-cli version "12.46.0"
Node.js version "12.20.0"
- **Cloud backend: balenaCloud?** 
- **Operating system version:** macOS 11.4
- **32/64 bit OS and processor:** 64-bit Intel processor
- **Install method:** executable installer

pdcastro:
Sounds like a duplicate of #825, and connects to #1018.

pipex:
I promised I searched the issues first. Closing this one then

Issue #1570, opened 2020-01-15
The tag option to balena build doesn't seem to work


CameronDiver:
I can't seem to get the tag variable to be respected by the `balena build` command:
```
[10:49] util-containers/network git:(master) ✗ λ balena build -t test-image -A amd64 -d test .
[Info]    Creating default composition with source: /home/cameron/Projects/util-containers/network
[Info]    Building for amd64/test
[Build]   Built 1 service in 0:33
[Build]   main Image size: 193.09 MB
[Success] Build succeeded!
[10:49] util-containers/network git:(master) ✗ λ docker images
REPOSITORY                 TAG                 IMAGE ID            CREATED             SIZE
network_main               latest              43a997b14d68        11 seconds ago      202MB
balenalib/amd64-debian     latest              328983ff8812        29 hours ago        113MB
```

- **balena CLI version:** 11.18.3
- **Operating system version:** Arch
- **32/64 bit OS and processor:** 64bit intel
- **Install method:** npm
- **If npm install, Node.js and npm version:**
```
[10:51] util-containers/network git:(master) ✗ λ node -v
v10.16.3
[10:51] util-containers/network git:(master) ✗ λ npm -v
6.9.0
```
---

*Please keep in mind that we try to use the issue tracker of this repository for specific bug
reports & CLI feature requests. General & troubleshooting questions are encouraged to be posted to
the [balena forums](https://forums.balena.io), which are monitored by balena's support team and
where the community can both contribute and benefit from the answers.*

*Before submitting this issue please check that this issue is not a duplicate. If there is another
issue describing the same problem or feature please add your information to the existing issue's
comments.*

pdcastro:
Related to:
* #825 [Build command ignores --tag argument](https://github.com/balena-io/balena-cli/issues/825)
* #1018 [Deploy command ignores --tag argument](https://github.com/balena-io/balena-cli/issues/1018)

CameronDiver:
My bad, I didn't search well enough. I'm going to close this issue and we can track it with the other - sorry for the noise!

pdcastro:
[...]

PR #2328, opened 2021-09-09
build, deploy: Fix processing of '--tag' option

pdcastro:
Resolves: #825
Resolves: #1018
Connects-to: #1570, #2327
Change-type: patch

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -118,6 +118,7 @@ export async function loadProject(
 	logger: Logger,
 	opts: ComposeOpts,
 	image?: string,
+	imageTag?: string,
 ): Promise<ComposeProject> {
 	const compose = await import('resin-compose-parse');
 	const { createProject } = await import('./compose');
@@ -156,7 +157,12 @@ export async function loadProject(
 		}
 	}
 	logger.logDebug('Creating project...');
-	return createProject(opts.projectPath, composeStr, opts.projectName);
+	return createProject(
+		opts.projectPath,
+		composeStr,
+		opts.projectName,
+		imageTag,
+	);
 }
 
 /**
@@ -411,6 +417,18 @@ async function installQemuIfNeeded({
 	return needsQemu;
 }
 
+export function makeImageName(
+	projectName: string,
+	serviceName: string,
+	tag?: string,
+) {
+	let name = `${projectName}_${serviceName}`;
+	if (tag) {
+		name = [name, tag].map((s) => s.replace(/:/g, '_')).join(':');
+	}
+	return name.toLowerCase();
+}
+
 function setTaskAttributes({
 	tasks,
 	buildOpts,
@@ -426,7 +444,7 @@ function setTaskAttributes({
 		const d = imageDescriptorsByServiceName[task.serviceName];
 		// multibuild (splitBuildStream) parses the composition internally so
 		// any tags we've set before are lost; re-assign them here
-		task.tag ??= [projectName, task.serviceName].join('_').toLowerCase();
+		task.tag ??= makeImageName(projectName, task.serviceName, buildOpts.t);
 		if (isBuildConfig(d.image)) {
 			d.image.tag = task.tag;
 		}
@@ -707,7 +725,7 @@ export async function getServiceDirsFromComposition(
  * Return true if `image` is actually a docker-compose.yml `services.service.build`
  * configuration object, rather than an "external image" (`services.service.image`).
  *
- * The `image` argument may therefore refere to either a `build` or `image` property
+ * The `image` argument may therefore refer to either a `build` or `image` property
  * of a service in a docker-compose.yml file, which is a bit confusing but it matches
  * the `ImageDescriptor.image` property as defined by `resin-compose-parse`.
  *
@@ -1666,8 +1684,9 @@ export const composeCliFlags: flags.Input<ComposeCliFlags> = {
 			"Don't convert line endings from CRLF (Windows format) to LF (Unix format).",
 	}),
 	projectName: flags.string({
-		description:
-			'Specify an alternate project name; default is the directory name',
+		description: stripIndent`\
+			Name prefix for locally built images. This is the 'projectName' portion
+			in 'projectName_serviceName:tag'. The default is the directory name.`,
 		char: 'n',
 	}),
 };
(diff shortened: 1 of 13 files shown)


lib/utils/docker.ts
```diff
@@ -72,7 +72,9 @@ export const dockerConnectionCliFlags: flags.Input<DockerConnectionCliFlags> = {
 
 export const dockerCliFlags: flags.Input<DockerCliFlags> = {
 	tag: flags.string({
-		description: 'The alias to the generated image',
+		description: `\
+Tag any built images with this value. This is the 'tag' portion in
+'projectName_serviceName:tag'. The default is 'latest'.`,
```
pdcastro, line 77: @dfunckt, re issue #825 that you created 3 years ago, and
@pipex re issue #2327 that you created a few hours ago,
you had the expectation that `--tag` could be used to specify the full / exact name of an image, e.g. `--tag my_img:my_tag`. The complication is that `balena build` or `balena deploy` may create several images, one for each service, and of course they cannot all have the same name. What this PR is currently implementing is using `--tag` only for the `:tag` part of the image name. The thinking is that it might be useful in the scenario where all service images should be tagged with, say, a common version for the app/fleet release. Example:

`$ balena build . --projectName dronecontrol --tag v2.2.4 ...`

Images created:
`dronecontrol_service1:v2.2.4`
`dronecontrol_service2:v2.2.4`
`dronecontrol_service3:v2.2.4`

What do you think? The `--tag` flag was being completely ignored for at least 3 years, so there are no worries about breaking compatibility at this point. I understand it might have worked a very a long time ago, before apps/fleets could have multiple services/containers, and `--tag` would then name the only image built.

Another option is to simply delete the `--tag` flag.

dfunckt: @pdcastro I agree that `--tag` should merely deal with the tag part of the image name. To address Felipe's use case, one would then have to just do:

```
$ balena build --projectName myfleet --tag v1.1.0
```

Which would produce something like:

```
myfleet_api:v1.1.0
myfleet_sensor:v1.1.0
myfleet_dashboard:v1.1.0
```

dfunckt: While we're at it, if it's not much of a complication I think it would be useful if `--tag` could be passed several times, to make several tags for images the same build, eg. both versioned *and* `latest`.

lib/utils/device/deploy.ts
```diff
@@ -310,7 +310,7 @@ function connectToDocker(host: string, port: number): Docker {
 	});
 }
 
-export async function performBuilds(
+async function performBuilds(
```
dfunckt, line 313: Good stuff 👍

dfunckt (commented):
Looks great to me.

lib/utils/docker.ts
pipex: This looks great @pdcastro, nice work. I think this definitely addresses the use case I saw on support this week.

My only worry would be with user confusion regarding diverging behavior between `balena build --tag` and `docker build --tag` (or `balena build` on balena engine). Even though that is addressed in the docs, I'm sure we'll get some confused users on support.

Also, what is the default value for `projectName` I assume is the folder name? We might want to clarify that somewhere if we haven't yet.

dfunckt: I see your point. Technically, the direct equivalent for the CLI's `build` command is no longer Docker CLI (since balena is no longer single-container) but Docker Compose, which has no `--tag` attribute. Not sure how apparent is that to users though.

pdcastro: > Also, what is the default value for projectName I assume is the folder name? We might want to clarify that somewhere if we haven't yet.

The help output for the `--projectName` flag reads:

Before this PR:
```
-n, --projectName <projectName>  Specify an alternate project name; default is the directory name
```
After this PR (as it stands):
```
-n, --projectName <projectName>  Prefix for the name of built images. This is the 'projectName' portion in
                                 'projectName_serviceName:tag'. The default is the directory name.

```

pdcastro: > My only worry would be with user confusion regarding diverging behavior between balena build --tag and docker build --tag (or balena build on balena engine).

Even if we wanted to, the `docker build --tag` semantics cannot be used for the balena CLI, because `balena build` produces multiple images while `docker build` produces a single image. Like Akis pointed out, `balena build`'s counterpart is be closer to `docker-compose build` than `docker build`. We would have to come up with a different solution / semantic, or remove `--tag` altogether.

pdcastro: > While we're at it, if it's not much of a complication I think it would be useful if `--tag` could be passed several times, to make several tags for images the same build, eg. both versioned _and_ latest.

I gather that the clean/proper way of doing this would be to change the interface between `balena-multibuild` and `balena-cli` and `balena-builder`, specifically changing the following `task.tag` attribute from `string` to `string[]` on this line: https://github.com/balena-io-modules/balena-multibuild/blob/v4.11.1/lib/build-task.ts#L55-L59

This is because it is `balena-multibuild` that sets the `t` attribute in the docker engine API request (which apparently does allow specifying multiple `t` attributes). It would then become 3 PRs plus testing, touching the builder. It could be done, yes, but I will play the _"is it a priority?"_ get-out-of-jail-free card, and leave it for the future. :-)

dfunckt: > change the interface between balena-multibuild and balena-cli and balena-builder

Ah, good call -- I didn't realise this was the case, thanks for checking 👍

I was looking to reap low-hanging fruit, but it doesn't seem to be the case unfortunately.

pdcastro:
Force push: Added a test case for `balena build`.

lib/utils/compose_ts.ts
```diff
@@ -411,6 +417,18 @@ async function installQemuIfNeeded({
...
 }
 
+export function makeImageName(
+	projectName: string,
+	serviceName: string,
+	tag?: string,
+) {
+	let name = `${projectName}_${serviceName}`;
+	if (tag) {
+		name = [name, tag].map((s) => s.replace(/:/g, '_')).join(':');
```
pdcastro, line 427: I've added this bit of logic as I have learned that, in the past, [we advised users to add a colon to the project name](https://jel.ly.fish/support-thread-discourse-323654), e.g. `--projectName myproject:v1.1.1`, as a workaround for the fact that `--tag` didn't work, even though the service name was then appended to the tag, e.g. `myproject:v1.1.1_main`. This logic ensures a valid image name by replacing the colon with an underscore if `--tag` is used, thus preserving compatibility.

pdcastro:
@balena-ci I self-certify!

Issue #2330, opened 2021-09-09
push to local device: "ECONNRESET: socket hang up" and process exits with non-zero error code


pdcastro:
```
$ balena push 192.168.1.138
[debug] new argv=[/home/xxx/balena-cli/balena,/snapshot/versioned-source/bin/balena,push,192.168.1.138] length=4
[Info]    Starting build on device 192.168.1.138
[...]
[Build]   [main] Success!

[Info]    Streaming device logs...
[Live]    Watching for file changes...
[Live]    Waiting for device state to settle...
[Live]    Cleaning up device...
[Warn]    Connection to device lost
Retrying "Streaming logs" after 1.0s (1 of 1000) due to: DeviceConnectionLostError: Connection to device lost
ECONNRESET: socket hang up

Error: socket hang up
    at connResetException (internal/errors.js:609:14)
    at Socket.socketOnEnd (_http_client.js:453:23)
    at Socket.emit (events.js:327:22)
    at Socket.EventEmitter.emit (domain.js:482:12)
    at endReadableNT (_stream_readable.js:1221:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)

For further help or support, visit:
https://www.balena.io/docs/reference/balena-cli/#support-faq-and-troubleshooting
```

Based on additional Sentry.io breadcrumbs, I suspect the ECONNRESET comes from a supervisor API query: `GET //192.168.1.138:48484/v2/state/status` with 'undefined' HTTP response status. 

The "root cause" would likely be a supervisor failure. The could, however, retry supervisor API requests with a delay between subsequent requests.

PR #2331, opened 2021-09-09
push: Await and retry supervisor API requests to a local device

pdcastro:
Connects-to: #2330
Change-type: patch

### DIFF
--- a/lib/utils/device/api.ts
+++ b/lib/utils/device/api.ts
@@ -20,6 +20,7 @@ import * as os from 'os';
 import * as request from 'request';
 import type * as Stream from 'stream';
 
+import { retry } from '../helpers';
 import Logger = require('../logger');
 import * as ApiErrors from './errors';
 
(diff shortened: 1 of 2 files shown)


lib/utils/device/api.ts
```diff
@@ -256,24 +261,35 @@ export class DeviceAPI {
...
+					}
+				});
 			});
+		};
+
+		return await retry({
+			func: doRequest,
+			initialDelayMs: 2000,
+			maxAttempts: 6,
+			label: `Supervisor API (${opts.method} ${(opts as any).url})`,
```
pdcastro, line 292: @20k-ultra, @cywang117, this "livepush" CLI PR proposes to retry failed requests to the supervisor API. For GET requests it should be fine, but it could also include a POST request to `v2/local/target-state`. I don't see a problem with retrying such requests if they fail, do you? The motivation is the following paid support thread: https://jel.ly.fish/support-thread-1-0-0-front-cnv-b4ord71

lib/utils/device/api.ts
```diff
@@ -211,17 +212,21 @@ export class DeviceAPI {
...
+						// Certain versions of typescript won't convert
+						// this automatically
+						const sock = res.socket as any as NodeJSSocketWithFileDescriptor;
+						// We send a tcp keepalive probe once every 5 seconds
+						NetKeepalive.setKeepAliveInterval(sock, 5000);
+						// After 5 failed probes, the connection is marked as
+						// closed
+						NetKeepalive.setKeepAliveProbes(sock, 5);
+					}
+				} catch (error) {
```
pdcastro, line 228: Just adding a try-catch around this block because, during testing, I managed to get it to throw an unhandled exception.

lib/utils/device/api.ts
```diff
@@ -256,24 +261,35 @@ export class DeviceAPI {
...
-					case 400:
-						return reject(new ApiErrors.BadRequestDeviceAPIError(body.message));
-					case 503:
-						return reject(
-							new ApiErrors.ServiceUnavailableAPIError(body.message),
-						);
-					default:
-						return reject(new ApiErrors.DeviceAPIError(body.message));
-				}
+		const doRequest = async () => {
```
pdcastro, line 264: Just wrapping existing code in a `doRequest` function that is then retried a number of times.

lib/utils/device/logs.ts
```diff
@@ -80,16 +80,19 @@ async function displayDeviceLogs(
 			jsonStream.on('error', (e) => {
 				logger.logWarn(`Error parsing NDJSON log chunk: ${e}`);
 			});
-			logs.once('error', reject);
-			logs.once('end', () => {
+			logs.once('error', handleError);
+			logs.once('end', handleError);
```
pdcastro, line 84: This change shares a common error handling function for these `error` and `end` events, but I don't think it will make any difference to the case described in the support thread.

lib/utils/device/api.ts
cywang117: The POST retries to /v2/local/target-state should be fine too @pdcastro.

I can't tell from the code changes, but I saw that on sentry, the CLI logs the status code of each Supervisor API response. Would it also log retries and errors?

pdcastro: Each retry would _not_ be logged on Sentry, but when/if `maxAttempts` is reached, the error of the last retry would be logged on Sentry. Is this OK?

pdcastro:
> https://ci.balena-dev.com/builds/1375164
> Error creating or testing standalone zip package
> Error: "build:standalone": pkg output does not match expected output from "tests\test-data\pkg\expected-warnings-win32.txt"

This build error is coincidental and actually unrelated to this PR. I've created another PR to fix it: #2332
When that's sorted, I'll rebase this PR.

lib/utils/device/api.ts
cywang117: That sounds good to me!

pdcastro:
@balena-ci I self-certify!

PR #2332, opened 2021-09-09
chore: Update net-keepalive dependency (fix CLI packaging errors)

pdcastro:
The pkg binary (standalone zip package) suddenly started [failing our tests on Windows](https://ci.balena-dev.com/teams/main/pipelines/github-events-balena-io/jobs/node-cli/builds/714) because the `net-keepalive` dependency changed its availability for Windows (it's now installable and no-op, rather than uninstallable). This PR updates the pkg tests and also takes the opportunity to update the `net-keepalive` dependency.

Change-type: patch

### DIFF
--- a/lib/utils/device/api.ts
+++ b/lib/utils/device/api.ts
@@ -211,17 +211,21 @@ export class DeviceAPI {
 					);
 					return;
 				}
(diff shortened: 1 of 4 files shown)


pdcastro:
@balena-ci I self-certify!

PR #2334, opened 2021-09-13
os versions, os download: Add support for balenaOS ESR versions

pdcastro:
Resolves: #2005
Depends-on: balena-io-modules/balena-image-manager/pull/54
See: https://www.flowdock.com/app/rulemotion/resin-devices/threads/BtGUlpKl9fyhDGpI1OJ4W32EfAD
Change-type: minor

### DIFF
--- a/lib/commands/os/download.ts
+++ b/lib/commands/os/download.ts
@@ -34,21 +34,22 @@ export default class OsDownloadCmd extends Command {
 	public static description = stripIndent`
 		Download an unconfigured OS image.
 
-		Download an unconfigured OS image for a certain device type.
-		Check available types with \`balena devices supported\`
+		Download an unconfigured OS image for the specified device type.
+		Check available device types with 'balena devices supported'.
 
 		Note: Currently this command only works with balenaCloud, not openBalena.
 		If using openBalena, please download the OS from: https://www.balena.io/os/
 
-		If version is not specified the newest stable (non-pre-release) version of OS
-		is downloaded (if available), otherwise the newest version (if all existing
-		versions for the given device type are pre-release).
+		The '--version' option is used to select the balenaOS version. If omitted,
+		the latest released version is downloaded (and if only pre-release versions
+		exist, the latest pre-release version is downloaded).
 
-		You can pass \`--version menu\` to pick the OS version from the interactive menu
-		of all available versions.
+		Use '--version menu' or '--version menu-esr' to interactively select the
+		OS version. The latter lists ESR versions which are only available for
+		download on Production and Enterprise plans. See also:
+		https://www.balena.io/docs/reference/OS/extended-support-release/
 
-		To download a development image append \`.dev\` to the version or select from
-		the interactive menu.
+		Development images can be selected by appending \`.dev\` to the version.
 `;
 	public static examples = [
 		'$ balena os download raspberrypi3 -o ../foo/bar/raspberry-pi.img',
@@ -78,11 +79,13 @@ export default class OsDownloadCmd extends Command {
 		}),
 		version: flags.string({
 			description: stripIndent`
-				exact version number, or a valid semver range,
+				version number (ESR or non-ESR versions),
+				or semver range (non-ESR versions only),
 				or 'latest' (includes pre-releases),
-				or 'default' (excludes pre-releases if at least one stable version is available),
+				or 'default' (excludes pre-releases if at least one released version is available),
 				or 'recommended' (excludes pre-releases, will fail if only pre-release versions are available),
-				or 'menu' (will show the interactive menu)
+				or 'menu' (interactive menu, non-ESR versions),
+				or 'menu-esr' (interactive menu, ESR versions)
 				`,
 		}),
 		help: cf.help,
--- a/lib/commands/os/versions.ts
+++ b/lib/commands/os/versions.ts
@@ -18,9 +18,10 @@
 import { flags } from '@oclif/command';
 import Command from '../../command';
 import * as cf from '../../utils/common-flags';
-import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+import { stripIndent } from '../../utils/lazy';
 
 interface FlagsDef {
+	esr?: boolean;
 	help: void;
 }
 
@@ -34,6 +35,9 @@ export default class OsVersionsCmd extends Command {
 
 		Show the available balenaOS versions for the given device type.
 		Check available types with \`balena devices supported\`.
+
+		balenaOS ESR versions can be listed with the '--esr' option. See also:
+		https://www.balena.io/docs/reference/OS/extended-support-release/
 	`;
 
 	public static examples = ['$ balena os versions raspberrypi3'];
@@ -50,16 +54,20 @@ export default class OsVersionsCmd extends Command {
 
 	public static flags: flags.Input<FlagsDef> = {
 		help: cf.help,
+		esr: flags.boolean({
+			description: 'select balenaOS ESR versions',
+			default: false,
+		}),
 	};
 
 	public async run() {
-		const { args: params } = this.parse<FlagsDef, ArgsDef>(OsVersionsCmd);
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			OsVersionsCmd,
+		);
 
-		const { versions: vs, recommended } =
-			await getBalenaSdk().models.os.getSupportedVersions(params.type);
+		const { getFormattedOsVersions } = await import('../../utils/cloud');
+		const vs = await getFormattedOsVersions(params.type, !!options.esr);
 
-		vs.forEach((v) => {
-			console.log(`v${v}` + (v === recommended ? ' (recommended)' : ''));
-		});
+		console.log(vs.map((v) => v.formattedVersion).join('\n'));
 	}
 }
--- a/lib/utils/cloud.ts
+++ b/lib/utils/cloud.ts
@@ -137,8 +137,6 @@ export async function downloadOSImage(
 		? await resolveOSVersion(deviceType, OSVersion)
 		: 'default';
 
-	const displayVersion = OSVersion === 'default' ? '' : ` ${OSVersion}`;
-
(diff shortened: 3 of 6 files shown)


lib/utils/cloud.ts
```diff
@@ -150,10 +148,17 @@ export async function downloadOSImage(
...
 
+	const displayVersion = await new Promise((resolve, reject) => {
+		stream.on('error', reject);
+		stream.on('version', resolve);
+	});
+
 	const visuals = getVisuals();
-	const bar = new visuals.Progress(`Downloading Device OS${displayVersion}`);
+	const bar = new visuals.Progress(
+		`Downloading balenaOS version ${displayVersion}`,
```
pdcastro, line 158: Is it always correct to say `balenaOS` here, instead of `Device OS`?
Sample output after these changes:
```
Downloading balenaOS version 2.83.21+rev1.prod [========                ] 34% eta 30s
```
Whereas previously it might have read:
```
Downloading Device OS recommended [=                       ] 3% eta 1m17s
```

lib/utils/cloud.ts
```diff
@@ -150,10 +148,17 @@ export async function downloadOSImage(
 	const manager = await import('balena-image-manager');
 	const stream = await manager.get(deviceType, OSVersion);
 
+	const displayVersion = await new Promise((resolve, reject) => {
+		stream.on('error', reject);
+		stream.on('version', resolve);
```
pdcastro, line 153: This relies on a new `'version'` event introduced in balena-io-modules/balena-image-manager/pull/54

shawaj:
Any idea when this will be available?

We need to download and preload ESR images

pdcastro:
> Any idea when this will be available?
> We need to download and preload ESR images

@shawaj, this PR depends on balena-io-modules/balena-image-manager/pull/54 , which is waiting on a code review. A colleague said they'll try to review it tomorrow. Meanwhile, are you able to download the ESR image through the web dashboard, and then preload it?

lib/commands/os/versions.ts
```diff
@@ -55,11 +55,21 @@ export default class OsVersionsCmd extends Command {
 	public async run() {
 		const { args: params } = this.parse<FlagsDef, ArgsDef>(OsVersionsCmd);
 
-		const { versions: vs, recommended } =
-			await getBalenaSdk().models.os.getSupportedVersions(params.type);
+		const vs =
+			(await getBalenaSdk().models.hostapp.getAllOsVersions([params.type]))[
```
thgreasi, line 59: If you bump to the latest, you can instead use `getAvailableOsVersions()`, since after the next major, `getAllOsVersions()` is also going to include invalidated versions as well.

thgreasi (approved):
:+1:

lib/utils/cloud.ts
```diff
@@ -214,3 +215,29 @@ async function resolveOSVersion(deviceType: string, version: string) {
...
+			deviceType
+		] ?? []
+	)
+		.filter((v: SDK.OsVersion) => v.osType === (esr ? 'esr' : 'default'))
+		.map((v: SDK.OsVersion) => {
+			const i = v.formattedVersion.indexOf(' ');
+			v.formattedVersion =
+				i < 0
+					? `v${v.rawVersion}`
+					: `v${v.rawVersion}${v.formattedVersion.substring(i)}`;
```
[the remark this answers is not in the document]
pdcastro: Initially I thought of doing that, but then I realised that the `formattedVersion` field for ESR versions had "richer details" than just "recommended":
```
$ ./bin/balena os download --version menu-esr -o os.img raspberrypi3
Getting device operating system for raspberrypi3
? Select the OS version: (Use arrow keys)
❯ v2021.10.2.prod (next, recommended)
  v2021.10.2.dev (next)
  v2021.07.1.prod (current)
  v2021.07.1.dev (current)
  v2021.04.0.prod (sunset)
  v2021.04.0.dev (sunset)
  v2021.01.0.prod (outdated)
(Move up and down to reveal more choices)
```
Which I thought could good for the interactive menu. If we actually wanted to hide/delete information such as "next", "current" and "sunset", then yes we could just check the `isRecommended` attribute.

The formatting code is now also being shared between `os download` and `os versions`:

```
$ ./bin/balena os versions --esr raspberrypi3
v2021.10.2.prod (next, recommended)
v2021.10.2.dev (next)
v2021.07.1.prod (current)
v2021.07.1.dev (current)
v2021.04.0.prod (sunset)
v2021.04.0.dev (sunset)
v2021.01.0.prod (outdated)
v2021.01.0.dev (outdated)
v2020.07.2.prod (outdated)
v2020.07.2.dev (outdated)
v2020.04.0.prod (outdated)
v2020.04.0.dev (outdated)
v2019.10.2.prod (outdated)
v2019.10.2.dev (outdated)
```

lib/commands/os/versions.ts
```diff
@@ -50,16 +54,20 @@ export default class OsVersionsCmd extends Command {
 
 	public static flags: flags.Input<FlagsDef> = {
 		help: cf.help,
+		esr: flags.boolean({
+			description: 'select balenaOS ESR versions',
+			default: false,
+		}),
```
pdcastro, line 60: In my first go at this PR, I did not have a `--esr` flag and all versions were listed together. But some issues mentioned for `os download` also applied here:

* Two versions labeled "recommended", one ESR and one non-ESR. We could amend the formatting, but:
  * If we didn't print "recommended" at all, it would strictly be a breaking change (re output) for non-ESR versions.
  * If we printed "recommended" only for non-ESR versions, then... are we implying that ESR versions are not recommended? :-)
  * If we did some other special formatting thing for `os versions`, then the formatting code could not be shared with the interactive menu of `os download`.

* Re sorting, having ESR first and non ESR later or vice versa, there was still an element of the issue mentioned in a comment for the `os download` command: whatever we list at the top _with a recommended label_ is not necessarily the best (actually recommended) option for the user. With a `--esr` flag, the "recommendation" always fits the user's choice of ESR or non ESR.

* And finally, the web dashboard has a separate dropdown box to select ESR versions, avoiding listing them together. That solution has a 1-1 mapping to a `--esr` flag, which allows me to claim "platform UI/UX consistency" as well. :-)

lib/commands/os/versions.ts
pdcastro: Done 👍

lib/utils/cloud.ts
```diff
@@ -214,3 +215,29 @@ async function resolveOSVersion(deviceType: string, version: string) {
...
+ * regard to ESR or non-ESR versions, and having the `formattedVersion` field
+ * reformatted for compatibility with the pre-existing output format of the
+ * `os versions` and `os download` commands.
+ */
+export async function getFormattedOsVersions(
+	deviceType: string,
+	esr: boolean,
+): Promise<SDK.OsVersion[]> {
+	return (
+		(await getBalenaSdk().models.hostapp.getAllOsVersions([deviceType]))[
```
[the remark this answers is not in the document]
pdcastro: Done 👍  (A version bump wasn't needed, SDK version `15.51.1` includes `getAvailableOsVersions`)

pdcastro:
@balena-ci I self-certify!

pdcastro:
@thgreasi, only after this PR was merged, I noticed that `sdk.models.hostapp.getAvailableOsVersions()` does not accept device type aliases like `'nuc'`, whereas `sdk.models.os.getSupportedVersions()` accepts aliases. As a result, the `balena os versions` command no longer accepts aliases. I seem to remember a previous conversation regarding aliases support for the hostapp model -- is this a thing? What would you suggest we did for `os versions` and `os download` to support device type aliases?

lib/commands/os/versions.ts
```diff
@@ -50,16 +54,20 @@ export default class OsVersionsCmd extends Command {
...
+		);
 
-		const { versions: vs, recommended } =
-			await getBalenaSdk().models.os.getSupportedVersions(params.type);
+		const { getFormattedOsVersions } = await import('../../utils/cloud');
+		const vs = await getFormattedOsVersions(params.type, !!options.esr);
 
-		vs.forEach((v) => {
-			console.log(`v${v}` + (v === recommended ? ' (recommended)' : ''));
-		});
```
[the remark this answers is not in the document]
thgreasi: @pdcastro it probably surfaces an issue with the original hostApp release migration/import. The SDK & CLI are doing the correct thing. Given how old those releases are though and our plan to deprecate such old releases, I wouldn't assign time on it unless a users asks about it.

thgreasi:
I'm indeed working on adding the aliases to the model, but I don't have an estimate & I'm also about to release a major of the SDK.
A workaround for the regression is to do `await sdk.models.device.getManifestBySlug('slugOrAlias').slug`.

lib/commands/os/download.ts
```diff
@@ -78,11 +79,13 @@ export default class OsDownloadCmd extends Command {
...
-				exact version number, or a valid semver range,
+				version number (ESR or non-ESR versions),
+				or semver range (non-ESR versions only),
 				or 'latest' (includes pre-releases),
-				or 'default' (excludes pre-releases if at least one stable version is available),
+				or 'default' (excludes pre-releases if at least one released version is available),
 				or 'recommended' (excludes pre-releases, will fail if only pre-release versions are available),
-				or 'menu' (will show the interactive menu)
+				or 'menu' (interactive menu, non-ESR versions),
+				or 'menu-esr' (interactive menu, ESR versions)
```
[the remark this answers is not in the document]
dfunckt: > If users typed any of the combinations below, surely they'd expect the "latest", "recommended" or "default" qualifiers to apply to ESR versions rather than non-ESR versions, even if the help output said that the --esr flag could only be used with --version menu, and even if the CLI printed an error if --esr was used with anything other than --version menu. As it happens, those qualifiers only apply to non-ESR versions as currently implemented in the SDK or API.
>
> - `--esr --version latest`
> - `--esr --version recommended`
> - `--esr --version default`

For all modifiers the version for ESR is the same @pdcastro so there is no conflict really and the `download` command could go with the `--esr` argument as well.

PR #2345, opened 2021-09-18
Add support for releases

toochevere:
To adding release handling to the cli

Tasks

- [x] Main code
- [x] Manuel Testing
- [x] Automated tests

Change-type: minor
See: https://www.flowdock.com/app/rulemotion/r-supervisor/threads/JN2-gnspQ-v6WaeWCvm9T8NYDY1
Signed-off-by: Paul Jonathan <pj@balena.io>

### DIFF
--- a/lib/commands/release/finalize.ts
+++ b/lib/commands/release/finalize.ts
@@ -1,0 +1,86 @@
+/**
+ * @license
+ * Copyright 2016-2020 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { flags } from '@oclif/command';
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+import { tryAsInteger } from '../../utils/validation';
+
+interface FlagsDef {
+	help: void;
+}
+
+interface ArgsDef {
+	commitOrId: string | number;
+}
+
+export default class ReleaseFinalizeCmd extends Command {
+	public static description = stripIndent`
+		Finalize a release.
+
+		Finalize a release. Releases can be "draft" or "final", and this command
+		changes a draft release into a final release. Draft releases can be created
+		with the \`--draft\` option of the \`balena build\` or \`balena deploy\`
+		commands.
+
+		Draft releases are not automatically deployed to devices tracking the latest application
+		release. For a draft release to be deployed to a device, the device should be
+		explicity pinned to that release. Conversely, final releases may trigger immediate
+		deployment to unpinned devices (subject to a device's  polling period) and, for
+		this reason, final releases cannot be changed back to draft status.
+`;
+	public static examples = [
+		'$ balena release finalize a777f7345fe3d655c1c981aa642e5555',
+		'$ balena release finalize 1234567',
+	];
+
+	public static usage = 'finalize <commitOrId>';
+
+	public static flags: flags.Input<FlagsDef> = {
+		help: cf.help,
+	};
+
+	public static args = [
+		{
+			name: 'commitOrId',
+			description: 'the commit or ID of the release to finalize',
+			required: true,
+			parse: (commitOrId: string) => tryAsInteger(commitOrId),
+		},
+	];
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(ReleaseFinalizeCmd);
+
+		const balena = getBalenaSdk();
+
+		const release = await balena.models.release.get(params.commitOrId, {
+			$select: ['id', 'is_final'],
+		});
+
+		if (release.is_final) {
+			console.log(`Release ${params.commitOrId} is already finalized!`);
+			return;
+		}
+
+		await balena.models.release.finalize(release.id);
+		console.log(`Release ${params.commitOrId} finalized`);
+	}
+}
--- a/lib/commands/release/index.ts
+++ b/lib/commands/release/index.ts
@@ -1,0 +1,128 @@
+/**
+ * @license
+ * Copyright 2016-2020 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { flags } from '@oclif/command';
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, getVisuals, stripIndent } from '../../utils/lazy';
+import type * as BalenaSdk from 'balena-sdk';
+import jsyaml = require('js-yaml');
+import { tryAsInteger } from '../../utils/validation';
+
+interface FlagsDef {
+	help: void;
+	composition?: boolean;
+}
+
+interface ArgsDef {
+	commitOrId: string | number;
+}
+
+export default class ReleaseCmd extends Command {
+	public static description = stripIndent`
+		Get info for a release.
+`;
+	public static examples = [
+		'$ balena release a777f7345fe3d655c1c981aa642e5555',
+		'$ balena release 1234567',
+	];
+
+	public static usage = 'release <commitOrId>';
+
+	public static flags: flags.Input<FlagsDef> = {
+		help: cf.help,
+		composition: flags.boolean({
+			default: false,
+			char: 'c',
+			description: 'Return the release composition',
+		}),
+	};
+
+	public static args = [
+		{
+			name: 'commitOrId',
+			description: 'the commit or ID of the release to get information',
+			required: true,
+			parse: (commitOrId: string) => tryAsInteger(commitOrId),
+		},
+	];
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
+			ReleaseCmd,
+		);
+
+		const balena = getBalenaSdk();
+		if (options.composition) {
+			await this.showComposition(params.commitOrId, balena);
+		} else {
+			await this.showReleaseInfo(params.commitOrId, balena);
+		}
+	}
+
+	async showComposition(
+		commitOrId: string | number,
+		balena: BalenaSdk.BalenaSDK,
+	) {
+		const release = await balena.models.release.get(commitOrId, {
+			$select: 'composition',
+		});
+
+		console.log(jsyaml.dump(release.composition));
+	}
+
+	async showReleaseInfo(
+		commitOrId: string | number,
+		balena: BalenaSdk.BalenaSDK,
+	) {
+		const fields: Array<keyof BalenaSdk.Release> = [
+			'id',
+			'commit',
+			'created_at',
+			'status',
+			'semver',
+			'is_final',
+			'build_log',
+			'start_timestamp',
+			'end_timestamp',
+		];
+
+		const release = await balena.models.release.get(commitOrId, {
+			$select: fields,
+			$expand: {
+				release_tag: {
+					$select: ['tag_key', 'value'],
+				},
+			},
+		});
+
+		const tagStr = release
+			.release_tag!.map((t) => `${t.tag_key}=${t.value}`)
+			.join('\n');
+
+		const _ = await import('lodash');
+		const values = _.mapValues(
+			release,
+			(val) => val ?? 'N/a',
+		) as Dictionary<string>;
+		values['tags'] = tagStr;
+
+		console.log(getVisuals().table.vertical(values, [...fields, 'tags']));
+	}
+}
--- a/lib/commands/releases.ts
+++ b/lib/commands/releases.ts
@@ -1,0 +1,86 @@
+/**
+ * @license
+ * Copyright 2016-2020 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { flags } from '@oclif/command';
+import Command from '../command';
+import * as cf from '../utils/common-flags';
+import { getBalenaSdk, getVisuals, stripIndent } from '../utils/lazy';
+import { applicationNameNote } from '../utils/messages';
+import type * as BalenaSdk from 'balena-sdk';
+
+interface FlagsDef {
+	help: void;
+}
+
+interface ArgsDef {
+	fleet: string;
+}
+
+export default class ReleasesCmd extends Command {
+	public static description = stripIndent`
+		List all releases of a fleet.
+
+		List all releases of the given fleet.
+
+		${applicationNameNote.split('\n').join('\n\t\t')}
+`;
+	public static examples = ['$ balena releases myorg/myfleet'];
+
+	public static usage = 'releases <fleet>';
+
+	public static flags: flags.Input<FlagsDef> = {
+		help: cf.help,
+	};
+
+	public static args = [
+		{
+			name: 'fleet',
+			description: 'fleet name or slug',
+			required: true,
+		},
+	];
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(ReleasesCmd);
+
+		const fields: Array<keyof BalenaSdk.Release> = [
+			'id',
+			'commit',
+			'created_at',
+			'status',
+			'semver',
+			'is_final',
+		];
+
+		const balena = getBalenaSdk();
+
+		const releases = await balena.models.release.getAllByApplication(
+			params.fleet,
+			{ $select: fields },
+		);
+
+		const _ = await import('lodash');
+		console.log(
+			getVisuals().table.horizontal(
+				releases.map((rel) => _.mapValues(rel, (val) => val ?? 'N/a')),
+				fields,
+			),
+		);
+	}
+}
--- a/completion/_balena
+++ b/completion/_balena
@@ -8,7 +8,7 @@ _balena() {
   local context state line curcontext="$curcontext"
 
   # Valid top-level completions
-  main_commands=( apps build deploy envs fleets join keys leave login logout logs note orgs preload push scan settings ssh support tags tunnel version whoami api-key app app config device device devices env fleet fleet internal key key local os tag util )
+  main_commands=( apps build deploy envs fleets join keys leave login logout logs note orgs preload push releases scan settings ssh support tags tunnel version whoami api-key app app config device device devices env fleet fleet internal key key local os release release tag util )
   # Sub-completions
   api_key_cmds=( generate )
   app_cmds=( create purge rename restart rm )
@@ -21,6 +21,7 @@ _balena() {
   key_cmds=( add rm )
   local_cmds=( configure flash )
   os_cmds=( build-config configure download initialize versions )
+  release_cmds=( finalize )
   tag_cmds=( rm set )
 
 
@@ -72,6 +73,9 @@ _balena_sec_cmds() {
       ;;
       "os")
         _describe -t os_cmds 'os_cmd' os_cmds "$@" && ret=0
+      ;;
+      "release")
+        _describe -t release_cmds 'release_cmd' release_cmds "$@" && ret=0
       ;;
       "tag")
         _describe -t tag_cmds 'tag_cmd' tag_cmds "$@" && ret=0
--- a/completion/balena-completion.bash
+++ b/completion/balena-completion.bash
@@ -7,7 +7,7 @@ _balena_complete()
   local cur prev
 
   # Valid top-level completions
-  main_commands="apps build deploy envs fleets join keys leave login logout logs note orgs preload push scan settings ssh support tags tunnel version whoami api-key app app config device device devices env fleet fleet internal key key local os tag util"
+  main_commands="apps build deploy envs fleets join keys leave login logout logs note orgs preload push releases scan settings ssh support tags tunnel version whoami api-key app app config device device devices env fleet fleet internal key key local os release release tag util"
   # Sub-completions
   api_key_cmds="generate"
   app_cmds="create purge rename restart rm"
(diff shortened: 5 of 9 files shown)


lib/commands/release/index.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+import { flags } from '@oclif/command';
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, getVisuals, stripIndent } from '../../utils/lazy';
+
+interface FlagsDef {
+	help: void;
+	finalize?: boolean;
```
toochevere, line 25: `finalize` is superfluous and will be removed.

lib/commands/releases.ts
```diff
@@ -0,0 +1,79 @@
...
+		if (!params.fleet) {
+			const { ExpectedError } = await import('../errors');
+			throw new ExpectedError('You must specify a fleet');
+		}
+
+		await this.listReleases(params);
+	}
+
+	protected async listReleases(params: ArgsDef) {
+		const fields = ['commit', 'created_at', 'status', 'semver', 'is_final'];
```
thgreasi, line 63: Let's also include the id, as a way to disambiguate releases with the same commit.
```suggestion
		const fields = ['id', 'commit', 'created_at', 'status', 'semver', 'is_final'];
```

lib/commands/release/index.ts
```diff
@@ -0,0 +1,108 @@
...
+			ReleaseCmd,
+		);
+		if (!params.commit) {
+			const { ExpectedError } = await import('../../errors');
+			throw new ExpectedError('You must specify a release commit');
+		}
+
+		const balena = getBalenaSdk();
+
+		const release = await balena.models.release.get(params.commit!);
```
thgreasi, line 70: In the latest TS you shouldn't need the `!` in the `params.commit!` :thinking:
You can probably get rid of it now as well by just destructuring it (`const { commit } = params;`) before checking whether it has a value and then using the `commit` directly everywhere else.

Let's also only fetch the fields that we are interested at:
```suggestion
		const release = await balena.models.release.get(params.commit!, { $select: fields }); // the fields need to be defined top level or in a static property for this to work
```
PS: I guess you will need to switch the selected fields, depending on whether the user provided that flag or not.

lib/commands/release/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+interface ArgsDef {
+	commit?: string;
+}
+
+export default class ReleaseCmd extends Command {
+	public static description = stripIndent`
+		Get info for a release.
+`;
+	public static examples = ['$ balena release commit'];
```
thgreasi, line 37:
```suggestion
	public static examples = ['$ balena release 20f0640'];
```
Right?

lib/commands/release/index.ts
```diff
@@ -0,0 +1,108 @@
...
+			default: false,
+			char: 'c',
+			description: 'Return the release composition',
+		}),
+	};
+
+	public static args = [
+		{
+			name: 'commit',
+			description: 'the commit of the release to get information',
```
thgreasi, line 53: Maybe we could also document using release IDs, since commits are not even unique in the same fleet.
(The commit is unique only across successful releases of a given fleet, hence you can get an AmbiguousReleaseError in case i) your first `git push` failed, or you are `git push`ing the same commit to more than one fleets.)

The same SDK method already supports using id or commit.
You can check how it's done for devices here:
See: https://github.com/balena-io/balena-cli/blob/88fc3f77146144eb56b906aa413ef2f8e6b84c95/lib/commands/device/index.ts#L68

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,74 @@
...
+	public static usage = 'finalize <commit>';
+
+	public static flags: flags.Input<FlagsDef> = {
+		help: cf.help,
+	};
+
+	public static args = [
+		{
+			name: 'commit',
+			description: 'the commit of the release to get information',
```
thgreasi, line 46: Like in the `balena release`, I think we need to also add support for using the release id here as well.

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,74 @@
...
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(ReleaseFinalizeCmd);
+		if (!params.commit) {
+			const { ExpectedError } = await import('../../errors');
+			throw new ExpectedError('You must specify a release commit');
+		}
+
+		const balena = getBalenaSdk();
+
+		const release = await balena.models.release.get(params.commit!);
```
thgreasi, line 61:
```suggestion
		const release = await balena.models.release.get(params.commit);
```

Like below

lib/commands/release/index.ts
```diff
@@ -0,0 +1,108 @@
...
+			throw new ExpectedError('You must specify a release commit');
+		}
+
+		const balena = getBalenaSdk();
+
+		const release = await balena.models.release.get(params.commit!);
+		if (!release) {
+			const { ExpectedError } = await import('../../errors');
+			throw new ExpectedError(`Release ${params.commit} not found!`);
+		}
```
thgreasi, line 74: This will probably never run, since the SDK throws `BalenaReleaseNotFound` on its own when a release is not found.

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,74 @@
...
+		if (!release) {
+			const { ExpectedError } = await import('../../errors');
+			throw new ExpectedError(`Release ${params.commit} not found!`);
+		}
```
thgreasi, line 65: > This will probably never run, since the SDK throws `BalenaReleaseNotFound` on its own when a release is not found.

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		if (release.is_final) {
+			const { ExpectedError } = await import('../../errors');
+			throw new ExpectedError(`Release ${params.commit} is not draft!`);
```
thgreasi, line 69: How about v?
```suggestion
			throw new ExpectedError(`Release ${params.commit} is already finalized!`);
```

lib/commands/release/index.ts
pipex: Does this mean that you can have a failed push and a successful push with same commit? I thought pushing a release from the same git commit would just update the release, still making it unique.

I'm not sure about useful is adding support for release ids here, as the only place the release id can be obtained is from the builder logs and from the API

lib/commands/release/finalize.ts
pipex: What is the advantage of throwing an error in this case? Is there a way to terminate early without throwing an error, saying ``Release ${params.commit} is already final. Nothing to do here`

thgreasi: You have a point that in case you are trying to finalize something that's already finalized, an info log might just be enough instead of throwing, b/c your command's target is already in the desired state :+1:

lib/commands/release/index.ts
thgreasi: I'm not sure whether that has changed recently, but for sure when I joined each git push created a new release and the rule that allows that hasn't been changed to anything stricter yet:
See: Rule: It is necessary that each application that owns a release1 that has a status that is equal to "success" and has a commit1, owns at most one release2 that has a status that is equal to "success" and has a commit2 that is equal to the commit1.

lib/commands/releases.ts
toochevere: I didn't realize that releases could actually have the same commit 🙄

lib/commands/release/index.ts
toochevere: Will remove

lib/utils/release-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type * as BalenaSdk from 'balena-sdk';
+
+export async function getRelease(
```
toochevere, line 20: Note:  `balena.models.release.get()` isn't smart enough if you pass a number as a string (`id`) to do the right thing.  In such a case, you always get "not found".  So we have to try to see if we can parse it as a number and if we can, attempt first to get it as an id.  In most cases, the `intParse` will fail, so we won't actually get multiple attempts.

lib/commands/release/index.ts
```diff
@@ -0,0 +1,120 @@
...
+		const release = await getRelease(commitOrId, fieldsForGet, balena);
+
+		const tagFields: Array<keyof BalenaSdk.ResourceTagBase> = [
+			'tag_key',
+			'value',
+		];
+
+		const tags = await balena.models.release.tags.getAllByRelease(release.id, {
+			$select: tagFields,
+		});
```
thgreasi, line 106: Pine allows you to combine those two requests into a single one, so that we are kinder with our API:
```js
await balena.models.release.get(commitOrId, {
	$select: fieldsForGet,
	$expand: {
		release_tag: {
			$select: tagFields,
		}
	}
});
```
One extra reason for using `tryAsInteger` or `tryIdOrStringKeyGet` since they will make it easier to do so.

lib/commands/release/index.ts
```diff
@@ -0,0 +1,120 @@
...
+		const tagStr = tags.map((t) => `${t.tag_key}=${t.value}`).join('\n');
+
+		const _ = await import('lodash');
+		const values = _.mapValues(
+			release,
+			(val) => val ?? 'N/a',
+		) as Dictionary<string>;
+		values['tags'] = tagStr;
+		const fieldsForDisplay = fieldsForGet.map((f) => f) as string[];
+		fieldsForDisplay.push('tags');
```
thgreasi, line 116:
```suggestion
		const fieldsForDisplay = [
			...fieldsForGet,
			'tags',
		];
```

lib/commands/release/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+interface ArgsDef {
+	commitOrId?: string;
+}
+
+export default class ReleaseCmd extends Command {
+	public static description = stripIndent`
+		Get info for a release.
+`;
+	public static examples = ['$ balena release commit', '$ balena release id'];
```
thgreasi, line 39: The `balena device` example though include valid example values
See https://github.com/balena-io/balena-cli/blob/88fc3f77146144eb56b906aa413ef2f8e6b84c95/lib/commands/device/index.ts#L62

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,79 @@
...
+interface ArgsDef {
+	commitOrId?: string;
+}
+
+export default class ReleaseFinalizeCmd extends Command {
+	public static description = stripIndent`
+		Finalize a release.
+`;
+	public static examples = [
+		'$ balena release finalize commit',
```
thgreasi, line 37: > The `balena device` example though include valid example values
See https://github.com/balena-io/balena-cli/blob/88fc3f77146144eb56b906aa413ef2f8e6b84c95/lib/commands/device/index.ts#L62

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,79 @@
...
+	public static usage = 'finalize <commitOrId>';
+
+	public static flags: flags.Input<FlagsDef> = {
+		help: cf.help,
+	};
+
+	public static args = [
+		{
+			name: 'commitOrId',
+			description: 'the commit or ID of the release to get information',
```
thgreasi, line 50:
```suggestion
			description: 'the commit or ID of the release to finalize',
```

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,79 @@
...
+			['id', 'is_final'],
+			balena,
+		);
+
+		if (release.is_final) {
+			console.log(`Release ${params.commitOrId} is already finalized!`);
+			return;
+		}
+
+		balena.models.release.finalize(release.id);
```
thgreasi, line 76:
```suggestion
		await balena.models.release.finalize(release.id);
```

lib/commands/release/index.ts
toochevere: Perfect.  I was wondering how to do this better.

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,79 @@
...
+export default class ReleaseFinalizeCmd extends Command {
+	public static description = stripIndent`
+		Finalize a release.
+`;
+	public static examples = [
+		'$ balena release finalize commit',
+		'$ balena release finalize id',
+	];
+
+	public static usage = 'finalize <commitOrId>';
```
pdcastro, line 41: Re `commitOrId`, what is a release ID? Is it the numeric database ID, like 1234567 ? I am considering #1843, although I think the web dashboard exposes the numeric database ID and maybe even the output of `balena push` exposes it at end of the build (?). I see pros and cons in accepting a database ID here, but a product decision was made in past to stop exposing and using numeric database IDs.

I understand that a "release commit" had its origins in git commits and `git push` to balenaCloud, when the release commit was actually exactly the same value as the git commit. `balena push` disregards the git commit, and a release UUID is then created and named a commit, for consistency and backwards compatibility. This explanation of mine is in part informed guessing, just to get us thinking about it. :-)

pdcastro: Ah, only now I noticed outdated / hidden review comments, where @thgreasi had requested numeric database IDs to be added despite #1843, almost the opposite of my review comment:

> Maybe we could also document using release IDs, since commits are not even unique in the same fleet.
> (The commit is unique only across successful releases of a given fleet, hence you can get an AmbiguousReleaseError in case i) your first `git push` failed, or you are `git push`ing the same commit to more than one fleets.)

Non uniqueness of release commits is indeed a strong argument for accepting numeric database IDs, and make for an uneasy, incorrect "semantic mapping" between commit hashes and UUIDs, in the sense that UUIDs are properly unique.

To the extent that commit hashes are not unique while past product call decisions were made to avoid exposing numeric database IDs, in theory one could think of a feature request where the API data model should add "proper" release UUIDs in addition to release commits and numeric database IDs. How messy! 😅

lib/commands/releases.ts
```diff
@@ -0,0 +1,84 @@
...
+	];
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(ReleasesCmd);
+		if (!params.fleet) {
+			const { ExpectedError } = await import('../errors');
+			throw new ExpectedError('You must specify a fleet');
+		}
```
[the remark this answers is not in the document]
pdcastro: Another consequence of `required: true` is that it changes the help output from:
```
ARGUMENTS
  [fleet]  fleet name or slug
```
to:
```
ARGUMENTS
  <fleet>  fleet name or slug
```

lib/commands/releases.ts
```diff
@@ -0,0 +1,84 @@
...
+import * as cf from '../utils/common-flags';
+import { getBalenaSdk, getVisuals, stripIndent } from '../utils/lazy';
+import type * as BalenaSdk from 'balena-sdk';
+
+interface FlagsDef {
+	help: void;
+}
+
+interface ArgsDef {
+	fleet?: string;
```
pdcastro, line 29:
```suggestion
	fleet: string;
```
Because the fleet is a required parameter.

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,79 @@
...
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+import { getRelease } from '../../utils/release-utils';
+
+interface FlagsDef {
+	help: void;
+}
+
+interface ArgsDef {
+	commitOrId?: string;
```
pdcastro, line 29:
```suggestion
	commitOrId: string;
```
Because it is a required parameter. I think this goes together with `required: true` in the parameter definition.

lib/commands/release/index.ts
```diff
@@ -0,0 +1,120 @@
...
+import jsyaml = require('js-yaml');
+import { getRelease } from '../../utils/release-utils';
+
+interface FlagsDef {
+	help: void;
+	composition?: boolean;
+}
+
+interface ArgsDef {
+	commitOrId?: string;
```
pdcastro, line 32:
```suggestion
	commitOrId: string;
```

lib/commands/releases.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export default class ReleasesCmd extends Command {
+	public static description = stripIndent`
+		List all releases of a fleet.
+
+		List all releases of the given fleet.
+
+		${applicationNameNote.split('\n').join('\n\t\t')}
```
pdcastro, line 39: I am contributing a commit that expands some descriptions. Here's how the help output looks like:
```
$ ./bin/balena help release finalize
Finalize a release.

USAGE
  $ balena finalize <commitOrId>

ARGUMENTS
  [commitOrId]  the commit or ID of the release to get information

OPTIONS
  -h, --help  show CLI help

DESCRIPTION
  Finalize a release. Releases can be "draft" or "final", and this command
  changes a draft release into a final release. Draft releases can be created
  with the `--draft` option of the `balena build` or `balena deploy`
  commands.

  Draft releases are not automatically deployed to "unpinned" devices, i.e.
  devices that track the latest available release. For a draft release to
  be deployed to a device, the release should be explicity pinned to that
  device. Conversely, final releases may trigger immediate deployment to
  unpinned devices (subject to a device's  polling period) and, for this
  reason, final releases cannot be changed back to draft status.

EXAMPLES
  $ balena release finalize a777f7345fe3d655c1c981aa642e5555
  $ balena release finalize 1234567
```

lib/commands/releases.ts
```diff
@@ -0,0 +1,84 @@
...
+			'created_at',
+			'status',
+			'semver',
+			'is_final',
+		];
+
+		const balena = getBalenaSdk();
+
+		const releases = await balena.models.release.getAllByApplication(
+			params.fleet!,
```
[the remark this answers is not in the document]
thgreasi: @pdcastro open balena app slugs work in the same exact way as in the cloud for some time now :)
See: https://github.com/balena-io/open-balena-api/blob/c728541da6087a92cde3801bfb5eff4820d87354/src/features/applications/hooks/slug.ts#L21-L24

lib/commands/release/finalize.ts
thgreasi: Given #1843 and how `balena push` is our suggested way of creating new releases, I think I'm fine supporting only commits (if @pdcastro insists) and then bringing up a selector for the user to select the target release. We could probably also filter out failed releases, since it doesn't offer anything to finalize them, but that would still be an issue for the `balena release` command.

pipex: I think filtering out failed releases is the correct approach for `balena release finalize` since, as I understood, the commit is unique for final releases right? Other alternative that could work for this and `balena release` is to sort by status (with successful releases first) and return just the first result. My guess is that in most cases that would return what the user wants.

pdcastro: I don't insist! 😁  Relying on human interaction (bringing up a selector for the user to select the target release) is an even more serious issue than exposing numeric database IDs in my opinion (I have only ever seen 1 user complaint about numeric database IDs as they thought it was somehow a security issue, while I have seen many complaints on difficulties to programmatically use the CLI because of colours, unicorns, non-JSON output, interactive menus, etc, e.g. #1089, for CLI use in scripts and/or CI environments).

The proper solution might actually be to add release UUIDs to the data model in addition to release commits, but that's outside the scope of this PR. On balance, I think in this case we are better off allowing numeric database IDs, as implemented.

lib/commands/releases.ts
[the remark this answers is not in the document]
pipex: Funny that the exclamation mark is required. I would expect the `if` a few lines up to act as a type guard. I guess it won't be needed if the fleet argument is set to required.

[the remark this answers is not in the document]
pdcastro: > Funny that the exclamation mark is required

Oops, just noticed that my comment read "does appear to be necessary" when I meant to write "does **not** appear to be necessary". I've edited it. `npm run build` was successful on my laptop without the exclamation mark.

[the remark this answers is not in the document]
toochevere: Thanks @pdcastro.  Not in reference to `fleet`, but rather in reference to `release`, database ID's were specifically added to potentially disambiguate a release.  FYI.

lib/commands/release/finalize.ts
toochevere: > what is a release ID? Is it the numeric database ID, like 1234567 ?

Correct.  These were specifically requested by @thgreasi to disambiguate releases.

lib/commands/releases.ts
toochevere: wow, really good.

lib/commands/release/finalize.ts
toochevere: I am hesitant about this.  This would cause a different list from the CLI that the dashboard shows.  I think we need to show them all, or else add a flag `--all`.

pdcastro: > I think filtering out failed releases is the correct approach

I suspect Felipe meant filtering out failed releases if the commit given as argument to `balena release finalize <commit>` was ambiguous. Say there were 3 releases with commit `a777f7345fe3d655c1c981aa642e5555`, which I think (not sure) you could try and reproduce by using the `git push` deployment method (rather than `balena push`) with a Dockerfile containing a RUN line that executes a script that fails (non zero exit code) if it is executed before midday, and succeeds after midday (like me, I work better in the afternoon). If you `git push` that git commit twice before midday, you get two failed releases. Then `git push` the same commit again after midday, and the release succeeds. Then try `balena release finalize <commit>` (by commit, not database ID). Presumably the SDK will throw some "Release is ambiguous" error. In this case, Thodoris suggest using an interactive menu for the user to select the right release to finalize, and Felipe reasoned: there can be many failed releases matching a given commit, but only one successful release with that same commit (enforced by the API schema), so we should filter out the failed releases and finalize the successful one, avoiding the need for an interactive menu and avoiding the need for user to use a database ID to disambiguate between releases that share the same commit.

 it was in reply to Thodoris' suggestion of an interactive menu for the user to disambiguate the commit in case they ran `balena release finalize a777f7345fe3d655c1c981aa642e5555` and there were multiple releases

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,89 @@
...
+	public static description = stripIndent`
+		Finalize a release.
+
+		Finalize a release. Releases can be "draft" or "final", and this command
+		changes a draft release into a final release. Draft releases can be created
+		with the \`--draft\` option of the \`balena build\` or \`balena deploy\`
+		commands.
+
+		Draft releases are not automatically deployed to "unpinned" devices, i.e.
+		devices that track the latest available release. For a draft release to
```
thgreasi, line 42:
```suggestion
		Draft releases are not automatically deployed to devices tracking
		the latest application release. For a draft release to
```

lib/commands/release/finalize.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params } = this.parse<FlagsDef, ArgsDef>(ReleaseFinalizeCmd);
+
+		const balena = getBalenaSdk();
+
+		const release = await balena.models.release.get(
+			tryAsInteger(params.commitOrId),
```
thgreasi, line 75: I think this should be as a `parse` prop in the definition of the `commitOrId` arg, like in the device command.
See: https://github.com/balena-io/balena-cli/blob/e098cdca171abf9b1df2534c59f39276833b84ad/lib/commands/device/index.ts#L68

lib/commands/release/index.ts
```diff
@@ -0,0 +1,121 @@
...
+			'created_at',
+			'status',
+			'semver',
+			'is_final',
+			'build_log',
+			'start_timestamp',
+			'end_timestamp',
+		];
+
+		const release = await balena.models.release.get(tryAsInteger(commitOrId), {
```
thgreasi, line 99: Can we convert this to a `parse` arg option?
See: https://github.com/balena-io/balena-cli/blob/e098cdca171abf9b1df2534c59f39276833b84ad/lib/commands/device/index.ts#L68


In that case make sure to change all `commitOrId` params to `string | number`.

lib/commands/release/finalize.ts
pipex: Sorry I had not seen Paulo's response. But what he describes is exactly what I meant, however much better explained :p

pipex: From discussion with Thodoris and PJ, it turns out disambiguation for `finalize` is not really necessary since releases created with the `git` flow are always final. However for consistency with the `release` info (and for the sake of time), we agreed  that keeping the `commitOrId` input of both is fine for now. 

We will need to discuss how to make release `commit` an unambiguous identier on the API at some point in the future.

thgreasi (approved):

Issue #2152, opened 2021-01-12
balena os download hangs on 83%, errors out later in a container


vipulgupta2048:
I am running balenaCLI v12.37.0 (Installed globally through NPM) inside a container with [fincm3-alpine-node](https://hub.docker.com/r/balenalib/fincm3-alpine-node) as the base image. On our testing runs, we see the `balena os download` command flaking out about 8 out of 10 times. 

# Expected Behavior

The image requested should download without errors.

# Actual Behavior

On running the command inside the container, `balena os download raspberrypi3 -o /tmp/raspberrypi3.img --version v2.58.3+rev1.prod` the download proceeds normally and gets stuck strangely at 83% (Every time). The ETA time does keep increasing after eventually failing with the error mentioned below. 

```
bash-5.0# balena os download raspberrypi3 -o /tmp/raspberrypi3.img --version v2.58.3+rev1.prod
Getting device operating system for raspberrypi3
Downloading Device OS 2.58.3+rev1.prod [========================] 100% eta 5s  
Error: unexpected end of file
    at Zlib.zlibOnError [as onerror] (zlib.js:180:17)
```

We did some troubleshooting of our own. Tried different connections, cli versions (previously v12.32), and base images (previously it was debian) `balena os download` works perfectly on x86 (my system, popOS 18.04) but not sure why it is having troubles inside a container. Anything you suggest, I be most happy to try out. We have tried using `--debug` flag as well but no change in output.

# Steps to Reproduce the Problem

This is the most important and helpful part of a bug report. If we cannot reproduce the
problem, it is difficult to tell what the fix should be, or whether code changes have
fixed it.

1. [Dockerfile.template](https://github.com/balena-io-playground/socks-proxy-tests/blob/master/exclave-control-center/Dockerfile.template) and [docker-compose.yml](https://github.com/balena-io-playground/socks-proxy-tests/blob/master/docker-compose.yml) here for the container  
2. Deployed on a balenaFin v1.1.1
3. Run the command on the container, `balena os download raspberrypi3 -o /tmp/raspberrypi3.img --version v2.58.3+rev1.prod`

# Specifications

- **balena CLI version:** balena-cli version "12.37.0", Node.js version "12.19.1"
- **Cloud backend: balenaCloud
- **Operating system version:** Alpine: latest
- **32/64 bit OS and processor:** armv7hf (balenaFin)
- **Install method:** npm 
[...]

PR #2346, opened 2021-09-20
Update balena-sdk to 15.51.1

nitishagar:
Update balena-sdk from 15.48.0 to 15.51.0

Change-type: patch
  
HQ: https://jel.ly.fish/e450713f-001c-4470-b571-71724aeb0869
See: https://www.flowdock.com/app/rulemotion/r-product/threads/DZt_LvQvMCOoN6p2EceSu6MG4MF
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/IPdnkCLH2G0o7DSlkSS9nRetA4f  
See: https://github.com/balena-io/balena-cli/pull/2294

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,7 +207,7 @@
     "balena-image-manager": "^7.0.3",
     "balena-preload": "^10.5.0",
     "balena-release": "^3.2.0",
-    "balena-sdk": "^15.48.0",
+    "balena-sdk": "^15.51.1",
     "balena-semver": "^2.3.0",
     "balena-settings-client": "^4.0.7",
     "balena-settings-storage": "^7.0.0",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1300,12 +1300,12 @@
       "integrity": "sha512-u86QDMtkpHLlvehs3Z+yHklXRhDPL5XGCO3BCSuaD61gKzrNDUIj03cz8T/PBPPUJqn7DfWkf9sKP9VwlvxKuw=="
     },
     "@balena/node-web-streams": {
-      "version": "0.2.3",
-      "resolved": "https://registry.npmjs.org/@balena/node-web-streams/-/node-web-streams-0.2.3.tgz",
-      "integrity": "sha512-WaFtrO5lQUAWmLVcBn7V0tLHOuX/S9JPxmhfcEc9drLZhNUKF/psnNwWGfhPxgRwwik4hK0AqKIbjqkLLBXhSg==",
+      "version": "0.2.4",
+      "resolved": "https://registry.npmjs.org/@balena/node-web-streams/-/node-web-streams-0.2.4.tgz",
+      "integrity": "sha512-Q9By3GPzANMZuf1i5i7Agyh6BUe6tTa+VCCZzsFzU32iXMcuDRXYHbNIKESrcjVXxiZScPB4u++WPw4LRyK1Gg==",
       "requires": {
         "is-stream": "^1.1.0",
-        "web-streams-polyfill": "^1.3.2"
+        "web-streams-polyfill": "^3.1.0"
       }
     },
     "@balena/udif": {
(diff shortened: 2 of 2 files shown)


pdcastro (commented):
Looks good @nitishagar 👍   Just a minor thing: The commit and PR specify `change-type: minor`, but from what I can tell, no new functionality is provided to CLI users. It looks to me like it fits in the category of "chore" dependency updates with no impact to end users, also like bug fixes or security updates, which I would normally classify as a "patch" (rather than "minor") update. 
If there is actually some new feature provided to users, then the commit message could perhaps be amended to reflect it (while keeping the commit body as it is).

klutchell:
I'm a little wary that the following unrelated upstream changes would also be included:
- https://github.com/balena-io/balena-sdk/pull/1120
- https://github.com/balena-io-modules/balena-request/pull/156

These were supposed to be done around the same time as some other fixes for https://github.com/balena-io/balena-cli/issues/2152. However, we haven't figured out those other fixes yet and the PR mentioned above actually masks the `Z_BUF_ERR` and appears to download the image successfully even though the final file is incomplete.

So while on it's own it seems like a good fix to have, it actually hides another issue that we haven't solved yet. I can revert balena-requests, or revert the change in both repos while moving them forward, but I worry we will start seeing incomplete file downloads unless we get to the root of the os download pipeline issues.

Thoughts on how to proceed and unblock this PR?

pdcastro:
> [...] and the PR mentioned above actually masks the `Z_BUF_ERR` and appears to download the image successfully even though the final file is incomplete.
> [...] I can revert balena-requests, or revert the change in both repos while moving them forward [...]

Thanks for the insight @klutchell 👍  Instead of reverting those changes, could the changes be enabled/disabled through a [feature switch](https://en.wikipedia.org/wiki/Feature_toggle)? If we could think of no better option, perhaps even an environment variable that the CLI could set, starting with this PR, say `process.env.MASK_Z_BUF_ERR=false`, either globally or in selected commands like `balena os download`. (I don't mean an environment variable that users would see/set, I mean simply CLI code assigning to `process.env.VAR_NAME`.)

nitishagar:
For the following bit:
> If there is actually some new feature provided to users, then the commit message could perhaps be amended to reflect it (while keeping the commit body as it is).

Among other changes, we are looking to enable addition of `provision-key` name as a param for CLI to consume. Ref: https://github.com/balena-io/balena-cli/pull/2294

I took the default message with the update script, but can detail it based on the above. Let me know.

pdcastro:
> If there is actually some new feature provided to users [...]
>> Among other changes, we are looking to enable addition of `provision-key` name as a param for CLI to consume. Ref: #2294

@nitishagar, but this PR (just bumping `balena-sdk`) does not add `provision-key` name as a param. I understand that this PR does not make any difference to CLI users (other than possibly introducing a bug in `balena os download` as mentioned by Kyle). In other words, I understand that there isn't anything that CLI users can do with the CLI after this PR is merged, that they could not do before it was merged. So this PR would be `Change-type: patch`, and #2294 would be `Change-type: minor`.

>> I took the default message with the update script, but can detail it based on the above. Let me know.

If my understanding above is correct, then the commit message is OK as it is, but `Change-type` should be patch.

The reason why it matters (a little bit) is that the commit message and new semver version will be visible in the [Changelog](https://github.com/balena-io/balena-cli/blob/master/CHANGELOG.md) that CLI end users and support agents refer to in order to find out what's new, and to decide whether to update the CLI.

klutchell:
I would like to get the following PRs merged before this one if possible:
- https://github.com/balena-io/balena-sdk/pull/1126 (and update this PR to use sdk `v15.51.1`)
- https://github.com/balena-io/balena-cli/pull/2347 (under test)

package.json
```diff
@@ -207,7 +207,7 @@
     "balena-image-manager": "^7.0.3",
     "balena-preload": "^10.5.0",
     "balena-release": "^3.2.0",
-    "balena-sdk": "^15.48.0",
+    "balena-sdk": "^15.51.0",
```
pdcastro, line 210: @nitishagar, Kyle's PRs have been merged and `balena-sdk` v15.51.1 is now available. Please select `balena-sdk` v15.51.1 here and change the commit `Change-type` to patch, then we can get this PR merged. Thanks! 🙏

pdcastro (approved):

PR #2353, opened 2021-09-29
Fix help output for 'release finalize' command

pdcastro:
* Add reference documentation for the `releases`, `release` and `release finalize` commands to `doc/cli.markdown`
* Fix `balena help -v` for `release finalize`
* Coincidentally, update `@sentry/node` and remove `mixpanel` to fix a Linux build test error, "RequestError: connect ECONNREFUSED 127.0.0.1:443". I believe it is the same issue reported in https://github.com/sindresorhus/got/issues/951, although I do not understand why it only started manifesting now.
* Clean up some test warning messages.

Change-type: patch


lib/app.ts
```diff
@@ -33,6 +33,7 @@ export const setupSentry = onceAsync(async () => {
 	const config = await import('./config');
 	const Sentry = await import('@sentry/node');
 	Sentry.init({
+		autoSessionTracking: false,
```
pdcastro, line 36: `@sentry/node` was updated from v5 to v6, and this added property preserves the v5 behavior. Otherwise, session HTTP requests are always made to Sentry.io even if no error occurred, which seemed unnecessary to me.

automation/capitanodoc/capitanodoc.ts
```diff
@@ -82,6 +82,14 @@ const capitanoDoc = {
 				'build/commands/device/shutdown.js',
 			],
 		},
+		{
+			title: 'Releases',
+			files: [
+				'build/commands/releases.js',
+				'build/commands/release/index.js',
+				'build/commands/release/finalize.js',
```
pdcastro, line 90: The `doc/cli.markdown` file was missing reference documentation for the `releases`, `release` and `release finalize` commands. Adding these lines fixes it.

lib/events.ts
```diff
@@ -109,16 +96,43 @@ export async function trackCommand(commandSignature: string) {
...
 		}
 	} catch {
 		// ignore
 	}
 }
+
+/**
+ * Make the event tracking HTTPS request to balenaCloud's '/mixpanel' endpoint.
+ */
+async function sendEvent(balenaUrl: string, event: string, username?: string) {
```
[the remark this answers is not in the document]
pdcastro: Also, although in the past we actually used `mixpanel.com`, currently we use `amplitude.com` instead. Either way, the requests are proxied through the balenaCloud API backend, which is why it didn't matter that the CLI was still using the `mixpanel` client library for this GET request.

pdcastro:
@balena-ci I self-certify!

pdcastro:
@toochevere @20k-ultra, I've self certified but post-mortem reviews are also welcome. 👻

PR #2357, opened 2021-10-04
Error message when renaming a fleet now mentions the target name.

cmfcruz:
The source name is shown in the error message instead of the target name.
This patch fixes the error message.

Change-type: patch
Signed-off-by: Carlo Miguel F. Cruz <carloc@balena.io>

---

### DIFF
--- a/lib/commands/app/rename.ts
+++ b/lib/commands/app/rename.ts
@@ -120,7 +120,7 @@ export class FleetRenameCmd extends Command {
 		} catch (e) {
 			// BalenaRequestError: Request error: "organization" and "app_name" must be unique.
 			if ((e.message || '').toLowerCase().includes('unique')) {
-				throw new ExpectedError(`Error: fleet ${params.fleet} already exists.`);
+				throw new ExpectedError(`Error: fleet ${newName} already exists.`);
 			}
 			throw e;
 		}


lib/commands/app/rename.ts
```diff
@@ -120,7 +120,9 @@ export class FleetRenameCmd extends Command {
 		} catch (e) {
 			// BalenaRequestError: Request error: "organization" and "app_name" must be unique.
 			if ((e.message || '').toLowerCase().includes('unique')) {
-				throw new ExpectedError(`Error: fleet ${params.fleet} already exists.`);
+				throw new ExpectedError(
+					`Error: fleet ${params.newName} already exists.`,
```
pdcastro, line 124:
```suggestion
					`Error: fleet ${newName} already exists.`,
```
If I am reading it right, `params.newName` may be `undefined` in which case `newName` (rather than `params.newName`) will store the new name interactively typed by the user (see definition of `const newName =` in line 108).

pdcastro (changes requested):
👍

lib/commands/app/rename.ts
cmfcruz: Thanks for reviewing, @pdcastro ! I'll amend the commit to apply that change.

pdcastro (approved):

PR #2255, opened 2021-04-21
Add message regarding deprecation of --buildArg option

srlowe:
Change-type: patch
Signed-off-by: Scott Lowe <scott@balena.io>

PR #2360, opened 2021-10-12
Add ability to push to remote device UUID

joshbwlng:
Change-type: minor

---

Add ability to push to a device using its UUID. This works for both local and remote devices by first tunneling the device's supervisor API (port `48484`) and the docker service (port `2375`) to `localhost` and then running build as normal but with the target device host being set as `localhost`. The device must be in local mode in order for this to work (otherwise the necessary ports won't be available for tunneling).

### DIFF
--- a/doc/cli.markdown
+++ b/doc/cli.markdown
@@ -2913,12 +2913,14 @@ Examples:
 	
 	$ balena push 23c73a1.local --system
 	$ balena push 23c73a1.local --system --service my-service
+	
+	$ balena push 9f0cc5e4-0707-487c-ba84-edd0c36ff1c9
 
 ### Arguments
 
 #### FLEETORDEVICE
 
-fleet name or slug, or local device IP address or ".local" hostname
+fleet name or slug, device UUID, or local device IP address or ".local" hostname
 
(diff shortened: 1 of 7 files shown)


pdcastro (commented):
General heads-up comment/question, and I don't mean to imply an answer either way, I really don't know! Does this need to go through a product call before further effort is put into it? (Brainstorm topic, discussion, then Product Improvement.) I have nothing against this new feature/ability, but we have a few painful examples of past PRs and features where significant dev effort was invested, and then when the feature was eventually mentioned or discussed in a product call, even post merge, the work had to be undone because it was deemed to be insufficiently justified "additional product surface". For example, PR #2110 (add `--buildArg`) that led to #2255 (deprecate `--buildArg`), or the `push --pull` flag (I was told off for adding it without a product discussion), or the Docker containers for the CLI that we had to remove from DockerHub after the feature was released and shared with CLI users.

I gather this is for a hack week project. While this is a draft PR, the CI will create release assets in the releases page that can be used for demo and realistic testing. Then perhaps when the work is validated / blessed at the product level, the PR could then be merged.

Issue #2349, opened 2021-09-23
`config inject` should not require an internet connection


pdcastro:
So that it can be used in air-gapped manufacturing environments.

Connects-to #2343 

Internal discussion: (restricted access)
https://jel.ly.fish/brainstorm-topic-8c32926f-b928-47a5-9e4f-d069cf06fc12

pdcastro:
This issue was resolved in CLI version 12.53.0.

Issue #2343, opened 2021-09-16
Remove connectivity requirement for commands that can work without API access


alexgg:
# About this issue tracker

*The balena CLI (Command Line Interface) is a tool used to interact with the balena platform.
This GitHub issue tracker is used for bug reports and feature requests regarding the CLI
tool. General and troubleshooting questions (such as setting up your project to work with a
balenalib base image) are encouraged to be posted to the [balena
forums](https://forums.balena.io), which are monitored by balena's support team and where the
community can both contribute and benefit from the answers.*

*Please also check that this issue is not a duplicate. If there is another issue describing
the same problem or feature please add comments to the existing issue.*

*Thank you for your time and effort creating the issue report, and helping us improve
the balena CLI!*

---

# Expected Behavior

If there is no need to access the API, a command should not access the internet. For example, `balena config inject` has no need to either connect to the internet or be delayed in its action by attempting to connect to the internet.

The use case here is using balena CLI on a manufacturing line without internet access and where each second it takes to perform an operation has a cost.

# Actual Behavior

Injecting  a pre-generated `config.json` into a pre-downloaded image fails if there is no network connection.

# Steps to Reproduce the Problem

1. Use an airgapped workstation
2. Attempt to inject  a pre-generated `config.json` into a pre-downloaded image

# Specifications
```
balena-cli version "12.48.12"
Node.js version "10.21.0"
```
Ubuntu 18.04 64bits

# Additional References

Product call https://jel.ly.fish/cb881f8f-7141-4310-bb26-1065778f2e27

pdcastro:
Regarding the fact that `balena config inject` queries the `/device-types/v1` API endpoint, possible solutions were discussed in an [FD thread](https://www.flowdock.com/app/rulemotion/s-sales/threads/uu0aIKbd-LZJpUJWo3ZIsew-nPs) (restricted access) and reference was made to https://github.com/balena-io/balena-cli/issues/1632#issuecomment-775998774.

pdcastro:
This issue was resolved in CLI version 12.53.0.

PR #2369, opened 2021-10-22
config read/write/inject: Avoid need for internet access

pdcastro:
As raised by [brainstorm topic](https://jel.ly.fish/brainstorm-topic-8c32926f-b928-47a5-9e4f-d069cf06fc12)
And as specified by [product improvement](https://jel.ly.fish/improvement-modify-balena-config-inject-related-cli-commands-work-without-internet-connection-9ca2e56)

Resolves: #2343 
Resolves: #2349 
Depends-on: balena-io-modules/balena-config-json/pull/21
Change-type: minor


pdcastro:
@balena-ci I self-certify!

PR #2374, opened 2021-11-11
Don't recursively scan ignored directories

LucianBuzzo:
This changes improves the speed that the project is tarballed by filtering out
ignored directories *during* the file scanning process instead of
afterwards, preventing the CLI from needlessly scanning potentially
large directories (e.g. node_modules).
Whilst testing with the Jellyfish repository, which contains a number of
sub directories, each with their own node_modules folder, I was able to
reduce the time taken to scan and tarball the project from 70s to 2s,
which is a massive improvement.

Change-type: patch

### DIFF
--- a/lib/utils/ignore.ts
+++ b/lib/utils/ignore.ts
@@ -301,7 +301,23 @@ export async function filterFilesWithDockerignore(
 	await new Promise((resolve, reject) => {
 		// Looking at klaw's source code, `preserveSymlinks` appears to only
 		// afect the `stats` argument to the `data` event handler
-		klaw(projectDir, { preserveSymlinks: false })
+		klaw(projectDir, {
+			preserveSymlinks: false,
+			filter: (value) => {
+				const relPath = path.relative(projectDir, value);
+				for (const dir of dockerignoreServiceDirs) {
+					if (relPath.startsWith(dir)) {
+						if (!ignoreByDir[dir].ignores(relPath.substring(dir.length))) {
+							return true;
+						}
+					}
+				}
+				if (!ignoreByDir[root].ignores(relPath)) {
+					return true;
+				}
+				return false;
+			},
+		})
 			.on('error', reject)
 			.on('end', resolve)
 			.on('data', (item: { path: string; stats: Stats }) => {
@@ -320,17 +336,7 @@ export async function filterFilesWithDockerignore(
 				if (path.basename(relPath) === '.dockerignore') {
 					dockerignoreFiles.push(fileInfo);
 				}
-				for (const dir of dockerignoreServiceDirs) {
-					if (relPath.startsWith(dir)) {
-						if (!ignoreByDir[dir].ignores(relPath.substring(dir.length))) {
-							filteredFileList.push(fileInfo);
-						}
-						return;
-					}
-				}
-				if (!ignoreByDir[root].ignores(relPath)) {
-					filteredFileList.push(fileInfo);
-				}
+				filteredFileList.push(fileInfo);
 			});
 	});
 	return { filteredFileList, dockerignoreFiles };
--- a/lib/utils/device/deploy.ts
+++ b/lib/utils/device/deploy.ts
@@ -197,14 +197,17 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 
 	await checkBuildSecretsRequirements(docker, opts.source);
 	globalLogger.logDebug('Tarring all non-ignored files...');
+	const tarStartTime = Date.now();
 	const tarStream = await tarDirectory(opts.source, {
 		composition: project.composition,
 		convertEol: opts.convertEol,
 		multiDockerignore: opts.multiDockerignore,
 		nogitignore: opts.nogitignore, // v13: delete this line
 	});
+	globalLogger.logDebug(`Tarring complete in ${Date.now() - tarStartTime} ms`);
 
 	// Try to detect the device information
+	globalLogger.logDebug('Fetching device information...');
 	const deviceInfo = await api.getDeviceInformation();
 
 	let buildLogs: Dictionary<string> | undefined;


lib/utils/ignore.ts
```diff
@@ -301,7 +301,23 @@ export async function filterFilesWithDockerignore(
...
+					if (relPath.startsWith(dir)) {
+						if (!ignoreByDir[dir].ignores(relPath.substring(dir.length))) {
+							return true;
+						}
+					}
+				}
+				if (!ignoreByDir[root].ignores(relPath)) {
+					return true;
+				}
+				return false;
```
pdcastro, line 318: If I recall correctly, when this `filterFilesWithDockerignore` function was originally written we considered early directory pruning ("don't visit a subtree if the subtree's 'root' directory is ignored") to improve performance, and then realised that the `.dockerignore` format (unlike `.gitignore`) makes this difficult because it allows patterns  (anywhere in the `.dockerignore` file) that re-add selected files or directories down a subtree where a parent directory was ignored.

There is this documentation:

https://github.com/balena-io-modules/dockerignore#whats-different-from-ignore
> _The .dockerignore spec has several differences from .gitignore:_
> _[...]_
> _With .gitignore, when a parent directory is ignored, subdirectories cannot be re-added (using `!`) since git simply avoids walking through the subtree as an optimization. With .dockerignore, a subdirectory can be re-added even if a parent directory has been ignored._

It may help to look at some `dockerignore` test cases:
https://github.com/balena-io-modules/dockerignore/blob/master/test/ignore.js
Such as:
```
[
    'It is possible to re-include a file if a parent directory of that file is excluded',
    [
      'abc/',
      '!abc/a.js'
    ],
    {
      'abc/a.js': 0,
      'abc/d/e.js': 1,
    }
  ],
```

I.e., if an `abc/` pattern determines the this directory is ignored, but some other `'!abc/a.js'` pattern adds exceptions down the `abc/` subtree (at any depth down the subtree), we cannot simply prune (filter out) the whole `abc/` subtree.

Also these:
```
[
    'unignore with 1 globstar',
    [
      'foo/*',
      '!foo/bar.js'
    ],
    {
      'foo/bar.js': 0,
      'foo/bar2.js': 1,
      'foo/bar/bar.js': 1,
    }
  ],
  [
    'unignore with 2 globstars',
    [
      'foo/**',
      '!foo/bar.js'
    ],
    {
      'foo/bar.js': 0,
      'foo/bar2.js': 1,
      'foo/bar/bar.js': 1,
    }
  ],
```
Or indeed "ignore EVERYTHING, except for these files":
```
[
    'wildcard with whitelisting (zeit/dockerignore/issues/11)', [
      '*',
      '!src',
      '!scripts',
      '!config',
      '!yarn.lock',
      '!package.json',
      '!.env*',
    ], {
      '.env.sample': 0,
      '.env/.sample': 0,
      '.env/other': 0,
      'Dockerfile': 1,
      '.dockerignore': 1,
    }
  ],
```

Like you, I suspect we could do something cleverer than testing every single file in the project whether that specific file is being docker-ignored. I can imagine a clever algorithm that analysed all dockerignore patterns and "compiled them" in an optimized, annotated data structure of patterns that would tell us, for a given directory, whether any exceptions exist in its subtree. Most of the time there'd be no exceptions, so the subtree could be pruned, achieving the 70s → 2s mentioned in the PR description. (The improvement multiplier is virtually unbounded, a matter of traversing an arbitrarily large subtree vs not traversing it at all.) It's probably doable, but it would not be as simple as the changes currently put forward in this PR.

By the way, the [balena-io-modules/dockerignore](https://github.com/balena-io-modules/dockerignore) repo is a Javascript port of Docker's Golang implementation of dockerignore file filtering (ref: [dockerignore.go](https://github.com/moby/moby/blob/v19.03.8/builder/dockerignore/dockerignore.go) and [fileutils.go](https://github.com/moby/moby/blob/v19.03.8/pkg/fileutils/fileutils.go)), _almost line by line to ensure maximum compatibility_ (regarding the relevant pattern matching algorithm). So, presumably, Docker itself suffers from the performance issues that motivated this PR, and if we came up with a cleverer algorithm, it would benefit Docker as well (unless they've already improved it since Docker v19).

lib/utils/ignore.ts
```diff
@@ -320,17 +336,7 @@ export async function filterFilesWithDockerignore(
...
-						if (!ignoreByDir[dir].ignores(relPath.substring(dir.length))) {
-							filteredFileList.push(fileInfo);
-						}
-						return;
-					}
-				}
-				if (!ignoreByDir[root].ignores(relPath)) {
-					filteredFileList.push(fileInfo);
-				}
+				filteredFileList.push(fileInfo);
```
pdcastro, line 339: The build is failing:
https://ci.product-os.io/builds/35125
```
...
expected vs. found diff:
{
    "src/file1.sh": {
        "fileSize": 12,
        "type": "file"
    }
}
...
expected { '.dockerignore': { fileSize: 12, type: 'file' },
  'Dockerfile-alt': { fileSize: 13, type: 'file' },
  'file2-crlf.sh': { fileSize: 14, type: 'file' } } to deeply equal { '.dockerignore': { fileSize: 12, type: 'file' },
  'file2-crlf.sh': { fileSize: 14, type: 'file' },
  'src/file1.sh': { fileSize: 12, type: 'file' },
  'Dockerfile-alt': { fileSize: 13, type: 'file' } }
```
I believe at this test case:
https://github.com/balena-io/balena-cli/blob/v12.51.1/tests/commands/build.spec.ts#L461
```ts
it('should create the expected tar stream (docker-compose, --multi-dockerignore)'
```

I think (just from reading, I didn't run/debug it to confirm) that the following test data `.dockerignore` file was not processed as it should:
https://github.com/balena-io/balena-cli/blob/v12.51.1/tests/test-data/projects/docker-compose/basic/service2/.dockerignore
```
**/file1.sh
```
Such that `src/file1.sh` was not ignored as it should have been.

pdcastro (changes requested):

LucianBuzzo:
@pdcastro Thankyou for the detailed response here, I'm closing this PR in favour of https://github.com/balena-io/balena-cli/pull/2376 which has noticeable speed improvements, but without the issues you describe here.

PR #2375, opened 2021-11-12
deploy: Ensure the release fails if an image's digest (hash) is missing

pdcastro:
This PR improves error handling for `balena deploy`, ensuring the release status is 'failed' and error message is printed for the user in the event that an image digest (content hash) cannot be extracted from the image push stream by `docker-progress`.

The issue was reported in the support thread linked below. We don't know exactly what the root cause was for a blank digest (so far we could not reproduce it), and therefore this PR does NOT fix the root cause as such, however this PR at least detects the situation, print an error message and marks the release as 'failed'. Before this PR, the release would be marked 'successful' despite the missing image hash, causing the supervisor to get in bad state on devices.

See: https://jel.ly.fish/support-thread-supervisor-running-may-unhealthy-bf05575
Change-type: patch

### DIFF
--- a/lib/utils/compose_ts.ts
+++ b/lib/utils/compose_ts.ts
@@ -1306,15 +1306,101 @@ async function getTokenForPreviousRepos(
 	return token;
 }
 
+async function pushAndUpdateServiceImages(
+	docker: Dockerode,
+	token: string,
+	images: TaggedImage[],
+	afterEach: (
+		serviceImage: import('balena-release/build/models').ImageModel,
+		props: object,
+	) => void,
+) {
+	const { DockerProgress } = await import('docker-progress');
+	const { retry } = await import('./helpers');
+	const { pushProgressRenderer } = await import('./compose');
+	const tty = (await import('./tty'))(process.stdout);
+	const opts = { authconfig: { registrytoken: token } };
+	const progress = new DockerProgress({ docker });
+	const renderer = pushProgressRenderer(
+		tty,
+		getChalk().blue('[Push]') + '    ',
+	);
+	const reporters = progress.aggregateProgress(images.length, renderer);
+
+	const pushImage = async (
+		localImage: Dockerode.Image,
+		index: number,
+	): Promise<string> => {
+		try {
+			// TODO 'localImage as any': find out exactly why tsc warns about
+			// 'name' that exists as a matter of fact, with a value similar to:
+			// "name": "registry2.balena-cloud.com/v2/aa27790dff571ec7d2b4fbcf3d4648d5:latest"
+			const imgName: string = (localImage as any).name || '';
+			const imageDigest: string = await retry({
+				func: () => progress.push(imgName, reporters[index], opts),
+				maxAttempts: 3, // try calling func 3 times (max)
+				label: imgName, // label for retry log messages
+				initialDelayMs: 2000, // wait 2 seconds before the 1st retry
+				backoffScaler: 1.4, // wait multiplier for each retry
+			});
+			if (!imageDigest) {
+				throw new ExpectedError(stripIndent`\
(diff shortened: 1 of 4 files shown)


pdcastro:
@balena-ci retest

pdcastro:
@balena-ci I self-certify!

lib/utils/compose_ts.ts
```diff
@@ -1306,15 +1306,101 @@ async function getTokenForPreviousRepos(
...
+				func: () => progress.push(imgName, reporters[index], opts),
+				maxAttempts: 3, // try calling func 3 times (max)
+				label: imgName, // label for retry log messages
+				initialDelayMs: 2000, // wait 2 seconds before the 1st retry
+				backoffScaler: 1.4, // wait multiplier for each retry
+			});
+			if (!imageDigest) {
+				throw new ExpectedError(stripIndent`\
+					Unable to extract image digest (content hash) from image upload progress stream for image:
+					${imgName}`);
```
pdcastro, line 1349: Testing whether `imageDigest` is empty and, if so, throwing an error is the key change of this PR in relation to addressing the linked support thread. The rest of the PR is just conversion from Javascript to Typescript, and replacement of `Bluebird.using` with try/catch/finally. cc: @dfunckt @thgreasi (I have self certified, but post mortem feedback is still welcome!)

dfunckt:
@pdcastro did you manage to reproduce/understand when a digest isn't returned? In principle, I think this is better handled in docker-progress -- did you consider this option?

pdcastro:
> @pdcastro did you manage to reproduce/understand when a digest isn't returned? In principle, I think this is better handled in docker-progress -- did you consider this option?

Yes, I agree, but I was unable to reproduce the error with the same version of docker and the same version of the CLI (which was by the way the standalone zip package with Node.js v10, thus ruling out a Node.js issue unless the user's installation is broken). More details in my comments on the linked thread. If we could reproduce the error, we could then identify why docker-progress could not extract the digest and then fix docker-progress as the case may be. I am not sure if the user themselves are able to reproduce the specific issue of an empty digest. The user provided "steps to reproduce"; however, from the user's point of view, the "visible issue" was that the supervisor was stuck and running `balena deploy` again would not get the supervisor unstuck, and overall this looks like reproducing the issue, but it only means that the supervisor remained stuck. It is possible that the empty digest occurrence was one-off even for the user, causing the supervisor to get stuck until support agents deleted the supervisor database.
The user was invited to try CLI v12.51.3 or later. If the error happens again, it will be before any device downloads the release, and the user themselves will then realise that there is a problem that is unrelated to any devices or the supervisor. Steps to reproduce would then hopefully be a lot more focused than what the user gave us last week.

dfunckt:
Thanks @pdcastro, that makes sense. Without being able to reproduce at the level docker-progress works we can’t even attempt to fix it properly and I think it’s fine to handle it by aborting in the CLI like this PR does.

As a future improvement, to increase the odds of finding the root cause, do you think it makes sense to change the error message to inform the user that this is an unexpected condition and encourage them to report it?

pdcastro:
> As a future improvement, to increase the odds of finding the root cause, do you think it makes sense to change the error message to inform the user that this is an unexpected condition and encourage them to report it?

Not sure. If the error happened consistently and reproducibly, it would presumably block the user and the user would be likely to report it anyway. If it happened sporadically and the error message made a point about the user reporting it to us, they would report it and we might not be able to fix it anyway, as it is hard to reproduce. Also, this may be a controversial opinion but _"it's not like we don't have enough bugs and features to worry about"_ (e.g., [these](https://github.com/balena-io/balena-cli/issues)), so I'd rather not be asked to investigate low ROI issues that cannot be reproduced. :-)

PR #2376, opened 2021-11-12
Improve directory scan speed prior to tarballing

LucianBuzzo:
This changes improves the speed that the project is tarballed by switching from
`klaw` to `fdir` and not running `lstat` on files that are ignored.
Whilst testing with the Jellyfish repository, which contains a number of
sub directories, each with their own node_modules folder, I was able to
reduce the time taken to scan and tarball the project from 70s to 11s,
which is a massive improvement.

Change-type: patch

### DIFF
--- a/lib/utils/ignore.ts
+++ b/lib/utils/ignore.ts
@@ -201,6 +201,48 @@ export interface FileStats {
 }
 
 /**
+ * Create a list of files for the filesystem subtree rooted at
+ * projectDir, excluding entries for directories themselves.
+ * @param projectDir Source directory (root of subtree to be listed)
+ */
+async function listFiles(projectDir: string): Promise<string[]> {
+	const dirs: string[] = [];
+	const files: string[] = [];
+	dirs.push(projectDir);
+	async function walk(currentDirs: string[]): Promise<string[]> {
+		if (!currentDirs.length) {
+			return files;
+		}
+
+		const foundDirs: string[] = [];
+
+		// Because `currentDirs` can be of arbitrary length, process them in smaller batches
+		// to avoid out of memory errors.
+		// This approach is significantly faster than using Bluebird.map with a
+		// concurrency setting
+		const chunks = _.chunk(currentDirs, 100);
+		for (const chunk of chunks) {
+			await Promise.all(
+				chunk.map(async (dir) => {
+					const _files = await fs.readdir(dir, { withFileTypes: true });
+					for (const entry of _files) {
+						const fpath = path.join(dir, entry.name);
+						if (entry.isDirectory()) {
+							foundDirs.push(fpath);
+							dirs.push(fpath);
+						} else {
+							files.push(fpath);
+						}
+					}
+				}),
+			);
+		}
+		return walk(foundDirs);
+	}
+	return walk([projectDir]);
+}
+
+/**
  * Return the contents of a .dockerignore file at projectDir, as a string.
  * Return an empty string if a .dockerignore file does not exist.
  * @param projectDir Source directory
@@ -211,7 +253,7 @@ async function readDockerIgnoreFile(projectDir: string): Promise<string> {
 	let dockerIgnoreStr = '';
 	try {
 		dockerIgnoreStr = await fs.readFile(dockerIgnorePath, 'utf8');
-	} catch (err) {
+	} catch (err: any) {
 		if (err.code !== 'ENOENT') {
 			throw new ExpectedError(
 				`Error reading file "${dockerIgnorePath}": ${err.message}`,
@@ -269,7 +311,10 @@ export async function filterFilesWithDockerignore(
 	projectDir: string,
 	multiDockerignore: boolean,
 	serviceDirsByService: ServiceDirs,
-): Promise<{ filteredFileList: FileStats[]; dockerignoreFiles: FileStats[] }> {
+): Promise<{
+	filteredFileList: FileStats[];
+	dockerignoreFiles: FileStats[];
+}> {
 	// path.resolve() also converts forward slashes to backslashes on Windows
 	projectDir = path.resolve(projectDir);
 	const root = '.' + path.sep;
@@ -294,45 +339,65 @@ export async function filterFilesWithDockerignore(
 	const dockerignoreServiceDirs: string[] = multiDockerignore
 		? Object.keys(ignoreByDir).filter((dir) => dir && dir !== root)
 		: [];
+	const files = await listFiles(projectDir);
+
 	const dockerignoreFiles: FileStats[] = [];
 	const filteredFileList: FileStats[] = [];
-	const klaw = await import('klaw');
 
-	await new Promise((resolve, reject) => {
-		// Looking at klaw's source code, `preserveSymlinks` appears to only
-		// afect the `stats` argument to the `data` event handler
-		klaw(projectDir, { preserveSymlinks: false })
-			.on('error', reject)
-			.on('end', resolve)
-			.on('data', (item: { path: string; stats: Stats }) => {
-				const { path: filePath, stats } = item;
-				// With `preserveSymlinks: false`, filePath cannot be a symlink.
-				// filePath may be a directory or a regular or special file
-				if (!stats.isFile()) {
-					return;
-				}
+	// Because `files` can be of arbitrary length, process them in smaller batches
+	// to avoid out of memory errors.
+	// This approach is significantly faster than using Bluebird.map with a
+	// concurrency setting
+	const chunks = _.chunk(files, 750);
+	for (const chunk of chunks) {
+		await Promise.all(
+			chunk.map(async (filePath) => {
 				const relPath = path.relative(projectDir, filePath);
-				const fileInfo = {
-					filePath,
-					relPath,
-					stats,
-				};
+
+				// .dockerignore files are always added to a list of known dockerignore files
 				if (path.basename(relPath) === '.dockerignore') {
-					dockerignoreFiles.push(fileInfo);
+					const diStats = await fs.stat(filePath);
+					dockerignoreFiles.push({
+						filePath,
+						relPath,
+						stats: diStats,
+					});
 				}
-				for (const dir of dockerignoreServiceDirs) {
-					if (relPath.startsWith(dir)) {
-						if (!ignoreByDir[dir].ignores(relPath.substring(dir.length))) {
-							filteredFileList.push(fileInfo);
-						}
+
+				// First check if the file is ignored by a .dockerignore file in a service directory
+				const matchingDir = dockerignoreServiceDirs.find((dir) => {
+					return relPath.startsWith(dir);
+				});
+
+				// If the file is ignore in a service directory, exit early, otherwise check if it is ignored by the root .dockerignore file.
+				// Crucially, if the file is in a known service directory, and isn't ignored, the root .dockerignore file should not be checked.
+				if (matchingDir) {
+					if (
+						ignoreByDir[matchingDir].ignores(
+							relPath.substring(matchingDir.length),
+						)
+					) {
 						return;
 					}
+				} else if (ignoreByDir[root].ignores(relPath)) {
+					return;
 				}
-				if (!ignoreByDir[root].ignores(relPath)) {
-					filteredFileList.push(fileInfo);
+
+				// At this point we can do a final stat of the file, and check if it should be included
+				const stats = await fs.stat(filePath);
+
+				// filePath may be a special file that we should ignore, such as a socket
+				if (stats.isFile()) {
+					filteredFileList.push({
+						filePath,
+						relPath,
+						stats,
+					});
 				}
-			});
-	});
+			}),
+		);
+	}
+
 	return { filteredFileList, dockerignoreFiles };
 }
 
--- a/lib/utils/device/deploy.ts
+++ b/lib/utils/device/deploy.ts
@@ -199,14 +199,17 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 
 	await checkBuildSecretsRequirements(docker, opts.source);
 	globalLogger.logDebug('Tarring all non-ignored files...');
+	const tarStartTime = Date.now();
 	const tarStream = await tarDirectory(opts.source, {
 		composition: project.composition,
 		convertEol: opts.convertEol,
 		multiDockerignore: opts.multiDockerignore,
 		nogitignore: opts.nogitignore, // v13: delete this line
 	});
+	globalLogger.logDebug(`Tarring complete in ${Date.now() - tarStartTime} ms`);
 
 	// Try to detect the device information
+	globalLogger.logDebug('Fetching device information...');
 	const deviceInfo = await api.getDeviceInformation();
 
 	let buildLogs: Dictionary<string> | undefined;
--- a/lib/utils/remote-build.ts
+++ b/lib/utils/remote-build.ts
@@ -317,12 +317,18 @@ async function getTarStream(build: RemoteBuild): Promise<Stream.Readable> {
 			Object.keys(build.opts.registrySecrets).length > 0
 				? preFinalizeCallback
 				: undefined;
-		return await tarDirectory(path.resolve(build.source), {
+		globalLogger.logDebug('Tarring all non-ignored files...');
+		const tarStartTime = Date.now();
+		const tarStream = await tarDirectory(path.resolve(build.source), {
 			preFinalizeCallback: preFinalizeCb,
 			convertEol: build.opts.convertEol,
 			multiDockerignore: build.opts.multiDockerignore,
 			nogitignore: build.nogitignore, // v13: delete this line
 		});
+		globalLogger.logDebug(
+			`Tarring complete in ${Date.now() - tarStartTime} ms`,
+		);
+		return tarStream;
 	} finally {
 		tarSpinner.stop();
 	}


LucianBuzzo:
@pdcastro What is this line checkinf for? https://github.com/balena-io/balena-cli/blob/master/tests/docker-build.ts#L74

pdcastro:
> What is this line checking for? https://github.com/balena-io/balena-cli/blob/master/tests/docker-build.ts#L74
```ts
    expect(foundFiles).to.not.have.property(header.name);
```

I think it checks that a file was not added twice to the tar stream.
Files found in the tar stream are added to the `foundFiles` objects.
If the `foundFiles` object already had such a file, it is an unexpected condition.

lib/utils/ignore.ts
```diff
@@ -295,44 +295,47 @@ export async function filterFilesWithDockerignore(
...
-				}
-			});
+
+	const { fdir } = await import('fdir');
+
+	// crawl the project files
+	const files = await new fdir()
+		.withFullPaths()
+		.crawl(projectDir)
+		.withPromise();
```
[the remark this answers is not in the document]
LucianBuzzo: @pdcastro The issue I'm finding is that `withSymLinks()` will follow the sim link and provide the resolved path, and the CLI tests expect the path to the symlink itself. I'm not sure which one is correct, though I think the tar functionality should probably be providing the path to the symlink, rather than the resolved path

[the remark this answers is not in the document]
LucianBuzzo: Switching to [`recursive-fs`](https://github.com/simov/recursive-fs) gives the same level of performance boost, but without the aforementioned issue with symlink resolution.

LucianBuzzo:
@balena-ci retest

lib/utils/ignore.ts
```diff
@@ -295,44 +295,40 @@ export async function filterFilesWithDockerignore(
...
-				if (!ignoreByDir[root].ignores(relPath)) {
-					filteredFileList.push(fileInfo);
-				}
-			});
+
+	const rfs = await import('recursive-fs');
+	const { files } = await rfs.read(projectDir);
+
+	const filteredFilePaths: string[] = (files as string[]).filter((filePath) => {
+		const relPath = path.relative(projectDir, filePath);
```
pdcastro, line 303:
```suggestion
		const relPath = path.relative(projectDir, filePath);
		if (path.basename(relPath) === '.dockerignore') {
			return true;
		}
```
I think (I didn't test) this change is needed to preserve existing behaviour in some corner cases, _in addition to_ the similar lines later in this same function:
```ts
		if (path.basename(relPath) === '.dockerignore') {
			dockerignoreFiles.push(fileInfo);
		}
```
Because the latter lines would be pointless if `.dockerignore` was not included in the `filteredFilePaths` array.

This may (or not) indeed be one of the reasons why some test cases are still failing, e.g.:

```
      -  "[Warn] * C:\\var\\lib\\concourse\\worker\\volumes\\live\\0ac9c696-c6d6-4eba-5984-b6b9c8547f60\\volume\\tests\\test-data\\projects\\no-docker-compose\\dockerignore2/lib/.dockerignore"
      +  "[Warn] * C:\\var\\lib\\concourse\\worker\\volumes\\live\\0ac9c696-c6d6-4eba-5984-b6b9c8547f60\\volume\\tests\\test-data\\projects\\no-docker-compose\\dockerignore2\\lib\\.dockerignore"
```
https://ci.product-os.io/builds/42293

pdcastro (commented):
Tests are passing on Linux and macOS, 🎉  but some tests are still failing on Windows (https://ci.product-os.io/builds/42293).  I have added a couple more review comments that may address at least some of the reasons why some tests are failing.

lib/utils/ignore.ts
pdcastro: Oh, actually, I should read my own comments, :-) the diff I pasted above shows that the difference is related to forward slashes vs backslashes, the latter the canonical form on Windows. It looks like the `recursive-fs` implementation assumes forward slashes on all platforms:

https://github.com/simov/recursive-fs/blob/master/lib/read.js#L19
```js
          var fpath = `${dir}/${entry.name}`
```
(Note how they hardcode the forward slash above, instead of using something like `path.join(dir, entry.name)` or  `path.sep`.)

Windows also understands forward slashes, so this is not to say that `recursive-fs` is necessarily broken on Windows, just not canonical and potentially mixing forward slashes with backslashes in a path string, which may be more of a matter of aesthetics than correctness. The current implementation in the balena CLI uses backslashes on Windows as the native / canonical form, e.g. the result of [path.normalize()](https://nodejs.org/docs/latest-v10.x/api/path.html#path_path_normalize_path). We could call `path.normalize()` on the paths in `filteredFilePaths`, or we could accept that Windows paths will have a mix of forward slashes and backslashes and amend the test cases instead.

pdcastro:
```
         "[Warn] -----------------------------------------------------------------------------------------------------------------------------------------------------------------"
         "[Warn] The following .dockerignore file(s) will not be used:"
      -  "[Warn] * C:\\var\\lib\\concourse\\worker\\volumes\\live\\0ac9c696-c6d6-4eba-5984-b6b9c8547f60\\volume\\tests\\test-data\\projects\\no-docker-compose\\dockerignore2/lib/.dockerignore"
      +  "[Warn] * C:\\var\\lib\\concourse\\worker\\volumes\\live\\0ac9c696-c6d6-4eba-5984-b6b9c8547f60\\volume\\tests\\test-data\\projects\\no-docker-compose\\dockerignore2\\lib\\.dockerignore"
         "[Warn] When --multi-dockerignore (-m) is used, only .dockerignore files at the"
         "[Warn] root of each service's build context (in a microservices/multicontainer"
         "[Warn] fleet), plus a .dockerignore file at the overall project root, are used."
         "[Warn] See \"balena help push\" for more details."
         "[Warn] -----------------------------------------------------------------------------------------------------------------------------------------------------------------"
```

To be honest it doesn't feel right to display warning messages with a mix of forward slashes and backslashes in a path string like `no-docker-compose\\dockerignore2/lib/.dockerignore` as shown above, so I would favour a solution where this was avoided, preferably displaying the platform's canonical path format (backslashes on Windows). This could mean converting paths at the point where they are displayed. However, an existing assumption in the codebase is that paths use the platform's canonical representation. Indeed, the documentation of the `filterFilesWithDockerignore` function reads:

```ts
/**
 ...
 * The service directory names/paths must be relative to the project root dir
 * and be "normalized" (path.normalize()) before the call to this function:
 * they should use backslashes on Windows, not contain '.' or '..' segments and
 * not contain multiple consecutive path separators like '//'. Also, relative
 * paths must not start with './' (e.g. 'a/b' instead of './a/b').
 */
export async function filterFilesWithDockerignore(
```

We could also take the view that `recursive-fs` 's approach of hardcoding forward slashes consists of "anti-social behaviour", and we could submit a PR to their repo to use `path.join()` instead.

By the way, just because saying this feels relevant, :-) the rules are different for the contents of tar streams. The tar format always uses forward slashes, never backslashes, regardless of the platform. It's like a platform-independent spec of a file format. Hence calls to `toPosixPath` in code such as:

```
async function newTarDirectory(
	...
		pack.entry(
			{
				name: toPosixPath(fileStats.relPath),
```

Also, and again just because this feels relevant, note that user input on Windows (command line arguments such as `--source`) may contain a mix of forward slashes and backslashes and the CLI currently copes with this by normalising user input. On Linux and macOS, however, backslashes are never understood as path component separators.

package.json
```diff
@@ -262,6 +262,7 @@
     "prettyjson": "^1.1.3",
     "progress-stream": "^2.0.0",
     "reconfix": "^1.0.0-v0-1-0-fork-46760acff4d165f5238bfac5e464256ef1944476",
+    "recursive-fs": "^2.1.0",
```
pdcastro, line 265: Something I should have remembered before, but better late than never... The current implementation using `klaw` is something we arrived at in PR #2093 from @joshbwlng (a year ago!) that was aimed at improving performance for Jellyfish with livepush -- not speed, but reducing high CLI memory usage of a previous ***recursive*** directory traversal in a custom `listFiles()` function:

> https://github.com/balena-io/balena-cli/pull/2093#issuecomment-727771295
> Been testing this branch today and it looks like there might be another problem that causes `filterFilesWithDockerignore()` to fail when too many files exist under the project root. Jellyfish currently has multiple `node_modules` directories, five when all `npm install`s have been ran (one for the project root along with one for each app). After adding a few debug logs, I was able to track down the problem to the `listFiles()` function in `lib/utils/ignore.ts`. This runs before dockerignore patterns are taken into account, also causing non-livepush `push`es to fail during `newTarDirectory()`. 

The previous recursive traversal in a `listFiles()` function (replaced with `klaw`) can be seen in commit: [push: Reduce memory usage when filtering files with dockerignore](https://github.com/balena-io/balena-cli/pull/2093/commits/4768f763856aa9f761988477f97ec872d226b004).

That previous `listFiles()` function probably had higher memory usage than the use of `recursive-fs` proposed in this PR, I think mainly because `listFiles()` executed `fs.stat` on _all files before filtering,_ which this PR cleverly avoids. Nonetheless, `recursive-fs` also performs _recursive_ directory traversal (the hint is in the package name...), as we see in the following `walk` function that calls itself:
https://github.com/simov/recursive-fs/blob/v2.1.0/lib/read.js#L30

I leave it to you to decide, but the thought that occurred to me (out of the blue while having a shower!), compounded by the fact that `recursive-fs` has the forward slash vs `path.join()` Windows issue, is to replace `recursive-fs` with a custom `listFiles()` function of a year ago, with a few key changes:

* The new `listFiles()` function would **not** call `fs.stat` on all files before filtering, which I understand is the key speed improvement of this PR.
* The new `listFiles()` function would have an iterative, **non-recursive** implementation, hoping that this saves some memory by avoiding the automated function call stack build-up. Ref: [Recursion to Iteration](https://blog.moertel.com/posts/2013-05-11-recursive-to-iterative.html)
* Like the old `listFiles()` function and like `recursive-fs`, the new `listFiles()` function would be based on `fs.readdir()`, and would take advantage of `entry.isDirectory()` and other [fs.Dirent](https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_class_fs_dirent) methods without needing a call to `fs.stat`.

LucianBuzzo: Thanks for all the extra information @pdcastro 
When you were looking to reduce memory, how were you testing memory consumption? I've been developing the solution using Jellyfish as a test bed, which generates a list of ~277k files, but I have quite a powerful workstation so don't really see any issues with memory.
I'd quite like to avoid re-inventing a file scanning script, so if we can find an existing solution that works that would be the best solution. 
As an additional thought, tarballing a directory for consumption by Docker seems to be a generally useful tool that could exist outside of the CLI - what are your thoughts on this?

lib/utils/ignore.ts
LucianBuzzo: Regardless of the approach we take here, I've sent a [PR to `recursive-fs`](https://github.com/simov/recursive-fs/pull/5) to adjust the paths generated on Windows.

package.json
LucianBuzzo: I also think that the big culprit of memory consumption was the call to `stat` rather than the recursive programming style.

lib/utils/ignore.ts
LucianBuzzo: For now, I'd like to push forward and use `path.normalize` on the list of filtered files.

package.json
pdcastro: > how were you testing memory consumption?

I don't remember having actually measured it (embarrassingly). 😊  I think we had isolated CLI versions where `push` worked and where it didn't, we found that `listFiles()` is something that had changed between those, and found that using `klaw` avoided the failure reported by Josh.

Re not re-inventing a file scanning script, yes I agree. 👍  At the same time, I don't feel strongly between the options of adding a small function to enumerate files around `fs.readdir()`, versus adding a new dependency to `package.json`. Adding dependencies also adds some maintenance cost over time, for example we regularly need to update dependencies (bumping them in `package.json`) for reasons such as updating the supported versions of Node.js or addressing `npm audit` concerns (because of changes made by 3rd parties that may or may not benefit us), and when doing so we need to check the dependencies' changelog for what's changed, especially new major versions.

> tarballing a directory for consumption by Docker seems to be a generally useful tool that could exist outside of the CLI - what are your thoughts on this?

Related packages that come to mind:

* Our own `tar-utils` package:
    * https://www.npmjs.com/package/tar-utils
    * https://github.com/balena-io-modules/node-tar-utils/
* [tar-stream](https://www.npmjs.com/package/tar-stream)
* [@balena/dockerignore](https://www.npmjs.com/package/@balena/dockerignore)
* Our own `docker-toolbelt`
    * https://www.npmjs.com/package/docker-toolbelt
    * https://github.com/balena-io-modules/docker-toolbelt
* [dockerode](https://www.npmjs.com/package/dockerode)

I think you were thinking about a new package that combined tar-stream with dockerignore. Our `docker-toolbelt` package could be a candidate for hosting the functionality. (`docker-toolbelt` could do with a full rewriting, not just converting from Coffeescript to Typescript but also a review of the functionality it provides and the public interface.)

A complexity that the module would have to address is where to expect the `.dockerignore` file(s) to be, i.e. the difference between providing and not providing the balena CLI's `--multi-dockerignore (-m)` flag, and compatibility with the "standard" `docker-compose build` and `docker build` behaviour.

lib/utils/ignore.ts
```diff
@@ -295,44 +295,44 @@ export async function filterFilesWithDockerignore(
...
-				}
-				if (!ignoreByDir[root].ignores(relPath)) {
-					filteredFileList.push(fileInfo);
+
+	const rfs = await import('recursive-fs');
+	const { files } = await rfs.read(projectDir);
+
+	const filteredFilePaths: string[] = (files as string[])
+		.filter((filePath) => {
+			const relPath = path.relative(projectDir, filePath);
```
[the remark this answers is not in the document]
LucianBuzzo: @pdcastro Is there a CLI test that checks for this behaviour?

LucianBuzzo:
@pdcastro I can't respond on your comment above so I will post here instead. I think you make a good point about maintenance overhead, so I will try an approach where I re-implement `listFiles` but without using `stat` internally.

lib/utils/ignore.ts
```diff
@@ -200,6 +200,41 @@ export interface FileStats {
...
+						dirs.push(fpath);
+					} else {
+						files.push(fpath);
+					}
+				}
+			}),
+		);
+		return walk(foundDirs);
+	}
+	return walk([projectDir]);
```
pdcastro, line 235:
```suggestion
		return await walk(foundDirs);
	}
	return await walk([projectDir]);
```
It is not incorrect to leave out `await` here, but I understand that adding `await` improves debug tracing (easier to follow stack traces) when the returned promised is rejected, as the rejection gets associated with the line of code where `await` is used. But this is just me parroting what I was told (or read) before, as I've never tested it myself to confirm.

In the case of the inner `walk` call specifically, as it is recursive, I wonder if adding `await` would reduce parallelisation by also reducing the depth of the recursive stack (also reducing memory usage). In which case, the `await` keyword would not be just a "good practice" matter but a performance matter, controlling a tradeoff between memory and speed.

lib/utils/ignore.ts
```diff
@@ -200,6 +200,41 @@ export interface FileStats {
...
+	dirs.push(projectDir);
+	async function walk(currentDirs: string[]): Promise<string[]> {
+		if (!currentDirs.length) {
+			return files;
+		}
+
+		const foundDirs: string[] = [];
+
+		await Promise.all(
+			currentDirs.map(async (dir) => {
```
pdcastro, line 220: If you don't mind checking, what would the speed impact be (in seconds) if we did **not** use `Promise.all` + `.map` to parallelise the calls to `fs.readdir`? I.e., some less "aggressive" coding like:
```suggestion
		for (const dir of currentDirs) {
```

And also adding the `await` keyword to `return await walk(foundDirs);` (see other comment too).

I suspect the aggressive parallelisation could be blamed for the memory exhaustion of `listFiles` a year ago. Does this parallelisation mean that `fs.readdir` would run simultaneously for potentially **ALL** directories in the project at once? E.g. 100,000 directories (exaggeration) would mean running `fs.readdir` 100,000 times in parallel with simultaneous stack memory use for 100,000 calls to `listFiles`. Or maybe I am misunderstanding it!

I gather that `recursive-fs` does **not** do aggressive parallelisation thanks to the following test in callback style:

https://github.com/simov/recursive-fs/blob/v2.1.0/lib/read.js#L29-L30
```js
        if (++complete === _dirs.length) {
          walk(__dirs)
        }
```

LucianBuzzo: @pdcastro This is an await/async version of what recursive-fs does, which prevents recursion into subdirectories until the current level has been scanned. I haven't tested a linear/non-async approach but based on what I've seen from other implementations I'd expect it to be about 30% slower.
The memory usage is almost certainly down to unnecessarily calling `stat` AFAICT.

LucianBuzzo:
@balena-ci rebase

pdcastro:
@balena-ci retest

lib/utils/ignore.ts
```diff
@@ -200,6 +200,41 @@ export interface FileStats {
...
+				const _files = await fs.readdir(dir, { withFileTypes: true });
+				for (const entry of _files) {
+					const fpath = `${dir}/${entry.name}`;
+					if (entry.isDirectory()) {
+						foundDirs.push(fpath);
+						dirs.push(fpath);
+					} else {
+						files.push(fpath);
```
pdcastro, line 228: To be confirmed:
I assume (didn't test) that `fpath` here could be a symbolic link or special file (like socket, block or character devices, named pipes, etc). If a symbolic link, it is not followed (resolved) here, and as currently implemented, I think the dockerignore tests are made against the unresolved symlinks rather than against the files they point to.

I believe the `klaw` implementation followed (resolved) symlinks, and then discarded entries that weren't a regular file, all _before_ the dockerignore tests were executed on the entries:
```ts
				if (!stats.isFile()) {
					return;
				}
```

Unless there are reasons not to (such as noticeably hurting performance), I would favour maintaining existing behaviour (otherwise, strictly, it would be a breaking change). Existing tests may or may not be catching all these scenarios and yes, it would be nice to improve the test cases. :-)

LucianBuzzo: @pdcastro I don't think that `klaw` resolves symlinks, if it did the tests that we have for symlinked files would fail. Could you create a test case that runs the scenario you are describing?

lib/utils/ignore.ts
```diff
@@ -200,6 +200,41 @@ export interface FileStats {
...
+			return files;
+		}
+
+		const foundDirs: string[] = [];
+
+		await Promise.all(
+			currentDirs.map(async (dir) => {
+				const _files = await fs.readdir(dir, { withFileTypes: true });
+				for (const entry of _files) {
+					const fpath = `${dir}/${entry.name}`;
```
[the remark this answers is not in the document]
LucianBuzzo: @pdcastro It's ~5% slower. On the Jellyfish repo, which is pretty huge in terms of number of files to scan, changing to the approach you suggest results in an extra 500ms of scanning. I think this is an acceptable tradeoff for having cleaner code here.

LucianBuzzo:
@pdcastro From what I understand of `.dockeringore` behaviour, if we were able to determine that negation is not used, we could speed up the scanning process even further by not recursing into ignored directories 🤔 
This would be something to do in a separate PR though.

pdcastro:
@balena-ci rebase

pdcastro:
I've added a test case and modified others in PR #2384, hence rebasing. While writing the tests, I also identified two pre-existing bugs that I've described in issues #2383 and #2385. To avoid delaying this PR, I suggest that those bugs be fixed in other PRs.
(Even though this PR, as is, potentially addresses #2385. The problem with addressing #2385 in this PR is that I only tested `docker build`'s behavior with socket files, and not all other possible special files such as block devices, character devices, FIFOs, etc. It would make sense to test all possible special files before fixing #2385. And as for #2383, the fix is arguably a breaking change that would ideally be delivered in CLI v13 - next major.)

lib/utils/ignore.ts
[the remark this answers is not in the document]
pdcastro: I've modified existing test cases in #2384 to cover this scenario.

lib/utils/ignore.ts
pdcastro: I've added a test case for Unix domain socket files in #2384, rebased this PR and now the build is failing - sorry! 😀 
See also this comment: https://github.com/balena-io/balena-cli/pull/2376#issuecomment-975017791

Re symlinks, I have not (yet) determined whether or not `klaw` resolves symlinks at this point, but potentially it does not matter\* because symlinks will eventually be resolved in [readFile(fileStats.filePath)](https://github.com/balena-io/balena-cli/blob/v12.52.1/lib/utils/compose_ts.ts#L812) if they are not resolved here.

\* Except for pre-existing bug #2383, which I propose be fixed in some other PR to avoid delaying this one.

LucianBuzzo:
@pdcastro Great stuff! It sounds like we should utilise/port the official test suite for dockerignore behaviour and use that here

LucianBuzzo:
@balena-ci retest

pdcastro:
> It sounds like we should utilise/port the official test suite for dockerignore behaviour and use that here

It looks like [that test suite](https://github.com/balena-io-modules/dockerignore/blob/v1.0.2/test/ignore.js) does not include tests for special files like symbolic links, unix domain sockets, block devices, etc. That suite has the somewhat impressive functionality of actually executing `docker build` and `docker run` to verify that pattern matching behaved just like docker, which is what would be have been useful here at the balena CLI level for symbolic links, sockets, etc. I believe the `dockerignore` library is really a "pattern matcher at the string level" (a bit like a regular expression engine). When we call the `ignores('some-path')` function, the implementation of `dockerignore` does not access the file system (the source code does not even import the 'fs' module) and it's OK if the given path does not even exist. As such, it actually makes sense that that test suite, for its purposes, does not include test cases for special files: it makes no difference at the string matching level.

The considerations of file type (socket, symbolic link, etc) are made by the balena CLI or the `docker` client at a "higher level" than what the `dockerignore` library concerns itself with.

Re "official test suite" -- the [@balena/dockerignore](https://www.npmjs.com/package/@balena/dockerignore) module is an improved fork of the archived [@zeit/dockerignore](https://www.npmjs.com/package/@zeit/dockerignore) module and a JS port of Docker's / Moby's source code where `.dockerignore` pattern matching is implemented. (Btw, improved to the point that AWS chose to depend on it - [@aws-cdk/core](https://www.npmjs.com/package/@aws-cdk/core), [aws-cdk-lib](https://www.npmjs.com/package/aws-cdk-lib).) As far as I am aware, the test cases in the test suite were written by the authors of the JS port of `dockerignore` pattern matching, hence not quite "official", but it is reliable because it is backed by automated comparison against the behaviour of executing `docker build` and `docker run`. The original, "official" dockerignore pattern matching source code is just part of Docker's / Moby's Go codebase; there isn't a "standalone" `dockerignore` library used by Docker / Moby (or at least there wasn't the last time I checked, for Docker v19). The `balena-io-modules/dockerignore` library's interface/API was borrowed from the npm [ignore](https://www.npmjs.com/package/ignore) package, which is an implementation of the `.gitignore` spec, replacing the matching engine with an implementation (JS port) of the `.dockerignore` spec.

It would be interesting to find Docker's / Moby's source code where the treatment of special files is implemented, specially or at least the treatment of symbolic links as observed in #2383. One should do that as part of the work of addressing #2383 and #2385 (in other PRs).

LucianBuzzo:
@pdcastro I've updated the code to reduce unnecessary looping and to check against special files like sockets, and the tests are now passing. This still shows a considerable speed improvement when testing it locally, which I'm super happy with 👍

lib/utils/ignore.ts
```diff
@@ -294,45 +332,57 @@ export async function filterFilesWithDockerignore(
...
-		klaw(projectDir, { preserveSymlinks: false })
-			.on('error', reject)
-			.on('end', resolve)
-			.on('data', (item: { path: string; stats: Stats }) => {
-				const { path: filePath, stats } = item;
-				// With `preserveSymlinks: false`, filePath cannot be a symlink.
-				// filePath may be a directory or a regular or special file
-				if (!stats.isFile()) {
+	await Promise.all(
+		files.map(async (filePath) => {
```
pdcastro, line 340: If I am reading it right, this will create an array of promises, the array size being equal to the number of files in the project, and then all promises are awaited in parallel. If the project had 500,000 files, this would produce 500,000 promises awaiting in parallel.

I am concerned about memory usage. Consider also some people's hack week projects where they wish to run `balena push` in low-memory IoT devices for the purpose of running an IDE on the device itself.

The following StackOverflow post suggests a hard Node.js limit of 2.1 million promises awaited in parallel, in which case a project with just over 2.1 million files would necessarily fail, even if RAM was not a problem: https://stackoverflow.com/questions/55753746/how-much-is-the-limit-of-promise-all

There is probably a number of concurrent promises above which further parallelisation yields no further performance improvement. I mean, running perhaps "100" (?) `fs.stat` calls in parallel might be faster than running them sequentially, but running "500,000" in parallel might be no faster than running "100" at a time. In fact, I would not be surprised if, as the number got large enough, performance worsened due to extra "bookkeeping" (epoll array iterations looking for fd's that are ready?) or in case the system had to resort to memory pagination in swap space (disk).

Bluebird has a fancy `promise.map` implementation that accepts a `concurrency` parameter:
> http://bluebirdjs.com/docs/api/promise.map.html
> The concurrency limit applies to Promises returned by the mapper function and it basically limits the number of Promises created. For example, if concurrency is 3 and the mapper callback has been called enough so that there are three returned Promises currently pending, no further callbacks are called until one of the pending Promises resolves. So the mapper function will be called three times and it will be called again only after at least one of the Promises resolves.

Perhaps you could test some concurrency numbers and find the sweet one that will hopefully be relatively low.

lib/utils/ignore.ts
```diff
@@ -294,45 +332,57 @@ export async function filterFilesWithDockerignore(
...
+			const relPath = path.relative(projectDir, filePath);
+
+			// .dockerignore files are always added to a list of known dockerignore files
+			if (path.basename(relPath) !== '.dockerignore') {
```
pdcastro, line 344:
```suggestion
			if (path.basename(relPath) === '.dockerignore') {
```
Some test cases are failing and I think the errors can be explained by the logic above:

```
         "[Warn] The following .dockerignore file(s) will not be used:"
         "[Warn] * /tmp/build/29c50989/versioned-source/tests/test-data/projects/no-docker-compose/basic/src/start.sh"
         "[Warn] * /tmp/build/29c50989/versioned-source/tests/test-data/projects/no-docker-compose/basic/src/windows-crlf.sh"
```

`start.sh` and `windows-crlf.sh` are not `.dockerignore` files!

lib/utils/ignore.ts
```diff
@@ -294,45 +332,57 @@ export async function filterFilesWithDockerignore(
...
+					filePath,
+					relPath,
+					stats: diStats,
+				});
+			}
+
+			// First check if the file is ignored by a .dockerignore file in a service directory
+			const matchingDir = _.find(dockerignoreServiceDirs, (dir) => {
+				return relPath.startsWith(dir);
+			});
```
pdcastro, line 356: If it was Page reviewing the PR, :-) he'd say this is an unnecessary use of Lodash. Is the following equivalent?
```suggestion
			const matchingDir = dockerignoreServiceDirs.find((dir) =>
				relPath.startsWith(dir)
			);
```

lib/utils/ignore.ts
LucianBuzzo: Oops, this is a typo, fixed!

lib/utils/ignore.ts
LucianBuzzo: @pdcastro This function will walk one layer of directories at a time, so even on the Jellyfish project, which is a very large multicontainer app, the maximum number of promises created is 8993.
Switching to `Bluebird.map` is a convenient way to prevent something crazy happening, and setting the concurrency to 750 has no impact on performance.

lib/utils/ignore.ts
LucianBuzzo: @pdcastro yep, good idea, I'll switch to native `.find()`

lib/utils/ignore.ts
pdcastro: > This function will walk one layer of directories at a time

Which function, `listFiles`? But it is being called as follows:
```ts
    const files = await listFiles(projectDir);
```
So even if `listFiles` builds the array on one layer of directories at a time, the `files` array above will be as large as all files in the project, no?

> setting the concurrency to 750 has no impact on performance.

So, shouldn't we at least do that?

LucianBuzzo: @pdcastro I should have been clearer, I've set the concurrency to 750 for listFiles, We can do something similar for the `ignore` processing as well.

LucianBuzzo: @pdcastro I did some more testing with both `listFiles` and the processing of the returned values and found that using `Bluebird` instead of native promises was significantly slower.
Using `Bluebird.map` without concurrency completed the tar process in 12387ms, after switching to native promises + [_.chunk](https://lodash.com/docs/4.17.15#chunk) to limit concurrency the process completed in 9057ms.
The actual tarballing step takes ~2s on my test setup, so it seems that using Bluebird is ~47% slower than a "native" approach!
I've now refactored both places I was using `Promise.all` to first chunk the arrays at a sensible size to prevent arbitrarily large file systems causing high memory consumption.

lib/utils/ignore.ts
```diff
@@ -294,45 +339,65 @@ export async function filterFilesWithDockerignore(
...
-				// filePath may be a directory or a regular or special file
-				if (!stats.isFile()) {
-					return;
-				}
+
+	// Because `files` can be of arbitrary length, process them in smaller batches
+	// to avoid out of memory errors.
+	// This approach is significantly faster than using Bluebird.map with a
+	// concurrency setting
+	const chunks = _.chunk(files, 750);
```
pdcastro, line 351: Nice 👍  In your testing where you found that 750 was a sweet number, how many (logical/virtual) CPUs did the machine have? I am thinking of making this number proportional to the CPU count, something like:
```suggestion
	const chunks = _.chunk(files, Math.round(750 * (os.cpus().length || 1) / LUCIANS_CPU_COUNT));
```
Replace `LUCIANS_CPU_COUNT` with the result of `os.cpus().length` on the workstation where you ran the tests.

If the count on your machine was 16, then on a weaker device with cpu count 1 the chunk number would become 47.

I understand that Node.js is single threaded so CPU count would not be relevant in "user-space code", but I believe that the promises would be blocked on syscalls (`fs.stat`) that are executed by the kernel, and I assume that the kernel would take advantage of all CPU cores.

A similar calculation could be done in `listFiles`.

LucianBuzzo: @pdcastro My machine has 24 cores, but as you mention I'm not sure if the count is super relevant. Do you have any idea how to verify that `fs.stat` is affected by the number of CPU cores?

LucianBuzzo: I also wonder how necessary this change is, considering the main concern is out of memory errors. Perhaps we could merge the change and see if CPU consumption becomes an issue for people?

pdcastro (approved):

Issue #2372, opened 2021-11-08
balena ssh: Remove unneeded communication with the API when using local mode


rmorillo24:
# Expected Behavior

when working in local mode with a device, doing `balena ssh` should work always when using a local network, even when there is no connection with balenaCloud.
I would expect local mode to work independently from the cloud, either because there is a firewall, or API unavailability.

# Actual Behavior

When we've experienced balenaCloud API outages, `balena ssh` would not work, not connecting to the device or returning any error. 

# Steps to Reproduce the Problem
With an API outage, balena ssh doesn´t work.
I´ve tried disconnecting the router from internet, and it does connect, but when I exit the session it remains waiting .

# Specifications

- **balena CLI version:** e.g. 1.2.3 (output of the `"balena version -a"` command)
- **Cloud backend: openBalena or balenaCloud?** If unsure, it will be balenaCloud
- **Operating system version:** e.g. Windows 10, Ubuntu 18.04, macOS 10.14.5
- **32/64 bit OS and processor:** e.g. 32-bit Windows on 64-bit Intel processor
- **Install method:** npm or zip package or executable installer
- **If npm install, Node.js and npm version:** e.g. Node v8.16.0 and npm v6.4.1

# Additional References

If applicable, please add additional links to GitHub projects, forums.balena.io threads,
gist.github.com, Google Drive attachments, etc.

srlowe:
@rmorillo24 Thanks for creating the issue.  

Just to clarify, 

> I´ve tried disconnecting the router from internet, and it does connect, but when I exit the session it remains waiting .

So now when you try without an internet connection, you can connect to local device via SSH?  (but the CLI doesn't return control of the command line afterwards?) 

 Just asking because this seems different to the symptoms you describe during the API outage:

> When we've experienced balenaCloud API outages, balena ssh would not work, not connecting to the device or returning any error.

So are you saying that the behaviour was different in the API outage and your simulated test (disconnected router)? 

I'll do some tests of my own now.

rmorillo24:
That is correct @srlowe 
While writing down the issue, I went to reproduce by disconnecting the router, and found the different behavior in nboth situations.
They are different symptoms, and I guess that an API outage is behaving differently than what I tried: it might be establishing the connection, but not having a response. In the use case I reproduced there can't be a connection at all. Thus the difference in behavior.

pdcastro:
CLI v12.53.3 addresses this issue in a couple of different ways / scenarios.

[...]

PR #2377, opened 2021-11-12
Improve offline usage experience

srlowe:
Resolves: #2343
Resolves: #2349
Resolves: #2372 
Change-type: minor

### DIFF
--- a/lib/commands/config/read.ts
+++ b/lib/commands/config/read.ts
@@ -21,7 +21,7 @@ import * as cf from '../../utils/common-flags';
 import { getVisuals, stripIndent } from '../../utils/lazy';
 
 interface FlagsDef {
-	type: string;
+	type?: string;
 	drive?: string;
 	help: void;
 }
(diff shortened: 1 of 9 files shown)


lib/commands/config/read.ts
```diff
@@ -61,7 +59,7 @@ export default class ConfigReadCmd extends Command {
 		await safeUmount(drive);
 
 		const config = await import('balena-config-json');
-		const configJSON = await config.read(drive, options.type);
+		const configJSON = await config.read(drive, '');
```
pdcastro, line 64: This commit, `config inject/read/write: Avoid need for internet connection, ignore device type`, depends on balena-io-modules/balena-config-json/pull/21 
When I suggested cherry picking PR #2369, I mean only commit [044fd5b Add request timeout (5s) for deprecation check and event tracking](https://github.com/balena-io/balena-cli/pull/2369/commits/044fd5b683f25d71dbb3981c96027bc89a94c821)
The rest wasn't ready yet.

srlowe: @pdcastro Right, gotcha, thanks.

PR #2378, opened 2021-11-12
Improve UX for offline usage

srlowe:
Change-type: patch
Connects-to: #2343
Resolves: #2372

@pdcastro  I set the timeout a little bit shorted (4s instead of 5s), let me know what you think.

### DIFF
--- a/lib/app.ts
+++ b/lib/app.ts
@@ -77,11 +77,13 @@ export function setMaxListeners(maxListeners: number) {
 /** Selected CLI initialization steps */
 async function init() {
 	if (process.env.BALENARC_NO_SENTRY) {
-		console.error(`WARN: disabling Sentry.io error reporting`);
+		if (process.env.DEBUG) {
+			console.error(`WARN: disabling Sentry.io error reporting`);
+		}
 	} else {
 		await setupSentry();
 	}
-	checkNodeVersion();
+	await checkNodeVersion();
 
 	const settings = new CliSettings();
 
@@ -91,7 +93,9 @@ async function init() {
 	setupBalenaSdkSharedOptions(settings);
 
 	// check for CLI updates once a day
-	(await import('./utils/update')).notify();
+	if (!process.env.BALENARC_OFFLINE_MODE) {
+		(await import('./utils/update')).notify();
+	}
 }
 
 /** Execute the oclif parser and the CLI command. */
@@ -149,7 +153,10 @@ async function oclifRun(command: string[], options: AppOptions) {
 /** CLI entrypoint. Called by the `bin/balena` and `bin/balena-dev` scripts. */
 export async function run(cliArgs = process.argv, options: AppOptions = {}) {
 	try {
-		const { normalizeEnvVars, pkgExec } = await import('./utils/bootstrap');
+		const { setOfflineModeEnvVars, normalizeEnvVars, pkgExec } = await import(
+			'./utils/bootstrap'
+		);
+		setOfflineModeEnvVars();
 		normalizeEnvVars();
 
 		// The 'pkgExec' special/internal command provides a Node.js interpreter
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -16,7 +16,11 @@
  */
 
 import Command from '@oclif/command';
-import { InsufficientPrivilegesError } from './errors';
+import {
+	InsufficientPrivilegesError,
+	NotAvailableInOfflineModeError,
+} from './errors';
+import { stripIndent } from './utils/lazy';
 
 export default abstract class BalenaCommand extends Command {
 	/**
@@ -39,6 +43,13 @@ export default abstract class BalenaCommand extends Command {
 	 * if user is not already logged in.
 	 */
 	public static authenticated = false;
+
+	/**
+	 * Require an internet connection to run.
+	 * When set to true, command will exit with an error
+	 * if user is running in offline mode (BALENARC_OFFLINE_MODE).
+	 */
+	public static offlineCompatible = false;
 
 	/**
 	 * Accept piped input.
@@ -98,6 +109,29 @@ export default abstract class BalenaCommand extends Command {
 	}
 
 	/**
+	 * Throw NotAvailableInOfflineModeError if in offline mode.
+	 *
+	 * Called automatically if `onlineOnly=true`.
+	 * Can be called explicitly by command implementation, if e.g.:
+	 *  - check should only be done conditionally
+	 *  - other code needs to execute before check
+	 *
+	 *  Note, currently public to allow use outside of derived commands
+	 *  (as some command implementations require this. Can be made protected
+	 *  if this changes).
+	 *
+	 * @throws {NotAvailableInOfflineModeError}
+	 */
+	public static checkNotUsingOfflineMode() {
+		if (process.env.BALENARC_OFFLINE_MODE) {
+			throw new NotAvailableInOfflineModeError(stripIndent`
+		This command requires an internet connection, and cannot be used in offline mode.
+		To leave offline mode, unset the BALENARC_OFFLINE_MODE environment variable.
+		`);
+		}
+	}
+
+	/**
 	 * Read stdin contents and make available to command.
 	 *
 	 * This approach could be improved in the future to automatically set argument
@@ -123,6 +157,10 @@ export default abstract class BalenaCommand extends Command {
 
 		if (ctr.authenticated) {
 			await BalenaCommand.checkLoggedIn();
+		}
+
+		if (!ctr.offlineCompatible) {
+			BalenaCommand.checkNotUsingOfflineMode();
 		}
 
 		if (ctr.readStdin) {
--- a/lib/commands/config/write.ts
+++ b/lib/commands/config/write.ts
@@ -70,6 +70,7 @@ export default class ConfigWriteCmd extends Command {
 	};
 
 	public static root = true;
+	public static offlineCompatible = true;
 
 	public async run() {
 		const { args: params, flags: options } = this.parse<FlagsDef, ArgsDef>(
--- a/lib/events.ts
+++ b/lib/events.ts
@@ -16,7 +16,7 @@
  */
 
 import * as packageJSON from '../package.json';
-import { getBalenaSdk } from './utils/lazy';
+import { getBalenaSdk, stripIndent } from './utils/lazy';
 
 interface CachedUsername {
 	token: string;
@@ -129,10 +129,20 @@ async function sendEvent(balenaUrl: string, event: string, username?: string) {
 		data: Buffer.from(JSON.stringify(trackData)).toString('base64'),
 	};
 	try {
-		await got(url, { searchParams, retry: 0 });
+		await got(url, { searchParams, retry: 0, timeout: 4000 });
 	} catch (e) {
 		if (process.env.DEBUG) {
 			console.error(`[debug] Event tracking error: ${e.message || e}`);
 		}
+
+		if (e instanceof got.TimeoutError) {
+			console.error(stripIndent`
+				Timeout submitting analytics event to balenaCloud/openBalena.
+				If you are using the balena CLI in an air-gapped environment with a filtered
+				internet connection, set the BALENARC_OFFLINE_MODE=1 environment variable
+				when using CLI commands that do not strictly require access to balenaCloud.
+			`);
+		}
+		// Note: You can simulate a timeout using non-routable address 10.0.0.0
 	}
 }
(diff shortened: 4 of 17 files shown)


lib/events.ts
```diff
@@ -129,7 +129,7 @@ async function sendEvent(balenaUrl: string, event: string, username?: string) {
 		data: Buffer.from(JSON.stringify(trackData)).toString('base64'),
 	};
 	try {
-		await got(url, { searchParams, retry: 0 });
+		await got(url, { searchParams, retry: 0, timeout: 4000 });
```
pdcastro, line 132: > pdcastro I set the timeout a little bit shorted (4s instead of 5s), let me know what you think.

That's a tricky one. Actually, I had thought of 5s as an intentionally "unreasonably long timeout" so that it would be hopefully clear to anyone, ourselves included, that "something is wrong/broken and needs to be fixed, this is not normal".

For the sake of argument, consider two alternative scenarios:

* The CLI always requires an internet connection of "reasonable quality" (not a 50Kbps dial-up modem!). But in the rare event that the internet is down, some commands (like `balena ssh <ip-address>`) will still work, despite that very noticeable 5s timeout.

* At least for selected commands, the CLI never requires an internet connection. In this case, a 4s or 5s timeout in arguably not acceptable.

These are probably not the only choices, we could go with something in between or some other reasoning. My point is that 4s or 5s timeout is not a good user experience if it was an ordinary behaviour, regular occurrence.

My hope for a solution that addressed #2343 / #2349 / #2372 would be that the 4s or 5s timeout would not be relied on, somehow.

It also depends on the meaning of "no internet connection". In some testing of mine, if I turned WiFi off in the laptop and it had no active network interface at all, that connection timeout would NOT be triggered / relied on. The CLI would actually run faster than usual and exit straight away. That's because, when the process attempts a connection, the OS replies straight away with "sorry, no can do" and the error handling swallows its (as it should), and all is good.

The problem happens when the machine actually has a network connection, but:

1. Some local firewall only allows local communication with servers "on the factory floor" (picture an air gapped manufacturing line, the subject of #2349), and drops network packets to the outside world, like the balena API.

2. The machine has a good, unfiltered internet connection but the balenaCloud backend (API) is down, the subject of #2372.

For the second case, we could argue that relying on a 4s or 5s timeout is OK because it is a very rare occurrence. But for the first case, it would be a regular occurrence and it is harder to say that always having a 4s or 5s timeout is OK. It does not sound OK to me...

But how to solve the first case?

An "obvious" (?) solution is to reduce the timeout, not just from 5s to 4s but reduce it to... hmm... 400ms, or someone's favourite definition of acceptable. :-)

In the old times of Mixpanel, even before Page implemented analytics proxying via the backend (which was then improved to exit "as soon as possible", I think thanks to a suggestion of yours), we had some timeouts in the hundreds of milliseconds that never actually worked properly (for complex implementation reasons related to the Mixpanel module). When we switched to the proxied analytics, we had a brief discussion of what timeout to set and I think it was Page (?) who dared suggest "no timeout at". The reasoning was that we care for analytics and if the balenaCloud backend is not answering fast enough, we need to improve the backend rather than have the CLI drop requests.

An air gapped (filtered internet) manufacturing environment is a different situation to a slow backend, but they share a possible timeout solution.

Long comment, food for thought. If the purpose of this PR is to solve #2372, then:

* Is it really true that a firewall that dropped packets would cause `balena ssh <ip-addr>` to rely on the 4s or 5s timeout?

* If we have to rely on the timeout for #2372, should we reduce it to 400ms or 500ms instead of 4s or 5s?

* Would it make sense to have different timeouts for different CLI commands? E.g. if we classified some commands as "requires internet" or "does not require internet", what about setting the timeouts to 500ms when internet is not required, and 5s when internet is required?
  * Oh, not just "per command" but dynamically, because `balena ssh` only requires an internet connection if the argument is not an IP address or `.local` hostname.

* If a command (as dynamically determined) does not require internet connection, why even attempt analytics and deprecation requests? Answer: because even if an internet connection is not strictly required, "95% of the time" (?!) it would still be available and could be used.

pdcastro: Hmm I got confused in my comment, mixing #2349 with #2372. For either issue, a 4s or 5s timeout would be OK if the cause of the timeout is that the balenaCloud backend (API) is down, because this is very rare. However, if the cause of the timeout is an air gapped network (filtered internet connection) which would apply "all of the time", then a 4s of 5s timeout is less clearly acceptable.

I don't know... I am OK with merging this PR as it is. What do you think? :-)

pdcastro: Just had another thought / idea for discussion.

I think we are fundamentally saying that `balena ssh <ip-addr>` will _optionally_ use an internet connection (for analytics, Sentry.io and deprecation checks) if one is available, but the difficulty is that we cannot detect whether or not an internet connection is available (quickly without waiting for a timeout that hurts UX) in the scenario of an air gapped network with a firewall that drops packets to/from the outside world.

The CLI understands the BALENARC_UNSUPPORTED=1, BALENARC_NO_SENTRY=1, and BALENARC_NO_ANALYTICS=1 to disable deprecation checks, Sentry.io reporting and analytics reporting respectively. But I think we would **not** want to "advertise" these variables because we could end up with a substantial fraction of users just **always** setting them for all commands. They could add the variables to their `.bashrc` file and make them part of their "standard system setup for the balena CLI"...

However, what about a new variable, say BALENARC_OFFLINE_MODE=1, (or maybe BALENARC_AIR_GAPPED=1) whose behavior would be:

* BALENARC_OFFLINE_MODE=1 tells the CLI never to attempt a connection "to the internet", i.e. balena API, Dockerhub, npm registry etc. Connections to local devices and a local Docker engine would be allowed.

* We would manually label all commands (in some cases, depending on arguments such as `balena ssh <UUID>` _vs._ `balena ssh <ip-addr | host.local>`) regarding whether or not they depend on an internet connection, i.e. whether or not they are compatible with BALENARC_OFFLINE_MODE=1.

* If a user sets BALENARC_OFFLINE_MODE=1, then:
    * If they attempt to run a command that requires an internet connection such as `balena ssh <UUID>`, the CLI would print an error message that that command requires an internet connection. They have to unset the env var.
    * If they attempt to run a command that does NOT require an internet connection such as `balena ssh <ip-addr>`, the CLI simply acts as if BALENARC_UNSUPPORTED=1, BALENARC_NO_SENTRY=1 and BALENARC_NO_ANALYTICS=1 had been set.

* If a user does NOT set BALENARC_OFFLINE_MODE=1, then:
    * If they attempt to run a command that does NOT require an internet connection such as `balena ssh <ip-addr>`, the CLI will attempt submitting analytics and making deprecation checks (and possibly Sentry.io) and the HTTP request may timeout after 5s. We would detect whether the request timed out. If it timed out, we would print a warning message to stderr advising the user, along the lines of _"Timeout submitting analytics event to balenaCloud. If you are using the balena CLI in an air gapped environment without an internet connection, set the BALENARC_OFFLINE_MODE=1 environment variable when using CLI commands that do not strictly require access to balenaCloud, such as `balena ssh <ip-addr>`"_  
I would classify such a warning message as "advertising" BALENARC_OFFLINE_MODE "where needed". Users would not simply always set BALENARC_OFFLINE_MODE=1 because most (?) CLI commands actually do require an internet connection for understandable purposes that users care about, e.g. `balena push`.
    * Otherwise it just works as usual.

pdcastro:
@balena-ci retest

lib/events.ts
srlowe: Thanks for the detailed thoughts. Yeah it's a subtly complex issue.  Yeah I think the solution you've described here with BALENARC_OFFLINE_MODE sounds reasonable.  I'll take a look at the practicalities of implementing that.

lib/events.ts
```diff
@@ -129,10 +130,18 @@ async function sendEvent(balenaUrl: string, event: string, username?: string) {
...
 	} catch (e) {
 		if (process.env.DEBUG) {
 			console.error(`[debug] Event tracking error: ${e.message || e}`);
 		}
+		if (e instanceof got.TimeoutError) {
+			throw new AnalyticsRequestTimeout(stripIndent`
+				Timeout submitting analytics event to balenaCloud/openBalena.
+				If you are using the balena CLI in an air-gapped environment without an internet connection,
+				set the BALENARC_OFFLINE_MODE=1 environment variable when using CLI commands that do not strictly
+				require access to balenaCloud, such as balena ssh <ip-addr>.
```
pdcastro, line 143: Mainly to limit it at 80 characters per line, but also some minor suggestions:
```suggestion
				Timeout submitting analytics event to balenaCloud/openBalena.
				If you are using the balena CLI in an air-gapped environment with a filtered
				internet connection, set the BALENARC_OFFLINE_MODE=1 environment variable
				when using CLI commands that do not strictly require access to balenaCloud.
```

lib/command.ts
```diff
@@ -97,6 +108,29 @@ export default abstract class BalenaCommand extends Command {
...
+	 *  (as some command implementations require this. Can be made protected
+	 *  if this changes).
+	 *
+	 * @throws {NotAvailableInOfflineModeError}
+	 */
+	public static checkNotUsingOfflineMode() {
+		if (process.env.BALENARC_OFFLINE_MODE) {
+			throw new NotAvailableInOfflineModeError(stripIndent`
+		This command requires an internet connection, and cannot be used in offline mode.
+		To leave offline mode, unset the BALENARC_OFFLINE_MODE environment variable.
```
pdcastro, line 129: 👍

lib/events.ts
pdcastro: Indeed! It doesn't look like it would help us in this case though. In the case of a firewall that selectively drops packets (affecting connections to balenaCloud), `lookup` may or may not be blocked, and `connect` (or `secureConnect`) would most likely be blocked. I understand that, if we don't use this object, the 4s timeout we specify applies to the sum of all phases "from lookup to response", which is probably what we want.

lib/command.ts
```diff
@@ -125,6 +159,17 @@ export default abstract class BalenaCommand extends Command {
...
 
+		if (!ctr.offlineCompatible) {
+			BalenaCommand.checkNotUsingOfflineMode();
+		}
+
+		if (process.env.BALENARC_OFFLINE_MODE) {
+			// Set implied env vars
+			process.env.BALENARC_UNSUPPORTED = '1';
+			process.env.BALENARC_NO_SENTRY = '1';
+			process.env.BALENARC_NO_ANALYTICS = '1';
```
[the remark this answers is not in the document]
srlowe: I haven't done proper testing yet, but could be. (It wasn't ready for review).

lib/app.ts
```diff
@@ -77,11 +77,13 @@ export function setMaxListeners(maxListeners: number) {
...
 	if (process.env.BALENARC_NO_SENTRY) {
-		console.error(`WARN: disabling Sentry.io error reporting`);
+		if (process.env.DEBUG) {
+			console.error(`WARN: disabling Sentry.io error reporting`);
+		}
 	} else {
 		await setupSentry();
 	}
-	checkNodeVersion();
+	await checkNodeVersion();
```
srlowe, line 86: Seemed to be missing.

lib/commands/config/write.ts
```diff
@@ -66,9 +66,8 @@ export default class ConfigWriteCmd extends Command {
 		help: cf.help,
 	};
 
-	public static authenticated = true;
```
srlowe, line 69: Seems this is not needed.

lib/command.ts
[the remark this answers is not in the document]
srlowe: Right, changed now in line with your suggestions.

lib/app.ts
```diff
@@ -77,11 +77,13 @@ export function setMaxListeners(maxListeners: number) {
...
 
 	const settings = new CliSettings();
```
pdcastro, line 88: A few lines below this line in the `init()` function, there is a call to `notify()`:
```ts
	// check for CLI updates once a day
	(await import('./utils/update')).notify();
```
We should not make that call if `BALENARC_OFFLINE_MODE` is set.

lib/commands/config/write.ts
```diff
@@ -66,9 +66,8 @@ export default class ConfigWriteCmd extends Command {
...
-
 	public static root = true;
+	public static offlineCompatible = true;
```
pdcastro, line 70: Strictly speaking, `config read/write/inject` will only be `offlineCompatible` after #2369 is merged. Until then, some dependencies make calls to balenaCloud API endpoints to retrieve device type details that contain partition numbers. But I suppose there is no harm in setting `offlineCompatible` already now, so no need to change it.

pdcastro: #2369 has been merged, so that's no longer a concern.

pdcastro:
@balena-ci rebase

pdcastro (approved):

PR #2390, opened 2021-11-24
Delete 'doc/automated-init.md' and improve 'balena help device init'

pdcastro:
Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2391, opened 2021-11-25
Stop creating an extra provisioning API key in each config generation

thgreasi:
Change-type: patch
Changelog-entry: Avoid creating an extra provisioning API key in os configure & config generate
See: https://github.com/balena-io/balena-cli/pull/2294#discussion_r756499196

---

### DIFF
--- a/lib/utils/config.ts
+++ b/lib/utils/config.ts
@@ -35,6 +35,9 @@ export interface ImgConfig {
 	wifiKey?: string;
 	initialDeviceName?: string;
 
+	apiKey?: string;
+	deviceApiKey?: string;
+
 	// props for older OS versions
 	connectivity?: string;
 	files?: {
@@ -51,7 +54,7 @@ export interface ImgConfig {
 	};
 }
 
-export async function generateBaseConfig(
+export async function generateApplicationConfig(
 	application: BalenaSdk.Application,
 	options: {
 		version: string;
@@ -70,9 +73,7 @@ export async function generateBaseConfig(
 	const config = (await getBalenaSdk().models.os.getConfig(
 		application.slug,
 		options,
-	)) as ImgConfig & { apiKey?: string };
-	// os.getConfig always returns a config for an app
-	delete config.apiKey;
+	)) as ImgConfig;
 
 	// merge sshKeys to config, when they have been specified
 	if (options.os && options.os.sshKeys) {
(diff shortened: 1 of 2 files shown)


lib/utils/config.ts
```diff
@@ -122,9 +109,14 @@ export function generateDeviceConfig(
...
-				await addApplicationKey(config, application.id);
+				config.apiKey = await sdk.models.application.generateApiKey(
+					application.id,
+				);
 			} else {
-				await addDeviceKey(config, device.uuid, deviceApiKey || true);
+				config.deviceApiKey =
+					typeof deviceApiKey === 'string' && deviceApiKey
+						? deviceApiKey
+						: await sdk.models.device.generateDeviceKey(device.uuid);
```
pdcastro, line 119: Much better! 👍 
(What was that `deviceApiKey || true` logic?! Good riddance!) 😂

pdcastro (commented):
Looking great, 💯  thanks Thodoris!

lib/utils/config.ts
thgreasi: Yea right?
Ideally I wanted to do just:
```
config.deviceApiKey = deviceApiKey ?? await sdk.models.device.generateDeviceKey(device.uuid);
```
but that's not 100% how it worked before.
An empty string deviceApiKey is probably bad input though anyway, but I really can't guess whether anyone could do that.
Let me know if you prefer me to include the additional simplification or not.

pdcastro: I think the simplification is not possible without additional refactoring. `deviceApiKey` is declared as `deviceApiKey: string | true | undefined` and it may indeed have boolean value `true` to request a device key to be generated, for example in this call: https://github.com/balena-io/balena-cli/blob/v12.52.2/lib/commands/config/generate.ts#L215

I am not a fan of that kind of mixed type API/interface where a value can be a string or a boolean or undefined to trigger different behaviours, but that's how it was implemented and I think it would be low ROI to change it further in this PR.

The PR is good as it is, the important thing was to avoid the key duplication, and as a bonus a bunch of smelly code was deleted and simplified. 👍

thgreasi: I think it's fine to do that simplification though since we call it like:
```
config = await generateDeviceConfig(
  device,
  options.deviceApiKey || options['generate-device-api-key'] || undefined,
  answers,
);
```
and then the original implementation of `generateDeviceConfig()` was later doing:
```
await addDeviceKey(config, device.uuid, deviceApiKey || true);
```
which afaict means that undefiend/true/false had no difference for the deviceApiKey param, since the final `if (customDeviceApiKey === true) {` check would give the exact same result in all these cases.

Which leads us to the conclusion that the `--generate-device-api-key` option for `balena config generate` has really no effect...

lib/utils/config.ts
```diff
@@ -86,24 +87,7 @@ export async function generateBaseConfig(
...
-
-	if (semver.satisfies(options.version, '<2.7.8')) {
-		await addApplicationKey(config, application.id);
-	} else {
-		await addProvisioningKey(config, application.id);
-	}
-
-	return config;
-}
+export const generateApplicationConfig = generateBaseConfig;
```
nitishagar, line 90: @thgreasi Is the `config.apiKey` from `generateBaseConfig` sufficient? Or to put it another way, the version `2.7.8` bifurcation isn't necessary for `generateApplicationConfig`?

I don't have history on this version change, that is why the ask.

thgreasi: The API (and as a result `os.getConfig()`) already handles that case and creates the correct type of apiKey depending on the version.

thgreasi:
@balena-ci I self-certify!

thgreasi:
This took more that I thought to fully test...

PR #2393, opened 2021-11-25
Fix mistake in `config generate` examples

srlowe:
Change-type: patch


srlowe:
@balena-ci I self-certify!

pdcastro:
@balena-ci rebase

PR #2395, opened 2021-11-25
Set the correct target state when using Compose "image" field

### LINKED ISSUE #2394: balena push fails when "image" is used in conjunction with "build" in docker-compose
# Expected Behavior

I should be able to `balena push` a multicontainer project, where the `docker-compose.yml` has services that specify both the `image` and `build` keywords. 
If you're wondering why you would want to do this, if you specify `image` as well as `build`, then Compose names the built image with the value found in `image`.

# Actual Behavior

The `balena push` command will pack, send and build the containers as expected, but once the images have finished building, it gets stuck in a loop, repeatedly printing:
```
[Debug]   Device state not settled, retrying in 1000ms
[Debug]   Device state not settled, retrying in 1000ms
```
and occasionally outputting something like:
```
[Logs]    [11/25/2021, 4:15:57 PM] Installing service 'livechat local_image_livechat:latest'
```

When I pull the supervisor logs from the running device I can see the supervisor failing to find the correct image:
```
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]   Scheduling another update attempt in 32000ms due to failure:  Error: Failed to apply state transition steps. (HTTP code 404) no such container - No such image: local_image_livechat:latest  Steps:["start"]
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]         at fn (/usr/src/app/dist/app.js:6:8594)
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]       at runMicrotasks (<anonymous>)
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]       at processTicksAndRejections (internal/process/task_queues.js:97:5)
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]   Device state apply error Error: Failed to apply state transition steps. (HTTP code 404) no such container - No such image: local_image_livechat:latest  Steps:["start"]
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]         at fn (/usr/src/app/dist/app.js:6:8594)
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]       at runMicrotasks (<anonymous>)
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [error]       at processTicksAndRejections (internal/process/task_queues.js:97:5)
Nov 25 16:16:30 db0c73f balena-supervisor[3070]: [api]     GET /v2/state/status 200 - 26.868 ms
Nov 25 16:16:31 db0c73f balena
```

If I then run `balena images` on the device, I can see that there is a built image on the device using the same name I set in the `image` field in `docker-compose.yml`, but for some reason, the supervisor is not being instructed to use that image, instead hardcoding it to `local_image_$SERVICE_NAME`.

# Steps to Reproduce the Problem

I've created a repo to reproduce the problem, which you can see here https://github.com/LucianBuzzo/balena-cli-push-image-mvr:

1. [Clone the repo](https://github.com/LucianBuzzo/balena-cli-push-image-mvr)
1. Set up a device and put it in local mode
1. `balena push` the repo to the device

# Specifications

- **balena CLI version:** 12.48.4
- **Cloud backend: openBalena or balenaCloud?** balenaCloud
- **Operating system version:** Windows 10
- **32/64 bit OS and processor:** 64-bit operating system, x64-based AMD processor
- **Install method:** npm
- **If npm install, Node.js and npm version:** e.g.  node v12.15.0 (npm v6.13.4)

LucianBuzzo:
Fixes #2394

When pushing to a device in local mode, if a service is not external, and uses
an `image` field, that value should be used for tags and target state, otherwise
it won't match the image name generated on the device by balenaEngine.

Change-type: patch

### DIFF
--- a/lib/utils/device/deploy.ts
+++ b/lib/utils/device/deploy.ts
@@ -34,6 +34,7 @@ import {
 	loadProject,
 	makeBuildTasks,
 	tarDirectory,
+	makeImageName,
 } from '../compose_ts';
 import Logger = require('../logger');
 import { DeviceAPI, DeviceInfo } from './api';
@@ -44,6 +45,7 @@ import { stripIndent } from '../lazy';
 
 const LOCAL_APPNAME = 'localapp';
 const LOCAL_RELEASEHASH = 'localrelease';
+const LOCAL_PROJECT_NAME = 'local_image';
 
 // Define the logger here so the debug output
 // can be used everywhere
@@ -375,7 +377,11 @@ async function performBuilds(
 				// We can be sure that localImage.name is set here, because of the failure code above
 				const image = docker.getImage(localImage.name!);
 				await image.tag({
-					repo: generateImageName(localImage.serviceName),
+					repo: makeImageName(
+						LOCAL_PROJECT_NAME,
+						localImage.serviceName,
+						'latest',
+					),
 					force: true,
 				});
 				imagesToRemove.push(localImage.name!);
@@ -533,7 +539,7 @@ async function assignDockerBuildOpts(
 					'io.resin.local.image': '1',
 					'io.resin.local.service': task.serviceName,
 				},
-				t: generateImageName(task.serviceName),
+				t: getImageNameFromTask(task),
 				nocache: opts.nocache,
 				forcerm: true,
 				pull: opts.pull,
@@ -550,8 +556,10 @@ async function assignDockerBuildOpts(
 	);
 }
 
-function generateImageName(serviceName: string): string {
-	return `local_image_${serviceName}:latest`;
+function getImageNameFromTask(task: BuildTask): string {
+	return !task.external && task.tag
+		? task.tag
+		: makeImageName(LOCAL_PROJECT_NAME, task.serviceName, 'latest');
 }
 
 export function generateTargetState(
@@ -587,6 +595,8 @@ export function generateTargetState(
 			contract = keyedBuildTasks[name].contract;
 		}
 
+		const task = keyedBuildTasks[name];
+
 		services[idx] = {
 			...defaults,
 			...opts,
@@ -595,7 +605,7 @@ export function generateTargetState(
 				imageId: idx,
 				serviceName: name,
 				serviceId: idx,
-				image: generateImageName(name),
+				image: getImageNameFromTask(task),
 				running: true,
 			},
 		};


LucianBuzzo:
@pdcastro @srlowe Are there any test cases that verify that a service is running on a device after pushing? If so I could use it as a template for a test to verify this behaviour.

lib/utils/device/deploy.ts
```diff
@@ -554,6 +555,12 @@ function generateImageName(serviceName: string): string {
 	return `local_image_${serviceName}:latest`;
 }
 
+function getImageNameFromTask(task: BuildTask): string {
+	return !task.external && task.tag
+		? task.tag
+		: generateImageName(task.serviceName);
```
pdcastro, line 561: Nice. 👍   This PR reminded me of the `makeImageName` function defined at https://github.com/balena-io/balena-cli/blob/v12.53.2/lib/utils/compose_ts.ts#L420-L430
How about centralising the image name generation logic (cloud and device) by **deleting** the `generateImageName` function from this file and importing `makeImageName` instead? It looks like this file already imports from `compose_ts`, so it might be just a matter of updating the import statement at the top of the file, and then making the call:
```ts
    makeImageName('local_image', task.serviceName, 'latest')
```

pdcastro (commented):
> Are there any test cases that verify that a service is running on a device after pushing? If so I could use it as a template for a test to verify this behaviour.

I don't think we have such tests. We mock the supervisor in the following test case, but it doesn't sound like that's what you are after:
https://github.com/balena-io/balena-cli/blob/v12.53.2/tests/commands/logs.spec.ts#L32

If you've manually tested these changes, I would not have problems with the PR being merged without automated tests, but it's up to you.

LucianBuzzo:
@pdcastro I tested this pretty extensively, and it worked without issue. The behaviour is completely unchanged if you aren't using the combination of `image` and `build`, so I think it's safe to merge.

lib/utils/device/deploy.ts
LucianBuzzo: Oh, nice idea, I'll check it out.

LucianBuzzo: Done! 👍

karaxuna (approved):

PR #2398, opened 2021-11-27
os configure, local configure: Reuse disk partition scanning logic

pdcastro:
This PR deletes duplicate code between `balena-cli` and `balena-config-json`. The `balena config read/write/inject` commands were using the disk partition scanning logic from `balena-config-json`, whereas the `balena os configure` and `balena local configure` commands were using similar logic coded in `lib/utils/helpers.ts`. This PR makes all those commands use the logic from `balena-config-json`, and deletes the duplicate logic. This way, any bug fixes or improvements to the disk scanning logic are shared between the CLI and any other repos / packages that use `balena-config-json`. As a bonus, this also allowed deleting a couple of dependencies from `package.json`.

Connects-to: balena-io-modules/balena-config-json/pull/21
Connects-to: #2369
Connects-to: #2267
Connects-to: https://jel.ly.fish/improvement-modify-balena-config-inject-related-cli-commands-work-without-internet-connection-9ca2e56
Connects-to: https://jel.ly.fish/brainstorm-topic-8c32926f-b928-47a5-9e4f-d069cf06fc12

Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2400, opened 2021-11-29
Update to TypeScript version of balena-sync

LucianBuzzo:
Change-type: patch
Depends-on: balena-io-modules/balena-sync/pull/167

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@aws-crypto/ie11-detection": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/ie11-detection/-/ie11-detection-2.0.0.tgz",
-      "integrity": "sha512-pkVXf/dq6PITJ0jzYZ69VhL8VFOFoPZLZqtU/12SGnzYuJOOGNfF41q9GxdI1yqC8R13Rq3jOLKDFpUJFT5eTA==",
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/ie11-detection/-/ie11-detection-1.0.0.tgz",
+      "integrity": "sha512-kCKVhCF1oDxFYgQrxXmIrS5oaWulkvRcPz+QBDMsUr2crbF4VGgGT6+uQhSwJFdUAQ2A//Vq+uT83eJrkzFgXA==",
       "requires": {
         "tslib": "^1.11.1"
       },
@@ -20,14 +20,13 @@
       }
     },
     "@aws-crypto/sha256-browser": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-browser/-/sha256-browser-2.0.0.tgz",
-      "integrity": "sha512-rYXOQ8BFOaqMEHJrLHul/25ckWH6GTJtdLSajhlqGMx0PmSueAuvboCuZCTqEKlxR8CQOwRarxYMZZSYlhRA1A==",
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-browser/-/sha256-browser-1.1.0.tgz",
+      "integrity": "sha512-VIpuLRDonMAHgomrsm/zKbeXTnxpr4aHDQmS4pF+NcpvBp64l675yjGA9hyUYs/QJwBjUl8WqMjh9tIRgi85Sg==",
       "requires": {
-        "@aws-crypto/ie11-detection": "^2.0.0",
-        "@aws-crypto/sha256-js": "^2.0.0",
-        "@aws-crypto/supports-web-crypto": "^2.0.0",
-        "@aws-crypto/util": "^2.0.0",
+        "@aws-crypto/ie11-detection": "^1.0.0",
+        "@aws-crypto/sha256-js": "^1.1.0",
+        "@aws-crypto/supports-web-crypto": "^1.0.0",
         "@aws-sdk/types": "^3.1.0",
         "@aws-sdk/util-locate-window": "^3.0.0",
         "@aws-sdk/util-utf8-browser": "^3.0.0",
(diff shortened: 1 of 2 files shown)


LucianBuzzo:
@srlowe Can you help me understand why these tests are failing?

srlowe:
@LucianBuzzo Looking into it.

srlowe:
`npm run build:standalone` is failing - looks like the `pkg.scripts` section in `package.json` needs amending.  Will check the details.

LucianBuzzo:
@srlowe Why does the uppercasing break the standalone packaging? Surely this will be a problem with other modules that we don't control as well?

srlowe:
> @srlowe Why does the uppercasing break the standalone packaging? Surely this will be a problem with other modules that we don't control as well?

@LucianBuzzo It looks like pkg (https://github.com/vercel/pkg/) is unable to properly detect dependencies if the import casing does not match the package name casing.  I don't know of this being an issue before, perhaps @pdcastro has seen a case.

pdcastro:
> perhaps @pdcastro has seen a case

Possible, but I don't remember other examples. As Scott said, it looks like a situation where [pkg](https://www.npmjs.com/package/pkg) is looking for a module name (`jsonstream`) in a case-sensitive way, and not finding it. I am actually quite happy with the error message:

```
Error creating or testing standalone zip package
Error: "build:standalone": pkg output does not match expected output from "tests/test-data/pkg/expected-warnings-linux.txt"
Diff:
================================================================================
+ > Warning Cannot find module 'jsonstream' from 'node_modules/balena-sync/build'
+ %1: node_modules/balena-sync/build/docker-utils.js
+ 
```

Compared to how it was 1 or 2 years ago... Before we implemented this `diff` test and output, that pkg warning/error would go completely unnoticed (the build would succeed) and eventually end users of the standalone zip package would file bug reports that the CLI was misbehaving in mysterious ways and usually only on Linux (where the standalone zip package is used most). At least now the build fails quite early!

By the way, it is in my to-do list to replace `pkg` and the standalone zip package with [caxa](https://www.npmjs.com/package/caxa), which is basically a self extracting zip archive that also executes an "entrypoint command" after extraction, and which we could effectively use as a good CLI installer for Linux. We'll then stop using `pkg` altogether (all platforms) and this will free us of all the complexity and issues related to `pkg`, including this `JSONstream` issue which really only the tip of the iceberg. :-)  I've done quite a bit of successful testing with `caxa` but I have not yet managed to get it over the line, including adapting the CI which by now might mean using Transformers instead of extending the current solution based on `balena-concourse` (part of the issue being our desire to provide CLI installers for ARM as well).

LucianBuzzo:
@pdcastro @srlowe The WIP PR on balena-cli is now passing! I think this Is now good to go 👍

package.json
```diff
@@ -214,7 +214,7 @@
     "balena-semver": "^2.3.0",
     "balena-settings-client": "^4.0.7",
     "balena-settings-storage": "^7.0.0",
-    "balena-sync": "^11.0.2",
+    "balena-sync": "11.0.3-lucianbuzzo-typescript-32279716b412d16d5eadacf35d18f6b10996ea28",
```
pdcastro, line 217: balena-io-modules/balena-sync/pull/167 was merged, now this line can be updated.

LucianBuzzo: I've updated the dep and will now move this PR out of draft

npm-shrinkwrap.json
```diff
@@ -2833,6 +2833,14 @@
         "@types/node": "*"
       }
     },
+    "@types/rsync": {
```
pdcastro, line 2836: I think these new `@types/*` dependencies (coming from `balena-sync`) should not have found their way into the `balena-cli` `npm-shrinkwrap.json` file, because `balena-cli` does not need them in either dev or prod installs of the balena CLI (e.g. the CLI itself does not import package `'rsync'`, and therefore does not need to install `@types/rsync`). Added a comment to the `balena-sync` PR: https://github.com/balena-io-modules/balena-sync/pull/167/files#r768840867

I suggest a patch PR to `balena-io-modules/balena-sync` to change those dependencies from prod to dev, then amending `package.json` in this PR.

pdcastro:
@balena-ci rebase

npm-shrinkwrap.json
pdcastro: It is not a big deal though, and we could merge this PR as it is. However, I just learned that we need to release CLI v13 in the next few days ([FD thread](https://www.flowdock.com/app/rulemotion/i-cli/threads/hdlxb8LTwYxQ8os3tFciOvh0CwD)), and I was reasoning that it might now be best for us to avoid making any non-essential changes to CLI v12, because it would be trickier to apply fixes to CLI v12 after CLI v13 was out (better safe than sorry). Which is to say, I would suggest waiting for CLI v13 to be out (next week) before merging this PR.

pdcastro:
It looks like the CI is not able to rebase this PR automatically. CLI v13 has now been released (thanks for waiting!) and a manual rebase and force push should fix it.

npm-shrinkwrap.json
```diff
@@ -5738,6 +5672,15 @@
         }
       }
     },
+    "cpu-features": {
```
pdcastro: I think that this newly added `cpu-features` dependency is the cause of the following test error:
```
make: Entering directory '/tmp/build/c694c2a4/versioned-source/node_modules/cpu-features/build'
  ACTION Configuring dependencies /tmp/build/c694c2a4/versioned-source/node_modules/cpu-features/deps/cpu_features/build/Makefile
/bin/sh: 1: cmake: Permission denied
config_deps.target.mk:13: recipe for target '/tmp/build/c694c2a4/versioned-source/node_modules/cpu-features/deps/cpu_features/build/Makefile' failed
make: *** [/tmp/build/c694c2a4/versioned-source/node_modules/cpu-features/deps/cpu_features/build/Makefile] Error 127
make: Leaving directory '/tmp/build/c694c2a4/versioned-source/node_modules/cpu-features/build'
```
`cpu-features` is an ***optional*** dependency (`"optional": true,`) of the pre-existing `ssh2` dependency, so perhaps this `cpu-features` block could be simply deleted without harmful side effects?

npm-shrinkwrap.json
```diff
@@ -3728,9 +3646,9 @@
       "integrity": "sha512-zg7Hz2k5lI8kb7U32998pRRFin7zJlkfezGJjUc2heaD4Pw2wObakCDVzkKztTm/Ln7eiVvYsjqak0Ed4LkMDA=="
     },
     "aws4-axios": {
-      "version": "2.4.6",
-      "resolved": "https://registry.npmjs.org/aws4-axios/-/aws4-axios-2.4.6.tgz",
-      "integrity": "sha512-MipHf2VZYwl5DVTKUf9Be+3e20ukCknQbHVhVw+ann2UJ0KKa7XgZtAPnBjnlvrlNvomB1AYOQX0+5I/XqwPEg==",
+      "version": "2.3.1",
```
pdcastro: This `aws4-axios` downgrade is a case in point re my other comment (several downgraded modules). Why does `npm` think that it needs to be downgraded to v2.3.1? It is a dependency of `etcher-sdk@6.2.1`, which is not touched by this PR as far as I can tell.

npm-shrinkwrap.json
```diff
@@ -90,892 +88,821 @@
...
+        "@aws-sdk/url-parser-native": "3.10.0",
+        "@aws-sdk/util-base64-browser": "3.10.0",
+        "@aws-sdk/util-base64-node": "3.10.0",
+        "@aws-sdk/util-body-length-browser": "3.10.0",
+        "@aws-sdk/util-body-length-node": "3.10.0",
+        "@aws-sdk/util-user-agent-browser": "3.10.0",
+        "@aws-sdk/util-user-agent-node": "3.10.0",
+        "@aws-sdk/util-utf8-browser": "3.10.0",
+        "@aws-sdk/util-utf8-node": "3.10.0",
+        "tslib": "^2.0.0"
```
pdcastro: These and several other modules are being downgraded in this PR. But even after running the `git` and `npm` commands that I mention in another comment, this still happens. I don't understand why. 🤔  So far I was not able to pinpoint a (transitive) dependency that requires the downgrade of all these packages.

pdcastro (changes requested):
The tests are failing with a few different errors, I think the most important being:
```
Error: "test-shrinkwrap": /.../automation/test-lock-deduplicated.sh failed with exit code 1:
"/.../automation/test-lock-deduplicated.sh" [undefined]
```
I suspect that something went wrong when updating the `npm-shrinkwrap.json` while rebasing this PR on the master branch. I would usually solve this is by running:
```
$ npm --version
6.14.15
$ git checkout lucianbuzzo/upgrade-balena-sync-ts
$ git fetch -t origin master:master
$ rm -rf node_modules/ && npm i && npm dedupe && npm i
$ git add npm-shrinkwrap.json
$ git commit -n --amend --no-edit
$ git push --force-with-lease origin lucianbuzzo/upgrade-balena-sync-ts
```

The key commands above mean: _"given the `npm-shrinkrwap.json` file as it exists in the master branch, a blank slate `node_modules` folder and this PR's changes to the `package.json` file, make the strictly necessary changes to the `npm-shrinkwrap.json` file and deduplicate them"._

(The reason for the deleting the local `node_modules` folder is that `npm install` takes it into account when updating `npm-shrinkwrap.json`, but sometimes this is not desirable.)

I am submitting a "Request changes" review mainly because of the seemingly unnecessary upgrade to `balena-sdk` that I think requires some testing and would belong to some other PR (see other comments).

PR #2401, opened 2021-11-30
chore: Bump multicast-dns to rebased commit

klutchell:
Otherwise npm install fails due to the missing commit in npm-shrinkwrap.json

Change-type: patch
See: https://github.com/balena-io-modules/multicast-dns/pull/1

I think this is the correct solution, and I've made identical PRs on a number of other projects as well. Let me know if there's a better way to approach this.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -12252,7 +12252,7 @@
       "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w=="
     },
     "multicast-dns": {
-      "version": "git+https://github.com/resin-io-modules/multicast-dns.git#a15c63464eb43e8925b187ed5cb9de6892e8aacc",
+      "version": "git+https://github.com/resin-io-modules/multicast-dns.git#db98d68b79bbefc936b6799de9de1038ba49f85d",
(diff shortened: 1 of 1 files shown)


npm-shrinkwrap.json
```diff
@@ -12252,7 +12252,7 @@
       "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w=="
     },
     "multicast-dns": {
-      "version": "git+https://github.com/resin-io-modules/multicast-dns.git#a15c63464eb43e8925b187ed5cb9de6892e8aacc",
+      "version": "git+https://github.com/resin-io-modules/multicast-dns.git#db98d68b79bbefc936b6799de9de1038ba49f85d",
```
pdcastro, line 12255: Presumably this also affects older versions of the CLI (installing older versions with `npm install`), and it will break the CLI installation again if PR balena-io-modules/multicast-dns/pull/1 is rebased again.
@LucianBuzzo, now that we have learned this, I suggest creating another branch off `listen-on-all-interfaces` and a new PR (in the `multicast-dns` repo) to continue the work of PR balena-io-modules/multicast-dns/pull/1,  and leaving the old PR alone until the new PR is merged and downstream dependencies like `balena-cli` can be updated to use a new proper version of the repo (I suppose that's the plan).

pdcastro (approved):

Issue #2387, opened 2021-11-22
os download: should not require '.prod' suffix for ESR versions


pdcastro:
In the case of non-ESR images, `os download` defaults to production images when the `'.prod'` suffix is not added to the version string:

```
$ balena os download intel-nuc -o os.img --version 2.83.18+rev1
Getting device operating system for intel-nuc
Downloading balenaOS version 2.83.18+rev1.prod [                        ] 1% eta 2m28s

$ balena os download intel-nuc -o os.img --version 2.83.18+rev1.prod
Getting device operating system for intel-nuc
Downloading balenaOS version 2.83.18+rev1.prod [                        ] 1% eta 3m13s
```

But in the case of ESR images, it fails unless `'.prod'` is added:

```sh
$ balena os download intel-nuc -o os.img --version 2021.10.0
Getting device operating system for intel-nuc
BalenaRequestError: Request error: "Server error"

$ balena os download intel-nuc -o os.img --version 2021.10.0.prod
Getting device operating system for intel-nuc
Downloading balenaOS version 2021.10.0.prod [=                       ] 3% eta 1m19s
```

Affects CLI v12.52.0, which added support for ESR versions.

While at it: `balena os download` should also add ESR usage examples to the `balena help os download` output.

floion:
This also happened on rpi4 with OS version `2.88.4`. Needed to append `.prod` or `.dev` to get it download the OS image successfully.

pdcastro:
Resolved in CLI v12.55.3 or later.

PR #2404, opened 2021-12-09
os download: Assume '.prod' suffix by default for all balenaOS versions

pdcastro:
Also add a better error message when the user is not logged in and attempts to download a balenaOS ESR version.

See also: (restricted access) https://www.flowdock.com/app/rulemotion/resin-devices/threads/bI-Uhq-pxSdxgN5lxAsYpd8GCpX

Resolves: #2387
Change-type: patch


pdcastro:
@balena-ci retest

pdcastro:
@balena-ci I self-certify!

Issue #2386, opened 2021-11-22
os download, os versions: should support device type aliases


pdcastro:
```
$ balena os download nuc -o os.img --version menu
Getting device operating system for nuc
Error: No balenaOS versions found for device type 'nuc'.
Double-check the device type slug with 'balena devices supported'.

$ balena os download intel-nuc -o os.img --version menu
Getting device operating system for intel-nuc
? Select the OS version: (Use arrow keys)
❯ v2.83.18+rev1.prod (recommended)
...

$ balena os versions nuc
Error: No balenaOS versions found for device type 'nuc'.
Double-check the device type slug with 'balena devices supported'.

$ balena os versions intel-nuc
v2.83.18+rev1.prod (recommended)
...
```

I.e., 'nuc' is not supported, even though it is an alias for 'intel-nuc':

```
$ balena devices supported
SLUG                                   ALIASES                ARCH    NAME
intel-nuc                              nuc                    amd64   Intel NUC
...
```

In the case of `os versions`, the issue was introduced in CLI v12.52.0 (PR #2334).
In the case of `os download`, it pre-existed.

pdcastro:
Connects-to: #2005, #2334

pdcastro:
Resolved in CLI v12.55.4.

PR #2405, opened 2021-12-10
os download, os versions: Accept device type aliases

pdcastro:
Resolves: #2386
Change-type: patch

### DIFF
--- a/lib/utils/cloud.ts
+++ b/lib/utils/cloud.ts
@@ -107,6 +107,16 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
 	(_sdk, deviceUUID) => deviceUUID,
 );
 
+/** Given a device type alias like 'nuc', return the actual slug like 'intel-nuc'. */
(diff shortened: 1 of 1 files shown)


lib/utils/cloud.ts
```diff
@@ -202,15 +212,15 @@ async function resolveOSVersion(
...
-	if (version[0] === 'v') {
-		version = version.slice(1);
-	}
-	// The version must end with either '.dev' or '.prod', as expected
-	// by `balena-image-manager` and the balena SDK. Note that something
-	// like '2.88.4.prod' is not valid semver (https://semver.org/),
-	// so we don't even attempt semver parsing here.
-	if (!version.endsWith('.dev') && !version.endsWith('.prod')) {
-		version += '.prod';
+	if (/^v?\d+\.\d+\.\d+/.test(version)) {
```
pdcastro, line 215: Bug fix: prevent '.prod' suffix being added to version specifications such as 'latest', 'default', 'recommended'.

PR #2408, opened 2021-12-13
Test balena push with docker-compose using symlinks

thgreasi:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/i-cli/threads/Yo53CeVqtfSoo4Dw_qO8B0ZD2Nu

Resolves: #   
Depends-on: <url>   
See: <url>   

---


lib/utils/ignore.ts
```diff
@@ -227,7 +227,12 @@ async function listFiles(projectDir: string): Promise<string[]> {
 					const _files = await fs.readdir(dir, { withFileTypes: true });
 					for (const entry of _files) {
 						const fpath = path.join(dir, entry.name);
-						if (entry.isDirectory()) {
+						const isDirectory =
+							entry.isDirectory() ||
+							(entry.isSymbolicLink() &&
+								(await fs.stat(await fs.realpath(fpath))).isDirectory());
```
pdcastro, line 233: Is the `realpath()` call necessary? I understand that `fs.stat()` will (recursively) follow symlinks if `fpath` is a symlink (as opposed to `fs.lstat()`), so it would not be necessary to explicitly resolve the symlink with `realpath()`.

```suggestion
								(await fs.stat(fpath)).isDirectory());
```

tests/test-data/projects/docker-compose/symbolic-links/service1
```diff
@@ -0,0 +1 @@
+../symlink-fixtures/service1/
```
pdcastro, line 1: Build is failing on Windows with:
```
glob error [Error: EPERM: operation not permitted, scandir 'C:\var\lib\concourse\worker\volumes\live\a9a3a529-20a6-4d2c-748a-bf5059dd242b\volume\tests\test-data\projects\docker-compose\symbolic-links\service1'] {
  errno: -4048,
  code: 'EPERM',
  syscall: 'scandir',
  path: 'C:\\var\\lib\\concourse\\worker\\volumes\\live\\a9a3a529-20a6-4d2c-748a-bf5059dd242b\\volume\\tests\\test-data\\projects\\docker-compose\\symbolic-links\\service1'
}
```
It rings a bell that Windows doesn't play nicely with soft links, however out of the 5 exists tests that touch the `dockerignore2` test project that uses soft links, I think only one is currently being skipped on Windows:

* `push`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/commands/push.spec.ts#L355
* `push`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/commands/push.spec.ts#L412
* `tarDirectory`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/utils/tarDirectory.spec.ts#L92
* `tarDirectory`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/utils/tarDirectory.spec.ts#L147
* "live push": https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/utils/device/live.spec.ts#L230

So maybe some investigation on how come that 4 other tests don't need to be skipped on Windows (maybe they don't really exercise the soft link aspect?) before skipping this test on Windows as well.

lib/utils/ignore.ts
thgreasi: TIL that `fs.stat()` does follow symbolic links. Nice catch :+1:

tests/test-data/projects/docker-compose/symbolic-links/service1
pdcastro: > how come that 4 other tests don't need to be skipped on Windows

* `tarDirectory`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/utils/tarDirectory.spec.ts#L147

This is the one that is completely skipped on Windows:

```js
const itSkipWindows = process.platform === 'win32' ? it.skip : it;
itSkipWindows('should produce a compatible tar stream (symbolic links)'
```

* `push`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/commands/push.spec.ts#L355

This test is not skipped but it cheats :-) with:
```
			...(isWindows
				? {
						'lib/src-a.txt': { fileSize: 5, type: 'file' },
						'src/src-a.txt': { fileSize: 5, type: 'file' },
				  }
				: {}),
```

* `push`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/commands/push.spec.ts#L412
* `tarDirectory`: https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/utils/tarDirectory.spec.ts#L92
* "live push": https://github.com/balena-io/balena-cli/blob/v12.55.4/tests/utils/device/live.spec.ts#L230

I think these 3 tests pass on Windows because `git clone` on Windows makes a real copy of the file that is pointed to by the symlink.

So, I think it would be fair to skip this `test-data/projects/docker-compose/symbolic-links/service1` test case on Windows, for example with the `itSkipWindows` trick.

thgreasi: Looks like it's related to symlinks to folders on windows :thinking: 
I will skip it for now since I don't have the bandwith to further investigate this, and it already adds value as is.

thgreasi: Oh wait, this might actually be a balena-lint bug :)

pdcastro:
Build failed quickly for all platforms because of missing `npm run build` that calls `npm run lint` that modifies the source code and causes uncommitted code changes:
```
Error: "catch-uncommitted": npx failed with exit code 1:
```

tests/test-data/projects/docker-compose/symbolic-links/service1
pdcastro: I don't think so, the balena-lint issue was caused by the commit that added the `itNoWin` call.

thgreasi: @pdcastro I mean that the error in the node-cli check shows up right after the `npm run lint` and before any test actually runs
```
> balena-lint -e ts -e js --fix automation/ completion/ lib/ typings/ tests/ bin/balena bin/balena-dev gulpfile.js .mocharc.js .mocharc-standalone.js

glob error [Error: EPERM: operation not permitted, scandir 'C:\var\lib\concourse\worker\volumes\live\b07a1719-74ca-483e-6b51-980733460019\volume\tests\test-data\projects\docker-compose\symbolic-links\service1'] {
  errno: -4048,
  code: 'EPERM',
  syscall: 'scandir',
  path: 'C:\\var\\lib\\concourse\\worker\\volumes\\live\\b07a1719-74ca-483e-6b51-980733460019\\volume\\tests\\test-data\\projects\\docker-compose\\symbolic-links\\service1'
}
```

See: https://github.com/isaacs/node-glob/issues/284

pdcastro: Oh I see! 👍   Interestingly it doesn't happen for the other soft links under the tests folder (pre-existing test cases). Maybe because the other links point to files rather than directories? The error mentions `scandir`.
How to solve or get around this? 🤔

thgreasi: This might work, based on the comments in that issue of `glob`:
See: https://github.com/balena-io-modules/node-balena-lint/pull/82

thgreasi: Agrrrrr. I now fails is the `test:source` step.
Should I just drop the tests to get this fixed sooner than later?

thgreasi: See: https://github.com/balena-io/balena-cli/pull/2410

And I will leave this PR having only the test case, just in case someone has the patience to fix them for Windows...

pdcastro: Before dropping the tests, how about:
* Do not commit the `service1` soft link to the git repo.
* Have the new test case (added by this PR) dynamically create (and clean up) the soft link with `fs.symlink` that you had suggested in Flowdock.
* Probably re-add `itNoWin` to the test case.

thgreasi: During my research I saw ppl facing such issues when creating symlinks as well, but at this point it sounds more promising than the committed one. I will add a reminder for later this week.

Issue #2411, opened 2021-12-14
`balena push` should report the actual URL the release is being pushed to.


pranavpeshwe:
# Expected Behavior
When  `balena push` is performed, a message saying `Uploading source package ` is displayed on the console. The message should contain the correct (actual) builder URL depending on whatever has been configured or set in the environment variable (such as `BALENARC_BALENA_URL` for example).

# Actual Behavior
`balena push` displays a message saying `Uploading source package to balenaCloud` no matter what URL the release is being pushed to.

# Steps to Reproduce the Problem
Perform `balena push` to balena-staging or balenaMachine and observe the messages displayed on the terminal.

# Specifications
```
$ balena version -a
balena-cli version "12.52.0"
Node.js version "12.18.1"

$ neofetch --stdout | egrep 'OS|Host|Kernel'
OS: macOS 11.6 20G165 x86_64
Host: MacBookPro16,1
Kernel: 20.6.0
```
# Additional References
Balena internal discussion: https://www.flowdock.com/app/rulemotion/i-cli/threads/mhGLGAlzXFkqNuHkHooJQN-Bpz7

PR #2412, opened 2021-12-14
Removed hardcoded 'balenaCloud' in console message.

pranavpeshwe:
Change-type: patch
Signed-off-by: Pranav Peshwe <pranav@balena.io>

Resolves: #2411


lib/utils/remote-build.ts
```diff
@@ -405,8 +405,10 @@ async function getRemoteBuildStream(
 	// We only show the spinner when outputting to a tty
 	if (process.stdout.isTTY) {
 		const visuals = getVisuals();
+		const uploadUrl =
+			process.env.BALENARC_BUILDER_URL || `https://builder.${build.baseUrl}`;
 		uploadSpinner = new visuals.Spinner(
-			'Uploading source package to balenaCloud',
+			`Uploading source package to ${uploadUrl}`,
```
pranavpeshwe, line 411: This results in a log statement that looks for example like:
```
Uploading source package to https://builder.1ae7c2471111a3b24f64da7f8286014f.bob.local
```
when `BALENARC_BALENA_URL` is exported with value _1ae7c2471111a3b24f64da7f8286014f.bob.local_

lib/utils/remote-build.ts
```diff
@@ -405,8 +405,10 @@ async function getRemoteBuildStream(
 	// We only show the spinner when outputting to a tty
 	if (process.stdout.isTTY) {
 		const visuals = getVisuals();
+		const uploadUrl =
+			process.env.BALENARC_BUILDER_URL || `https://builder.${build.baseUrl}`;
```
pdcastro, line 409: Minor point: this duplicates the URL forming logic from the `getBuilderEndpoint` function, such that if someone changed either of them, they'd have to remember to change it in the other place too. A short alternative that comes to mind would be:
```js
    `Uploading source package to ${new URL(builderUrl).origin}`
```
Just a suggestion, no worries if you prefer to do it as you've coded it.

pdcastro (commented):
Regarding the build check failure:
```
No commits were annotated with a change type since version 12.55.6
```

What happens is that the relevant commit is missing a blank line before the `Change-type` line:

```
    Removed hardcoded 'balenaCloud' in console message.
    Change-type: patch
    Signed-off-by: Pranav Peshwe <pranav@balena.io>
```

It should instead be:

```
    Removed hardcoded 'balenaCloud' in console message.

    Change-type: patch
    Signed-off-by: Pranav Peshwe <pranav@balena.io>
```

While at it, by convention, we use the imperative in the commit messages (see [Changelog](https://github.com/balena-io/balena-cli/blob/master/CHANGELOG.md)), so it would be `Remove` rather than `Removed`, and we don't end it with a period. Also by convention, I would include the name of the affected CLI command for the benefit of anyone reading the Changelog, so:

```
    push: Remove hardcoded 'balenaCloud' in console message

    Change-type: patch
    Signed-off-by: Pranav Peshwe <pranav@balena.io>
```

Finally, I see the PR currently has two commits, one of them being a merge commit:

* Removed hardcoded 'balenaCloud' in console message.
* Merge branch 'master' into pp/issue-2411

That merge commit should not be there - please delete it. Refresh the master branch and rebase on it instead of merging. There several different way of achieving it (google "git rebase"), one my favourites is:

```
$ git checkout pp/issue-2411
$ git fetch origin master:master
$ git rebase master
```

After the above I expect the merge commit will disappear. If it is still there, do an "interactive rebase" (google it) and explicitly `drop` the merge commit.

pranavpeshwe:
Thanks much for inputs regarding the error and wording convention. Shall abide by the convention.
 
> Finally, I see the PR currently has two commits, one of them being a merge commit:
> 
> * Removed hardcoded 'balenaCloud' in console message.
> * Merge branch 'master' into pp/issue-2411
> 
> That merge commit should not be there - please delete it. Refresh the master branch and rebase on it instead of merging.

I thought just a `rebase` won't help (because local & remote will have different views of the same branch). Thought I should try before posting here and indeed I had to force the remote to accept my push.

lib/utils/remote-build.ts
pranavpeshwe: What you say makes perfect sense - such repetition is undesirable at best. The snippet you've suggested looks nice and I think I will go with it. Thanks much!
Do you think this change deserves adding a test case(s)? Willing to do that if required.

pdcastro:
> I thought just a `rebase` won't help (because local & remote will have different views of the same branch). Thought I should try before posting here and indeed I had to force the remote to accept my push.

Note the need for refreshing master before the rebase, otherwise indeed the remote master and the local master may differ.
Then, force push after rebasing is fine. If you wanted to be really careful, you could use `--force-with-lease`:

```
git push --force-with-lease origin pp/issue-2411
```

This causes the push to fail in case someone else (maybe a reviewer) contributed a commit to your PR, thus avoiding unintentionally overwriting someone else's changes. Not usually an issue though.

lib/utils/remote-build.ts
pdcastro: No test case required for this change in my opinion, but test it manually just to confirm that the output is still as expected.

pranavpeshwe:
Even with a refreshing master rebase wasn't enough.
And yes, as a matter of habit I too prefer `with-lease`, just in case.

pranavpeshwe:
Have updated (squashed) the commit(s) and altered the commit message as suggested. Feel free to suggest any other improvements (even nitpicks are welcome! :) ).

pdcastro:
@balena-ci retest

PR #2422, opened 2021-12-21
Include version info when installed on windows

Page-:
This allows the balena-cli version to show up in the control panel and also be available for `winget` when listing upgrades (right now it's `unknown` and always shows up in the upgrade list)


Page-:
@balena-ci I self-certify!

PR #2423, opened 2021-12-21
Update dev dependencies

Page-:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


lib/utils/compose.js
```diff
@@ -443,7 +443,10 @@ export class BuildProgressUI {
 		this._startTime = Date.now();
 	}
 
-	end(summary = null) {
+	/**
+	 * @param {Dictionary<string> | undefined} summary
```
Page-, line 447: This is to make it match the typing it ends up getting assigned to, the typescript update recognized that they were different (presumably it was defaulting to `any` before)

lib/utils/docker.ts
```diff
@@ -193,19 +193,7 @@ export async function createClient(
 	opts: dockerode.DockerOptions,
 ): Promise<dockerode> {
 	const Docker = await import('dockerode');
-	const docker = new Docker(opts);
-	const { modem } = docker;
-	// Workaround for a docker-modem 2.0.x bug where it sets a default
```
Page-, line 198: This was causing problems for typings with the @types/dockerode update and the issue it works around was fixed in https://github.com/apocas/docker-modem/commit/6c346d131f1d8d172909f5b4d517a6425d8d2282 and is definitely included as we're running 3.x

lib/utils/helpers.ts
```diff
@@ -89,10 +89,10 @@ export async function sudo(
 	await executeWithPrivileges(command, stderr, isCLIcmd);
 }
 
-export function runCommand<T>(commandArgs: string[]): Promise<T> {
+export async function runCommand<T>(commandArgs: string[]): Promise<T> {
 	const { isSubcommand } =
 		require('../preparser') as typeof import('../preparser');
-	if (isSubcommand(commandArgs)) {
+	if (await isSubcommand(commandArgs)) {
```
Page-, line 95: This was previously always truthy, the typescript update caught this bug but I wonder if the unconditional behavior might actually be necessary, @pdcastro ?

pdcastro: Maybe, needs testing. Currently focusing on CLI v13. cc: @srlowe

Page-: I checked and it's actually only used in 3 places and they're all sub commands, which explains why this was working before

Page-:
@balena-ci rebase

Page-:
@balena-ci I self-certify!

PR #2420, opened 2021-12-21
Release CLI v13

pdcastro:
See: https://www.flowdock.com/app/rulemotion/i-cli/threads/hdlxb8LTwYxQ8os3tFciOvh0CwD
Change-type: major

PR #2425, opened 2021-12-22
Fix some app/fleet terminology issues

srlowe:


pdcastro:
Commit moved from here to PR #2420 because the CI tests only run against master.
See also: https://www.flowdock.com/app/rulemotion/i-cli/threads/hdlxb8LTwYxQ8os3tFciOvh0CwD

PR #2429, opened 2021-12-24
os versions, os download: Replace deprecated version fields 

pdcastro:
Replace deprecated `rawVersion` and `formattedVersion` fields and use alternative overload of `getAvailableOsVersions`. As a result, the word 'recommended' is no longer printed next to any OS versions. The SDK was simply always labelling  the `'latest'` version `'recommended'`. Using the latest OS version is just common sense anyway, except for very conservative users who might prefer using a slightly older version (?) and who might question the recommendation.

See: https://www.flowdock.com/app/rulemotion/r-architecture/threads/YN4RClMOz6v6W-VCfoqBaNoliFZ
Change-type: patch

### DIFF
--- a/lib/utils/cloud.ts
+++ b/lib/utils/cloud.ts
@@ -225,53 +225,45 @@ async function selectOSVersionFromMenu(
 	deviceType: string,
 	esr: boolean,
 ): Promise<string> {
-	const vs = await getFormattedOsVersions(deviceType, esr);
+	const vs = await getOsVersions(deviceType, esr);
 
 	const choices = vs.map((v) => ({
-		value: v.rawVersion,
-		name: v.formattedVersion,
+		value: v.raw_version,
+		name: formatOsVersion(v),
 	}));
 
 	return getCliForm().ask({
 		message: 'Select the OS version:',
 		type: 'list',
 		choices,
-		default: (vs.find((v) => v.isRecommended) ?? vs[0])?.rawVersion,
+		default: vs[0]?.raw_version,
 	});
 }
 
 /**
- * Return the output of sdk.models.hostapp.getAvailableOsVersions(), filtered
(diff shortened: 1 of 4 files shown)


lib/utils/cloud.ts
```diff
@@ -280,3 +272,9 @@ export async function getFormattedOsVersions(
 	}
 	return versions;
 }
+
+export function formatOsVersion(osVersion: SDK.OsVersion): string {
+	return osVersion.line
+		? `v${osVersion.raw_version} (${osVersion.line})`
+		: `v${osVersion.raw_version}`;
```
pdcastro, line 279: Before this PR:
```
$ balena os versions raspberrypi4-64
v2.88.4+rev0.prod (recommended)
v2.88.4+rev0.dev
v2.87.16+rev1.prod
...

$ balena os versions --esr raspberrypi4-64
v2021.10.2.prod (next, recommended)
v2021.10.2.dev (next)
v2021.07.1.prod (current)
...
```
After this PR:
```
 $ balena os versions raspberrypi4-64
v2.88.4+rev0.prod
v2.88.4+rev0.dev
v2.87.16+rev1.prod
...

$ balena os versions --esr raspberrypi4-64
v2021.10.2.prod (next)
v2021.10.2.dev (next)
v2021.07.1.prod (current)
...
```

pdcastro:
@balena-ci I self-certify!

lib/utils/cloud.ts
pdcastro: A note about this was added to the [v13 release notes](https://github.com/balena-io/balena-cli/wiki/CLI-v13-Release-Notes) as a potentially/arguably breaking change. (This PR was released as version 13.0.1 in less than 24 hours after the release of 13.0.0. Extended [five-second rule](https://en.wikipedia.org/wiki/Five-second_rule) applies?) 😁

PR #2432, opened 2022-01-03
Convert lib/utils/compose to typescript

Page-:
This is focused on getting one of the last pieces of javascript converted to typescript, I'm leaving the merging of `compose_types`/`compose_ts`/`compose` for a future exercise, but at least it'll be significantly easier due to using a single language


Page-:
@balena-ci I self-certify!

PR #2434, opened 2022-01-04
Build: switch from using inline-source via gulp to using it directly

Page-:
This allows us to remove gulp and associated dependencies whilst generating identical output


package.json
```diff
@@ -174,12 +174,12 @@
     "ent": "^2.2.0",
     "filehound": "^1.17.5",
     "fs-extra": "^9.1.0",
-    "gulp": "^4.0.2",
-    "gulp-inline-source": "^4.0.0",
     "http-proxy": "^1.18.1",
     "husky": "^4.3.8",
+    "inline-source-cli": "^2.0.0",
```
pdcastro, line 179: Nice to finally get rid of gulp 🎉  I wonder if we even actually need the inline source functionality. It would require some testing, but if it is just for the `error.ejs` and `success.ejs` files, I think they are only used by the `balena login` command and performance in loading those pages isn't really a concern. (Is the inlining just for performance though?) Either way, good job. 👍

Page-: I have no idea if it's for performance or not, having it be a single page does remove any possible worries about relative links/whatever on localhost though

Page-:
@balena-ci I self-certify!

PR #2438, opened 2022-01-05
Migrate tabular commands to new output framework

srlowe:
This PR updates all the commands that output tabular data to use the new output framework (behind v14 flag).  The output framework has been updated to include extra features that we needed in the commands. You can test v14 using the global `--v-next` flag.

**A few other things:**

`balena release` has highlighted an output artifact that we probably should discuss how to solve (regarding formatting of multi-line values) - see comments.

`balena key` output has been modified a bit - it's now IMO a bit cleaner

`balena os versions` is not included, since it requires some discussion.

We still have the issue with oclif's buggy json output.  I've been looking into it, and it might be difficult to solve without more extensive changes to the oclif code (though I haven't reached a conclusion yet).  If it turns out to be especially problematic, one options is to add our own json output implementation to the output framework (which will probably mean temporarily excluding other output flags when using the json flag).  This only affects multi-record output.   All output changes are behind the v14 flag though, so it's safe to merge this in the mean time.

There's currently an unresolved issue with the `hideNullOrUndefinedValues` option, due to the way oclif structures it's output.  (see skipped tests).


lib/command.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2020 Balena Ltd.
+ * Copyright 2016-2022 Balena Ltd.
```
srlowe, line 3: I've set the copyright notices to be the same across the board, since there doesn't seem to be any value in tracking individual start dates.

lib/command.ts
```diff
@@ -171,4 +171,5 @@ export default abstract class BalenaCommand extends Command {
 
 	protected outputMessage = output.outputMessage;
 	protected outputData = output.outputData;
+	protected printTitle = output.printTitle;
```
srlowe, line 174: This allows commands to output a title explicitly (in the same format we get in the tabular data).

lib/commands/envs.ts
```diff
@@ -181,24 +184,59 @@ export default class EnvsCmd extends Command {
...
-			fields.push(options.json ? 'serviceName' : 'serviceName => SERVICE');
-		}
+		if (isV14()) {
+			const results = [...varArray] as any;
 
-		if (options.json) {
-			const { pickAndRename } = await import('../utils/helpers');
-			const mapped = varArray.map((o) => pickAndRename(o, fields));
-			this.log(JSON.stringify(mapped, null, 4));
+			// Rename fields
```
srlowe, line 190: Dropping support for the old cli-visuals field renaming syntax required a bit more code here, but it's only used in a couple of places in the codebase.

lib/framework/output.ts
```diff
@@ -1,33 +1,50 @@
...
 export interface DataSetOutputOptions extends DataOutputOptions {
 	filter?: string;
 	'no-header'?: boolean;
 	'no-truncate'?: boolean;
 	sort?: string;
 }
 
+// Not exposed to user
+export interface InternalOutputOptions {
+	displayNullValuesAs?: string;
```
srlowe, line 42: ..lets us specify how null/undefined values should be displayed.

lib/preparser.ts
```diff
@@ -53,6 +53,12 @@ export async function preparseArgs(argv: string[]): Promise<string[]> {
 		if (extractBooleanFlag(cmdSlice, '--debug')) {
 			process.env.DEBUG = '1';
 		}
+		// support global --v-next flag
```
srlowe, line 56: We now have a `--v-next` that can be used anywhere, similar to `--debug`, which makes new major version testing much more convenient.

lib/commands/key/index.ts
```diff
@@ -1,6 +1,6 @@
 /**
```
srlowe, line 1: I removed the ID from the output, since you have to provide the ID in the command params it's redundant and just clutters the output.

Old:
```
ID:   155617
NAME: GitHub Key 3

AAAAB3NzaC1yc2EAAAADAQABAAABAQCwUlR+9FZuQnctoUnhr/Gud+cR10vVg7GaB3C2uukezPFrAvYBBNwFX5E5ICcCFJh6q/75kThgsgSTh8y1cQIZqlStyh3vPvQ8VL1BcnJW/ZXecMrCPyup8CDnvqTNDJsVpm6yy1TgZCWDsHBA19uPnMn3gNoehDynAnZ8QVXDS+KuGMq3WN8e9J10OxLLnIJcM5y7TXdrNAKm+GODCAK2LEnD5MqQfY9Pzv1Zcmmlvp4y34XjoDK38z/C97NLW92MtCqAERPZZBnPypoBhWh2G3we/JTQwh3xgffMgF+VAxRFvnhdZLFvFudcxl0wxPG4uL1+yo4ph7SgIHyUxkv9
```

New (without Id)
```
 GitHub Key 3
 ────────────
AAAAB3NzaC1yc2EAAAADAQABAAABAQCwUlR+9FZuQnctoUnhr/Gud+cR10vVg7GaB3C2uukezPFrAvYBBNwFX5E5ICcCFJh6q/75kThgsgSTh8y1cQIZqlStyh3vPvQ8VL1BcnJW/ZXecMrCPyup8CDnvqTNDJsVpm6yy1TgZCWDsHBA19uPnMn3gNoehDynAnZ8QVXDS+KuGMq3WN8e9J10OxLLnIJcM5y7TXdrNAKm+GODCAK2LEnD5MqQfY9Pzv1Zcmmlvp4y34XjoDK38z/C97NLW92MtCqAERPZZBnPypoBhWh2G3we/JTQwh3xgffMgF+VAxRFvnhdZLFvFudcxl0wxPG4uL1+yo4ph7SgIHyUxkv9
```

lib/commands/release/index.ts
```diff
@@ -1,6 +1,6 @@
 /**
```
srlowe, line 1: The output framework currently doesn't display multi-line values very nicely (see below).  I think our options here are to either: 

1. copy current behaviour, and indent subsequent lines to same column (Looks nice, but may hinder copy/paste)
2. add newline before outputting multi-line values (better copy/paste support)

Current:
```
IS FINAL:        true
BUILD LOG:       [Info]     Starting build for xxx, user xx
                 [Info]     Dashboard link: https://dashboard.balena-cloud.com/apps/xxx/devices
                 [Info]     Building on arm01
```
New framework (needs changing):
```
 Is final        : true
 Build log       : [Info]     Starting build for xxx, user xx
[Info]     Dashboard link: https://dashboard.balena-cloud.com/apps/xxx/devices
[Info]     Building on arm01
```

lib/commands/envs.ts
pdcastro: I was going to suggest just dropping the renaming, like I think you had suggested before, as it is more friendly to code autogeneration and this change only applies to v14 (OK to be a breaking change).
But then I tested it and realised that `cli-ux` does not play well with camelCase, e.g. `deviceUUID` became `Deviceuuid` in the table header:
```
$ ./bin/balena-dev envs -d a459e1677ba853e10af2f655faaca503
 Id      Name          Value            Fleet           Deviceuuid           Service
...
```
I think this means we need to add some code somewhere for `cli-ux` (or the CLI codebase) to automatically convert camelCase to spaces, e.g. `deviceUUID` becomes `Device UUID`. And if we were to do that, then we might as well consider (optional) copying the existing `pickAndRename` logic (`helpers.ts`) that understands renaming specifications like 'deviceUUID => DEVICE'.

What do you think?

pdcastro (commented):
Good work! 👍 
Submitting some initial comments, will probably add some more later.

lib/commands/device/index.ts
```diff
@@ -163,37 +168,52 @@ export default class DeviceCmd extends Command {
...
+			'memory_total_mb',
+			'memory_usage_percent',
+			'storage_block_device',
+			'storage_usage_mb',
+			'storage_total_mb',
+			'storage_usage_percent',
+			'undervoltage_detected',
+		];
+
+		if (isV14()) {
```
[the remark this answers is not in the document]
pdcastro: It turns out that the master branch was broken for v14, and I've fixed it in PR #2441.
After this PR is rebased on master, let's try running the tests for v14:
```
$ export BALENA_CLI_VERSION_OVERRIDE='14.0.0'
$ npm test
```

pdcastro (commented):
> Submitting some initial comments, will probably add some more later.

Done, no more comments pending. :-)

lib/command.ts
srlowe: Right, just using the first date sounds sensible.

lib/commands/envs.ts
srlowe: Yeah ok, I think dropping the renaming is best if it works.  Adding that camelCase processing will be quite straightforward.  I think perhaps it would be best not to carry over the old conventions (at least initially) to keep things lean.  Since we don't do renames much, I think renaming manually in the logic is a bit more explicit.  It might also push us to modify the model as needed rather than patching in the view.  We can always add support for the rename conventions later (though the pickAndRename logic won't be a drop-in, I did take a look at it and it will need some modification).

lib/commands/release/index.ts
srlowe: Ok, let's go with option 2 then.

lib/commands/device/index.ts
```diff
@@ -163,37 +168,52 @@ export default class DeviceCmd extends Command {
...
+			});
+		} else {
+			// Old output implementation
+			outputFields.unshift(`$${device.device_name}$`);
+			console.log(
+				getVisuals().table.vertical(
+					device,
+					outputFields.filter((f) => f !== 'device_name'),
+				),
+			);
```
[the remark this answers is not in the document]
srlowe: Ok I'll take a look at that.

lib/commands/devices/index.ts
```diff
@@ -21,19 +21,21 @@ import * as cf from '../../utils/common-flags';
 import { expandForAppName } from '../../utils/helpers';
 import { getBalenaSdk, getVisuals, stripIndent } from '../../utils/lazy';
 import { applicationIdInfo, jsonInfo } from '../../utils/messages';
+import { type Application } from 'balena-sdk';
+import { type DataSetOutputOptions } from '../../framework';
```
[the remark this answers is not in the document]
srlowe: Yeah.  In one place it looked tidier, so I carried on with it.   But probably we should keep it as it was I guess.

lib/commands/devices/index.ts
```diff
@@ -99,31 +100,52 @@ export default class DevicesCmd extends Command {
...
-			'os_version',
-			'dashboard_url',
-		];
-
-		if (options.json) {
-			const { pickAndRename } = await import('../../utils/helpers');
-			const mapped = devices.map((device) => pickAndRename(device, fields));
-			console.log(JSON.stringify(mapped, null, 4));
+		if (isV14()) {
+			const outputFields = [
```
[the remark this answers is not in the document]
srlowe: Ok.  My strategy was to modify the old code as little as possible, even if this led to code duplication (since any duplication would only be temporary).

lib/framework/output.ts
```diff
@@ -125,34 +151,91 @@ async function outputDataItem(
...
+ * @param data The data object to process
+ * @param options Output options
+ */
+function processNullValues(data: any, options: InternalOutputOptions) {
+	Object.entries(data).forEach(([k, v]) => {
+		if (v == null) {
+			if (options.displayNullValuesAs) {
+				data[k] = options.displayNullValuesAs;
+			} else if (options.hideNullOrUndefinedValues) {
+				delete data[k];
```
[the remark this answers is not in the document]
srlowe: Yeah this crossed my mind too, but as you say, we would not expect to use the data after the output logic was called, so I didn't worry about it too much.  But you're right - it would be better to not modify the input.

lib/commands/device/index.ts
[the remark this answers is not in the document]
srlowe: Right thanks - Looks like I do need to make some modifications in the tests for v14 support.

LucianBuzzo:
@srlowe Why is this change hidden behind an `is v14` check?

srlowe:
> Why is this change hidden behind an `is v14` check?

@LucianBuzzo Because it's a breaking change that will require a new major version, and this is how we have handled major versions to date.

LucianBuzzo:
@srlowe Right, so do you consider any change in visual output a breaking change?

srlowe:
> Right, so do you consider any change in visual output a breaking change?

@LucianBuzzo Paulo has quite strong views on not breaking things for people who may be scripting against the CLI, so we've agreed a joint approach that significant output changes would only happen during major releases.  (And smaller changes on a case by case basis).

LucianBuzzo:
@srlowe Have you discussed supporting a --json flag for commands to support users who may script against the CLI, allowing more freedom in the visual presentation? Right now it's analogous to not changing a website in case people are scraping it.

srlowe:
@LucianBuzzo Right, one of the things this PR adds is a standard --json flag across all commands that output structured data.  So this should make scripting much easier and frees us from worrying about changes to the presentation so much.

pdcastro:
> Right now it's analogous to not changing a website in case people are scraping it.

I don't think this is a fair analogy, especially in relation to the `--json` option and other data output formatting options. Web dashboards are not generally considered scripting tools and scraping them is very much "at your own risk", but command line interfaces are often used to automate tasks, piping stdin / stderr / stdout in shell scripts.

I think it is reasonable to use major version releases for "breaking changes" to data output format, but this should not prevent improvements from being made. Feature switches like `isV14()` are a way of grouping breaking changes to be released together. We could/should also release major versions more often than we have historically done. I would not have any problems with a new major version every month, and this could even avoid the need of feature switches in some cases, like I suppose this PR. (Cameron used to object to frequent major version releases, but he's not around anymore.) :-)

We have a [GDoc](https://docs.google.com/document/d/1uHE-_LLu5gJUQZICQJtlo76EQCm90z0CMAwpkfw6s8w/edit) where we keep track of a "wish list" of changes that we would like to group together for the next major. Perhaps we could move that GDoc to a public wiki page in this repo.

> Paulo has quite strong views on not breaking things for people who may be scripting against the CLI,

It is true :-) and it only comes up here because I was asked to wear the loosely defined hat of "balena CLI maintainer" -- alongside Scott -- which makes me feel responsible for what is done to the balena CLI. Take that hat away from me, make the balena CLI "someone else's problem", and my views will not be nearly as strong. :-)

LucianBuzzo:
Thanks for the response @pdcastro . If every command supported `--json` and we made guarantees about the exit codes do you think that would be enough to support users scripting against the CLI, whilst decoupling data from aesthetics?

pdcastro:
> If every command supported `--json` and we made guarantees about the exit codes do you think that would be enough to support users scripting against the CLI, whilst decoupling data from aesthetics?

That would go a long way, and I think this PR delivers some of it. Other considerations regarding scripting are mentioned in #1089 - e.g. _when in "non-interactive mode",_ avoid colours and unicorns that make it harder for a script to determine whether an operation was successful, fail with an exit code instead of interactively asking for missing or ambiguous input\* (which either makes scripting significantly more complex, or causes script to hang while CLI waits on stdin), better separation of stdout and stderr as per [traditional CLI conventions/guidelines](https://clig.dev/), avoiding debugging messages when `--json` is used (#2447).

Most of the above is the opposite of what would be considered user friendly if a person was interactively running CLI commands, which I think makes it necessary for the balena CLI to have two modes of operation: interactive mode and non-interactive (or scripting) mode. Both modes "reduce user friction", just in different use case scenarios.

ab77:
@balena-ci retest

pdcastro:
@balena-ci rebase

PR #2441, opened 2022-01-08
v14 preparations: Fix TypeError produced by 'npx oclif manifest'

pdcastro:
Before this PR:
```
$ export BALENA_CLI_VERSION_OVERRIDE='14.0.0'
$ npx tsc && npx oclif manifest
    TypeError Plugin: balena-cli: Cannot read property 'type' of undefined
```

... which are part of `npm run build`. I.e., the build was broken when the next major (v14) feature switch was turned on.

Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2444, opened 2022-01-14
patch: Tell user that balena leave command does not remove the device…

taromurao:
Tell user that balena leave command does not remove the device, but take it out of balena cloud's controll
Change-type: patch

### DIFF
--- a/lib/utils/promote.ts
+++ b/lib/utils/promote.ts
@@ -80,7 +80,12 @@ export async function leave(
 	logger.logDebug('Deconfiguring...');
 	await deconfigure(deviceHostnameOrIp);
 
-	logger.logSuccess('Device successfully left the platform.');
+	logger.logSuccess(stripIndent`
+		Device successfully left the platform. The device will still be listed as part
+		of the fleet, but changes to the fleet will no longer affect the device and its
+		status will eventually be reported as 'Offline'. To irrecoverably delete the
+		device from the fleet, use the 'balena device rm' command or delete it through
+		the balenaCloud web dashboard.`);
 }
 
 async function execCommand(


lib/utils/promote.ts
```diff
@@ -80,7 +80,10 @@ export async function leave(
 	logger.logDebug('Deconfiguring...');
 	await deconfigure(deviceHostnameOrIp);
 
-	logger.logSuccess('Device successfully left the platform.');
+	logger.logSuccess(
+		'Device successfully left the platform. The device will still be visible in the fleet, \
+		but changes to the fleet will no longer affect this device (Heartbeat only).',
+	);
```
pdcastro, line 86: This prints without a line break on my console, with a lot of blank spaces in the middle of a single line:
```
$ ./bin/balena-dev leave 192.168.1.36
...
[Success] Device successfully left the platform. The device will still be visible in the fleet, 		but changes to the fleet will no longer affect this device (Heartbeat only).
```

Suggestion:

```suggestion
	logger.logSuccess(stripIndent`
		Device successfully left the platform. The device will still be listed as part
		of the fleet, but changes to the fleet will no longer affect the device and its
		status will eventually be reported as 'Offline'. To irrecoverably delete the
		device from the fleet, use the 'balena device rm' command or delete it through
		the balenaCloud web dashboard.`);
```
I chose "listed" instead of "visible" because openBalena does not have a web dashboard and I think that "listed" is a more neutral term between the CLI and the web dashboard. I believe there are discussions regarding changing the "heartbeat only" and "VPN only" statuses, so I suggest "eventually offline" because "offline" is less likely to change as a status.

taromurao:
@balena-ci I self-certify!

joshbwlng:
@balena-ci republish

Issue #2445, opened 2022-01-16
Building from source: ffi-napi requires node >= 12.8


bernhardkaindl:
On Ubuntu 20.04 an Node 10 or older errors like these will occur:
```
  CXX(target) Release/obj.target/ffi_bindings/src/ffi.o
In file included from /home/balena/balena-cli/node_modules/ffi-napi/node_modules/node-addon-api/napi.h:2725,
                 from ../src/ffi.h:2,
                 from ../src/ffi.cc:3:
/home/balena/balena-cli/node_modules/ffi-napi/node_modules/node-addon-api/napi-inl.h: In member function ‘T* Napi::Env::GetInstanceData()’:
/home/balena/balena-cli/node_modules/ffi-napi/node_modules/node-addon-api/napi-inl.h:471:24: error: there are no arguments to ‘napi_get_instance_data’ that depend on a template parameter, so a declaration of ‘napi_get_instance_data’ must be available [-fpermissive]
  471 |   napi_status status = napi_get_instance_data(_env, &data);
      |                        ^~~~~~~~~~~~~~~~~~~~~~
/home/balena/balena-cli/node_modules/ffi-napi/node_modules/node-addon-api/napi-inl.h:471:24: note: (if you use ‘-fpermissive’, G++ will accept your code, but allowing the use of an undeclared name is deprecated)
../src/ffi.cc: In static member function ‘static FFI::InstanceData* FFI::InstanceData::Get(Napi::Env)’:
../src/ffi.cc:11:24: error: ‘napi_get_instance_data’ was not declared in this scope; did you mean ‘napi_new_instance’?
   11 |   napi_status status = napi_get_instance_data(env, &d);
      |                        ^~~~~~~~~~~~~~~~~~~~~~
      |                        napi_new_instance
../src/ffi.cc: In function ‘Napi::Object BindingHook(Napi::Env, Napi::Object)’:
../src/ffi.cc:348:24: error: ‘napi_set_instance_data’ was not declared in this scope; did you mean ‘napi_new_instance’?
  348 |   napi_status status = napi_set_instance_data(
      |                        ^~~~~~~~~~~~~~~~~~~~~~
      |                        napi_new_instance
make: *** [ffi_bindings.target.mk:116: Release/obj.target/ffi_bindings/src/ffi.o] Error 1
make: Leaving directory '/home/balena/balena-cli/node_modules/ffi-napi/build'
gyp ERR! build error 
[...]

PR #2446, opened 2022-01-16
Update docs and package.json re min Node.js supported version (12.8.0)

pdcastro:
Resolves: #2445
Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2450, opened 2022-01-19
chore: Update @types/node to v12.20.42

pdcastro:
Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2452, opened 2022-01-28
added balena ssh troubleshoot

inwenc:
Resolves: added possible troubleshooting solution in troubleshooting.md for ssh error

See: https://www.flowdock.com/app/rulemotion/r-beginners/threads/OnmNsYyN00_HEnFpterKJ9mGvUm

---

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -84,7 +84,7 @@ $ sudo chown -R <user> $HOME/.balena
 Users sometimes come across broken line wrapping or cursor behavior in text terminals, for example
 when long command lines are typed in a `balena ssh` session, or when using text editors like `vim`
 or `nano`. This is not something specific to the balena CLI, being also a commonly reported issue
-with standard remote terminal tools like `ssh` or `telnet`.  It is often a remote shell
+with standard remote terminal tools like `ssh` or `telnet`. It is often a remote shell
 configuration issue (files like `/etc/profile`, `~/.bash_profile`, `~/.bash_login`, `~/.profile`
 and the like on the remote machine), including UTF-8 misconfiguration, the use of unsupported ASCII
 control characters in shell prompt formatting (e.g. the `$PS1` env var) or the output of tools or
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -101,16 +101,32 @@ bind 'set horizontal-scroll-mode off'
...
+- Mac: Terminal, iTerm2
+- Windows: PowerShell, PuTTY, WSL (Windows Subsystem for Linux)
+
+## "balena ssh" ends with permission denied when using RSA keys
+
+RSA keys generated with this command from github documentation: `ssh-keygen -t rsa -b 4096 -C "your_email@example.com"`
+
+It failed with this error:
+`Offering public key: /Users/<username>/.ssh/id_rsa RSA`
+`send_pubkey_test: no mutual signature algorithm`
```
pdcastro, line 123: I gather that GitHub's documentation no longer advises `-t rsa`:
https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

For this reason, I suggest not mentioning GitHub here:
```suggestion
Some operating systems or SSH packages have started disabling the RSA key type on security grounds. As a result, `balena ssh -v` may fail with error:
`Offering public key: /Users/<username>/.ssh/id_rsa RSA`
`send_pubkey_test: no mutual signature algorithm`
```

TROUBLESHOOTING.md
```diff
@@ -101,16 +101,32 @@ bind 'set horizontal-scroll-mode off'
...
+
+Possible Solutions
+
+- Either add PubkeyAcceptedKeyTypes +ssh-rsa to SSH configuration or use a different algorithm for the key (ecdsa or ed25519 should work fine)
+
+- Change `.ssh/config` to refer to the custom named identify file? e.g. id_balena / id_balena.pub
```
pdcastro, line 129: * Did you mean a comma or a colon instead of a question mark? I.e. `file?` → `file,`
* Typo: `identify file` → `identity file`
* How does a custom identity file help? I guess that the custom identity file could not be RSA, and would have to be ecdsa or ed25519, just like the primary / default identity file, right? If so, we may want to clarify that a custom identity file may be helpful if the user wishes to keep their their primary / default RSA identity file unchanged, and create an alternative file with a different key type for use with balena.
  * I also think we would need to provide an example of how the `Host` section should look like in the `.ssh/config` file. What value should the user use for `Host`? Does it work with `Host *.balena`? I assume that the custom identity file configuration would have to fall under some such `Host` configuration. Otherwise, it would effectively replace the primary / default identity file for all hosts, not just balena.

TROUBLESHOOTING.md
```diff
@@ -101,16 +101,32 @@ bind 'set horizontal-scroll-mode off'
...
+
+RSA keys generated with this command from github documentation: `ssh-keygen -t rsa -b 4096 -C "your_email@example.com"`
+
+It failed with this error:
+`Offering public key: /Users/<username>/.ssh/id_rsa RSA`
+`send_pubkey_test: no mutual signature algorithm`
+
+Possible Solutions
+
+- Either add PubkeyAcceptedKeyTypes +ssh-rsa to SSH configuration or use a different algorithm for the key (ecdsa or ed25519 should work fine)
```
pdcastro, line 127: Adding backticks around `PubkeyAcceptedKeyTypes +ssh-rsa` would look nice.

pdcastro (changes requested):
On the PR's Conversation tab, towards the end of the page, there is a section that shows the build test results. Currently that section reads _"Some checks were not successful"._ One the of entries is `VersionBot/generate-version` and there is a `Details` link next to it. Clicking that link opens a build results page that reads:
```
Command failed: versionist --config=versionist.tmp.js
No commits were annotated with a change type since version 13.1.11
```
This would be because the commit is missing the `Change-type` footer. Have a look at the Contributing document for more details: https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md
Also relevant: https://github.com/people-os/process/blob/master/process/Commit_and_PR_Guidelines.md

While at it, the commit line reads _"added balena ssh troubleshoot",_ and this line will show in the Changelog file after the PR is merged: [added balena ssh troubleshoot](https://github.com/balena-io/balena-cli/blob/master/CHANGELOG.md)
As you can see in the Changelog, by convention we use the imperative form, e.g. _"Add, Update, Remove, Drop, Bump, Standardize, etc",_ instead of _"Added, Updated, Removed, Dropped, Bumped, Standardized, etc"._

ab77:
@inwenc reminder to close or merge this.

Issue #2458, opened 2022-02-18
`balena ssh <ip-addr> <service>`: Could not access docker daemon on device


pdcastro:
```
$ balena ssh 192.168.1.36 bar
Could not access docker daemon on device 192.168.1.36.
Please ensure the device is in local mode.
```

* Only happens with production variants of balenaOS, because TCP port 2375 (balenaEngine) is blocked.
* Only happens when a service name is used, because the CLI attempts to resolve the service name through a call to balenaEngine on port 2375.
* Didn't happen before CLI v13.2.0 because those older versions didn't allow `balena ssh <ip-addr>` at all with production variants of balenaOS. See:
  * #2292 
  * #2457

# Known workarounds

* ssh to the host OS and, from there, enter the container through a `balena-engine exec` command:
```
$ balena ssh 192.168.1.36

root@eb90fbf:~# balena-engine ps --format '{{.Names}}'
bar_4587562_2078151_3261c9d4c22f2c53a5267be459c89990
foo-tests_4587563_2078151_3261c9d4c22f2c53a5267be459c89990

root@eb90fbf:~# balena-engine exec -it bar_4587562_2078151_3261c9d4c22f2c53a5267be459c89990 sh
/ #
```

* Use `balena ssh <device-UUID>` instead of `balena ssh <ip-addr>`

# Possible fix

Instead of connecting to balenaEngine through TCP port 2375, `balena ssh` could obtain the container ID or full name over ssh as well. An efficient way of doing so (that avoids multiple round trip ssh connections) was implemented in the proof-of-concept [ssh-uuid project](https://github.com/pdcastro/ssh-uuid): use a single ssh connection that remotely executes a bash function to both determine the container ID and open the shell (or execute a command) on the target container.

# Specifications

- **balena CLI version:** 13.2.0

Issue #1560, opened 2020-01-09
Remove unnecessary supervisor dependency for 'balena ssh'


pdcastro:
If the supervisor is stopped (or broken) on a device, attempting to open a shell in an app container results in the error:
```
$ balena ssh <uuid> main
BalenaRequestError: Request error: tunneling socket could not be established, cause=socket hang up
```
And yet, the web dashboard succeeds in this case. @sradevski has found that:
> _the dashboard calls sdk.model.device.getWithServiceDetails, which returns all the running services, which also contains each service’s ID. Then that ID is used to ssh into the service’s container. In the CLI, however, it makes a call to the supervisor to retrieve the ID, which is unnecessary._

To improve reliability, the CLI should be modified to do the same as the web dashboard.

pdcastro:
For reference, the dashboard code that handles ssh sessions is at:
https://github.com/balena-io/balenacloud-dashboard/tree/master/src/scripts/modules/tty

PR #2461, opened 2022-02-24
ssh: Allow ssh to service with IP address and production balenaOS image

pdcastro:
Resolves: #1560
Resolves: #2458
Change-type: minor


pdcastro:
@balena-ci I self-certify!

Issue #2469, opened 2022-04-06
build: Unable to establish remote docker engine TLS/HTTPS connection with --dockerPort 2376


pdcastro:
When the following `balena build` options are used, the balena CLI should establish a HTTPS / TLS connection to the remote docker engine where [TLS is configured as per Docker's documentation](https://docs.docker.com/engine/security/protect-access/#use-tls-https-to-protect-the-docker-daemon-socket):

```
$ balena build my-project --dockerHost 192.168.1.101 --dockerPort 2376 --ca ca.pem --key key.pem --cert cert.pem  ...
```

Above, `192.168.1.101` is the IP address of the machine where the Docker engine is running, and `2376` is Docker's standard port number for encrypted connections.

However, the connection fails with an error:

```
Docker seems to be unavailable. Is it installed and running?
Error: (HTTP code 400) unexpected - Client sent an HTTP request to an HTTPS server.
```

The server configuration is correct as confirmed by a Docker client command:

```
$ docker -H tcp://192.168.1.101:2376 --tlsverify image ls
```

# Known workarounds

* Use the `DOCKER_HOST` environment variable instead of the `--dockerHost` and `--dockerPort` options:

```
$ DOCKER_HOST=tcp://192.168.1.101:2376 balena build --ca ca.pem --key key.pem --cert cert.pem ...
```

* Depending on your requirements, ssh port forwarding could be used instead of TLS, in which case the ca/cert/key options are not required as ssh takes care of encryption and authentication. For example, the following command forwards local port `127.0.0.1:2375` on the client side (balena CLI) to Unix socket `/var/run/docker.sock` on the server side (Docker engine):

```
$ ssh -NL 127.0.0.1:2375:/var/run/docker.sock user@192.168.1.101

$ balena build my-project --dockerHost 127.0.0.1 --dockerPort 2375 --deviceType raspberrypi3 --arch armv7hf
...
[Success] Build succeeded!
```

Port number 2375 is the standard Docker engine port for unencrypted API connections. If it is already in use on the client side, any other free port number (other than 2376) could be used instead.

This issue was observed with balena CLI v13.3.1 on macOS and Windows, but I think it applies to older versions and platforms as well.

PR #2470, opened 2022-04-06
build: Ensure HTTPS is used with dockerPort 2376 or with ca/cert/key

pdcastro:
Resolves: #2469 
Change-type: patch


pdcastro:
@balena-ci retest

pdcastro:
@balena-ci I self-certify!

Issue #2500, opened 2022-07-14
Should drop support for numeric fleet id parameters & flags


thgreasi:
https://github.com/balena-io/balena-cli/blob/a92f58134f54755b36f105b22039e2cf720108bb/lib/utils/common-flags.ts#L28
https://github.com/balena-io/balena-cli/blob/a92f58134f54755b36f105b22039e2cf720108bb/lib/utils/common-args.ts#L21

PR #2508, opened 2022-07-15
Drop support for numeric fleet id parameters from all commands

myarmolinsky:
Drop support for fleet id in favor of fleet name and slug

Resolves: #2500 
Change-type: major
Signed-off-by: Matthew Yarmolinsky <matthew-timothy@balena.io>

---

### DIFF
--- a/lib/commands/preload.ts
+++ b/lib/commands/preload.ts
@@ -288,7 +288,7 @@ Can be repeated to add multiple certificates.\
 				preloader.on('error', reject);
 				resolve(
 					this.prepareAndPreload(preloader, balena, {
-						appId: fleetSlug,
+						slug: fleetSlug,
 						commit,
 						pinDevice,
 					}),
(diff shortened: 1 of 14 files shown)


lib/utils/sdk.ts
```diff
@@ -24,71 +24,46 @@ import type {
...
-	}
-	// Not a slug and not a numeric database ID: must be an app name.
+	// Not a slug: must be an app name.
 	// TODO: revisit this logic when we add support for fleet UUIDs.
 	return await sdk.models.application.getAppByName(
-		nameOrSlugOrId,
+		nameOrSlug,
 		options,
 		'directly_accessible',
 	);
```
thgreasi, line 48: We could still allow this to accept numbers, but I see that it was a forcing function surfacing more commands that might be using IDs.

Let's just drop the unnecessary string check and move the special case handling first:
```suggestion
	const { looksLikeFleetSlug } = await import('./validation');
	if (!looksLikeFleetSlug(nameOrSlug)) {
		// Not a slug: must be an app name.
		// TODO: revisit this logic when we add support for fleet UUIDs.
		return await sdk.models.application.getAppByName(
			nameOrSlug,
			options,
			'directly_accessible',
		);
	}

	return await sdk.models.application.getDirectlyAccessible(
		nameOrSlug,
		options,
	);
```

Applies to `getFleetSlug()` as well.

lib/commands/preload.ts
```diff
@@ -503,15 +503,15 @@ Would you like to disable automatic updates for this fleet now?\
 		preloader: Preloader,
 		balenaSdk: BalenaSDK,
 		options: {
-			appId?: string;
```
thgreasi, line 506: This was just bad var naming, right?

thgreasi (commented):
LGTM :+1: self-certify-ble

thgreasi:
How about naming it like v ?
> Drop support for numeric fleet id parameters from all commands

lib/commands/preload.ts
myarmolinsky: Yep

thgreasi (approved):

Issue #2503, opened 2022-07-14
Cleanup all isV14 switches


thgreasi:
See: https://github.com/balena-io/balena-cli/search?q=isV14

PR #2511, opened 2022-07-20
Cleanup all isV14 switches

myarmolinsky:
Cleanup all isV14 switches

Resolves: #2503 
Change-type: major
Signed-off-by: Matthew Yarmolinsky <matthew-timothy@balena.io>

---

### DIFF
--- a/docs/balena-cli.md
+++ b/docs/balena-cli.md
@@ -367,6 +367,14 @@ fleet name, slug (preferred), or numeric ID (deprecated)
 
 open fleet dashboard page
 
+#### --fields FIELDS
+
+only show provided fields (comma-separated)
+
+#### -j, --json
+
+output in json format
+
 ## fleet create &#60;name&#62;
 
 Create a new balena fleet.
--- a/docs/balena-cli.md
+++ b/docs/balena-cli.md
@@ -333,6 +333,30 @@ Examples:
 
 ### Options
 
+#### --fields FIELDS
+
+only show provided fields (comma-separated)
+
+#### -j, --json
+
+output in json format
+
+#### --filter FILTER
+
+filter results by substring matching of a given field, eg: --filter field=foo
+
+#### --no-header
+
+hide table header from output
+
+#### --no-truncate
+
+do not truncate output to fit screen
+
+#### --sort SORT
+
+field to sort by (prepend '-' for descending order)
+
 ## fleet &#60;fleet&#62;
 
 Display detailed information about a single fleet.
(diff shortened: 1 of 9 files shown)


lib/utils/common-flags.ts
```diff
@@ -97,18 +96,7 @@ export const deviceType = flags.string({
...
-		: {
-				type: flags.string({
-					description: 'ignored - no longer required',
-					char: 't',
-					required: false,
-					hidden: true,
-				}),
-		  }),
-};
+export const deviceTypeIgnored = {};
```
thgreasi, line 99: Can we completely drop this, or is it used anywhere else?

docs/balena-cli.md
```diff
@@ -333,6 +333,30 @@ Examples:
 
 ### Options
 
+#### --fields FIELDS
```
thgreasi, line 336: Maybe we need to split these in separate commits to communicate to users what changed (rather than the not-that-useful "Cleanup all isV14 switches"). I'm guessing that using "blame" on GH to find the PR that added those version checks should help find a better commit message.

lib/utils/common-flags.ts
myarmolinsky: It's used but it's pointless to keep, just doing `...{}` pretty much, dropping

docs/balena-cli.md
```diff
@@ -367,6 +367,14 @@ fleet name, slug (preferred), or numeric ID (deprecated)
 
 open fleet dashboard page
 
+#### --fields FIELDS
+
+only show provided fields (comma-separated)
```
thgreasi, line 396: Let's change the commit to `fleet: Add --fields and --json options`, to be consistent w/ how the changelog looks.

docs/balena-cli.md
```diff
@@ -333,6 +333,30 @@ Examples:
...
+
+#### -j, --json
+
+output in json format
+
+#### --filter FILTER
+
+filter results by substring matching of a given field, eg: --filter field=foo
+
+#### --no-header
```
thgreasi, line 348: Let's change the title to `fleets: Add --filter, --no-header, --no-truncate, and --sort options` so that we are consistent with the older changelog entries.

thgreasi (commented):
It would be good to also mention the oclif output change on its own, as a heads-up that this might break output parsing. Maybe we could have a single commit that changes fleet & fleets to use this.outputData w/o passing the options, and we then add the options as extras?

PR #2530, opened 2022-09-29
Add to description that command is device specific

hraftery:
It seems that "initialize" is not well defined or referenced elsewhere, and it turns out that what happens is defined in the device type manifest. The `fincm3` runs "burn", which is an etcher-image-write (which, incidentally, is not sufficient to flash a Fin unless `rpiboot` is run first). So this is the best "clarification" I could come up with.

Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/345890-loop.2Fbalena-io/topic/cli.20-.20initialize.20command.3F/near/301237302

### DIFF
--- a/docs/balena-cli.md
+++ b/docs/balena-cli.md
@@ -2261,7 +2261,9 @@ expiry date assigned to generated provisioning api key (format: YYYY-MM-DD)
 ## os initialize &#60;image&#62;
 
 Initialize an os image for a device with a previously
-		configured operating system image.
+		configured operating system image and flash the
+		an external storage drive or the device's storage
+		medium depending on the device type.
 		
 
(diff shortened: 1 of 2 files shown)


thgreasi (approved):

thgreasi:
@balena-ci retest

hraftery:
For the record, lest I doubt my memory in the future, I didn't write this commit.

thgreasi:
@hraftery I changed the commit title during summit. Are concerned about the commit title or the previous PR title?

hraftery:
Not so much the title @thgreasi , but that GitHub says I authored the commit at all. Just a quirk of the UI I think, but the only thing I wrote here is the first comment in the PR, so it's eerie to see my name as the author of what is your work.

PR #2532, opened 2022-10-11
Adding .nvmrc so we can use nvm use instead of hunting for version

zoobot:
We should move this repo to node v16 at least!
Change-type: minor

Signed-off-by: zoobot <rommims@gmail.com>

Resolves: #   
Depends-on: <url>   
See: <url>   

---


ab77:
@zoobot Change-type needs to be the last line.. I believe, otherwise no-bueno. Also, if the tests pass - just self-certify it, otherwise it will never get merged. Especially since this is cosmetical.

.nvmrc
```diff
@@ -0,0 +1 @@
+12.22.12
```
thgreasi, line 1: How about just 12, so that it gets the latest version rather than a pinned one?
```suggestion
12
```

thgreasi (commented):
Let's also change this to be `Change-type: patch` since the convention of using `minor` in end-user facing changelogs is for when the PR actually adds any new feature for them.

thgreasi:
@balena-ci rebase

thgreasi (approved):

thgreasi:
@zoobot can you please rebase over master so that we avoid the extra merge commit and can keep the history linear?
image

ab77:
@zoobot LGTM

ab77:
> @balena-ci rebase

@thgreasi @zoobot I suspect versionbot rebases are, under certain circumstances, won't work. Suggest you use built in functionality in Git UI from the drop down:

image

PR #2539, opened 2022-10-13
changes analytics endpoint to analytics-backend

otaviojacobi:
change-type: minor

This changes the endpoint from `balena-api` to `analytics-backend`. This should be only merged after: https://github.com/balena-io/analytics-backend/pull/64 is deployed, otherwise could impact performance of the CLI.


aethernet (commented):
Is the updated backend compatible with both versions ?

otaviojacobi:
> Is the updated backend compatible with both versions ?

Hello @aethernet the new endpoint supports the old-format too (altough I think for this particular endpoint it won't reply instantly). However, because previous versions of the CLI were built using the `api.balenaCloudUrl` instead of `data.balenaCloudUrl` they will still call the the balena-api code, in that sense, the new backend does not need to be backward compactible.

aethernet (approved):
makes sense !

thgreasi:
@otaviojacobi is the new version of the analytics-backed already deployed? The v0.5.9.
When a PR depends on a different PR being merged (and if a component, then also deployed), then the convention is to mark the PR as draft so that an approval doesn't merge it and add a
```
Depends-on: https://github.com/balena-io/analytics-backend/pull/64
```
line in the commit description.
Also the commits should have been in present tense eg:
* Keep events loggiging with default message [Otávio Jacobi]
* Use amplitude data events format [Otávio Jacobi]
* Change the analytics endpoint to analytics-backend [Otávio Jacobi]

After reviewing the commits I feel that they should have been squashed down to 1 commit before merging to have the minimum number of isolated changes that describe the PR's changes, especially since the first commit still uses localhost.

Since the SDK, CLI and many other components have their CHANGELOGs as the primary method for surfacing changes to users, not having tons of commit is kind of important. Moreover, since the analytics backend change is not a user facing one, this would have ideally been a `patch` change, since from the user's point of view this didn't add any feature.

Let me point you to our commit guidelines
See: https://github.com/people-os/process/blob/master/process/Commit_and_PR_Guidelines.md#commit-guidelines

PR #2548, opened 2022-10-20
Allow most printable ASCII chars for object label key (b-cli)

kb2ma:
Implemented as a balena-compose dependency update. See balena-io-modules/balena-compose#13 . Extends acceptable label characters to printable ASCII characters except space, single/double quotes and backtick.

**Draft PR until balena-compose PR merged so uses the generated version number.**

* Updates minimum version of @balena/compose and @types/dockerode to correspond to balena-compose PR

As of 2023-03-22 this PR still is blocked on an overall system solution between clients and server. This PR is useful; however older clients likely will fail in some way if they receive a service containing the extra characters.

### Updates `npm-shrinkwrap.json`

**Major**

@types/ssh2 0.5.52 --> 1.11.6, which removed its unnecessary dependency on ssh2-streams. See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60982 , from 2022-06.

**Minor**
* fp-ts 2.12.2 --> 2.13.1

**Patch**
* docker-modem
* io-ts
* nan
* pinejs-client-core
* pinejs-client-request
* @types/lodash

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1295,9 +1295,9 @@
       }
     },
     "@balena/compose": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/@balena/compose/-/compose-2.1.1.tgz",
-      "integrity": "sha512-Ra+nrR+wAuWYZsKFAtHpVZOpig19rcMvBtdoPuCjFgcpSCqGNLFaaHqRvg/kr9nN6sFXyaiYKpsb9C/varmy3Q==",
+      "version": "2.1.2",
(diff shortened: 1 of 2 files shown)


kb2ma:
linux-ubuntu-x86_64-node14 build fails 2 tests, both due to timeout failures > 12000 ms. However, my local tests pass on node 12, which currently is the required node version. Notice time to run tests below is < 1000 ms in both cases.

```
  balena build
[Info] Interceptor server listening on 127.0.0.1:33601
[Info] Proxy server listening on 127.0.0.1:41493
    ✓ should create the expected tar stream (single container) (546ms)

  balena os configure
    ✓ should inject a valid config.json file (992ms)
```

kb2ma:
@balena-ci retest

Issue #2563, opened 2022-11-16
preloading broken with docker engine 20.10.17


aethernet:
# Expected Behavior

preloading a fleet

# Actual Behavior

Stuck at "Reading image information" 

# Steps to Reproduce the Problem

1. $ balena os download raspberrypi3 -o /tmp/raspberry-pi.img  
2. $ DEBUG=1 balena preload /tmp/raspberry-pi.img --fleet balenasound --commit latest 

(i'm testing with balenasound but other combination of base image and fleet gave same results)

# Specifications

- **balena CLI version: balena-cli version "14.5.10"**
- **Cloud backend: balenaCloud
- **Operating system version: macos 12.3**
- **64 bit OS and processor: m1**
- **Install method: executable installer** 
- **If npm install, Node.js and npm version:** e.g. Node v8.16.0 and npm v6.4.1

# Additional References
https://forums.balena.io/t/cli-balena-preload-hangs-at-reading-image-information/365436/11

PR #2570, opened 2022-12-12
patch: update balena-preload to 12.2.0

aethernet:
Resolves: #2563

Change-type: patch


aethernet:
@balena-ci I self-certify!

thgreasi:
@aethernet you missed to do an `npm i` to actually get the shrinkwrap updated.
See: https://github.com/balena-io/balena-cli/pull/2574

thgreasi:
This v make tests fail if we forget in the future to update the shrinkwrap as well.
See: https://github.com/balena-io/balena-cli/pull/2576

PR #2571, opened 2022-12-16
Update to Node 14

pipex:
This PR includes one memory usage improvement that may help with #2165, which seems to be the main blocker for updating the node dependency.

Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/We.20need.20to.20talk.20about.20the.20CLI

### DIFF
--- a/lib/utils/ssh.ts
+++ b/lib/utils/ssh.ts
@@ -158,17 +158,16 @@ export async function runRemoteCommand({
 					resolve([code ?? undefined, signal ?? undefined]),
 				);
 
-				if (ps.stdin && stdin && typeof stdin !== 'string') {
-					stdin.pipe(ps.stdin);
-				}
-				if (ps.stdout && stdout && typeof stdout !== 'string') {
-					ps.stdout.pipe(stdout);
-				}
-				if (ps.stderr && stderr && typeof stderr !== 'string') {
-					ps.stderr.pipe(stderr);
-				}
-			},
-		);
+			if (ps.stdin && stdin && typeof stdin !== 'string') {
+				stdin.pipe(ps.stdin);
+			}
+			if (ps.stdout && stdout && typeof stdout !== 'string') {
+				ps.stdout.pipe(stdout);
+			}
+			if (ps.stderr && stderr && typeof stderr !== 'string') {
+				ps.stderr.pipe(stderr);
+			}
+		});
 	} catch (error) {
 		const msg = [
 			`ssh failed with exit code=${exitCode} signal=${exitSignal}:`,
--- a/lib/utils/ssh.ts
+++ b/lib/utils/ssh.ts
@@ -151,11 +151,12 @@ export async function runRemoteCommand({
 	let exitCode: number | undefined;
 	let exitSignal: NodeJS.Signals | undefined;
 	try {
-		[exitCode, exitSignal] = await new Promise<[number, NodeJS.Signals]>(
-			(resolve, reject) => {
-				const ps = spawn(program, args, { stdio })
-					.on('error', reject)
-					.on('close', (code, signal) => resolve([code, signal]));
+		[exitCode, exitSignal] = await new Promise((resolve, reject) => {
+			const ps = spawn(program, args, { stdio })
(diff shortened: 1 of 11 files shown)


ab77 (approved):

klutchell (changes requested):
I would prefer all the linting changes be done in a separate commit. It's nearly impossible to tell what has changed in these commits.

pipex:
I didn't realize the linter had ran, let me see how I can split the changes

package.json
```diff
@@ -89,7 +89,7 @@
   "author": "Balena Inc. (https://balena.io/)",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=12 <16"
+    "node": ">=14 <16"
```
klutchell, line 92: Do we also need to update `    "@types/node": "^12.20.42",` on line 148?

pipex: Makes sense

klutchell: I also found some other code comments that refer to `12.20.42` that can probably be cleaned up.

pipex:
> Where do we change the version of node packaged with standalone?

It's the one in the custom actions yaml config

klutchell (approved):

package.json
pipex: Converted back to draft to fix the code comments

tests/commands/ssh.spec.ts
```diff
@@ -178,7 +178,7 @@ async function startMockSshServer(): Promise<[Server, number]> {
 	});
 
 	return await new Promise<[Server, number]>((resolve, reject) => {
-		// TODO: remove 'as any' below. According to @types/node v12.20.42, the
+		// TODO: remove 'as any' below. According to @types/node v14.18.36, the
```
pipex, line 181: I tested removing the `as any` with the new version of `@types/node` but the code also fails to complain because of the error in the callback. I just updated the comment to match the node version

package.json
pipex: Resolved here https://github.com/balena-io/balena-cli/pull/2571#discussion_r1060770099

lib/utils/ssh.ts
```diff
@@ -151,11 +151,12 @@ export async function runRemoteCommand({
...
-		[exitCode, exitSignal] = await new Promise<[number, NodeJS.Signals]>(
-			(resolve, reject) => {
-				const ps = spawn(program, args, { stdio })
-					.on('error', reject)
-					.on('close', (code, signal) => resolve([code, signal]));
+		[exitCode, exitSignal] = await new Promise((resolve, reject) => {
+			const ps = spawn(program, args, { stdio })
+				.on('error', reject)
+				.on('close', (code, signal) =>
+					resolve([code ?? undefined, signal ?? undefined]),
```
pipex, line 158: The function signature for the `close` event changed in Node 14. Both `code` and `signal` can optionally be null. This updates the resolve call so the return type matches the variable definitions

pipex:
@Page- would you like to do a sanity test with this branch to see if you can replicate issue #2165 ?

klutchell:
🎉

micwoj92:
Is there reason why 15.0.1 is still not on npmjs? https://www.npmjs.com/package/balena-cli

klutchell:
> Is there reason why 15.0.1 is still not on npmjs? https://www.npmjs.com/package/balena-cli

Unfortunately the publish job failed as the artifacts have expired. We have a task to extend the artifact lifetimes to avoid this.
https://github.com/balena-io/balena-cli/pull/2571#issuecomment-1377595136

I'll try to re-run the job.

klutchell:
@micwoj92 it's there now, thanks for flagging this!

https://www.npmjs.com/package/balena-cli/v/15.0.1

micwoj92:
@klutchell 15.0.2 isnt uploaded too, do you know a longtime solution for that?

klutchell:
@micwoj92 15.0.2 is published now, and this error shouldn't pop up anymore.

Thanks again!

PR #2573, opened 2022-12-22
Upgrade balena-preload to 13.0.0

aethernet:
Drops ability to preload Intel Edison (EOL 2017)

Change-type: major

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3830,9 +3830,9 @@
       }
     },
     "balena-preload": {
-      "version": "12.2.0",
-      "resolved": "https://registry.npmjs.org/balena-preload/-/balena-preload-12.2.0.tgz",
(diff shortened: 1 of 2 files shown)


klutchell (changes requested):
Can you please change your commit format to look like [this](https://github.com/balena-io/balena-cli/commit/2b2c40c22d15d765f19b421976113939ec4fc455)?

The last few versions were not formatted correctly and this breaks nested changelogs.

Unfortunately that process is not documented well, if at all.

klutchell:
Have we asked CS how many Intel Edison devices are still registered to balenaCloud?

What about openBalena and balenaMachine users that might have these devices?

aethernet:
Context is here :
- https://balena.zulipchat.com/#narrow/stream/345889-loop.2Fbalena-os/topic/Intel.20Edison/near/310245867

Both balenaCloud and openBalena users can use the latest version with Edison support.

IMHO if we don't clean up exceptions when deprecating DT, we'll never do it. Should be part of the process.

klutchell (approved):

PR #2574, opened 2022-12-28
Update the npm-shrinkwrap.json dependencies to match the package.json

thgreasi:
Change-type: patch
See: https://github.com/balena-io/balena-cli/pull/2570


thgreasi:
@balena-ci I self-certify!

PR #2575, opened 2022-12-28
Stop using the deprecated balena-sync module

thgreasi:
Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #2580, opened 2023-01-13
Fix push --nolive doc typo

joshbwlng:
Change-type: patch


joshbwlng:
@balena-ci I self certify!

joshbwlng:
@balena-ci i self-certify!

PR #2597, opened 2023-03-13
Devices: explicitly fetches only used fields

otaviojacobi:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


thgreasi (commented):
other than that feel free to self-certify

lib/utils/helpers.ts
```diff
@@ -445,6 +445,18 @@ export const expandForAppName = {
...
+export const devicesSelectFields = {
+	$select: [
+		'id',
+		'uuid',
+		'device_name',
+		'status',
+		'is_online',
+		'supervisor_version',
+		'os_version',
+	] as any,
```
thgreasi, line 457:
```suggestion
	] satisfies keyof<BalenaSdk.Device>,
```

otaviojacobi: I don't think we can do this as CLI is locked to TS 4.6.4 right now

lib/utils/helpers.ts
```diff
@@ -445,6 +445,18 @@ export const expandForAppName = {
 	},
 } as const;
 
+export const devicesSelectFields = {
```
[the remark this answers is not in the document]
otaviojacobi: @thgreasi if we make `const fields = ...` as you suggested, this is a breaking change because the order of the output actually changes (for example, the device_type property should be the fourth element on the array for output not change....

otaviojacobi:
LGTM

PR #2614, opened 2023-04-27
Use valid release uuid for local releases

pipex:
On local push, the CLI uses `localrelease` as the `commit` property for the development application. This is not a valid uuid and will not be read properly by the supervisor, as seen in

https://github.com/balena-os/balena-supervisor/blob/master/src/compose/service.ts#L652

While this is not a problem right now, the commit is becoming the main way to identify a service release (replacing `releaseId` and `imageId`), and the invalid release uuid could cause update issues when pushing a local release on when using some API endpoints.

Change-type: patch
Relates-to: balena-os/balena-supervisor#2136

### DIFF
--- a/lib/utils/device/deploy.ts
+++ b/lib/utils/device/deploy.ts
@@ -44,7 +44,7 @@ import { displayBuildLog } from './logs';
 import { stripIndent } from '../lazy';
 
 const LOCAL_APPNAME = 'localapp';
-const LOCAL_RELEASEHASH = 'localrelease';
+const LOCAL_RELEASEHASH = '10ca12e1ea5e';
 const LOCAL_PROJECT_NAME = 'local_image';
 
 // Define the logger here so the debug output


pipex:
@klutchell what do you think about this? With this change and https://github.com/balena-os/balena-supervisor/pull/2136 local services will have the name `serviceName_10ca12e1ea5e`. I wonder if this might be end up being confusing for users. 

The alternative is to hardcode `localrelease` as a valid alternative in the regex here https://github.com/balena-os/balena-supervisor/blob/master/src/compose/service.ts#L652, which I don't love either but it will also be compatible with any CLI version

klutchell:
I don't think the name `serviceName_10ca12e1ea5e` is the way to go. It's confusing and barely resembles the original meaning.
We should either pick a new, short, uuid, that is easy to identify (e,g. `abc123`, `ba1e4a) or do as you suggested and hardcode `localrelease` as valid.

pipex:
I'm divided on the relevance of the container name, on one side, people should not be relying on the container name in their apps as it's a hidden interface, on the other hand, we know that people do (including me) to see at a glance what's happening with the device. 

My fear about hardcoding the `localrelease` on the supervisor is that once it's there, it will probably have to be there pretty much forever

klutchell:
Fair enough, the container name in local mode really shouldn't matter. I
think we are the exceptions not the rule.

Might as well go with your first instinct on this one and use what you have.

What happens when a new CLI connects to an old supervisor in local mode?

pipex:
> What happens when a new CLI connects to an old supervisor in local mode?

@klutchell nothing really, the old supervisor will just configure the container names with the new release commit

A more problematic case is an old CLI with the newer supervisor on PR #2136, the regex won't match the commit and the supervisor may fall into a loop of `updateMetadata failed renaming container to the same name`. The solution there is to update the CLI, but we might see some tickets on support

klutchell:
> A more problematic case is an old CLI with the newer supervisor

This makes me lean towards just hardcoding `localrelease` in the SV. Yeah it will probably end up being there forever, but it's not a big chunk of code that would significantly impact maintenance IMO.

pipex:
> it's not a big chunk of code that would significantly impact maintenance IMO.

Not by itself, but there are 10s of these type of code pieces in the supervisor that make refactoring harder. 

In any case I'm siding with you on this, I think we can do both. We merge this so the new name starts making it to devices and support `localrelease` on the supervisor, and remove it in the future once more users are running an updated CLI

pipex:
I self-certify!

PR #2615, opened 2023-04-27
Remove nvmrc

pipex:
There is not a lot of benefit to using `.nvmrc` as it still requires `nvm use`, and not everybody uses `nvm`. The call to `npm install` will already warn about using the wrong version.

Change-type: patch


pipex:
I self-certify!

thgreasi:
@pipex I actually liked being able to blindly use `nvm use` whenever I go to the cli directory, and have it pick the correct version, w/o me having to check the engines in package.json of follow the changelog.
The reason that we don't have them in many other repos, is that in most other places we are following the latest LTS.

pipex:
@thgreasi I wasn't sure anyone used that command. I removed it mostly because it's yet another place we have to change the version when bumping Node. Since that doesn't happen often, it's easy to miss one or two pieces.

pipex:
We can re-add it if it's really useful

PR #2618, opened 2023-05-02
Update to Node 16

pipex:
Change-type: patch


tests/test-data/pkg/expected-warnings-darwin.txt
```diff
@@ -92,3 +84,8 @@
   The file must be distributed with executable as %2.
   %1: node_modules/opn/xdg-open
   %2: path-to-executable/xdg-open
+prebuild-install warn install No prebuilt binaries found (target=v16.16.0 runtime=node arch=x64 libc= platform=darwin)
```
pipex, line 87: I suspect we need to solve this warning instead of adding it to the expected list. I don't understand enough yet about how `prebuild-install` works

pipex: apparently not, the warning is harmless

lib/commands/tunnel.ts
```diff
@@ -155,17 +155,17 @@ export default class TunnelCmd extends Command {
...
 								remotePort,
 							);
 						} catch (err) {
 							logConnection(
 								client.remoteAddress || '',
 								client.remotePort || 0,
-								client.localAddress,
-								client.localPort,
+								client.localAddress || '',
+								client.localPort || 0,
```
Page-, line 168: Should this be
```suggestion
								client.localAddress ?? '',
								client.localPort ?? 0,
```
?

lib/commands/tunnel.ts
```diff
@@ -155,17 +155,17 @@ export default class TunnelCmd extends Command {
...
```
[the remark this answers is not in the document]
pipex: Probably, I just went with what was already there. I'll change all to `??`

pipex:
I self-certify!

PR #2619, opened 2023-05-15
os configure, config generate: Add '--secureBoot' option to opt-in secure boot

alexgg:
Allow to generate a config file with `installer.secureboot` set so that a secure boot and disk encrypted system can be installed.

Change-type: minor

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/utils/config.ts
+++ b/lib/utils/config.ts
@@ -52,6 +52,10 @@ export interface ImgConfig {
 	os?: {
 		sshKeys?: string[];
 	};
+
+	installer?: {
+		secureboot?: boolean;
+	};
 }
 
 export async function generateApplicationConfig(
@@ -63,6 +67,7 @@ export async function generateApplicationConfig(
 		os?: {
 			sshKeys?: string[];
 		};
+		secureBoot?: boolean;
 	},
 ): Promise<ImgConfig> {
 	options = {
@@ -82,6 +87,12 @@ export async function generateApplicationConfig(
 		config.os.sshKeys = config.os.sshKeys
 			? [...config.os.sshKeys, ...options.os.sshKeys]
 			: options.os.sshKeys;
+	}
+
+	// configure installer secure boot opt-in if specified
+	if (options.secureBoot) {
+		config.installer ??= {};
+		config.installer.secureboot = options.secureBoot;
 	}
 
 	return config;
@@ -164,4 +175,59 @@ export async function validateDevOptionAndWarn(
 		Please note that development mode allows unauthenticated, passwordless root ssh access
 		and exposes network ports such as 2375 that allows unencrypted access to balenaEngine.
 		Therefore, development mode should only be used in private, trusted local networks.`);
+}
+
+/**
+ * Chech whether the `--secureBoot` option of commands related to OS configuration
+ * such as `os configure` and `config generate` is compatible with a given
+ * OS release, and print a warning regarding the consequences of using that
+ * option.
+ */
+export async function validateSecureBootOptionAndWarn(
+	secureBoot?: boolean,
+	slug?: string,
+	version?: string,
+	logger?: import('./logger'),
+) {
+	if (!secureBoot) {
+		return;
+	}
+	const { ExpectedError } = await import('../errors');
+	if (!version) {
+		throw new ExpectedError(`Error: No version provided`);
+	}
+	if (!slug) {
+		throw new ExpectedError(`Error: No device type provided`);
+	}
+	const sdk = getBalenaSdk();
+	const [OsRelease]: BalenaSdk.OsVersion[] =
+		await sdk.models.os.getAllOsVersions(`${slug}`, {
+			$select: 'id',
+			$filter: { raw_version: `${version.replace(/^v/, '')}` },
+		});
+	if (!OsRelease) {
+		throw new ExpectedError(`Error: No ${version} release for ${slug}`);
+	}
+	const OsContract = await sdk.models.release.get(OsRelease.id, {
+		$select: 'contract',
+	});
+	const yaml = require('js-yaml');
+	const contract = yaml.load(OsContract['contract']);
+	if (
+		contract.provides.some((entry: Dictionary<string>) => {
+			return entry.type === 'sw.feature' && entry.slug === 'secureboot';
+		})
+	) {
+		if (!logger) {
+			const Logger = await import('./logger');
+			logger = Logger.getLogger();
+		}
+		logger.logInfo(stripIndent`
+			The '--secureBoot' option is being used to configure a balenaOS installer image
+			into secure boot and full disk encryption.`);
+	} else {
+		throw new ExpectedError(
+			`Error: The '--secureBoot' option is not supported for ${slug} in ${version}`,
+		);
+	}
 }
--- a/lib/utils/config.ts
+++ b/lib/utils/config.ts
@@ -200,21 +200,25 @@ export async function validateSecureBootOptionAndWarn(
 		throw new ExpectedError(`Error: No device type provided`);
 	}
 	const sdk = getBalenaSdk();
-	const [OsRelease]: BalenaSdk.OsVersion[] =
-		await sdk.models.os.getAllOsVersions(`${slug}`, {
-			$select: 'id',
-			$filter: { raw_version: `${version.replace(/^v/, '')}` },
-		});
-	if (!OsRelease) {
+	const releasePineOpts = {
+		$select: 'contract',
+		$filter: { raw_version: `${version.replace(/^v/, '')}` },
+	} satisfies BalenaSdk.PineOptions<BalenaSdk.Release>;
+	// TODO: Remove the added type casting when the SDK returns the fully typed result
+	const [osRelease] = (await sdk.models.os.getAllOsVersions(
+		slug,
+		releasePineOpts,
+	)) as Array<
+		BalenaSdk.OsVersion &
+			BalenaSdk.PineTypedResult<BalenaSdk.Release, typeof releasePineOpts>
+	>;
+	if (!osRelease) {
 		throw new ExpectedError(`Error: No ${version} release for ${slug}`);
 	}
-	const OsContract = await sdk.models.release.get(OsRelease.id, {
-		$select: 'contract',
-	});
-	const yaml = require('js-yaml');
-	const contract = yaml.load(OsContract['contract']);
+
+	const contract = osRelease.contract ? JSON.parse(osRelease.contract) : null;
 	if (
-		contract.provides.some((entry: Dictionary<string>) => {
+		contract?.provides.some((entry: Dictionary<string>) => {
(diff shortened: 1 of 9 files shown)


lib/utils/config.ts
```diff
@@ -84,6 +89,12 @@ export async function generateApplicationConfig(
 			: options.os.sshKeys;
 	}
 
+	// configure installer secure boot opt-in if specified
+	if (options.secureBoot) {
+		config.installer = config.installer ? config.installer : {};
```
otaviojacobi, line 94: I think this can be simplified as 
```suggestion
		config.installer = config.installer ?? {};
```

lib/utils/config.ts
```diff
@@ -165,3 +176,48 @@ export async function validateDevOptionAndWarn(
...
+
+/**
+ * Chech whether the `--secureBoot` option of commands related to OS configuration
+ * such as `os configure` and `config generate` is compatible with a given
+ * device type and balenaOS version, and print a warning regarding the
+ * consequences of using that option.
+ */
+export async function validateSecureBootOptionAndWarn(
+	secureBoot?: boolean,
+	deviceManifest?: any,
```
otaviojacobi, line 188: @alexgg  Do you think there is anyway we can make this strongly typed? 
If I understand the code correctly, we only care about the "yocto" and nested "machine" properties. Looking on where where this is called, I see that this should be of type `BalenaSdk.DeviceTypeJson.DeviceType` however, I don't see the "machine" property.  @thgreasi do you think we might be missing this yocto.machine property on the SDK typings https://github.com/balena-io/balena-sdk/blob/master/lib/types/device-type-json.ts#L30-L33 ?

otaviojacobi: These should help get the typing in place for this one: https://github.com/balena-io/open-balena-api/pull/1351 and https://github.com/balena-io/balena-sdk/pull/1330

lib/utils/config.ts
thgreasi: Or even shorter:
```suggestion
		config.installer ??= {};
```

lib/utils/config.ts
```diff
@@ -165,3 +176,48 @@ export async function validateDevOptionAndWarn(
...
+) {
+	if (!secureBoot) {
+		return;
+	}
+
+	const deviceManifestObj: { [index: string]: any } = deviceManifest;
+	const deviceType = deviceManifestObj['yocto']['machine'];
+
+	const { ExpectedError } = await import('../errors');
+	const supportedDeviceTypes = ['generic-amd64'];
```
thgreasi, line 200: Ideally this should come from the DT contract (or the device-type.json :disappointed: ), so that we don't need to release new versions of the CLI every time we add support for secure boot to an extra DT.

For now, let's just move this const definition outside of the function at the top level to raise its importance vs other vars.

lib/utils/config.ts
```diff
@@ -165,3 +176,48 @@ export async function validateDevOptionAndWarn(
...
+	if (!supportedDeviceTypes.includes(`${deviceType}`)) {
+		throw new ExpectedError(
+			`Error: The '--secureBoot' option is not supported for the '${deviceType}' device Type. Supported device types include ${supportedDeviceTypes.join(
+				', ',
+			)}.`,
+		);
+	}
+
+	const supportedFromVersion = '2.114.21';
```
thgreasi, line 209: > Ideally this should come from the DT contract (or the device-type.json :disappointed: ), so that we don't need to release new versions of the CLI every time we add support for secure boot to an extra DT.
> 
> For now, let's just move this const definition outside of the function at the top level to raise its importance vs other vars.

lib/utils/config.ts
```diff
@@ -165,3 +176,48 @@ export async function validateDevOptionAndWarn(
...
+	if (!semver.satisfies(version, `>=${supportedFromVersion}`)) {
```
thgreasi, line 210: This would give false positives in case the OS version is an ESR.
For that case we need
* to use `sdk.models.os.getAllOsVersions(deviceType)` to fetch the OS versions for the DT
* find the OS version that was provided
* use the `basedOnVersion` for this comparison.

lib/utils/config.ts
alexgg: Done, thanks.

lib/utils/config.ts
alexgg: hey @otaviojacobi I agree - the correct place is the future equivalent to the `device-type.json` as there are two sides:

* The device type contract specified hw capabilities - whether the hardware has the secure boot capability
* The OS contract should state that the software supports it.

So the right way to do it would be for the CLI to query the OS release for its contract and use that.

The options were:

1) Hardcoding the list - this would allow us to merge this quickly and continue with the SB release
2) Extend `device-type.json` - this would be quick but we must not use this file any more
3) Extend the OS contract to add the secure boot setting, and have the CLI use it. For me this is the right thing, but it will take longer and will delay the SB feature.

So I will do as you say and move it outside the function, merge it, and work on 3) so we don't have to update the CLI in the future.

lib/utils/config.ts
alexgg: thanks for raising this @thgreasi . We do not release4 ESR releases at this moment for the only DT that supports this (generic-amd64), but I agree the solution needs to cater for ESR too.

I was wondering whether we could make use of the `meta-balena-base` label to compare when dealing with ESR, that is, we know the feature was introduced in `2.114.21` so any ESR whose `meta-balena-base` is above that will support it.

lib/utils/config.ts
alexgg: right, not sure if we should expose `device-type.json` via typings, but I see both @thgreasi and yourself agree so I'm fine with it. I'll update the SDK version once the PR is merged and tagged.

lib/utils/config.ts
thgreasi: Yea I agree. Having
* the DT contract on whether the HW can support secure boot (so that automation can use it)
* AND a field in the hostApp release contract is probably the best option, as the clearest way to communicate whether a given OS release supports.

Even though it would make it harder to find the min OS version that added support for secure boot, I don't think it matters that much. A generic error like "Secure boot is not supported by this OS Version" sgtm.

lib/utils/config.ts
thgreasi: @alexgg are you referring to the `meta-balena-base` release tag? B/c that's exactly what the `basedOnVersion` holds in my above suggestion (as returned by `sdk.models.os.getAllOsVersions(deviceType)`).

If this is the only blocker, we can just for now switch `>=` to `^` in the `satisfies` call and generalize this post-MVP

lib/utils/config.ts
alexgg: thanks, I updated the SDK and the code.

lib/utils/config.ts
alexgg: actually @thgreasi I think I going to dedicate a few hours to this approach as it's the right one and simplifies things - no need to check for DT or version compatibility, if the OS contract says it's supported that's it.

lib/utils/config.ts
alexgg: I used this in the update, thanks.

lib/utils/config.ts
alexgg: hi @thgreasi I pushed an implementation of the above. It's not completely tested as I  want to try an use the OS contract instead as I mentioned above. Will come back to this if I don't manage to make the contract work in reasonable time.

lib/utils/config.ts
```diff
@@ -165,3 +176,55 @@ export async function validateDevOptionAndWarn(
...
+	const sdk = getBalenaSdk();
+	const [OsRelease]: BalenaSdk.OsVersion[] =
+		await sdk.models.os.getAllOsVersions(`${slug}`, {
+			$select: 'id',
+			$filter: { raw_version: `${version.replace(/^v/, '')}` },
+		});
+	if (!OsRelease) {
+		throw new ExpectedError(`Error: No ${version} release for ${slug}`);
+	}
+	const OsContract = await sdk.models.release.get(OsRelease.id);
```
thgreasi, line 211: You can actually combine the two requests, but atm the SDK typings will not include the added properties :(
Working on this in the next major, but for now it's better to cast it rather than doing 2 requests.
```suggestion
	// TODO: Remove the added type casting when the SDK returns the fully typed result
	const [osRelease] = (await sdk.models.os.getAllOsVersions(`${slug}`, releasePineOpts)) as Array<BalenaSdk.OsVersion & BalenaSdk.PineTypedResult<BalenaSdk.Release, typeof releasePineOpts>>;
	if (!osRelease) {
		throw new ExpectedError(`Error: No ${version} release for ${slug}`);
	}
	const osContract = osRelease.contract;
```
If you still prefer 2 requests, then we should at least add a $select in the second request, so that it avoids fetching unnecessary field (which in this resource can be quite large).

thgreasi (commented):
Other than these LGTM :+1: :tada:

lib/utils/config.ts
alexgg: thanks @thgreasi I'll try with one request as soon as I get back to this PR

lib/utils/config.ts
```diff
@@ -165,3 +176,55 @@ export async function validateDevOptionAndWarn(
...
+	if (!OsContract) {
+		throw new ExpectedError(
+			`Error: No release contract for release ${OsRelease.id}`,
+		);
+	}
+	if (JSON.stringify(OsContract).search('secureboot')) {
```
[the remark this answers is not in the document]
alexgg: I'll use the exact location once the OS contract change is merged - the code above was just to see it working. I will mark it as `ready to review` once I feel the code is back to a reviewable state.

package.json
```diff
@@ -207,7 +207,7 @@
     "balena-image-fs": "^7.0.6",
     "balena-image-manager": "^8.0.0",
     "balena-preload": "^13.0.0",
-    "balena-sdk": "^16.40.0",
+    "balena-sdk": "^16.44.2",
```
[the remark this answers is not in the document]
alexgg: I split it into a different commit

lib/utils/config.ts
alexgg: The code now uses the OS release contract

lib/utils/config.ts
alexgg: @thgreasi I wasn't able to do this in one request, and I don't quite follow the code sample typings unfortunately. I have left this in two requests with the additional select, but I also think one request is best. Do you know how the single request would look like?

lib/utils/config.ts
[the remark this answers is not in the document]
alexgg: The code now parses the Os yaml contract and searches for the exact entry.

thgreasi (commented):
LGTM feel free to self-certify whenever you want

alexgg:
lgtm

PR #2641, opened 2023-06-30
Pin dockerode and drop docker-toolbelt

dfunckt:
- **Pin dockerode to v3.3.3:** v3.3.4 introduces a regression that is fixed by https://github.com/apocas/dockerode/pull/695 but Dockerode has not published a version that includes the fix yet. Pin the dependency to ensure we don’t ever update to a broken version.
- **Remove redundant dependency on docker-toolbelt:** nothing in the code seems to depend on functionality provided by toolbelt, so dropping the dependency.

Change-type: patch


dfunckt:
I self-certify!

PR #2648, opened 2023-07-14
ssh: use the new container name format

pipex:
The supervisor is changing the container name format from

`<serviceName>_<imageId>_<releaseId>_<commit>` to
`<serviceName>_<commit>`. This updates the SSH command to work with both
formats.

Change-type: minor
Depends-on: balena-os/balena-supervisor#2136

### DIFF
--- a/lib/utils/device/ssh.ts
+++ b/lib/utils/device/ssh.ts
@@ -76,13 +76,21 @@ export async function getContainerIdForService(
 		.filter((l) => l);
 
 	const { escapeRegExp } = await import('lodash');
-	const regex = new RegExp(`(?:^|\\/)${escapeRegExp(opts.service)}_\\d+_\\d+`);
+	const regex = new RegExp(
+		`(?:^|\\/)${escapeRegExp(
+			opts.service,
+		)}_(?:\\d+_)?(?:\\d+_)?(?:[a-f0-9]*)?$`,
+	);
 	// Old balenaOS container name pattern:
 	//    main_1234567_2345678
 	// New balenaOS container name patterns:
 	//    main_1234567_2345678_a000b111c222d333e444f555a666b777
 	//    main_1_1_localrelease
-	const nameRegex = /(?:^|\/)([a-zA-Z0-9_-]+)_\d+_\d+(?:_.+)?$/;
+	// Newer balenaOS container names just use commit
+	//    main_a000b111c222d333e444f555a666b777
+	//    main_10ca12e1ea5e
+	const nameRegex =
+		/(?:^|\/)([a-zA-Z0-9_-]+?)_(?:\d+_)?(?:\d+_)?(?:[a-f0-9]*)?$/;
 
 	const serviceNames: string[] = [];
 	const containerNames: string[] = [];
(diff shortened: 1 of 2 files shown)


lib/utils/device/ssh.ts
```diff
@@ -76,13 +76,21 @@ export async function getContainerIdForService(
...
 	//    main_1234567_2345678
 	// New balenaOS container name patterns:
 	//    main_1234567_2345678_a000b111c222d333e444f555a666b777
 	//    main_1_1_localrelease
-	const nameRegex = /(?:^|\/)([a-zA-Z0-9_-]+)_\d+_\d+(?:_.+)?$/;
+	// Newer balenaOS container names just use commit
+	//    main_a000b111c222d333e444f555a666b777
+	//    main_10ca12e1ea5e
+	const nameRegex =
+		/(?:^|\/)([a-zA-Z0-9_-]+?)_(?:\d+_)?(?:\d+_)?(?:[a-f0-9]*)?$/;
```
thgreasi, line 93: How about v which is slightly stricter?
```suggestion
		/(?:^|\/)([a-zA-Z0-9_-]+?)_(?:(?:\d+_\d+(?:_[a-f0-9]+)?)|(?:[a-f0-9]+))$/;
```
As far as I can tell though a user naming a service `december_12_2023` would make the name ambiguous :thinking:

lib/utils/device/ssh.ts
```diff
@@ -76,13 +76,21 @@ export async function getContainerIdForService(
 		.filter((l) => l);
 
 	const { escapeRegExp } = await import('lodash');
-	const regex = new RegExp(`(?:^|\\/)${escapeRegExp(opts.service)}_\\d+_\\d+`);
+	const regex = new RegExp(
+		`(?:^|\\/)${escapeRegExp(
+			opts.service,
+		)}_(?:\\d+_)?(?:\\d+_)?(?:[a-f0-9]*)?$`,
```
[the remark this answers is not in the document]
pipex: Is not really an OR, it can have numeric ids, numeric ids with commit or just commit. I think I can merge the numeric matching though

lib/utils/device/ssh.ts
pipex: Nice catch, unfortunately that doesn't work either :(

```
'december_12_2023_11_22'.match(/(?:^|\/)([a-zA-Z0-9_-]+?)_(?:(?:\d+_\d+(?:_[a-f0-9]+)?)|(?:[a-f0-9]+))$/)
[
  'december_12_2023_11_22',
  'december_12',
  index: 0,
  input: 'december_12_2023_11_22',
  groups: undefined
]
```

pipex: Without greedy `?` it matches the first number for some reason 

```
'december_12_2023_11_22'.match(/(?:^|\/)([a-zA-Z0-9_-]+)_(?:(?:\d+_\d+(?:_[a-f0-9]+)?)|(?:[a-f0-9]+))$/)
[
  'december_12_2023_11_22',
  'december_12_2023_11',
  index: 0,
  input: 'december_12_2023_11_22',
  groups: undefined
]
```

pipex: Testing alternatives

pipex: Hmm, 'december_12_2023_11_22` matches both syntaxes 

name 'december_12'
first id '2023'
second id '11'
commit '22'

It also matches 

name 'december_12_2023'
first id '2023'
second id '2023'
commit 'none'

pipex: I suspect the first one is the least of evils as container names have a release commit for a while

pipex: What do you think @thgreasi ?

pipex: Alternatively since at this point we alredy have access to the device, we could test depending on SV version

PR #2667, opened 2023-08-25
Update to Node 18

otaviojacobi:
Change-type: major

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/utils/remote-build.ts
+++ b/lib/utils/remote-build.ts
@@ -167,7 +167,7 @@ async function handleHeadlessBuildStream(
 	// been started
 	let message: HeadlessBuilderMessage;
 	try {
-		const response = await streamToPromise(stream as NodeJS.ReadWriteStream);
+		const response = await streamToPromise(stream as NodeJS.ReadStream);
 		message = JSON.parse(response.toString());
 	} catch (e) {
 		if (e.code === 'SIGINT') {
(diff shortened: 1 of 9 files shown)


lib/utils/remote-build.ts
```diff
@@ -419,7 +419,7 @@ async function getRemoteBuildStream(
 	if (build.opts.headless) {
 		stream = buildRequest;
 	} else {
-		stream = buildRequest.pipe(JSONStream.parse('*'));
+		stream = buildRequest.pipe(JSONStream.parse('*')) as NodeJS.ReadStream;
```
pipex, line 422: I don't love this extra casts, but the types from request are a bit weird to start with

otaviojacobi: @pipex I also don't link this extra one, the other one already existed and is now a little bit better (as a ReadStream is slightly more conservative than the ReadWriteStream) but I don't think we have many other options

lib/utils/remote-build.ts
```diff
@@ -167,7 +167,7 @@ async function handleHeadlessBuildStream(
 	// been started
 	let message: HeadlessBuilderMessage;
 	try {
-		const response = await streamToPromise(stream as NodeJS.ReadWriteStream);
+		const response = await streamToPromise(stream as NodeJS.ReadStream);
```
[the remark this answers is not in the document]
otaviojacobi: Well there is a ReadWriteStream, it is just that Stream.Stream is not compatcible with NodeJS.ReadWriteStream but only with NodeJS.ReadStream

PR #2668, opened 2023-08-25
patch: package all dependencies in the single binary

aethernet:
pkg now seems to support including the node native deps inside the bundled executable. 
This will need deep testing as there's no guarantee it will work.

The deps are most probably required for the etcher/flashing functionality so that one needs to be double checked.


aethernet:
Not that simple...

aethernet:
false direction.

This doesn't work at all. An alternative to make it work would be to switch from pkg to nexe, but the complexity of the migration is ... unknown.

The fact that it works with "some" packages bundled, is probably because they're either useless on apple silicon or used only on very specific actions.

PR #2671, opened 2023-09-05
Stop depending on old oclif libs and migrate to @oclif/core directly

otaviojacobi:
Change-type: major

Context on this PR: We are currently using [@oclif/command](https://github.com/oclif/command) [@oclif/config](https://github.com/oclif/config) and [@oclif/parser](https://github.com/oclif/parser) libs which were all deprecated in favor of [@oclif/core](https://github.com/oclif/core) there are several things we get from bumping our version such as [improved CLI performance](https://oclif.io/blog/2022/01/12/announcing-oclif-v2/#async-command-parsing) and a much much better support to Typescript (which allowed us to drop several interfaces we had to manually keep).

This PR moves us away from using deprecated code. 99% of it is done following [this migration guide](https://github.com/oclif/core/blob/main/MIGRATION.md) which changes `flags` to `Flags` and the whole structure of `Arguments`.

The patches we did on [@oclif/parser](https://github.com/balena-io/balena-cli/blob/master/patches/all/%40oclif%2Bparser%2B3.8.17.patch) have been ported to @oclif/core patch. The patch on @oclif/config could be dropped as this was fixed upstream and also we could also remove one patch on oclif itself as support for windows --targets build was fixed on upstream https://github.com/oclif/oclif/pull/1181 too.

There are no changes to any of the tests. There is one small change in order to make the tests run with the new @oclif/core format [here](https://github.com/balena-io/balena-cli/pull/2671/files#diff-2e593ff8544019efbea6d351823335105f80cb2d9a734486fc9d6ba5d7b1cf9aR120)

---


docs/balena-cli.md
```diff
@@ -1660,7 +1660,7 @@ Examples:
 
 ### Arguments
 
-#### COMMAND
+#### 0
```
otaviojacobi, line 1663: TODO: fix it

lib/commands/config/generate.ts
```diff
@@ -26,26 +27,6 @@ import {
 } from '../../utils/messages';
 import type { BalenaSDK, PineDeferred } from 'balena-sdk';
 
-interface FlagsDef {
```
otaviojacobi, line 29: @thgreasi @myarmolinsky This is one example of one of the big wins in this new version. the static "flags" types now are inffered from what you added on your flags so you get type checking working fine on your flags ("options") without manually needing to cast to what are the types of the flags

lib/commands/envs.ts
```diff
@@ -23,14 +24,7 @@ import * as cf from '../utils/common-flags';
...
 
-interface FlagsDef {
-	fleet?: string;
-	config: boolean;
-	device?: string; // device UUID
-	json: boolean;
-	help: void;
-	service?: string; // service name
-}
+type FlagsDef = Interfaces.InferredFlags<typeof EnvsCmd.flags>;
```
otaviojacobi, line 27: @thgreasi @myarmolinsky  Also here is one example in case we need to create a function that receives the parsed type of the flags we can use the new (and very userfull) `Interfaces.InferredFlags` instead of manually mantaining types, we let oclif do the dirty job for us.

lib/utils/normalization.ts
```diff
@@ -78,6 +78,6 @@ export async function disambiguateReleaseParam(
 /**
  * Convert to lowercase if looks like slug
  */
-export function lowercaseIfSlug(s: string) {
+export async function lowercaseIfSlug(s: string) {
```
thgreasi, line 81: Do custom `parse`rs need to be async and fail otherwise?

thgreasi (commented):
looks fine :+1:

otaviojacobi:
LGTM

PR #2672, opened 2023-09-05
Bump @balena/lint

myarmolinsky:
Bump @balena/lint

---


lib/app.ts
```diff
@@ -24,6 +24,8 @@ import {
 } from './preparser';
 import { CliSettings } from './utils/bootstrap';
 import { onceAsync } from './utils/lazy';
+import events = require('events');
+import * as BalenaSdk from 'balena-sdk';
```
otaviojacobi, line 28: @myarmolinsky I don't think we want/can move these imports to top level. This means that this code will be loaded everytime (even when it is not used) as long as any function is imported, which can impact the load time of the CLI. We want to have this as an import or await import but inside the function where it is used.

lib/utils/compose_ts.ts
```diff
@@ -1312,7 +1311,7 @@ async function pushAndUpdateServiceImages(
 			serviceImage.status = 'failed';
 			throw error;
 		} finally {
-			await afterEach(serviceImage, props);
+			afterEach(serviceImage, props);
```
thgreasi, line 1314: If you check how this function is called, you will realize the await here is necessary, so let's add the await back
```suggestion
			await afterEach(serviceImage, props);
```
and change the function args to:
```ts
	afterEach: (
		serviceImage: import('@balena/compose/dist/release/models').ImageModel,
		props: object,
	) => Promise<void>,
```

automation/deploy-bin.ts
```diff
@@ -41,7 +41,11 @@ export async function createGitHubRelease() {
 			assets: finalReleaseAssets[process.platform],
 		}),
 	);
-	console.log(`Release ${version} successful: ${ghRelease.html_url}`);
+	console.log(
+		`Release ${version} successful: ${
+			(ghRelease as { html_url: any; [key: string]: any }).html_url
+		}`,
+	);
```
thgreasi, line 48: Prefer to type as early as possible
```suggestion
	// const ghRelease = (await Bluebird.fromCallback(
	// ...
	)) as { html_url: any }; // (X)OR Record<string, string>, but not both, since the later one would supersede the rirst
	console.log(`Release ${version} successful: ${ghRelease.html_url}`);
```

thgreasi (commented):
Other than these I don't see any blocker :+1:

PR #2688, opened 2023-10-20
Support empty string values for `env add` and `tag set`

thgreasi:
Chang-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/CLI.3A.20Unexpected.20argument.20when.20setting.20empty.20env.20var.20value


thgreasi:
I self-certify!

otaviojacobi (approved):

PR #2689, opened 2023-10-25
Add device start-service and stop-service commands

mlveggo:
Resolves: #2690
Change-type: minor

---

### DIFF
--- a/lib/commands/device/start-service.ts
+++ b/lib/commands/device/start-service.ts
@@ -1,0 +1,139 @@
+/**
+ * @license
+ * Copyright 2016-2020 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Args } from '@oclif/core';
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, getCliUx, stripIndent } from '../../utils/lazy';
+import type { BalenaSDK } from 'balena-sdk';
+
+export default class DeviceStartServiceCmd extends Command {
+	public static description = stripIndent`
+		Start containers on a device.
+
+		Start containers on a device.
+
+		Multiple devices and services may be specified with a comma-separated list
+		of values (no spaces).
+		`;
+	public static examples = [
+		'$ balena device start-service 23c73a1 myService',
+		'$ balena device start-service 23c73a1 myService1,myService2',
+	];
+
+	public static args = {
+		uuid: Args.string({
+			description: 'comma-separated list (no blank spaces) of device UUIDs',
+			required: true,
+		}),
+		service: Args.string({
+			description: 'comma-separated list (no blank spaces) of service names',
+			required: true,
+		}),
+	};
+
+	public static usage = 'device start-service <uuid>';
+
+	public static flags = {
+		help: cf.help,
+	};
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params } = await this.parse(DeviceStartServiceCmd);
+
+		const balena = getBalenaSdk();
+		const ux = getCliUx();
+
+		const deviceUuids = params.uuid.split(',');
+		const serviceNames = params.service.split(',');
+
+		// Iterate sequentially through deviceUuids.
+		// We may later want to add a batching feature,
+		// so that n devices are processed in parallel
+		for (const uuid of deviceUuids) {
+			ux.action.start(`Starting services on device ${uuid}`);
+			await this.startServices(balena, uuid, serviceNames);
+			ux.action.stop();
+		}
+	}
+
+	async startServices(
+		balena: BalenaSDK,
+		deviceUuid: string,
+		serviceNames: string[],
+	) {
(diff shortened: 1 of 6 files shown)


maxekman:
This would be really useful. Thanks @mlveggo!

otaviojacobi:
Hello @mlveggo and thanks for the contribution. Could you please `git commit --amend` your commit adding the Change-type footer so CI can run? This will look something like
```
Add device startservice and stopservice commands

Change-type: minor
```

Note that there is a blank line between the commit message and the footer. Thanks!

lib/commands/device/startservice.ts
```diff
@@ -0,0 +1,155 @@
...
+		`;
+	public static examples = [
+		'$ balena device startservice 23c73a1 --service myService',
+		'$ balena device startservice 23c73a1 -s myService1,myService2',
+	];
+
+	public static args = {
+		uuid: Args.string({
+			description:
+				'comma-separated list (no blank spaces) of device UUIDs to start',
```
myarmolinsky, line 47: nit:
```suggestion
				'comma-separated list (no blank spaces) of device UUIDs',
```

lib/commands/device/startservice.ts
```diff
@@ -0,0 +1,155 @@
...
+
+		const balena = getBalenaSdk();
+		const ux = getCliUx();
+
+		const deviceUuids = params.uuid.split(',');
+		const serviceNames = options.service?.split(',');
+
+		// Iterate sequentially through deviceUuids.
+		// We may later want to add a batching feature,
+		// so that n devices are processed in parallel
```
myarmolinsky, line 78: +1
I have a PR from a while ago that aims to begin adding that support here: https://github.com/balena-io/pinejs/pull/675
Haven't had the time to pushed it to the end, hopefully will find the time soon

Related roadmap item in case you are interested: https://roadmap.balena.io/posts/19/allow-changes-to-variables-to-be-batched-and-submitted-all-at-once

lib/commands/device/startservice.ts
```diff
@@ -0,0 +1,155 @@
...
+				},
+			});
+		} catch (e) {
+			const { BalenaDeviceNotFound } = await import('balena-errors');
+			if (instanceOf(e, BalenaDeviceNotFound)) {
+				throw new ExpectedError(`Device ${deviceUuid} not found.`);
+			} else {
+				throw e;
+			}
+		}
```
myarmolinsky, line 111: Probably better we just wrap the call to `this.startServices()` in the trycatch. Also I don't think we need to explicitly catch the `BalenaDeviceNotFound` error, do we? An error of that instance should already have a relevant error message. Furthermore, we probably don't need the trycatch at all unless we want to output some specific text for the user (I expect you added it to manipulate the thrown error, right? Which as noted I don't think is necessary unless I'm missing something 🤔 )
```suggestion
		const device = await balena.models.device.getWithServiceDetails(deviceUuid, {
				$expand: {
					is_running__release: { $select: 'commit' },
				},
			});
```

lib/commands/device/startservice.ts
```diff
@@ -0,0 +1,155 @@
...
+
+		const activeRelease = getExpandedProp(device.is_running__release, 'commit');
```
myarmolinsky, line 113: To be clearer about what this const contains:
```suggestion
		const activeReleaseCommit = getExpandedProp(device.is_running__release, 'commit');
```

myarmolinsky (commented):
Great PR and feature addition, happy to have a community contribution! I've left some comments I would apply to both the start and stop service commands. My colleague will review as well when he has a chance

lib/commands/device/startservice.ts
```diff
@@ -0,0 +1,155 @@
...
+
+	public static args = {
+		uuid: Args.string({
+			description:
+				'comma-separated list (no blank spaces) of device UUIDs to start',
+			required: true,
+		}),
+	};
+
+	public static usage = 'device startservice <uuid>';
```
myarmolinsky, line 52: Would strongly prefer for this command to be `start-service` instead of `startservice` as we do with all our other commands

lib/commands/device/startservice.ts
mlveggo: Nice. Thanks for the information. Good to get a feel about the roadmap ahead.

lib/commands/device/start-service.ts
```diff
@@ -0,0 +1,144 @@
...
+		}),
+	};
+
+	public static usage = 'device start-service <uuid>';
+
+	public static flags = {
+		service: Flags.string({
+			description: 'comma-separated list (no blank spaces) of service names',
+			char: 's',
+		}),
```
thgreasi, line 52: Should we actually mark the service flag as required (both here in in the stop command), given that we do not support this command without it afaict?

```suggestion
		service: Flags.string({
			description: 'comma-separated list (no blank spaces) of service names',
			char: 's',
			required: true,
		}),
```
and then we can assume that `serviceNames` & `options.service` can't be undefined.

mlveggo: good suggestion. have made the changes.

lib/commands/device/startservice.ts
mlveggo: fixed

lib/commands/device/start-service.ts
```diff
@@ -0,0 +1,141 @@
...
+
+		const deviceUuids = params.uuid.split(',');
+		const serviceNames = params.service.split(',');
+
+		// Iterate sequentially through deviceUuids.
+		// We may later want to add a batching feature,
+		// so that n devices are processed in parallel
+		for (const uuid of deviceUuids) {
+			ux.action.start(`Starting services on device ${uuid}`);
+			if (serviceNames) {
```
otaviojacobi, line 71: @mlveggo I think this can be dropped now since params.service is mandatory serviceNames will always be truthy

mlveggo: true, leftover from flags. fixed.

myarmolinsky:
Seems an `npm dedupe` is needed to continue

myarmolinsky (approved):

PR #2692, opened 2023-10-27
device: Add `--json` option for JSON output

bbugh:
Change-type: minor

Continuation of #2685 where we're using the CLI to automate targeting releases to certain devices based on the device tag, this adds `--json` to the `device` command, and includes `device_tag`.

### DIFF
--- a/lib/commands/device/index.ts
+++ b/lib/commands/device/index.ts
@@ -20,6 +20,7 @@ import Command from '../../command';
 import * as cf from '../../utils/common-flags';
 import { expandForAppName } from '../../utils/helpers';
 import { getBalenaSdk, getVisuals, stripIndent } from '../../utils/lazy';
+import { jsonInfo } from '../../utils/messages';
 
 import type { Application, Release } from 'balena-sdk';
 
@@ -45,10 +46,13 @@ export default class DeviceCmd extends Command {
 		Show info about a single device.
 
 		Show information about a single device.
+
+		${jsonInfo.split('\n').join('\n\t\t')}
 		`;
 	public static examples = [
 		'$ balena device 7cf02a6',
 		'$ balena device 7cf02a6 --view',
+		'$ balena device 7cf02a6 --json',
 	];
 
 	public static args = {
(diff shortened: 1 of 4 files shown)


otaviojacobi:
Hello @bbugh sorry for the delay on this, I am adding this to my to do list for this week. Can you please rebase and fix the docs conflict?

bbugh:
Fixed the docs conflict

lib/commands/device/index.ts
```diff
@@ -76,33 +81,45 @@ export default class DeviceCmd extends Command {
...
+							'storage_block_device',
+							'storage_usage',
+							'storage_total',
+							'cpu_usage',
+							'cpu_temp',
+							'cpu_id',
+							'is_undervolted',
+						],
+						...expandForAppName,
+				  },
```
thgreasi, line 121: Tiny suggestion so that the `$select` is still used when the `--json` flag is used.
```suggestion
			{
				$select: [
					'device_name',
					'id',
					'overall_status',
					'is_online',
					'ip_address',
					'mac_address',
					'last_connectivity_event',
					'uuid',
					'supervisor_version',
					'is_web_accessible',
					'note',
					'os_version',
					'memory_usage',
					'memory_total',
					'public_address',
					'storage_block_device',
					'storage_usage',
					'storage_total',
					'cpu_usage',
					'cpu_temp',
					'cpu_id',
					'is_undervolted',
				],
				$expand: {
					...expandForAppName.$expand,
					...(options.json && {
						device_tag: {
							$select: ['tag_key', 'value'],
						},
					})
				},
			},
```

thgreasi: Alternatively in case not including the $select was intentional, we could make clear that it was intentional by wrapping it in a `...(!options.json && { $select: [/*...*/] }),` along w/ a comment.

bbugh: @thgreasi thanks for the review. 

It was intentional not to limit the json response based on the fields used by the table printer. JSON format in other tools is a raw data dump that allows the consumer to determine what data is important, which is one of the primary use cases for it. (See also similar discussion in my first PR). 

I wouldn't personally add anything to the code because I think the intent of the code is clear based on the use case and experience, but I respect that it's y'all's codebase and I think you can commit that comment to it if you want. 

Hope that helps.

thgreasi (approved):

PR #2693, opened 2023-10-27
devices: add device_tag to JSON output

bbugh:
change-type: minor

Hello again! 👋

Related to #2692 and #2685, adding `device_tag` expansion to the json output of the `devices` command, so we can query for devices with a specific device tag to pin a specific release to it.


tests/commands/device/devices.spec.ts
```diff
@@ -59,4 +59,30 @@ describe('balena devices', function () {
...
+				'/v6/device?$orderby=device_name%20asc&$select=id,uuid,device_name,status,is_online,supervisor_version,os_version&$expand=device_tag($select=tag_key,value),belongs_to__application($select=app_name,slug),is_of__device_type($select=slug),is_running__release($select=commit)',
+			)
+			.replyWithFile(200, path.join(apiResponsePath, 'devices.json'), {
+				'Content-Type': 'application/json',
+			});
+
+		const { out, err } = await runCommand('devices --json');
+		expect(err).to.be.empty;
+		const json = JSON.parse(out.join(''));
+		console.log(json);
```
otaviojacobi, line 75: hello @bbugh This is very minor but would you mind removing this `console.log` ? I would commit amend over it but because the origin is your fork I don't have permissions.

Other than that, looks good :)

lib/commands/devices/index.ts
```diff
@@ -116,7 +123,9 @@ export default class DevicesCmd extends Command {
 
 		if (options.json) {
 			const { pickAndRename } = await import('../../utils/helpers');
-			const mapped = devices.map((device) => pickAndRename(device, fields));
+			const mapped = devices.map((device) =>
+				pickAndRename(device, [...fields, 'device_tag']),
```
otaviojacobi, line 127: @thgreasi You probably have more context than I do on this, should the, now being expanded, device_tag have the pickAndRename applied to it? AFAIK this only applies in case of something being formatted with 'a => b' and we probably wouldn't want to change it if this is the tag value of the client?

tests/commands/device/devices.spec.ts
bbugh: Oh, sorry about that! That was definitely a mistake to leave in, thanks for catching it. I pushed a fix.

bbugh:
Coming back around to needing this in CI soon, is there anything else I can do to support this process? Thank you!

lib/commands/devices/index.ts
thgreasi: @otaviojacobi I think that we only use `pickAndRename` in this file for consistency reasons and that in this file using `pick` or `pickAndRename` has the same outcome since there is no `=>` used in the defined fields.

thgreasi (approved):

thgreasi:
@bbugh could you please also squash all commits and rebase your branch on top of master before merging so that the changelog only includes just one entry?

thgreasi (changes requested):
Formally requesting changes for squashing the commits just so that CI doesn't accidentally merge this before you do so.

bbugh:
> @bbugh could you please also squash all commits and rebase your branch on top of master before merging so that the changelog only includes just one entry?

Done, squashed and rebased to master @ 801a25995c06bce2c279cb7cb496e90c5a131bd7

PR #2712, opened 2023-12-11
deploy: Add rate-limiting aware retries for failed requests

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Improvement/Add-rate-limiting-aware-retries-to-the-CLI,-builder-etc-1381

### DIFF
--- a/lib/utils/compose.ts
+++ b/lib/utils/compose.ts
@@ -20,6 +20,7 @@ import type * as SDK from 'balena-sdk';
 import type Dockerode = require('dockerode');
 import * as path from 'path';
 import type { Composition, ImageDescriptor } from '@balena/compose/dist/parse';
+import type { RetryParametersObj } from 'pinejs-client-core';
 import type {
 	BuiltImage,
 	ComposeOpts,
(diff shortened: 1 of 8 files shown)


lib/utils/compose.ts
```diff
@@ -94,22 +95,60 @@ export function createProject(
...
+			require('@balena/env-parsing') as typeof import('@balena/env-parsing');
+		return intVar(varName);
+	};
+	// We use the BALENARCTEST namespace and only parse the env vars while in test mode
+	// since we plan to switch all pinejs clients with the one of the SDK and might not
+	// BALENARCTEST to have to support these env vars.
+	return {
+		minDelayMs: testIntVar('BALENARCTEST_API_RETRY_MIN_DELAY_MS') ?? 1000,
+		maxDelayMs: testIntVar('BALENARCTEST_API_RETRY_MAX_DELAY_MS') ?? 60000,
+		maxAttempts: testIntVar('BALENARCTEST_API_RETRY_MAX_ATTEMPTS') ?? 7,
```
thgreasi, line 117: Since on CI the tests run the cli on a separate sub-process, we can't mock these values, so the easiest way to workaround that was to add & start parsing a "tests only" env var (which we can later choose to surface to users if we want, but check the comment on why I think we shouldn't).

lib/utils/compose.ts
```diff
@@ -94,22 +95,62 @@ export function createProject(
...
+			minDelayMs: intVar('BALENARCTEST_API_RETRY_MIN_DELAY_MS'),
+			maxDelayMs: intVar('BALENARCTEST_API_RETRY_MAX_DELAY_MS'),
+			maxAttempts: intVar('BALENARCTEST_API_RETRY_MAX_ATTEMPTS'),
+		};
+	}
+
+	return {
+		minDelayMs: 1000,
+		maxDelayMs: 60000,
+		maxAttempts: 7,
```
otaviojacobi, line 119: Shouldn't we allow for customizing it? e.g. `minDelayMs: intVar('BALENARC_API_RETRY_MIN_DELAY_MS') ?? 1000,`

thgreasi: The comment right above is trying to explain that part:
```
		// We use the BALENARCTEST namespace and only parse the env vars while in test mode
		// since we plan to switch all pinejs clients with the one of the SDK and might not
		// want to have to support these env vars.
```
given that we will probably adopt the sdk's retry mechanism, and I'm not sure whether at that point we would want to support these env vars, I though it might be better for now to not expose them. If we change our mind later and decide that we do want them, we can do so easily in a minor.

tests/commands/deploy.spec.ts
```diff
@@ -289,6 +289,8 @@ describe('balena deploy', function () {
 		api.expectPatchImage({
 			replyBody: errMsg,
 			statusCode: 500,
+			// b/c failed requests are retried
+			persist: true,
```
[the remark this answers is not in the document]
thgreasi: Fine @otaviojacobi ... You made me write all tests that I wanted to leave for after this building cycle...

otaviojacobi (approved):

PR #2718, opened 2024-01-05
Add note about global flag in advanced install.

pipex:
Change-type: patch


pipex:
Not sure if the `--global` flag actually uses `npmshrinkwrap` versus non global

Issue #2715, opened 2024-01-01
Balena push does not build in local mode when compose file specifies a build context


joelmarty:
I created a project using a docker-compose file to deploy on a raspberry pi zero w device.
The compose file specifies one service which is build from a subdirectory in my repo.
When switching the device to local mode and deploying using `balena push`, the command hangs indefinitely and the image is never built.

The folder structure is as follows:
```
repo
|_ pihole
  |_ Dockerfile
  |_ ...
|_ docker-compose.yml
```

# Expected Behavior

`balena push <my device ip>` builds the project

# Actual Behavior

`balena -push x.x.x.x` hangs.
Here is a log trace with debug logging enabled:
```
$ balena --debug push x.x.x.x 
[debug] new argv=[/home/joel/opt/balena-cli/balena,/snapshot/balena-cli/bin/balena,push,x.x.x.x] length=4
[debug] Deprecation check: 3.06398 days since last npm registry query for next major version release date.
[debug] Will not query the registry again until at least 7 days have passed.
[Debug]   Using build source directory: . 
[Debug]   Pushing to local device: x.x.x.x
[Debug]   Checking we can access device
[Debug]   Sending request to http://x.x.x.x:48484/ping
[debug] Event tracking error: Timeout awaiting 'response' for 0ms
[Debug]   Checking device supervisor version: 15.0.4
[Info]    Starting build on device x.x.x.x
[Debug]   Loading project...
[Debug]   Resolving project...
[Debug]   docker-compose.yml file found at "."
[Debug]   Creating project...
[Debug]   Tarring all non-ignored files...
[Debug]   Tarring complete in 1090 ms
[Debug]   Fetching device information...
[Debug]   Sending request to http://x.x.x.x:48484/v2/local/device-info
[Debug]   Found build tasks:
[Debug]       pihole: build [./pihole]
[Debug]   Resolving services with [raspberry-pi|rpi]
[Debug]   Found project types:
[Debug]       pihole: Standard Dockerfile
[Debug]   Probing remote daemon for cache images
[Debug]   Using 3 on-device images for cache...
[Debug]   Starting builds...
[Debug]   pihole: Using platform option for build: linux/arm/v6
```
The build does not seem to be actually starting on the device and the command hangs there.

# Steps to Reproduce the Problem

Use a compose file defining a build context instead of referencing an external image:
```
version: "2.4"

services:
  pihole:
    build: pihole
```
[...]

PR #2719, opened 2024-01-13
Revert upgrade to dockerode to avoid regression

kb2ma:
Regression is `balena push` hangs in local mode. The problem was introduced with 0ba3522 in v17.4.7.

The ultimate fix is to upgrade use of dockerode to v4.0.x in balena modules that use v3.3.x. v4.0 of dockerode includes a [fix](https://github.com/apocas/dockerode/commit/f0a5338418de6292eb37d1a3ccc3f093eaedd506) for the dockerode regression introduced in v3.3.5 that is the source of the failure described in the GH issue below.

However, that upgrade requires some coordination among modules. This PR removes the immediate regression for balenaCLI.

Resolves: #2715  
Change-type: patch

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -40,7 +40,7 @@
         "denymount": "^2.3.0",
         "docker-modem": "3.0.0",
         "docker-progress": "^5.1.3",
-        "dockerode": "3.3.5",
+        "dockerode": "3.3.3",
         "ejs": "^3.1.6",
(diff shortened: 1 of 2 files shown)


kb2ma:
Also a test target build requested an `npm dedupe`, and that is included in the `npm-shrinkwrap.json`.

kb2ma:
Build OK now. I'll perform some manual tests before removing draft status.

pipex:
LGTM

Issue #2724, opened 2024-01-27
TypeError when Livepush rebuilds one service of several


kb2ma:
# Expected Behavior
Livepush should not generate an error when modifying a Dockerfile in the attached setup.

[type-error.zip](https://github.com/balena-io/balena-cli/files/14070736/type-error.zip)

# Actual Behavior
When modify one of the Dockerfiles, see a message like below.

```
$ balena push 192.168.1.1
...
[Live]    Device state settled
```
Modify Dockerfile
```
[Live]    Detected Dockerfile change, performing full rebuild of service alpine
...
[Build]   [alpine] Successfully built 33d946754ae7
[Build]   [alpine] Successfully tagged local_alpine:latest
[Logs]    [2024-01-27T02:07:40.884Z] Service exited 'alpine sha256:387ac3ddc40b1254e4107bdf7ea520df7673de4767a8a0ce1b6e6059e8263c7e'
[Error]   There was an error rebuilding the service: TypeError: Cannot read properties of undefined (reading 'external')
```
Collected a stack trace for the error:
```
[Debug]   TypeError: Cannot read properties of undefined (reading 'external')
[Debug]       at getImageNameFromTask (/home/kbee/dev/balena-cli/repo/build/utils/device/deploy.js:349:18)
[Debug]       at /home/kbee/dev/balena-cli/repo/build/utils/device/deploy.js:380:24
[Debug]       at /home/kbee/dev/balena-cli/repo/node_modules/lodash/lodash.js:4967:15
[Debug]       at baseForOwn (/home/kbee/dev/balena-cli/repo/node_modules/lodash/lodash.js:3032:24)
[Debug]       at /home/kbee/dev/balena-cli/repo/node_modules/lodash/lodash.js:4936:18
[Debug]       at Function.forEach (/home/kbee/dev/balena-cli/repo/node_modules/lodash/lodash.js:9410:14)
[Debug]       at generateTargetState (/home/kbee/dev/balena-cli/repo/build/utils/device/deploy.js:357:7)
[Debug]       at LivepushManager.handleServiceRebuild (/home/kbee/dev/balena-cli/repo/build/utils/device/live.js:231:76)
[Debug]       at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[Debug]       at async LivepushManager.handleFSEvents (/home/kbee/dev/balena-cli/repo/build/utils/device/live.js:170:13)
```

# Analysis
[...]

PR #2725, opened 2024-01-30
Fix target state construction with livepush

pipex:
When constructing the target state after a reported change from livepush, the handler function would not pass all build tasks to the function that constructs the target state, causing a TypeError when trying to obtain the target image name for each service. This updates the handler to pass all build tasks, ensuring the information is available to construct the target state.

Relates-to: #2724
Change-type: patch


kb2ma (approved):
Seems like a straightforward fix. Tested with 1 and 2 Dockerfiles. Nice commit comment.

PR #2738, opened 2024-03-08
Remove oclif patches that are no longer needed

otaviojacobi:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---


.github/actions/publish/action.yml
```diff
@@ -113,6 +119,7 @@ runs:
             PATH="/c/Program Files (x86)/Windows Kits/10/bin/${runner_arch}:${PATH}"
         fi
 
+        npm i --no-save oclif
```
otaviojacobi, line 122: The reason this is here and not in a devDependency is because: https://github.com/oclif/oclif/issues/1324 and shrinkwrap is included in the final bundle, so this avoids +5MB on the final bundle and our CI breaking because the file ends up being too large, it can go into devDependency oclif @aws-sdk is made optional

automation/build-bin.ts
```diff
@@ -60,10 +61,10 @@ const standaloneZips: PathByPlatform = {
 	win32: dPath(`balena-cli-${version}-windows-${arch}-standalone.zip`),
 };
 
-const oclifInstallers: PathByPlatform = {
-	darwin: dPath('macos', `balena-${version}.pkg`),
-	win32: dPath('win32', `balena-${version}-${arch}.exe`),
-};
+const getOclifInstallersOriginalNames = (sha: string): PathByPlatform => ({
```
otaviojacobi, line 64: Instead of patching oclif to make the naming the ones we want, we just get the name they create by default, this avoids a lot of mess up with the inner workings of oclif

completion/generate-completion.js
```diff
@@ -33,7 +33,7 @@ const commandsJson = JSON.parse(fs.readFileSync(commandsFilePath, 'utf8'));
 
 const mainCommands = [];
 const additionalCommands = [];
-for (const key of Object.keys(commandsJson.commands)) {
+for (const key of Object.keys(commandsJson.commands).sort()) {
```
otaviojacobi, line 36: Necessary to avoid intermitent fails

package.json
```diff
@@ -111,7 +111,7 @@
     ],
     "macos": {
       "identifier": "io.balena.cli",
-      "sign": "Developer ID Installer: Balena Ltd (66H43P8FRG)"
+      "sign": "\"Developer ID Installer: Balena Ltd (66H43P8FRG)\""
```
otaviojacobi, line 114: Removes one patching that was basically escaping strings

PR #2742, opened 2024-03-11
Use standard oclif run.js & dev.js

otaviojacobi:
Change-type: patch

This removes one more patch & allows us to more easily use libraries that expect the entrypoint to be `bin/run.js` (e.g auto update library). Tested with standalone, installers & npm installs.

The initial reason why we have renamed is mostly in here https://github.com/balena-io/balena-cli/pull/1285/files because before using oclif we had a `./bin/balena` which I just made become a symlink to the original run.js file.

Resolves: #   
Depends-on: <url>   
See: <url>   

---


bin/balena
otaviojacobi, line 1: Symlink so when working/debugging locally we can still do `./bin/balena`

bin/balena-dev
otaviojacobi, line 1: Symlink so when working/debugging locally we can still do `./bin/balena-dev`

bin/dev.js
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env node
```
otaviojacobi, line 1: This is just the exact same content of previous bin/balena-dev

bin/run.js
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
```
otaviojacobi, line 1: This is just the exact same content of previous bin/balena

myarmolinsky (approved):

PR #2743, opened 2024-03-11
Move macos binary signing to oclif pretarball lifecycle

otaviojacobi:
Change-type: patch

This is mostly removing the oclif patch to sign binaries and doing it on the `pretarball` step from oclif lifecycle, as recommended by oclif. This doesn't effectively change anything on any final binary and it was not done this way before because when it was created the `pretarball` life cycle step did not exist and no one realized it could be used after it was created. I manually tested on MacOS, linux & windows and everything seemed ok (including the use of the signed binaries)

---

### DIFF
--- a/.github/actions/publish/action.yml
+++ b/.github/actions/publish/action.yml
@@ -44,6 +44,12 @@ runs:
         node-version: ${{ inputs.NODE_VERSION }}
         cache: npm
 
+    - name: Set up Python 3.11
+      if: runner.os == 'macOS'
+      uses: actions/setup-python@d27e3f3d7c64b4bbf8e4abfb9b63b83e846e0435 # v4
+      with:
+        python-version: "3.11"
+
     - name: Install additional tools
       if: runner.os == 'Windows'
       shell: bash
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "scripts": {
     "postinstall": "node patches/apply-patches.js",
     "prebuild": "rimraf build/ build-bin/",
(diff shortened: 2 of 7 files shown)


.github/actions/publish/action.yml
```diff
@@ -44,6 +44,12 @@ runs:
         node-version: ${{ inputs.NODE_VERSION }}
         cache: npm
 
+    - name: Set up Python 3.11
```
otaviojacobi, line 47: This is needed because of an issue with python 3.12 and node-gyp https://github.com/nodejs/node-gyp/issues/2869

I am still doing the real fix so we are able to drop it bubbling up PRs bumping node-gyp https://github.com/balena-io-modules/drivelist/pull/453

But because we still have a dependency on etcher-image-writer (which requires a major with breaking changes to remove) we still have old node-gyp stuff around, but being worked on

package.json
```diff
@@ -48,6 +48,7 @@
   "scripts": {
     "postinstall": "node patches/apply-patches.js",
     "prebuild": "rimraf build/ build-bin/",
+    "pretarball": "ts-node --transpile-only ../../automation/run.ts sign:binaries",
```
otaviojacobi, line 51: The path here needs to be relative to where oclif will be when running the BuildTar step (which will always be `<root>/tmp/balena`

otaviojacobi: Moving this into this step instead of patching the library was oclif's mantainer suggestion for our workflow: https://github.com/oclif/oclif/issues/1331

klutchell (approved):

thgreasi (approved):

PR #2748, opened 2024-04-01
Enable npm dedupe as part of Renovate postUpdateOptions

klutchell:
See: https://docs.renovatebot.com/configuration-options/#postupdateoptions
Change-type: patch

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4038,9 +4038,9 @@
       }
     },
     "node_modules/@types/node": {
-      "version": "20.12.6",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-20.12.6.tgz",
-      "integrity": "sha512-3KurE8taB8GCvZBPngVbp0lk5CKi8M9f9k1rsADh0Evdz5SzJ+Q+Hx9uHoFGsLnLnd1xmkDQr2hVhlA0Mn0lKQ==",
+      "version": "20.12.7",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-20.12.7.tgz",
+      "integrity": "sha512-wq0cICSkRLVaf3UGLMGItu/PtdY7oaXaI/RVU+xliKVOtRna3PRY57ZDfztpDL0n11vfymMUnXv8QwYCO7L1wg==",
       "dependencies": {
         "undici-types": "~5.26.4"
       }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,4 @@
+{
+	"extends": ["github>balena-io/renovate-config"],
+	"postUpdateOptions": ["npmDedupe"]
+}


thgreasi:
I think that the tests fail b/c the dependencies in the repo are atm outdated. If you `npm i && npm dd && npm i` (like the test-dedupe.sh does) and push the diff, then this PR should pass.
An possible alternative I spotted after reading the renovate docs could be to set `npm config set prefer-dedupe true`, which will end up in the .npmrc, and will result all `npm install`s to use `--prefer-dedupe`, potentially saving us from the extra `npm dd` command & from augmenting the renovate config.

thgreasi:
See: https://github.com/balena-io/balena-ui/pull/6518

thgreasi (approved):

klutchell:
@thgreasi would we prefer this renovate approach, or the package.json setting used over https://github.com/balena-io/balena-ui/pull/6518? I suspect the latter?

thgreasi:
@klutchell I think we need to test to which degree the UI PR (--prefer-dedupe) works.
For sure it makes sense when devs do a manual `npm i` of a new/updated package to have npm to try dedupe upfront, but I think we need to wait to check whether that will be enough for renovate PRs as well.

I'm also happy getting this in as is (but you need to do an `npm i && npm dd && npm i` to match what the CLI does in its tests - I don't know why it uses an extra `npm i` and it might make sense to remove it even) and wait and compare how those two approaches work for the UI & CLI.

klutchell:
When running `rm -rf node_modules && npm i && npm dd && npm i` I get the following error:

```text
Failed to apply some patches:
Error: 
**ERROR** Failed to apply patch for package @oclif/core at path
  
    node_modules/@oclif/core

  This error was caused because @oclif/core has changed since you
  made the patch file for it. This introduced conflicts with your patch,
  just like a merge conflict in Git when separate incompatible changes are
  made to the same piece of code.

  Maybe this means your patch file is no longer necessary, in which case
  hooray! Just delete it!

  Otherwise, you need to generate a new patch file.

  To generate a new one, just repeat the steps you made to generate the first
  one.

  i.e. manually make the appropriate file changes, then run 

    patch-package @oclif/core

  Info:
    Patch file: patches/@oclif+core+3.25.0.patch
    Patch was made for version: 3.25.0
    Installed version: 3.26.0

---
patch-package finished with 1 error(s).
```

thgreasi:
Let me do that for you @klutchell 
See: https://github.com/balena-io/balena-cli/pull/2749/files

otaviojacobi:
rebasing as this should've been fixed with https://github.com/balena-io/balena-cli/pull/2751

thgreasi:
I think the issue is that for some reason in the cli the maintainers had chosen to do `npm install && npm dedupe && npm install`. It might be worth it exploring what dropping the last `npm install` would result (like we have it in the UI).

thgreasi:
@klutchell ends up that you just need to do an `npm dd` and that is picking up an updated version of `@types/node`. Just commit that and rebase this and it should be fine.

npm-shrinkwrap.json
thgreasi, line 1: @klutchell it seems that your `npm dd` reverted the shrinkwrap back to locafileVersion 2. Can you please drop that commit, bump your npm and try `npm dd` again?

thgreasi (changes requested):

npm-shrinkwrap.json
thgreasi: let me do that for you b/c it's been sad how long this PR has taken to get merged...

otaviojacobi (approved):

thgreasi (approved):

PR #2754, opened 2024-04-15
Upgrade dockerode and docker-modem dependencies

kb2ma:
According to apocas/dockerode#751, the major version bumps to dockerode and docker-modem were due to a mistake in code that was merged. So there is not a significant change to functionality. However, this PR does include test fixes due to an interface change in docker-modem.

Froze @types/dockerode dependency at v3.3.23 due to incompatibility with changes in v3.3.26. DefinitelyTyped/DefinitelyTyped#68935 adds specification of ImageTagOptions. However, they do not include some options used by balenaCLI. This incompatibility should be addressed in that repo.


tests/utils/docker.spec.ts
```diff
@@ -38,8 +38,19 @@ describe('getDefaultDockerModemOpts() function', function () {
...
 		});
+		if (typeof defaultOps.socketPath === 'function') {
+			// Function is always findDefaultUnixSocket(), which returns a promise.
+			// Must override type since @types/dockerode not updated yet.
+			const socketPath: () => Promise<string> = defaultOps.socketPath;
+			socketPath()
+				.then((socketPathRes) =>
+					expect(socketPathRes).to.equal(defaultSocketPath),
+				)
+				.catch((e) => assert.fail(`socketPath() failed: ${e}`));
```
Page-, line 50: I don't think this will work as you expect since it's in a promise that is neither awaited nor returned, I would simply rewrite this with async/await to avoid the issue:
```suggestion
			try {
				expect(await socketPath()).to.equal(defaultSocketPath),
			} catch (e) {
				throw new Error(`socketPath() failed: ${e}`));
			}
```
I also switched `assert.fail` to `throw new Error` as it should be equivalent and avoids the need for the extra import. I would also remove the `catch` block altogether if the custom error message isn't necessary but that doesn't matter to me either way

tests/utils/docker.spec.ts
```diff
@@ -131,7 +142,18 @@ describe('generateConnectOpts() function', function () {
...
 		});
+		if (typeof connectOpts.socketPath === 'function') {
+			// Function is always findDefaultUnixSocket(), which returns a promise.
+			// Must override type since @types/dockerode not updated yet.
+			const socketPath: () => Promise<string> = connectOpts.socketPath;
+			socketPath()
+				.then((socketPathRes) =>
+					expect(socketPathRes).to.equal(defaultSocketPath),
+				)
+				.catch((e) => assert.fail(`socketPath() failed: ${e}`));
```
Page-, line 154: Same here with async/await

tests/utils/docker.spec.ts
kb2ma: I agree, much simpler. Also removed catch block.

tests/utils/docker.spec.ts
kb2ma: Updated here as well.

kb2ma:
Rebased, simplified tests, and one more dedupe. Let me know if I need to squash the simplification. Otherwise should be ready to go @Page- .

otaviojacobi (approved):

PR #2765, opened 2024-05-14
Pluralize command categories in docs

dfunckt:
Pluralize command categories, eg. "Device" -> "Devices”.

Results in much fewer categories in our docs pages and we no longer have the somewhat silly separate singular and plural categories for commands such as `devices` and `device <command>`.

-------

Here's it looks now: https://github.com/balena-io/balena-cli/blob/c28039a3f2dbcf8ad9b7e7586382043296cd7b38/docs/balena-cli.md

Here's [how docs](https://docs.balena.io/reference/balena-cli/) look currently:

Screenshot 2024-05-14 at 17 26 49

### DIFF
--- a/automation/capitanodoc/capitanodoc.ts
+++ b/automation/capitanodoc/capitanodoc.ts
@@ -54,15 +54,15 @@ interface Documentation {
 
 // Mapping folders names to custom headings in the docs
 const commandHeadings: { [key: string]: string } = {
-	'api-key': 'API Key',
+	'api-key': 'API Keys',
 	'api-keys': 'API Keys',
 	login: 'Authentication',
 	whoami: 'Authentication',
 	logout: 'Authentication',
-	env: 'Environment Variable',
+	env: 'Environment Variables',
 	envs: 'Environment Variables',
 	help: 'Help and Version',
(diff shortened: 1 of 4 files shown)


shaunmulligan (approved):
Over all looks like a great improvement. I think we probably need to look at a wider UX improvement and potentially remove things like:
local config and list available drives.

I also think having `create application` and `create fleet` here in the CLI is confusing, but hopefully we can improve that when we properly have an app and fleet split.

automation/capitanodoc/capitanodoc.ts
```diff
@@ -54,25 +54,33 @@ interface Documentation {
...
 // Mapping folders names to custom headings in the docs
 const commandHeadings: { [key: string]: string } = {
-	'api-key': 'API Key',
+	'api-key': 'API Keys',
 	'api-keys': 'API Keys',
-	login: 'Authentication',
-	whoami: 'Authentication',
-	logout: 'Authentication',
-	env: 'Environment Variable',
+	app: 'Applications',
```
myarmolinsky, line 59:
```suggestion
	app: 'Apps',
```

PR #2779, opened 2024-07-01
Add commands for exporting and importing app/fleet releases.

cmfcruz:
Change-type: minor
See: https://balena.fibery.io/Work/Project/227

---

### DIFF
--- a/src/commands/release/export.ts
+++ b/src/commands/release/export.ts
@@ -29,20 +29,22 @@ export default class ReleaseExportCmd extends Command {
 		Exports a release into a file.
 
 		Exports a release to a file that you can use to import an exact 
-		copy of the original release into another app.
+		copy of the original release into another app, block, or fleet.
 
 		If the SemVer of a release is provided using the --version option,
-		the first argument is assumed to be the fleet's slug.
+		the first argument is assumed to be the app's slug.
 
 		Only successful releases can be exported.
+
+		To import a release to an app, block, or fleet, use 'balena release import'.
 `;
 	public static examples = [
 		'$ balena release export a777f7345fe3d655c1c981aa642e5555 -o ../path/to/release.tar',
 		'$ balena release export myOrg/myFleet --version 1.2.3 -o ../path/to/release.tar',
-		'$ balena release export myFleet --version 1.2.3 -o ../path/to/release.tar',
+		'$ balena release export myApp --version 1.2.3 -o ../path/to/release.tar',
 	];
 
-	public static usage = 'release export <commitOrFleet>';
+	public static usage = 'release export <commitOrApplication>';
 
 	public static flags = {
 		output: Flags.string({
@@ -51,15 +53,16 @@ export default class ReleaseExportCmd extends Command {
 			required: true,
 		}),
 		version: Flags.string({
-			description: 'version of the release to export from the specified fleet',
+			description:
+				'version of the release to export from the specified app, block, or fleet',
 		}),
 		help: cf.help,
 	};
 
 	public static args = {
-		commitOrFleet: Args.string({
+		commitOrApplication: Args.string({
 			description:
-				'release commit or fleet if used in conjunction with the --version option',
+				'release commit or app if used in conjunction with the --version option',
 			required: true,
 		}),
 	};
@@ -69,44 +72,46 @@ export default class ReleaseExportCmd extends Command {
 	public async run() {
 		const { args: params, flags: options } = await this.parse(ReleaseExportCmd);
 
-		const balena = getBalenaSdk();
-
 		let versionInfo = '';
+
 		try {
-			let releaseDetails:
-				| string
-				| number
-				| { application: string | number; rawVersion: string }; // ReleaseRawVersionApplicationPair
+			const balena = getBalenaSdk();
+
+			let releaseDetails: string | { application: number; rawVersion: string }; // ReleaseRawVersionApplicationPair
+
 			if (options.version != null) {
 				versionInfo = ` version ${options.version}`;
 				const parsedVersion = semver.parse(options.version);
 				if (parsedVersion == null) {
 					throw new ExpectedError(`version must be valid SemVer`);
-				} else {
-					const { getApplication } = await import('../../utils/sdk');
-					const application = (
-						await getApplication(balena, params.commitOrFleet)
-					).id;
-					releaseDetails = { application, rawVersion: parsedVersion.raw };
 				}
+				const { getApplication } = await import('../../utils/sdk');
+				const { id: application } = await getApplication(
+					balena,
+					params.commitOrApplication,
+				);
+				releaseDetails = { application, rawVersion: parsedVersion.raw };
 			} else {
-				releaseDetails = params.commitOrFleet;
+				releaseDetails = params.commitOrApplication;
 			}
 
-			const release = await balena.models.release.get(releaseDetails, {
-				$select: ['id'],
-			});
+			const { id: releaseId } = await balena.models.release.get(
+				releaseDetails,
+				{
+					$select: ['id'],
+				},
+			);
 			const releaseBundle = await create({
 				sdk: balena,
-				releaseId: release.id,
+				releaseId,
 			});
 			await fs.writeFile(options.output, releaseBundle);
 			console.log(
-				`Release ${params.commitOrFleet}${versionInfo} has been exported to ${options.output}.`,
+				`Release ${params.commitOrApplication}${versionInfo} has been exported to ${options.output}.`,
 			);
 		} catch (error) {
 			throw new ExpectedError(
-				`Release ${params.commitOrFleet}${versionInfo} could not be exported: ${error.message}`,
+				`Release ${params.commitOrApplication}${versionInfo} could not be exported: ${error.message}`,
 			);
 		}
 	}
--- a/src/commands/release/import.ts
+++ b/src/commands/release/import.ts
@@ -26,26 +26,26 @@ import { ExpectedError } from '../../errors';
 
 export default class ReleaseImportCmd extends Command {
 	public static description = stripIndent`
-		Imports a release from a file to an app or fleet.
+		Imports a release from a file to an app, block, or fleet.
 
-		Imports a release from a file to an app or fleet. The revision field of the release
-		is automatically omitted when importing a release. The backend will auto-increment
-		the revision field of the imported release if a release exists with the same semver.
-		A release will not be imported if a successful release with the same commit already
-		exists.
-
-		To export a release to a file, use 'balena release export'.
+		Imports a release from a file to an app, block, or fleet. The revision field of the
+		release is automatically omitted when importing a release. The balena API will
+		auto-increment the revision field of the imported release if a release exists with
+		the same semver. A release will not be imported if a successful release with the
+		same commit already exists.
 
 		Use the --override-version option to specify the version 
 		of the imported release, overriding the one saved in the file.
+
+		To export a release to a file, use 'balena release export'.
 `;
 	public static examples = [
-		'$ balena release import ../path/to/release.tar myFleet',
+		'$ balena release import ../path/to/release.tar myApp',
 		'$ balena release import ../path/to/release.tar myOrg/myFleet',
 		'$ balena release import ../path/to/release.tar myOrg/myFleet --override-version 1.2.3',
 	];
 
-	public static usage = 'release import <file> <fleet>';
+	public static usage = 'release import <file> <applicapplication>';
 
 	public static flags = {
 		'override-version': Flags.string({
@@ -61,10 +61,10 @@ export default class ReleaseImportCmd extends Command {
 			required: true,
 			description: 'path to a file, e.g. "./release.tar"',
 		}),
-		fleet: Args.string({
+		application: Args.string({
 			required: true,
 			description:
-				'fleet that the release will be imported to, e.g. "myOrg/myFleet"',
+				'app, block, or fleet that the release will be imported to, e.g. "myOrg/myFleet"',
 		}),
 	};
 
@@ -73,23 +73,24 @@ export default class ReleaseImportCmd extends Command {
 	public async run() {
 		const { args: params, flags: options } = await this.parse(ReleaseImportCmd);
 
-		const balena = getBalenaSdk();
-
-		let bundle: ReadStream;
 		try {
+			const balena = getBalenaSdk();
+
+			let bundle: ReadStream;
 			try {
 				const fileHandle = await fs.open(params.bundle);
 				bundle = fileHandle.createReadStream();
 			} catch (error) {
-				throw new ExpectedError(
-					`${params.bundle} does not exist or is not accessible`,
-				);
+				throw new Error(`${params.bundle} does not exist or is not accessible`);
 			}
 
 			const { getApplication } = await import('../../utils/sdk');
-			const application = (await getApplication(balena, params.fleet)).id;
+			const { id: application } = await getApplication(
+				balena,
+				params.application,
+			);
 			if (application == null) {
-				throw new ExpectedError(`Fleet ${params.fleet} not found`);
+				throw new ExpectedError(`Fleet ${params.application} not found`);
 			}
 			await apply({
 				sdk: balena,
@@ -98,11 +99,11 @@ export default class ReleaseImportCmd extends Command {
 				version: options['override-version'],
 			});
 			console.log(
-				`Release bundle ${params.bundle} has been imported to ${params.fleet}.`,
+				`Release bundle ${params.bundle} has been imported to ${params.application}.`,
 			);
 		} catch (error) {
 			throw new ExpectedError(
-				`Could not import release bundle ${params.bundle} to ${params.fleet}: ${error.message}`,
+				`Could not import release bundle ${params.bundle} to ${params.application}: ${error.message}`,
 			);
 		}
 	}
--- a/tests/commands/release/export.spec.ts
+++ b/tests/commands/release/export.spec.ts
@@ -23,6 +23,7 @@ describe('balena release export', function () {
 
 	this.beforeEach(async function () {
 		api = new BalenaAPIMock();
+		api.expectGetWhoAmI();
 		releaseFileBuffer = await fs.readFile(
 			path.join('tests', 'test-data', 'release.tar'),
 		);
@@ -38,7 +39,6 @@ describe('balena release export', function () {
 	});
 
 	itSS('should export a release to a file', async () => {
-		api.expectGetWhoAmI();
 		api.expectGetRelease();
 		releaseBundleCreateStub.resolves(stream.Readable.from(releaseFileBuffer));
 
@@ -54,7 +54,6 @@ describe('balena release export', function () {
 	});
 
 	itSS('should fail if the create throws an error', async () => {
-		api.expectGetWhoAmI();
 		api.expectGetRelease();
 		const expectedError = `BalenaReleaseNotFound: Release not found: ${appCommit}`;
 		releaseBundleCreateStub.rejects(new Error(expectedError));
@@ -69,7 +68,6 @@ describe('balena release export', function () {
 	});
 
 	itSS('should parse with application slug and version', async () => {
-		api.expectGetWhoAmI();
 		api.expectGetRelease();
 		api.expectGetApplication({ times: 2 });
 		releaseBundleCreateStub.resolves(stream.Readable.from(releaseFileBuffer));
@@ -86,7 +84,6 @@ describe('balena release export', function () {
 	});
 
 	it('should fail if the app slug is provided without the release version', async () => {
-		api.expectGetWhoAmI();
 		api.expectGetRelease({ notFound: true });
 		const expectedError = `Release not found: ${appSlug}`;
 
@@ -100,7 +97,6 @@ describe('balena release export', function () {
 	});
 
 	it('should fail if the semver is invalid', async () => {
-		api.expectGetWhoAmI();
 		const expectedError = 'version must be valid SemVer';
 
 		const { err } = await runCommand(
--- a/tests/commands/release/import.spec.ts
+++ b/tests/commands/release/import.spec.ts
@@ -11,13 +11,13 @@ const itSS = process.env.BALENA_CLI_TEST_TYPE === 'standalone' ? it.skip : it;
 describe('balena release import', function () {
 	const appCommit = '4c8becf0780ca69d33b638ea8fa163d7';
 	const appSlug = 'myOrg/myFleet';
-	// const appVersion = '1.2.3+rev1';
 	const releasePath = path.join('tests', 'test-data', 'release.tar');
 	let api: BalenaAPIMock;
 	const releaseBundleApplyStub = sinon.stub();
 
 	this.beforeEach(async function () {
 		api = new BalenaAPIMock();
+		api.expectGetWhoAmI();
 		mock('@balena/release-bundle', {
 			apply: releaseBundleApplyStub,
 		});
@@ -30,7 +30,6 @@ describe('balena release import', function () {
 	});
 
 	itSS('should import a release to an app', async () => {
-		api.expectGetWhoAmI();
 		api.expectGetApplication();
 		releaseBundleApplyStub.resolves(123);
 
@@ -48,7 +47,6 @@ describe('balena release import', function () {
 	itSS(
 		'should import a release to an app with a version override',
 		async () => {
-			api.expectGetWhoAmI();
 			api.expectGetApplication();
 			releaseBundleApplyStub.resolves(123);
 
@@ -65,7 +63,6 @@ describe('balena release import', function () {
 	);
 
 	it('should fail if release file does not exist', async () => {
-		api.expectGetWhoAmI();
 		const nonExistentFile = path.join(
 			'tests',
 			'test-data',
@@ -83,7 +80,6 @@ describe('balena release import', function () {
 	});
 
 	itSS('should fail if overriding version is not a valid semver', async () => {
-		api.expectGetWhoAmI();
 		api.expectGetApplication();
 		const expectedError = `Manifest is malformed: Expected version to be a valid semantic version but found '${appCommit}'`;
 		releaseBundleApplyStub.rejects(new Error(expectedError));
@@ -100,7 +96,6 @@ describe('balena release import', function () {
 	itSS(
 		'should fail if a successful release with the same commit already exists',
 		async () => {
-			api.expectGetWhoAmI();
 			api.expectGetApplication();
 			const expectedError = `A successful release with commit ${appCommit} (1.2.3) already exists; nothing to do`;
 			releaseBundleApplyStub.rejects(new Error(expectedError));
--- a/docs/balena-cli.md
+++ b/docs/balena-cli.md
@@ -282,9 +282,9 @@ are encouraged to regularly update the balena CLI to the latest version.
 
 - Releases
 
-	- [release export &#60;commitorfleet&#62;](#release-export-commitorfleet)
+	- [release export &#60;commitorapplication&#62;](#release-export-commitorapplication)
 	- [release finalize &#60;commitorid&#62;](#release-finalize-commitorid)
-	- [release import &#60;file&#62; &#60;fleet&#62;](#release-import-file-fleet)
+	- [release import &#60;file&#62; &#60;applicapplication&#62;](#release-import-file-applicapplication)
 	- [release &#60;commitorid&#62;](#release-commitorid)
 	- [release invalidate &#60;commitorid&#62;](#release-invalidate-commitorid)
 	- [release validate &#60;commitorid&#62;](#release-validate-commitorid)
@@ -3347,27 +3347,29 @@ The notes for this release
 
 # Releases
 
-## release export &#60;commitOrFleet&#62;
+## release export &#60;commitOrApplication&#62;
 
 Exports a release to a file that you can use to import an exact 
-copy of the original release into another app.
+copy of the original release into another app, block, or fleet.
 
 If the SemVer of a release is provided using the --version option,
-the first argument is assumed to be the fleet's slug.
+the first argument is assumed to be the app's slug.
 
 Only successful releases can be exported.
+
+To import a release to an app, block, or fleet, use 'balena release import'.
 
 Examples:
 
 	$ balena release export a777f7345fe3d655c1c981aa642e5555 -o ../path/to/release.tar
 	$ balena release export myOrg/myFleet --version 1.2.3 -o ../path/to/release.tar
-	$ balena release export myFleet --version 1.2.3 -o ../path/to/release.tar
+	$ balena release export myApp --version 1.2.3 -o ../path/to/release.tar
 
 ### Arguments
 
-#### COMMITORFLEET
+#### COMMITORAPPLICATION
 
-release commit or fleet if used in conjunction with the --version option
+release commit or app if used in conjunction with the --version option
 
 ### Options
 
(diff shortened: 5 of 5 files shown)


lib/commands/release/export.ts
```diff
@@ -0,0 +1,98 @@
...
+
+		const balena = getBalenaSdk();
+
+		let release: balenaSdk.Release;
+		if (
+			typeof params.commitOrId === 'string' &&
+			params.commitOrId.includes(':')
+		) {
+			const fleet = params.commitOrId.split(':')[0];
+			const rawVersion = params.commitOrId.split(':')[1];
```
otaviojacobi, line 70: This can be rewritten as 
```suggestion
const [fleet, rawVersion] = params.commitOrId.split(':');
```

lib/commands/release/export.ts
```diff
@@ -0,0 +1,98 @@
...
+			release = await balena.models.release.get(
+				{ application: fleet, rawVersion: rawVersion },
```
otaviojacobi, line 72: you can omit `rawVersion: rawVersion` with a single `rawVersion` as:
```suggestion
                                            { application: fleet, rawVersion },
```

lib/commands/release/export.ts
```diff
@@ -0,0 +1,98 @@
...
+				{ $select: ['id'], $top: 1 },
```
otaviojacobi, line 73: is `$top: 1` necessary?

lib/commands/release/export.ts
```diff
@@ -0,0 +1,98 @@
...
+			);
+		} else {
+			release = await balena.models.release.get(params.commitOrId, {
+				$select: ['id'],
+				$top: 1,
```
otaviojacobi, line 78: is `$top: 1` necessary?

src/commands/release/import.ts
```diff
@@ -0,0 +1,99 @@
...
+import Command from '../../command';
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+import { apply } from '@balena/release-bundle';
+import { createReadStream } from 'fs';
+import { ExpectedError } from '../../errors';
+
+export default class ReleaseImportCmd extends Command {
+	public static description = stripIndent`
+		Imports a release from a file to an application or fleet.
```
myarmolinsky, line 28: nit, the terminology we try to use is `app`
```suggestion
		Imports a release from a file to an app or fleet.
```

src/commands/release/import.ts
```diff
@@ -0,0 +1,99 @@
...
+
+		To save a release into a file, use 'balena release export'.
```
myarmolinsky, line 30: nit
```suggestion
		To save a release to a file, use 'balena release export'.
```

Personally I also would have gone with `export` instead of `save`, but I don't know if this was discussed anywhere

src/commands/release/import.ts
```diff
@@ -0,0 +1,99 @@
...
+		'$ balena release import ../path/to/release.tar -f 1234567',
+		'$ balena release import ../path/to/release.tar -f myFleet',
+		'$ balena release import ../path/to/release.tar -f myOrg/myFleet',
+		'$ balena release import ../path/to/release.tar -f myOrg/myFleet -V 1.2.3',
+	];
+
+	public static usage = 'release import <file>';
+
+	public static flags = {
+		fleet: { ...cf.fleet, exclusive: ['device'] },
```
myarmolinsky, line 45: Flags should be for adjusting the default of a command if IMO and if I understand CLIs correctly. So `override-version` makes sense as a flag, but `fleet` does not. So the command should look like `$ balena release import ../path/to/release.tar 1234567`

src/commands/release/import.ts
```diff
@@ -0,0 +1,99 @@
...
+		'override-version': Flags.string({
+			description:
+				'Imports this release with the specified version overriding the version in the file.',
+			char: 'V',
```
myarmolinsky, line 49: I'm hesitant to include a char for the command at the moment, as we need to discuss and figure out how to determine what char to use for commands, especially without creating confusion/overlap with out commands

WDYT @otaviojacobi , would you also skip the short version and remove `char` for now?

src/commands/release/import.ts
myarmolinsky: > Flags should be for adjusting the default of a command if IMO and if I understand CLIs correctly

I wouldn't mind if you could confirm my thinking on this @otaviojacobi 😅

src/commands/release/import.ts
```diff
@@ -0,0 +1,99 @@
...
+
+		try {
+			if (
+				typeof options.fleet !== 'number' &&
+				typeof options.fleet !== 'string'
+			) {
+				throw new ExpectedError('Fleet must be a number or slug.');
+			}
+
+			// TODO: validate if the path to the release bundle exists
```
myarmolinsky, line 79: Don't you just need to add `.on('error', () => throw new Error('foo'))`?

src/commands/release/import.ts
```diff
@@ -0,0 +1,99 @@
...
+	public static description = stripIndent`
+		Imports a release from a file to an application or fleet.
+
+		To save a release into a file, use 'balena release export'.
+
+		Use the --override-version option to specify the version 
+		of the imported release, overriding the one saved in the file.
+`;
+	public static examples = [
+		'$ balena release import ../path/to/release.tar -f 1234567',
```
myarmolinsky, line 36: I believe we have been trying to avoid allowing `id` to be used for identifying fleets

src/commands/release/import.ts
```diff
@@ -0,0 +1,99 @@
...
+
+		const bundle = createReadStream(params.bundle);
+
+		try {
+			if (
+				typeof options.fleet !== 'number' &&
+				typeof options.fleet !== 'string'
+			) {
+				throw new ExpectedError('Fleet must be a number or slug.');
+			}
```
myarmolinsky, line 77: > I believe we have been trying to avoid allowing id to be used for identifying fleets, I would remove this example

As I noted here ^

This leaves us with just string as a type anyway. We can remove this if altogether.

Then after the 
```
const application = await balena.models.application.get(options.fleet, {
				$select: ['id'],
			});
			```
			We can add this if you really want to check this:
```
// Check app exists
		if (!application) {
			throw new ExpectedError(`Error: fleet ${params.fleet} not found.`);
		}
		```

src/commands/release/export.ts
```diff
@@ -0,0 +1,110 @@
...
+		Saves a release into a file.
+
+        Saving a release to a file allows you to import an exact 
+		copy of the original release into another application.
+
+		Only successful releases can be saved into a file.
+`;
+	public static examples = [
+		'$ balena release export a777f7345fe3d655c1c981aa642e5555 -o ../path/to/release.tar',
+		'$ balena release export 1234567 -o ../path/to/release.tar',
```
myarmolinsky, line 39: Again I would not include `id`

src/commands/release/export.ts
```diff
@@ -0,0 +1,110 @@
...
+		'$ balena release export myOrg/myFleet:1.2.3 -o ../path/to/release.tar',
```
myarmolinsky, line 40: This looks very unintuitive 😕 🤔 @otaviojacobi what would you say is the correct/better way to ask the user for a semver which is bound to a specific fleet?

src/commands/release/export.ts
```diff
@@ -0,0 +1,110 @@
...
+				sdk: balena,
+				releaseId: release.id,
+			});
+			await fs.writeFile(options.output, releaseBundle);
+			console.log(
+				`Release ${params.commitOrId} has been exported to ${options.output}.`,
+			);
+		} catch (error) {
+			throw new ExpectedError(
+				`Release ${params.commitOrId} could not be exported: ${error.message}`,
```
myarmolinsky, line 106: Any way to get more explicit than this and share a reason(s) why the release could not be exported?

src/commands/release/export.ts
```diff
@@ -0,0 +1,110 @@
...
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params, flags: options } = await this.parse(ReleaseExportCmd);
+
+		const balena = getBalenaSdk();
+
+		let release: balenaSdk.Release;
+		if (
+			typeof params.commitOrId === 'string' &&
```
myarmolinsky, line 70: Isn't typeof always `string` for params coming in?

otaviojacobi: Not necessarily, check the definition of the `commitOrIdArg` parser:

```
export const commitOrIdArg = Args.custom({
	parse: async (commitOrId: string) => tryAsInteger(commitOrId),
});
```

It tries to parse as an integer, and in case it fails, it returns a string

src/commands/release/import.ts
otaviojacobi: I agree with this - If it is mandatory, it shouldn't be a flag, but rather a mandatory arg.

src/commands/release/import.ts
otaviojacobi: I am fine with it if there is a strong reason to have it, but I would also prefer not to have it because it is very easily to mistake with the much more common `v` char we already have for several commands which serves for verbose

src/commands/release/export.ts
cmfcruz: @myarmolinsky , is it still unintuitive? I believe many users would be already used to how a docker tag format (`<repo|tag>:<version>`) looks like and we are just going with a similar convention.  I assume this concern is because we are referring to a fleet and not an image repository/tag.

Alternatively, we can go with the similar convention of `balena os download <device_type> --version <version>` if that provides consistency across our UI. cc @dfunckt just to let him know I am suggesting an alternative.

src/commands/release/export.ts
```diff
@@ -0,0 +1,110 @@
...
+import { ExpectedError } from '../../errors';
+
+export default class ReleaseExportCmd extends Command {
+	public static description = stripIndent`
+		Saves a release into a file.
+
+        Saving a release to a file allows you to import an exact 
+		copy of the original release into another application.
+
+		Only successful releases can be saved into a file.
```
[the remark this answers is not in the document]
cmfcruz: We may as well match the terminology of the command.

src/commands/release/import.ts
cmfcruz: Thanks for the suggestion.  I did just that.

src/commands/release/export.ts
cmfcruz: Removed.

src/commands/release/import.ts
cmfcruz: Thanks for the suggestion.  I included it.

src/commands/release/import.ts
cmfcruz: We just need an option to override the version.  I think we could do without using a character to represent the option.

lib/commands/release/export.ts
cmfcruz: I updated the code to use this convention.

src/commands/release/import.ts
cmfcruz: Noted.  Updated the description.

src/commands/release/export.ts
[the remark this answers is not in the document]
cmfcruz: Descriptions updated.

src/commands/release/import.ts
cmfcruz: Removed.

src/commands/release/import.ts
cmfcruz: I changed it into the second argument instead.

src/commands/release/import.ts
cmfcruz: Descriptions updated.

src/commands/release/export.ts
cmfcruz: The release bundle module will throw an exception indicating why a release could not be exported.  We have a list of possible reasons here: https://github.com/balena-io-modules/balena-release-bundle/blob/master/src/create.ts#L30-L48

src/commands/release/export.ts
cmfcruz: @myarmolinsky @otaviojacobi , I changed the approach to use the flag `--version` instead similar to how he use the flag for OS downloads.  I also removed the example that uses the release ID since we don't recommend this approach anymore.

src/commands/release/import.ts
cmfcruz: I removed the character flag for the option.  Users will just need to use `--override-version` to use a different version than the one that the release have.

src/commands/release/export.ts
```diff
@@ -0,0 +1,116 @@
...
+			required: true,
+		}),
+		version: Flags.string({
+			description: 'version of the release to export from the specified fleet',
+		}),
+		help: cf.help,
+	};
+
+	public static args = {
+		commitOrId: commitOrIdArg({
```
otaviojacobi, line 60: As far as I understand we removed the id right? so should it be called commitOrSlug?

src/commands/release/export.ts
```diff
@@ -0,0 +1,116 @@
...
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params, flags: options } = await this.parse(ReleaseExportCmd);
+
+		const balena = getBalenaSdk();
+
+		let release: balenaSdk.Release;
+		if (typeof options.version === 'string') {
```
otaviojacobi, line 74: @cmfcruz is there any reason why you are explicitly testing "typeof" being string instead of `options.version != null`? I think the later is prefereable as it is more clear

src/commands/release/export.ts
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		try {
+			const releaseBundle = await create({
+				sdk: balena,
+				releaseId: release.id,
+			});
+			await fs.writeFile(options.output, releaseBundle);
+			const versionInfo =
+				typeof options.version === 'string'
```
otaviojacobi, line 104: same as above, any reason not to use 

```suggestion
options.version != null
```

src/commands/release/export.ts
```diff
@@ -0,0 +1,116 @@
...
+		let release: balenaSdk.Release;
+		if (typeof options.version === 'string') {
+			const application = params.commitOrId;
+			const parsedVersion = semver.parse(options.version);
+			if (parsedVersion == null) {
+				throw new ExpectedError(
+					`Release of ${application} with version ${options.version} could not be exported; version must be valid SemVer.`,
+				);
+			} else {
+				const rawVersion =
```
otaviojacobi, line 82: Is there any reason we can't just use `parsedVersion.raw`?

src/commands/release/export.ts
```diff
@@ -0,0 +1,116 @@
...
+					parsedVersion.build.length === 0
+						? parsedVersion.version
+						: `${parsedVersion.version}+${parsedVersion.build[0]}`;
```
otaviojacobi, line 85: Is there any reason we can't just use `parsedVersion.raw`?  e.g. we just do 

```
				release = await balena.models.release.get(
					{ application, rawVersion: parsedVersion.raw },
					{ $select: ['id'] },
				);
```

bellow and all these 4 lines of code can go away

src/commands/release/export.ts
```diff
@@ -0,0 +1,113 @@
...
+import { create } from '@balena/release-bundle';
+import * as fs from 'fs/promises';
+import * as semver from 'balena-semver';
+import { ExpectedError } from '../../errors';
+
+export default class ReleaseExportCmd extends Command {
+	public static description = stripIndent`
+		Exports a release into a file.
+
+        Exporting a release to a file allows you to import an exact
```
myarmolinsky, line 32: Missing an indent here?

myarmolinsky: Just for my better understanding of these commands, you say that this allows you to import an "exact" copy of the original release. Is this true? I remember hearing that simply re-building/pushing a release can yield 2 different outcomes/releases. Does importing a release not build it at all and just use everything from the previously built release exactly as it is?

cmfcruz: Yes, it pulls the images the selected release from the registry and packages them into the release bundle tar file with the release metadata. The release bundle contains enough information to create the same release in another fleet including notes and tags.

myarmolinsky: Wow, nice!

cmfcruz: Indentation fixed in latest commit.

src/commands/release/export.ts
cmfcruz: Not really.  I was just not familiar about the format of parsedVersion.raw.  Code updated.

tests/commands/release/export.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+import { promises as fs } from 'fs';
+
+// "itSS" means "it() Skip Standalone"
+const itSS = process.env.BALENA_CLI_TEST_TYPE === 'standalone' ? it.skip : it;
+
+describe('balena release export', function () {
+	const appCommit = 'badc0ffe';
+	const appSlug = 'testOrg/testApp';
+	const appVersion = '1.2.3+rev1';
+	const outputPath = '/tmp/release.tar';
```
otaviojacobi, line 16: @cmfcruz I suppose these tests are breaking on windows because this is not a valid path on windows (escaping using `/` vs `\\`). You can use nodejs `path` library with a join function to solve this, e.g.

```suggestion
import * as path from 'node:path';
...
const outputPath = path.join('tmp', 'release.tar');
```

cmfcruz: Code updated.  Thanks for the clue, @otaviojacobi .

src/commands/release/export.ts
```diff
@@ -0,0 +1,109 @@
...
+		the first argument is assumed to be the fleet's slug.
+
+		Only successful releases can be exported.
+`;
+	public static examples = [
+		'$ balena release export a777f7345fe3d655c1c981aa642e5555 -o ../path/to/release.tar',
+		'$ balena release export myOrg/myFleet --version 1.2.3 -o ../path/to/release.tar',
+	];
+
+	public static usage = 'release export <commitOrId>';
```
myarmolinsky, line 44: I think I saw a comment about this somewhere, maybe I missed your response

`commitOrId` is incorrect now, right? It should be `commitOrSlug`?

myarmolinsky: Ah or `commitOrFleet` as you wrote below

src/commands/release/export.ts
cmfcruz: No reason.  Thanks for the suggestion.  I updated the code to make it clear that the condition is for when the version (which is always a string) is provided.

src/commands/release/export.ts
cmfcruz: I used `commitOrFleet` instead.

src/commands/release/export.ts
```diff
@@ -0,0 +1,113 @@
...
+			let releaseDetails:
+				| string
+				| number
+				| { application: string | number; rawVersion: string }; // ReleaseRawVersionApplicationPair
+			if (options.version != null) {
+				versionInfo = ` version ${options.version}`;
+				const parsedVersion = semver.parse(options.version);
+				if (parsedVersion == null) {
+					throw new ExpectedError(`version must be valid SemVer`);
+				} else {
```
otaviojacobi, line 85: NIT: this else is not necessary as it will throw on the line above

src/commands/release/export.ts
```diff
@@ -0,0 +1,113 @@
...
+	public async run() {
+		const { args: params, flags: options } = await this.parse(ReleaseExportCmd);
+
+		const balena = getBalenaSdk();
+
+		let versionInfo = '';
+		try {
+			let releaseDetails:
+				| string
+				| number
```
myarmolinsky, line 78: When is this a `number`?

myarmolinsky: Is this a leftover from when `id` was an option? `commit` and `fleet` should each just be strings

src/commands/release/export.ts
```diff
@@ -0,0 +1,113 @@
...
+				| { application: string | number; rawVersion: string }; // ReleaseRawVersionApplicationPair
```
myarmolinsky, line 79: `application` should just be a `number` since it's always `id`, right?

src/commands/release/export.ts
```diff
@@ -0,0 +1,113 @@
...
+				versionInfo = ` version ${options.version}`;
+				const parsedVersion = semver.parse(options.version);
+				if (parsedVersion == null) {
+					throw new ExpectedError(`version must be valid SemVer`);
+				} else {
+					const { getApplication } = await import('../../utils/sdk');
+					const application = (
+						await getApplication(balena, params.commitOrFleet)
+					).id;
+					releaseDetails = { application, rawVersion: parsedVersion.raw };
```
myarmolinsky, line 90: nit:
```suggestion
					const { id }= (
						await getApplication(balena, params.commitOrFleet)
					);
					releaseDetails = { application: id, rawVersion: parsedVersion.raw };
```

Just because it confused me at first, making me think we were passing the whole application for a second there. Not sure if it's just me. Ideally the `application` property should be named `applicationId` as well

src/commands/release/export.ts
```diff
@@ -0,0 +1,113 @@
...
+			} else {
+				releaseDetails = params.commitOrFleet;
+			}
+
+			const release = await balena.models.release.get(releaseDetails, {
+				$select: ['id'],
+			});
+			const releaseBundle = await create({
+				sdk: balena,
+				releaseId: release.id,
```
myarmolinsky, line 101: nit:
```suggestion
			const { id } = await balena.models.release.get(releaseDetails, {
				$select: ['id'],
			});
			const releaseBundle = await create({
				sdk: balena,
				releaseId: id,
```

Since we don't use the `release` object as a whole anywhere

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+import * as cf from '../../utils/common-flags';
+import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+import { apply } from '@balena/release-bundle';
+import type { ReadStream } from 'fs';
+import { promises as fs } from 'fs';
+import { ExpectedError } from '../../errors';
+
+export default class ReleaseImportCmd extends Command {
+	public static description = stripIndent`
+		Imports a release from a file to an app or fleet.
```
myarmolinsky, line 29: Not to a `block` too?

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+
+		Imports a release from a file to an app or fleet. The revision field of the release
+		is automatically omitted when importing a release. The backend will auto-increment
+		the revision field of the imported release if a release exists with the same semver.
```
myarmolinsky, line 33: nit because I'm not sure how I feel about us saying "the backend" to the user:
```suggestion
		is automatically omitted when importing a release. The revision field of the imported release is auto-incremented
		if a release exists with the same semver.
```

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+		A release will not be imported if a successful release with the same commit already
```
myarmolinsky, line 34: Unrelated to the PR itself but I'm curious why is this the case? What were the downsides of assigning a new commit to the release if the same commit already existed? (in which case we would output the new commit as well at the end of this command)

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+		exists.
+
+		To export a release to a file, use 'balena release export'.
```
myarmolinsky, line 37: Perhaps we should add this same line (but for `import` of course) to the `export` command description?

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+
+		Use the --override-version option to specify the version 
+		of the imported release, overriding the one saved in the file.
```
myarmolinsky, line 40: nit: This sounds a bit confusing (maybe just to me) because it sounds like you may be overriding the version of the release you are importing (the upstream one)

Perhaps something like this?
```suggestion
		Use the --override-version option to use a different version 
		in the saved file than that of the original release.
```

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+		'$ balena release import ../path/to/release.tar myOrg/myFleet',
+		'$ balena release import ../path/to/release.tar myOrg/myFleet --override-version 1.2.3',
+	];
+
+	public static usage = 'release import <file> <fleet>';
+
+	public static flags = {
+		'override-version': Flags.string({
+			description:
+				'Imports this release with the specified version overriding the version in the file.',
```
myarmolinsky, line 53: See above

src/commands/release/import.ts
myarmolinsky: Ignore what I said, crossed wires in my brain from `export` to `import` 😖

src/commands/release/import.ts
myarmolinsky: Ignore what I said, crossed wires in my brain from `export` to `import` 😖

src/commands/release/export.ts
cmfcruz: [balena-semver's parse](https://github.com/balena-io-modules/balena-semver?tab=readme-ov-file#parse) returns a `null` if the semver could not be parsed.

tests/test-data/release.tar
myarmolinsky, line 1: Probably a lack of understanding on my part of how the testing works: Why do we need to have our own `release.tar` committed to the test-data instead of using the one from the `export` tests for the `import` tests?

tests/commands/release/export.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+	});
+
+	this.afterEach(() => {
+		// Check all expected api calls have been made and clean up.
+		api.done();
+		mock.stop('@balena/release-bundle');
+	});
+
+	itSS('should export a release to a file', async () => {
+		api.expectGetWhoAmI();
```
myarmolinsky, line 41: Is this something we can move to the `beforeEach`?

myarmolinsky: Looks like we indeed put it in `beforeEach` for other tests

src/commands/release/export.ts
cmfcruz: I see what you mean. Let me update that in my next commit.

tests/commands/release/import.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as sinon from 'sinon';
+import * as path from 'node:path';
+
+// "itSS" means "it() Skip Standalone"
+const itSS = process.env.BALENA_CLI_TEST_TYPE === 'standalone' ? it.skip : it;
+
+describe('balena release import', function () {
+	const appCommit = '4c8becf0780ca69d33b638ea8fa163d7';
+	const appSlug = 'myOrg/myFleet';
+	// const appVersion = '1.2.3+rev1';
```
myarmolinsky, line 14: comment?

src/commands/release/export.ts
cmfcruz: Noted.

src/commands/release/export.ts
cmfcruz: Noted.

src/commands/release/import.ts
cmfcruz: Good point.  Yes, it works with blocks as well.

src/commands/release/import.ts
cmfcruz: Let's just refer to it as the balena API.

tests/test-data/release.tar
cmfcruz: We do use the same file in both the `export` and `import` tests.  The output path of the `export` test is set to a path in a temporary directory while the `import` tests use the one in `test-data`.  It is also a valid release bundle file in case an engineer wants to do actual testing with a test fleet.

tests/commands/release/import.spec.ts
cmfcruz: Share eyes.  Nice catch!

tests/commands/release/export.spec.ts
cmfcruz: Good call. It is used in all tests in both `export` and `import` tests.

src/commands/release/import.ts
cmfcruz: We want to emphasize that the imported release is an exact copy of another release.  Since users are allowed to override the version of an imported release, the commit helps confirm if the release is a copy of a specific build from another fleet.  We also do not get any real advantage for allowing imports of the same copy of the release if a successful release with the same commit already exists.

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+
+		const balena = getBalenaSdk();
+
+		let bundle: ReadStream;
+		try {
+			try {
+				const fileHandle = await fs.open(params.bundle);
+				bundle = fileHandle.createReadStream();
+			} catch (error) {
+				throw new ExpectedError(
```
myarmolinsky, line 84: Does this really need to be an `ExpectedError` instead of just a plain `Error` if we're only using it for the message in the outer trycatch's catch?

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+
+	public static args = {
+		bundle: Args.string({
+			required: true,
+			description: 'path to a file, e.g. "./release.tar"',
+		}),
+		fleet: Args.string({
+			required: true,
+			description:
+				'fleet that the release will be imported to, e.g. "myOrg/myFleet"',
```
myarmolinsky, line 67: This only specified fleet, not app and block as well

src/commands/release/export.ts
```diff
@@ -0,0 +1,115 @@
...
+	public static usage = 'release export <commitOrFleet>';
+
+	public static flags = {
+		output: Flags.string({
+			description: 'output path',
+			char: 'o',
+			required: true,
+		}),
+		version: Flags.string({
+			description: 'version of the release to export from the specified fleet',
```
myarmolinsky, line 54: As above

src/commands/release/export.ts
```diff
@@ -0,0 +1,115 @@
...
+		}),
+		help: cf.help,
+	};
+
+	public static args = {
+		commitOrFleet: Args.string({
+			description:
+				'release commit or fleet if used in conjunction with the --version option',
```
myarmolinsky, line 62: As above

src/commands/release/export.ts
```diff
@@ -0,0 +1,115 @@
...
+			required: true,
+		}),
+		version: Flags.string({
+			description: 'version of the release to export from the specified fleet',
+		}),
+		help: cf.help,
+	};
+
+	public static args = {
+		commitOrFleet: Args.string({
```
myarmolinsky, line 60: Ideally this should be called `commitOrApplication` actually since it could be an app or block as well

src/commands/release/import.ts
cmfcruz: Good suggestion.

src/commands/release/import.ts
```diff
@@ -0,0 +1,109 @@
...
+	};
+
+	public static authenticated = true;
+
+	public async run() {
+		const { args: params, flags: options } = await this.parse(ReleaseImportCmd);
+
+		const balena = getBalenaSdk();
+
+		let bundle: ReadStream;
```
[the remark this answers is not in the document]
cmfcruz: Noted.

src/commands/release/import.ts
cmfcruz: Yeah, it should be a plain `Error` since this is something unexpected anyway.

src/commands/release/import.ts
cmfcruz: To match the change for `export`, I am changed this to `application` and updated the description.

myarmolinsky (approved):

PR #2796, opened 2024-07-10
Replace resin-discoverable-services with bonjour-service

otaviojacobi:
Instead of using the more generic resin-discoverable-services lib which is unmantained and currently has several vulnerabilities and forks for fixing issues (that were later on fixed upstream) we directly talk with mDNS using standard (and currently mantained) bonjour-service.

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/utils/discover.ts
+++ b/lib/utils/discover.ts
@@ -1,4 +1,5 @@
-import { enumerateServices, findServices } from 'resin-discoverable-services';
+import Bonjour from 'bonjour-service';
+import type { Service } from 'bonjour-service';
 
 interface LocalBalenaOsDevice {
 	address: string;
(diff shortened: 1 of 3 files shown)


lib/utils/discover.ts
```diff
@@ -7,34 +8,36 @@ interface LocalBalenaOsDevice {
...
-	const serviceDefinitions = Array.from(availableServices)
-		.filter((s) => Array.from(s.tags).includes(avahiBalenaSshTag))
-		.map((s) => s.service);
-
-	if (serviceDefinitions.length === 0) {
-		throw new Error(
-			`Could not find any available '${avahiBalenaSshTag}' services`,
-		);
-	}
+	const bonjour = new Bonjour();
```
dfunckt, line 22: I think it would be good to pass an error handler and report errors properly instead of letting them crash the process: https://github.com/onlxltd/bonjour-service#initializing

lib/utils/discover.ts
```diff
@@ -7,34 +8,36 @@ interface LocalBalenaOsDevice {
...
+	const browser = await new Promise<Browser>((resolve) => {
+		const browser = bonjour.find(avahiBalenaSshConfig);
+		setTimeout(() => {
+			resolve(browser);
+		}, timeout);
```
dfunckt, line 27: This is very strange code -- are you sure `.find` will always return within `timeout` interval? What if it doesn't? I think you might be misusing the API.

Also, seeing the docs (https://github.com/onlxltd/bonjour-service#var-browser--bonjourfindoptions-onup) it looks like you can have a callback called every time it finds a service.

And it also looks like it supports passing the subtype to filter upon as an option?

Overall, this is slightly confusing code and I'd like to understand it better as the behavior isn't fully apparent from looking at the code. Can you clarify please?

otaviojacobi: Hey @dfunckt thanks for the review.
So what this library does is that when you call bonjour find, it will have a callback for each service found and will internally store all the services found on the `services` getter. I can alternatively use the callback to store them myself with something like:

```ts
const services = new Promise<Service[]>((resolve) => {
    const services: Service[] = [];
    const browser = bonjour.find(avahiBalenaSshConfig, (service) => services.push(service));
    setTimeout(() => {
         browser.stop();
         resolve(services);
    }, timeout);
})
```

This was in fact my first version but I decided to move into directly using the services from the browser. It does however require you to have some kind of "timeout" or at least some kind of await sleep(timeout) between the calls of find/service to have it populated. Another alternative rewrite would be:

```ts
const browser = bonjour.find(avahiBalenaSshConfig);
await sleep(timeout); // has to be done
browser.stop();

const { services } = browser;
```

Do you have any preferences/reasoning why one would be better?

dfunckt: In terms of readability/understandability I think the first one is much more straightforward even if it is perhaps more LoC -- these extra lines of code make clear how found services are collected and when the whole process ends (moving `browser.stop()` in the timeout callback is a good idea).

Did you check the `subtype` option? Is it usable or you still need to do your own filtering?

otaviojacobi: Yes - I did, sadly, it doesn't work - it doesn't do anything really for `find` see: https://github.com/onlxltd/bonjour-service/issues/38 (maybe bumping the issue they addres it?)

But we can be slightly sure that the subtype will be correct because of the filter on `name` e.g. it is filtering via subtype `_resin-device._sub` as the protocol for zeroconf will use _{subbtype}._sub on the name, so I can have the filtering in there. I do the filtering on the client side for `subtypes` as a second check, but in reality, this should always already have the correct subtype.

lib/utils/discover.ts
otaviojacobi: I did that using the standard CLI error handling. I did force a simple error which looked fine with it.

lib/utils/discover.ts
```diff
@@ -7,34 +8,41 @@ interface LocalBalenaOsDevice {
...
+		const services: Service[] = [];
+		const browser = bonjour.find(avahiBalenaSshConfig, (service) =>
+			services.push(service),
 		);
-	}
+		setTimeout(() => {
+			browser.stop();
+			resolve(services);
+		}, timeout);
+	});
```
dfunckt, line 34: You're not using Bonjour besides for getting the services, so creating and destroying the instance within that promise, so that its lifetime is fully contained, might be a good idea.

Other than that, the PR looks great to me and feel free to merge when you're ready.

otaviojacobi: Ah! yes I understand what you mean now. Yes, this is a great suggestion.

myarmolinsky (approved):

PR #2814, opened 2024-07-15
Improve discover balena os across different networks

otaviojacobi:
This is an improvement over the `scan`, `join` and `leave` commands removing flakiness when searching over different networks. In short, instead of leaving bonjour to search across all interfaces, we forcebly conduct a search on each interface, this requires mDNS binding any ipv4 interface (0.0.0.0), otherwise it would bind over the interface itself, which is not desired as it causes services to only be able to receive information over that interface, see https://github.com/mafintosh/multicast-dns/issues/53. This targeted approach enhances the reliability and accuracy of network searches, reducing instances of missed connections or network errors typically caused by flakiness when relying on bonjour's default behavior.

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/lib/utils/discover.ts
+++ b/lib/utils/discover.ts
@@ -1,5 +1,6 @@
 import Bonjour from 'bonjour-service';
 import type { Service } from 'bonjour-service';
+import * as os from 'os';
 
 interface LocalBalenaOsDevice {
 	address: string;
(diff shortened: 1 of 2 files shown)


lib/utils/discover.ts
```diff
@@ -46,3 +34,60 @@ export async function discoverLocalBalenaOsDevices(
 			port,
 		}));
 }
+
+async function searchBalenaDevicesOverNetworks(
+	timeout: number,
+): Promise<Service[]> {
+	const networks = os.networkInterfaces();
+	const validNics: os.NetworkInterfaceInfo[] = [];
```
dfunckt, line 42: It's a bit subjective, but I'd either inline this function into `discoverLocalBalenaOsDevices`'s body, or make this function take `networks` as an argument. Right now, this function doesn't have any reason to exist besides the function name that clearly states what it does.

I'd personally err towards the former (to discourage our future selves to reuse something that is not meant to be reused), with a brief comment that it's doing a search over all networks (which is information currently conveyed by this function's name`)

dfunckt (approved):

PR #2816, opened 2024-07-17
v19

otaviojacobi:
This effectively removes the ability to filter/sort/customize the output table. The cli cannot properly handle this operations on all models and this one was inconsistent. For now we recommend that users that require parsing the CLI output use the output json format and do any kind of necessary parsing on it.

See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/balenaCLI.20bumps/near/452080642

Change-type: major

Resolves: #   
Depends-on: <url>   
See: <url>   

---


tests/commands/deploy.spec.ts
```diff
@@ -87,7 +87,6 @@ describe('balena deploy', function () {
 		api.expectGetRelease();
 		api.expectGetUser();
 		api.expectGetService({ serviceName: 'main' });
-		api.expectPostService409();
```
otaviojacobi, line 90: Previous `balena-compose` own baked 'getOrCreate' was actually a `createOrGet` (see: https://github.com/balena-io-modules/balena-compose/pull/54/files#diff-f061fd1a4a8154a1ce81603ac2d2bdd64d98b27b59757f802e391bfc25581b15L102-L118) . New approach is following pinejs-client `getOrCreate` (see: https://github.com/balena-io-modules/balena-compose/pull/54/files#diff-1b4d40e2938c5e5f9c316bc5dca572406d9285437650669024f871882a053cd2R211 and https://github.com/balena-io-modules/pinejs-client-js/blob/master/src/index.ts#L1562-L1577)

Therefore, this call doesn't happen anymore and does not need to be mocked

lib/commands/fleet/index.ts
```diff
@@ -78,16 +78,28 @@ export default class FleetCmd extends Command {
...
-			outputApplication,
-			['app_name', 'id', 'device_type', 'slug', 'commit'],
-			options,
+		if (options.json) {
+			console.log(JSON.stringify(applicationToDisplay, null, 4));
+			return;
+		}
+
+		// Emulate table.vertical title output, but avoid uppercasing and inserting spaces
+		console.log(`== ${applicationToDisplay.app_name}`);
```
thgreasi, line 95: Is it maybe the time to start using the slug as the header? Since that's what we advertise that ppl should be using :thinking:

lib/commands/fleets/index.ts
```diff
@@ -77,17 +77,29 @@ export default class FleetsCmd extends Command {
...
 				'id',
-				'app_name',
+				'app_name => NAME',
 				'slug',
 				'device_type',
-				'device_count',
 				'online_devices',
-			],
-			options,
+				'device_count',
```
thgreasi, line 101: Any reason that you moved this last, which in the json it is in the reverse order?
In the way it was before it was also more consistent w/ the order that we have those counts in the dashboard as well, which I'm guessing is where we got the idea to add `online_devices` to the cli as wel.

thgreasi (approved):

lib/commands/fleet/index.ts
otaviojacobi: I don't think so - we advertise people that the slug is the only unique thing but both can be used and are and will (as per previous discussions) still be supported, so I don't think this is really necessary

lib/commands/fleet/index.ts
```diff
@@ -78,16 +78,28 @@ export default class FleetCmd extends Command {
...
+		console.log(
+			getVisuals().table.vertical(applicationToDisplay, [
+				'id',
```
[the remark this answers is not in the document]
otaviojacobi: Same commment as before, I don't see a reasoning for it considering we already display the name

thgreasi:
:tada:

PR #2820, opened 2024-08-30
Added ability to SSH over tunnel.

robbe-haesendonck:
To allow for a more easy device management experience for the openbalena users, I "added" the ability to use the tunnel for getting SSH access.

This can be enabled by simply adding the `--usetunnel` flag, which will tunnel to the device and then open an ssh session over it.

Change-type: minor

### DIFF
--- a/src/commands/ssh/index.ts
+++ b/src/commands/ssh/index.ts
@@ -55,6 +55,7 @@ export default class SshCmd extends Command {
 	public static examples = [
 		'$ balena ssh MyFleet',
 		'$ balena ssh f49cefd',
+		'$ balena ssh f49cefd --usetunnel',
 		'$ balena ssh f49cefd my-service',
 		'$ balena ssh f49cefd --port <port>',
 		'$ balena ssh 192.168.0.1 --verbose',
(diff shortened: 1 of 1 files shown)


robbe-haesendonck:
Hello balena team!

I was wondering if this is something you'd be willing to pull/merge upstream?

If not I'll close this PR to reduce noise

myarmolinsky:
Hi, apologies for not responding to you sooner, somehow this PR missed our radar. We'll review and discuss it and get back to you as soon as possible

robbe-haesendonck:
No problem!
Thanks for the reply :)

myarmolinsky:
Hi, apologies for the delay. We have discussed your PR and we have concluded that the ideal solution would actually be to use a tunnel by default for the `ssh` command, and if the tunnel fails fallback on the current behavior (so no --usetunnel flag should be needed at all for this solution).

If you update your PR with this solution, we would happily review and test it when we have a chance. If you do not, we will look into it at a later date, but we do not have an ETA on when that would be.

robbe-haesendonck:
> Hi, apologies for the delay. We have discussed your PR and we have concluded that the ideal solution would actually be to use a tunnel by default for the `ssh` command, and if the tunnel fails fallback on the current behavior (so no `--usetunnel` flag should be needed at all for this solution).
> 
> If you update your PR with this solution, we would happily review and test it when we have a chance. If you do not, we will look into it at a later date, but we do not have an ETA on when that would be.

Thanks for the reply!

I'll pick it up asap and make tunneling the default behaviour and fallback on the current behaviour when tunneling fails.

PR #2834, opened 2024-09-18
Embed balena-image-manager instead of having it as a dependency

myarmolinsky:
Embed balena-image-manager instead of having it as a dependency

---


src/utils/cloud.ts
```diff
@@ -107,6 +107,302 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
...
+		? await getImage(deviceType, version)
+		: await doDownload({ ...options, deviceType, version });
+	// schedule the 'version' event for the next iteration of the event loop
+	// so that callers have a chance of adding an event handler
+	setImmediate(() =>
+		$stream.emit('balena-image-manager:resolved-version', version),
+	);
+	return $stream;
+};
+
```
otaviojacobi, line 405: Can we move this into a separated file (it can be under utils) like `image-manager.ts` or something along this lines? Otherwise this `cloud.ts` modules will just grow forever

src/commands/os/download.ts
```diff
@@ -19,6 +19,7 @@ import { Flags, Args } from '@oclif/core';
 import Command from '../../command';
 import * as cf from '../../utils/common-flags';
 import { stripIndent } from '../../utils/lazy';
+import { isESR } from '../../utils/cloud';
```
otaviojacobi, line 22: .e.g `../../utils/image-manager;`

myarmolinsky: Sure, makes sense

src/commands/os/download.ts
```diff
@@ -19,6 +19,7 @@ import { Flags, Args } from '@oclif/core';
 import Command from '../../command';
 import * as cf from '../../utils/common-flags';
 import { stripIndent } from '../../utils/lazy';
+import { isESR } from '../../utils/image-manager';
```
thgreasi, line 22: Let's make this an async import (exactly where we had the `await import('balena-image-manager');`), so that node only loads the "image-manager" code when we are in a command that actually uses it.

myarmolinsky:
> I would like a rephrase of `Remove balena-image-manager`. Maybe something `Embed the balena-image-manager instead of having it as a dependency` might be more accurate.

Done, but then I removed the `change-type` from the last commit for less redundancy in the changelog

src/utils/image-manager.ts
```diff
@@ -0,0 +1,312 @@
...
+	const { mkdirp } = await import('mkdirp');
+	await mkdirp(path.dirname(imagePath));
+
+	// Append .inprogress to streams, move them to the right location only on success
+	const inProgressPath = imagePath + '.inprogress';
+	const { promises, createWriteStream } = await import('fs');
+	const stream = createWriteStream(inProgressPath) as ReturnType<
+		typeof createWriteStream
+	> &
+		Record<'persistCache' | 'removeCache', () => Promise<void>>;
```
thgreasi, line 211: Would appreciate typings this explicitly since I think it justifies a bit more why we whose those pros:
```suggestion
	type ImageWritableStream = ReturnType<typeof createWriteStream> &
		Record<'persistCache' | 'removeCache', () => void>;
	const stream = createWriteStream(inProgressPath) as ImageWritableStream;
```
This way you can also add `Promise<ImageWritableStream>` as the return type of the function :)

tests/utils/image-manager/test-setup.js
```diff
@@ -0,0 +1,21 @@
+import { enable, registerMock, deregisterMock } from 'mockery';
```
thgreasi, line 1: Rename to `test-setup.ts`, in order to fix the v error:
```
 Exception during run: TypeError: Unknown file extension ".ts" for /home/runner/work/balena-cli/balena-cli/tests/utils/image-manager/image-manager.spec.ts
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:160:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:203:36)
    at defaultLoad (node:internal/modules/esm/load:143:22)
    at async ModuleLoader.load (node:internal/modules/esm/loader:553:7)
    at async ModuleLoader.moduleProvider (node:internal/modules/esm/loader:434:45)
    at async link (node:internal/modules/esm/module_job:87:21) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```

tests/utils/image-manager/image-manager.spec.ts
```diff
@@ -0,0 +1,547 @@
...
+		cleanupMocks();
+	});
+
+	describe('.getStream()', () => {
+		describe('given the existing image', function () {
+			beforeEach(function () {
+				this.image = tmp.fileSync();
+				fs.writeSync(this.image.fd, 'Cache image', 0, 'utf8');
+
+				this.cacheGetImagePathStub = stub(getImagePath);
```
thgreasi, line 44: This used to be `m.sinon.stub(cache, 'getImagePath')` with `cache = require('../build/cache')`, so I guess you need to do:
```suggestion
				this.cacheGetImagePathStub = stub(imageManager, 'getImagePath');
```
where `imageManager`  is something like:
```
import * as imageManager from '../???/build??/utils/image-manager';
```

myarmolinsky: Why wouldn't what I have now work? It looks like a valid overload of `stub` 🤔

myarmolinsky: Ah wait I see, it came from the *build* folder before

myarmolinsky: Changed it to
```
import {
	getFileCreatedDate,
	getImage,
	getImagePath,
	getImageWritableStream,
	getStream,
	isImageFresh,
} from '../../../build/utils/image-manager';
```

Kept using the stub overload I have for now though. If there's a specific reason one is better than the other, please let me know

thgreasi: I don't think that the tests will pass in the way you call the stub.
Afaik, the first arg needs to be an object, and the second arg needs to be the function that will get stubbed.
You can think of it like:
```
let originalFunc;
stub = (obj, methodName) => {
    originalFunc = obj[methodName];
    obj[methodName] = new StubbedFunc();
    return obj[methodName]
};
```

myarmolinsky: Yup, you are right

thgreasi (approved):

PR #2840, opened 2024-09-27
patch: change the name of the certificate

aethernet:
Now that we use the new certs from Apple, we need an Installer cert to sign a pkg.


thgreasi (approved):

myarmolinsky:
Closing as this is no longer necessary

PR #2843, opened 2024-10-08
patch: dedupe on ci

aethernet:
run `npm dd` as a postinstall step

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     ]
   },
   "scripts": {
-    "postinstall": "node patches/apply-patches.js",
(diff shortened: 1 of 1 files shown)


thgreasi:
I don't think that's a good idea, since postinstall runs on consumer's side as well
See: https://docs.npmjs.com/cli/v9/using-npm/scripts

aethernet:
@thgreasi why is that a problem?

thgreasi:
@aethernet for first it will slow down `npm i -g balena-cli` considerably, and it will also potentially update the local dependencies thus having them diverge from the ones that the shrinkwrap included.

PR #2895, opened 2024-12-16
Update @balena/lint to v9.1.3

otaviojacobi:
Update @balena/lint from 8.0.0 to 9.1.3

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,0 +1,32 @@
+const { FlatCompat } = require('@eslint/eslintrc');
+
+const compat = new FlatCompat({
+	baseDirectory: __dirname,
+});
+module.exports = [
+	...require('@balena/lint/config/eslint.config'),
+	...compat.config({
+		parserOptions: {
+			project: 'tsconfig.dev.json',
+		},
+		ignorePatterns: ['**/generate-completion.js', '**/bin/**/*'],
+		rules: {
+			ignoreDefinitionFiles: 0,
+			'@typescript-eslint/no-non-null-assertion': 'off',
+			'@typescript-eslint/no-shadow': 'off',
+			'@typescript-eslint/no-var-requires': 'off',
+			'@typescript-eslint/no-require-imports': 'off',
+			'@typescript-eslint/no-unnecessary-type-assertion': 'off',
+			'@typescript-eslint/prefer-nullish-coalescing': 'warn',
+
+			'no-restricted-imports': ['error', {
+				paths: ['resin-cli-visuals', 'chalk', 'common-tags', 'resin-cli-form'],
+			}],
+
+			'@typescript-eslint/no-unused-vars': ['error', {
+				argsIgnorePattern: '^_',
+				caughtErrorsIgnorePattern: '^_',
+			}],
+		},
+	}),
+];
--- a/src/commands/device/detect.ts
+++ b/src/commands/device/detect.ts
@@ -91,7 +91,7 @@ export default class DeviceDetectCmd extends Command {
 				try {
 					await docker.ping();
 					return true;
-				} catch (err) {
+				} catch {
 					return false;
 				}
 			}),
--- a/src/commands/device/init.ts
+++ b/src/commands/device/init.ts
@@ -155,7 +155,7 @@ export default class DeviceInitCmd extends Command {
 			try {
 				logger.logDebug(`Process failed, removing device ${device.uuid}`);
 				await balena.models.device.remove(device.uuid);
-			} catch (e) {
+			} catch {
 				// Ignore removal failures, and throw original error
 			}
 			throw e;
--- a/src/commands/device/logs.ts
+++ b/src/commands/device/logs.ts
@@ -135,7 +135,7 @@ export default class DeviceLogsCmd extends Command {
 			logger.logDebug('Checking we can access device');
 			try {
 				await deviceApi.ping();
-			} catch (e) {
+			} catch {
 				const { ExpectedError } = await import('../../errors');
 				throw new ExpectedError(
 					`Cannot access device at address ${params.device}.  Device may not be in local mode.`,
--- a/src/commands/device/register.ts
+++ b/src/commands/device/register.ts
@@ -76,6 +76,6 @@ export default class DeviceRegisterCmd extends Command {
 			options.deviceType,
 		);
 
-		return result && result.uuid;
+		return result.uuid;
 	}
 }
--- a/src/commands/device/ssh.ts
+++ b/src/commands/device/ssh.ts
@@ -82,7 +82,7 @@ export default class DeviceSSHCmd extends Command {
 				SSH server port number (default 22222) if the target is an IP address or .local
 				hostname. Otherwise, port number for the balenaCloud gateway (default 22).`,
 			char: 'p',
-			parse: async (p) => parseAsInteger(p, 'port'),
+			parse: (p) => parseAsInteger(p, 'port'),
 		}),
 		tty: Flags.boolean({
 			default: false,
@@ -110,13 +110,14 @@ export default class DeviceSSHCmd extends Command {
 		// Local connection
 		if (validateLocalHostnameOrIp(params.fleetOrDevice)) {
 			const { performLocalDeviceSSH } = await import('../../utils/device/ssh');
-			return await performLocalDeviceSSH({
+			await performLocalDeviceSSH({
 				hostname: params.fleetOrDevice,
 				port: options.port || 'local',
 				forceTTY: options.tty,
 				verbose: options.verbose,
 				service: params.service,
 			});
+			return;
 		}
 
 		// Remote connection
@@ -132,7 +133,7 @@ export default class DeviceSSHCmd extends Command {
 		const useProxy = !!proxyConfig && !options.noproxy;
 
 		// this will be a tunnelled SSH connection...
-		await checkNotUsingOfflineMode();
+		checkNotUsingOfflineMode();
 		await checkLoggedIn();
 		const deviceUuid = await getOnlineTargetDeviceUuid(
 			sdk,
--- a/src/commands/env/rename.ts
+++ b/src/commands/env/rename.ts
@@ -41,7 +41,7 @@ export default class EnvRenameCmd extends Command {
 		id: Args.integer({
 			required: true,
 			description: "variable's numeric database ID",
-			parse: async (input) => parseAsInteger(input, 'id'),
+			parse: (input) => parseAsInteger(input, 'id'),
 		}),
 		value: Args.string({
 			required: true,
--- a/src/commands/env/rm.ts
+++ b/src/commands/env/rm.ts
@@ -46,7 +46,7 @@ export default class EnvRmCmd extends Command {
 		id: Args.integer({
 			required: true,
 			description: "variable's numeric database ID",
-			parse: async (input) => parseAsInteger(input, 'id'),
+			parse: (input) => parseAsInteger(input, 'id'),
 		}),
 	};
 
--- a/src/commands/preload/index.ts
+++ b/src/commands/preload/index.ts
@@ -109,7 +109,7 @@ https://github.com/balena-io-examples/staged-releases\
 		'additional-space': Flags.integer({
 			description:
 				'expand the image by this amount of bytes instead of automatically estimating the required amount',
-			parse: async (x) => parseAsInteger(x, 'additional-space'),
+			parse: (x) => parseAsInteger(x, 'additional-space'),
 		}),
 		'add-certificate': Flags.string({
 			description: `\
@@ -126,7 +126,7 @@ Can be repeated to add multiple certificates.\
 		dockerPort: Flags.integer({
 			description:
 				'Docker daemon TCP port number (hint: 2375 for balena devices)',
-			parse: async (p) => parseAsInteger(p, 'dockerPort'),
+			parse: (p) => parseAsInteger(p, 'dockerPort'),
 		}),
 	};
 
@@ -155,7 +155,7 @@ Can be repeated to add multiple certificates.\
 					------------------------------------------------------------------------------
 					`);
 			}
-		} catch (error) {
+		} catch {
 			throw new ExpectedError(
 				`The provided image path does not exist: ${params.image}`,
 			);
@@ -192,11 +192,11 @@ Can be repeated to add multiple certificates.\
 				event.name,
 			));
 			if (event.action === 'start') {
-				return spinner.start();
-			} else {
-				console.log();
-				return spinner.stop();
+				spinner.start();
+				return;
 			}
+			console.log();
+			spinner.stop();
 		};
 
 		const commit = this.isCurrentCommit(options.commit || '')
--- a/src/commands/release/index.ts
+++ b/src/commands/release/index.ts
@@ -24,7 +24,7 @@ import { tryAsInteger } from '../../utils/validation';
 import { jsonInfo } from '../../utils/messages';
 
 export const commitOrIdArg = Args.custom({
-	parse: async (commitOrId: string) => tryAsInteger(commitOrId),
+	parse: tryAsInteger,
 });
 
 type FlagsDef = Interfaces.InferredFlags<typeof ReleaseCmd.flags>;
--- a/src/commands/ssh-key/rm.ts
+++ b/src/commands/ssh-key/rm.ts
@@ -40,7 +40,7 @@ export default class SSHKeyRmCmd extends Command {
 	public static args = {
 		id: Args.integer({
 			description: 'balenaCloud ID for the SSH key',
-			parse: async (x) => parseAsInteger(x, 'id'),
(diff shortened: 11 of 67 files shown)


src/commands/device/detect.ts
```diff
@@ -91,7 +91,7 @@ export default class DeviceDetectCmd extends Command {
 				try {
 					await docker.ping();
 					return true;
-				} catch (err) {
+				} catch (_e) {
```
myarmolinsky, line 94:
```suggestion
				} catch {
```

src/commands/device/init.ts
```diff
@@ -155,7 +155,7 @@ export default class DeviceInitCmd extends Command {
 			try {
 				logger.logDebug(`Process failed, removing device ${device.uuid}`);
 				await balena.models.device.remove(device.uuid);
-			} catch (e) {
+			} catch (_e) {
```
myarmolinsky, line 158:
```suggestion
			} catch {
```

src/commands/device/logs.ts
```diff
@@ -135,7 +135,7 @@ export default class DeviceLogsCmd extends Command {
 			logger.logDebug('Checking we can access device');
 			try {
 				await deviceApi.ping();
-			} catch (e) {
+			} catch (_e) {
```
myarmolinsky, line 138:
```suggestion
			} catch {
```

src/commands/device/register.ts
```diff
@@ -76,6 +76,6 @@ export default class DeviceRegisterCmd extends Command {
 			options.deviceType,
 		);
 
-		return result && result.uuid;
+		return result.uuid;
```
myarmolinsky, line 79: not `result?.uuid`?

eslint.config.js
```diff
@@ -0,0 +1,32 @@
...
+		},
+		ignorePatterns: ['**/generate-completion.js', '**/bin/**/*'],
+		rules: {
+			ignoreDefinitionFiles: 0,
+			'@typescript-eslint/no-non-null-assertion': 'off',
+			'@typescript-eslint/no-shadow': 'off',
+			'@typescript-eslint/no-var-requires': 'off',
+			'@typescript-eslint/no-require-imports': 'off',
+			'@typescript-eslint/no-unnecessary-type-assertion': 'off',
+			'@typescript-eslint/prefer-nullish-coalescing': 'warn',
```
myarmolinsky, line 20: Why do we need these? I'm okay with keeping these two but I think we can just remove the others
```suggestion
			'@typescript-eslint/no-var-requires': 'off',
			'@typescript-eslint/no-require-imports': 'off',
```

myarmolinsky: > warn

I would prefer to address everything for that matter, warnings included

src/commands/device/ssh.ts
```diff
@@ -82,7 +82,7 @@ export default class DeviceSSHCmd extends Command {
 				SSH server port number (default 22222) if the target is an IP address or .local
 				hostname. Otherwise, port number for the balenaCloud gateway (default 22).`,
 			char: 'p',
-			parse: async (p) => parseAsInteger(p, 'port'),
+			parse: async (p) => Promise.resolve(parseAsInteger(p, 'port')),
```
myarmolinsky, line 85: nit: I see you kept the `async` here but removed it in other spots. We should pick and stick to one :)

myarmolinsky: But do we even need the `Promise.resolve` in all these cases? Does `parse` require a promise? I would assume we can just remove the `async` and not have a `Promise.resolve` at all

src/commands/env/rename.ts
```diff
@@ -41,7 +41,7 @@ export default class EnvRenameCmd extends Command {
 		id: Args.integer({
 			required: true,
 			description: "variable's numeric database ID",
-			parse: async (input) => parseAsInteger(input, 'id'),
+			parse: (input) => Promise.resolve(parseAsInteger(input, 'id')),
```
myarmolinsky, line 44: Do we need the Promise.resolve in all these cases? Does parse require a promise? I would assume we can just remove the async and not have a Promise.resolve at all

src/commands/env/rm.ts
```diff
@@ -46,7 +46,7 @@ export default class EnvRmCmd extends Command {
 		id: Args.integer({
 			required: true,
 			description: "variable's numeric database ID",
-			parse: async (input) => parseAsInteger(input, 'id'),
+			parse: (input) => Promise.resolve(parseAsInteger(input, 'id')),
```
myarmolinsky, line 49: As above

src/commands/preload/index.ts
```diff
@@ -109,7 +109,7 @@ https://github.com/balena-io-examples/staged-releases\
 		'additional-space': Flags.integer({
 			description:
 				'expand the image by this amount of bytes instead of automatically estimating the required amount',
-			parse: async (x) => parseAsInteger(x, 'additional-space'),
+			parse: (x) => Promise.resolve(parseAsInteger(x, 'additional-space')),
```
myarmolinsky, line 112: As above

src/commands/preload/index.ts
```diff
@@ -126,7 +126,7 @@ Can be repeated to add multiple certificates.\
 		dockerPort: Flags.integer({
 			description:
 				'Docker daemon TCP port number (hint: 2375 for balena devices)',
-			parse: async (p) => parseAsInteger(p, 'dockerPort'),
+			parse: (p) => Promise.resolve(parseAsInteger(p, 'dockerPort')),
```
myarmolinsky, line 129: As above

src/commands/preload/index.ts
```diff
@@ -155,7 +155,7 @@ Can be repeated to add multiple certificates.\
 					------------------------------------------------------------------------------
 					`);
 			}
-		} catch (error) {
+		} catch (_e) {
```
myarmolinsky, line 158:
```suggestion
		} catch {
```

src/commands/preload/index.ts
```diff
@@ -192,10 +192,12 @@ Can be repeated to add multiple certificates.\
 				event.name,
 			));
 			if (event.action === 'start') {
-				return spinner.start();
+				spinner.start();
+				return;
 			} else {
```
myarmolinsky, line 197: I know this was here before but we don't even need the `else` since we `return`. We can just call `spinner.stop()` right after. We also don't even need that last return after the `spinner.stop` I think 🤔

src/commands/release/index.ts
```diff
@@ -24,7 +24,7 @@ import { tryAsInteger } from '../../utils/validation';
 import { jsonInfo } from '../../utils/messages';
 
 export const commitOrIdArg = Args.custom({
-	parse: async (commitOrId: string) => tryAsInteger(commitOrId),
+	parse: (commitOrId: string) => Promise.resolve(tryAsInteger(commitOrId)),
```
myarmolinsky, line 27: See other `parse` comments

src/commands/ssh-key/rm.ts
```diff
@@ -40,7 +40,7 @@ export default class SSHKeyRmCmd extends Command {
 	public static args = {
 		id: Args.integer({
 			description: 'balenaCloud ID for the SSH key',
-			parse: async (x) => parseAsInteger(x, 'id'),
+			parse: (x) => Promise.resolve(parseAsInteger(x, 'id')),
```
myarmolinsky, line 43: See other `parse` comments

src/errors.ts
```diff
@@ -229,7 +229,7 @@ async function sentryCaptureException(error: Error) {
 	Sentry.captureException(error);
 	try {
 		await Sentry.close(1000);
-	} catch (e) {
+	} catch (_e) {
```
myarmolinsky, line 232:
```suggestion
	} catch {
```

src/utils/cloud.ts
```diff
@@ -165,7 +165,8 @@ export async function downloadOSImage(
 		if (state != null) {
 			return bar.update(state);
 		} else {
-			return spinner.start();
+			spinner.start();
+			return;
 		}
```
myarmolinsky, line 170: We can remove the `else` and just call `spinner.start` after. And I think we can remove the `return;`

src/utils/compose.ts
```diff
@@ -476,9 +476,9 @@ export class BuildProgressUI implements Renderer {
...
 						return `${bar} ${status}`;
 					}
-					return `${bar}`;
+					return bar;
 				} else if (status) {
-					return `${status}`;
+					return status;
 				} else {
 					return 'Waiting...';
 				}
```
myarmolinsky, line 484: We should be able to remove the elses because of the returns:
```
}
if (status) {
  return status;
}
return 'Waiting...';
```

src/utils/compose.ts
```diff
@@ -607,7 +607,7 @@ export class BuildProgressInline implements Renderer {
 			if (error) {
 				return `${error}`;
 			} else if (status) {
```
myarmolinsky, line 609: As above

src/utils/device/api.ts
```diff
@@ -252,21 +252,26 @@ export class DeviceAPI {
...
-							);
-						case 503:
-							return reject(
-								new ApiErrors.ServiceUnavailableAPIError(body.message),
-							);
-						default:
-							return reject(new ApiErrors.DeviceAPIError(body.message));
+						case 200: {
+							resolve(body);
+							return;
```
myarmolinsky, line 261: nit: I think more common practice is to use `break` in a switch, but maybe I'm mistaken

src/utils/device/deploy.ts
```diff
@@ -143,7 +143,7 @@ export async function deployToDevice(opts: DeviceDeployOptions): Promise<void> {
 	try {
 		globalLogger.logDebug('Checking we can access device');
 		await api.ping();
-	} catch (e) {
+	} catch (_e) {
```
myarmolinsky, line 146:
```suggestion
	} catch {
```

src/utils/device/live.ts
```diff
@@ -252,7 +252,7 @@ export class LivepushManager {
 					try {
 						// sync because chokidar defines a sync interface
 						stats = fs.lstatSync(filePath);
-					} catch (err) {
+					} catch (_e) {
```
myarmolinsky, line 255:
```suggestion
					} catch {
```

src/utils/docker.ts
```diff
@@ -57,7 +57,7 @@ export const dockerConnectionCliFlags = {
 		description:
 			'Docker daemon TCP port number (hint: 2375 for balena devices)',
 		char: 'p',
-		parse: async (p) => parseAsInteger(p, 'dockerPort'),
+		parse: (p) => Promise.resolve(parseAsInteger(p, 'dockerPort')),
```
myarmolinsky, line 60: See other `parse` comments

src/utils/normalization.ts
```diff
@@ -78,14 +78,16 @@ export async function disambiguateReleaseParam(
 /**
  * Convert to lowercase if looks like slug
  */
-export async function lowercaseIfSlug(s: string) {
-	return s.includes('/') ? s.toLowerCase() : s;
+export function lowercaseIfSlug(s: string) {
+	return s.includes('/')
+		? Promise.resolve(s.toLowerCase())
+		: Promise.resolve(s);
```
myarmolinsky, line 84: What is the purpose of making these `Promise.resolve` instead of just letting the function return the strings?

src/utils/promote.ts
```diff
@@ -390,13 +390,12 @@ async function createApplication(
 	try {
 		const userInfo = await sdk.auth.getUserInfo();
 		username = userInfo.username;
-	} catch (err) {
+	} catch (_e) {
```
myarmolinsky, line 393:
```suggestion
	} catch {
```

src/utils/promote.ts
```diff
@@ -418,11 +417,13 @@ async function createApplication(
 						'You already have a fleet with that name; please choose another.',
 					);
 					continue;
-				} catch (err) {
-					return resolve(appName);
+				} catch (_e) {
```
myarmolinsky, line 420:
```suggestion
				} catch {
```

src/utils/tunnel.ts
```diff
@@ -61,11 +61,11 @@ export const tunnelConnectionToDevice = (
 					client.pipe(remote);
 					remote.pipe(client);
 					remote.on('error', (err) => {
-						console.error('Remote: ' + err);
+						console.error('Remote: ' + `${err}`);
```
myarmolinsky, line 64:
```suggestion
						console.error(`Remote: ${err}`);
```

src/utils/tunnel.ts
```diff
@@ -61,11 +61,11 @@ export const tunnelConnectionToDevice = (
...
 						client.end();
 					});
 					client.on('error', (err) => {
-						console.error('Client: ' + err);
+						console.error('Client: ' + `${err}`);
```
myarmolinsky, line 68:
```suggestion
						console.error(`Client: ${err}`);
```

tests/docker-build.ts
```diff
@@ -184,12 +182,13 @@ export async function testDockerBuildStream(o: {
...
 		o.dockerMock.expectPostBuild({
 			...o,
-			checkURI: async (uri: string) => {
+			checkURI: (uri: string) => {
 				const url = new URL(uri, 'http://test.net/');
 				const queryParams = Array.from(url.searchParams.entries());
 				expect(deepJsonParse(queryParams)).to.have.deep.members(
 					deepJsonParse(expectedQueryParams),
 				);
+				return Promise.resolve();
```
myarmolinsky, line 191: Why `Promise.resolve`?

tests/docker-build.ts
```diff
@@ -241,12 +240,13 @@ export async function testPushBuildStream(o: {
...
 	o.builderMock.expectPostBuild({
 		...o,
-		checkURI: async (uri: string) => {
+		checkURI: (uri: string) => {
 			const url = new URL(uri, 'http://test.net/');
 			const queryParams = Array.from(url.searchParams.entries());
 			expect(deepJsonParse(queryParams)).to.have.deep.members(
 				deepJsonParse(expectedQueryParams),
 			);
+			return Promise.resolve();
```
myarmolinsky, line 249: `Promise.resolve`

tests/projects.ts
```diff
@@ -50,7 +50,7 @@ export async function exists(fPath: string) {
 	try {
 		await fs.stat(fPath);
 		return true;
-	} catch (e) {
+	} catch (_e) {
```
myarmolinsky, line 53:
```suggestion
	} catch {
```

src/commands/version/index.ts
```diff
@@ -69,10 +69,9 @@ export default class VersionCmd extends Command {
 		const { flags: options } = await this.parse(VersionCmd);
 		const versions: JsonVersions = {
 			'balena-cli': (await import('../../../package.json')).version,
-			'Node.js':
-				process.version && process.version.startsWith('v')
-					? process.version.slice(1)
-					: process.version,
+			'Node.js': process.version.startsWith('v')
```
myarmolinsky, line 72: not `process.version?.startsWith('v')`

src/commands/device/register.ts
otaviojacobi: Typing says uuid will be present (API confirmed the behavior) so this check was/should not be needed

eslint.config.js
otaviojacobi: @myarmolinsky Yes, I would prefer address everything but nullish-coalescing would be huge and potential to breaking a lot of stuff. I would rather have them as a warning for now and migrate step by step. The others are historical reasons, I didn't add them they just the ones currently used for the CLI. The only one I added was `'@typescript-eslint/no-require-imports'` for obvious reasons.

src/commands/device/ssh.ts
otaviojacobi: The expected and enforced interface by oclif is to return a promise. But I agree, the Promise.resolve should be inside the parseAsInteger function and any other parse function. Fixed.

src/commands/preload/index.ts
otaviojacobi: Sure. This was auto corrected by the linter, but fixed anyway

src/utils/device/api.ts
otaviojacobi: Would rather avoid touching potentially breaking runtime code on a linter bum

tests/docker-build.ts
otaviojacobi: The expected interface (which I would rather not change on a linter bump) expects a promise. Beforehand we were using the syntax of a `async` function without doing any async work inside of it. The linter no longer likes this syntax and prefer the more explicit `Promise.resolve()` for a `Promise<void>` type

src/commands/version/index.ts
otaviojacobi: Nope. process.version is a mandatory property from nodejs runtime.

myarmolinsky (approved):

PR #2908, opened 2025-01-06
Enable dual stack support for balena tunnel

pipex:
The `balena tunnel`commands uses `localhost` as the default listening address. The problem is that on some dual stack networks (IPv4/IPv6), this can result on the tunneling server only binding to `::1` and not to `127.0.0.1`, potentially causing issues as this behavior is not intuitive.

This PR replaces the default server name with `::` which setups the tunneling server in dual-stack mode.

Change-type: patch

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2304,6 +2304,73 @@
         "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       }
     },
+    "node_modules/@grpc/grpc-js": {
+      "version": "1.12.5",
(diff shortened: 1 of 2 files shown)


pipex:
Note that this may have some security implications as `::` is equivalent to binding to `0.0.0.0`. Putting this in draft pending some [internal discussion](https://balena.zulipchat.com/#narrow/channel/403752-channel.2Fsupport-help/topic/balena.20tunnel.20waiting.20for.20connections/near/492157065)

pipex:
This implementation creates a security risk exposing the tunneled device to the internal network of the client device. Instead of this, the CLI should give the user the choice of stack (or autodetect) and tunnel to `localhost` or `::1` accordingly

PR #2914, opened 2025-02-21
Deny preload for an image with secure boot enabled

kb2ma:
Deny an attempt to preload an app into an image with secure boot enabled, and provide a message to the user. BalenaOS does not support this action.

Implementation includes a generic, callback-driven capability to explore a partition in `utils/explore-contents` module. Allows for the caller to determine success or failure.

### DIFF
--- a/src/commands/preload/index.ts
+++ b/src/commands/preload/index.ts
@@ -37,6 +37,7 @@ import type {
 	Release,
 } from 'balena-sdk';
 import type { Preloader } from 'balena-preload';
+import type * as Fs from 'fs';
 
 export default class PreloadCmd extends Command {
 	public static description = stripIndent`
@@ -159,6 +160,37 @@ Can be repeated to add multiple certificates.\
 			throw new ExpectedError(
 				`The provided image path does not exist: ${params.image}`,
 			);
+		}
+
+		// Verify that image is not enabled for secure boot. First, confirm it
+		// is a secure boot image with an /opt/*.sig file in the rootA partition.
+		const { explorePartition, BalenaPartition } = await import(
+			'../../utils/image-contents'
+		);
+		const isSecureBoot = await explorePartition<boolean>(
+			params.image,
+			BalenaPartition.ROOTA,
+			async (fs: typeof Fs): Promise<boolean> => {
+				try {
+					const files = await fs.promises.readdir('/opt');
+					return files.some((el) => el.endsWith('balenaos-img.sig'));
+				} catch {
+					// Typically one of:
+					// - Error: No such file or directory
+					// - Error: Unsupported filesystem.
+					// - ErrnoException: node_ext2fs_open ENOENT (44) args: [5261576,5268064,"r",0]
+					return false;
+				}
+				return false;
+			},
+		);
+		// Next verify that config.json enables secureboot.
+		if (isSecureBoot) {
+			const { read } = await import('balena-config-json');
+			const config = await read(params.image, '');
+			if (config.installer?.secureboot === true) {
+				throw new ExpectedError("Can't preload image with secure boot enabled");
+			}
 		}
 
 		// balena-preload currently does not work with numerical app IDs
--- a/src/utils/image-contents.ts
+++ b/src/utils/image-contents.ts
@@ -1,0 +1,69 @@
+/**
+ * @license
+ * Copyright 2025 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Utilities to explore the contents in a balenaOS image.
+
+import * as imagefs from 'balena-image-fs';
+import * as filedisk from 'file-disk';
+import { getPartitions } from 'partitioninfo';
+import type * as Fs from 'fs';
+
+/**
+ * @summary IDs for the standard balenaOS partitions
+ * @description Values are the base name for a partition on disk
+ */
+export enum BalenaPartition {
+	BOOT = 'boot',
+	ROOTA = 'rootA',
+	ROOTB = 'rootB',
+	STATE = 'state',
+	DATA = 'data',
+}
+
+/**
+ * @summary Allow a provided function to explore the contents of one of the well-known
+ * partitions of a balenaOS image
+ *
+ * @param {string} imagePath - pathname of image for search
+ * @param {BalenaPartition} partitionId - partition to find
+ * @param {(fs) => Promise<T>} - function for exploration
+ * @returns {T}
+ */
+export async function explorePartition<T>(
+	imagePath: string,
+	partitionId: BalenaPartition,
+	exploreFn: (fs: typeof Fs) => Promise<T>,
+): Promise<T> {
+	return await filedisk.withOpenFile(imagePath, 'r', async (handle) => {
+		const disk = new filedisk.FileDisk(handle, true, false, false);
+		const partitionInfo = await getPartitions(disk, {
+			includeExtended: false,
+			getLogical: true,
+		});
+
+		const findResult = await imagefs.findPartition(disk, partitionInfo, [
+			`resin-${partitionId}`,
+			`flash-${partitionId}`,
+			`balena-${partitionId}`,
+		]);
+		if (findResult == null) {
+			throw new Error(`Can't find partition for ${partitionId}`);
+		}
+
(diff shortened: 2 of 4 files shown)


src/commands/preload/index.ts
```diff
@@ -161,6 +163,40 @@ Can be repeated to add multiple certificates.\
...
+		// is a secure boot image with an /opt/*.sig file in the rootA partition.
+		const { explorePartition, BalenaPartition } = await import(
+			'../../utils/image-contents'
+		);
+		const isSecureBoot = await explorePartition<boolean>(
+			params.image,
+			BalenaPartition.ROOTA,
+			async (fs: typeof Fs): Promise<boolean> => {
+				try {
+					const promiseDir = promisify(fs.readdir);
```
rcooke-warwick, line 176: Maybe instead of using `promisify` here, we can use the native promises api: https://nodejs.org/docs/latest-v20.x/api/fs.html#fspromisesreaddirpath-options to avoid having to do this?

src/commands/preload/index.ts
```diff
@@ -161,6 +163,40 @@ Can be repeated to add multiple certificates.\
...
 		}
 
+		// Verify that image is not enabled for secure boot. First, confirm it
+		// is a secure boot image with an /opt/*.sig file in the rootA partition.
+		const { explorePartition, BalenaPartition } = await import(
+			'../../utils/image-contents'
+		);
+		const isSecureBoot = await explorePartition<boolean>(
+			params.image,
+			BalenaPartition.ROOTA,
```
rcooke-warwick, line 173: Not familiar with the `BalenaPartition` abstraction, but what happens here if this is a non flasher image, , with `flash-boot`, `resin-rootA` etc - does it just continue, then get caught on `line 179`?

src/commands/preload/index.ts
```diff
@@ -161,6 +163,40 @@ Can be repeated to add multiple certificates.\
...
+					// - Error: No such file or directory
+					// - Error: Unsupported filesystem.
+					// - ErrnoException: node_ext2fs_open ENOENT (44) args: [5261576,5268064,"r",0]
+					return false;
+				}
+				return false;
+			},
+		);
+		// Next verify that config.json enables secureboot.
+		if (isSecureBoot) {
```
rcooke-warwick, line 190: Not relevant to this PR, but maybe elsewhere we can add a check to the CLI that detects if someone has tried to enable secureboot on a non signed image, and warn them that this won't work, to avoid any confusion / chance of them thinking they've deployed a secure device but haven't!

src/commands/preload/index.ts
thgreasi: @rcooke-warwick this is not the `fs` namespace imported at the top level, but it's the virtual `fs` "namespace" that balena-image-fs gives us.
That being said, since we recently added support for promises in balena-image-fs, @kb2ma should be able to just use v directly.
```suggestion
					const files = await fs.promises.readdir('/opt');
```

src/utils/image-contents.ts
```diff
@@ -0,0 +1,122 @@
...
+				partNumber = innerPartNumber;
+			}
+		}
+
+		return await imagefs.interact<T>(disk, partNumber, exploreFn);
+	});
+}
+
+// Find partition, by partition name on GPT or by filesystem label on MBR.
+async function findPartition(
```
thgreasi, line 87: Since that's soooo similar to `findBootPartitionByName()` of balena-config-json, how about avoiding the duplication by creating and exposing a more generic function like the one you did here?
We can then have `findBootPartitionByName()` use this new `findPartition()` by wiring in the names of interest for the partition.

kb2ma: I thought about placing that function in balena-config-json, but it seemed out of scope to expose a generic partition finding function in a repo that's just about the boot partition.

Would it work better to put it in balena-image-fs? We also could rework the `findBootPartitionByName()` in balena-config-json to use that as well.

src/commands/preload/index.ts
kb2ma: I see the CLI already has a [utility function](https://github.com/balena-io/balena-cli/blob/4f665f43d2a91e2e4a5b4c2a98840424e6a0bc3c/src/utils/config.ts#L181) for such a check for `os configure`. Is that sufficient? 

In that more generic case, is it better to check the image itself as we do here? Actually AIUI there are older versions of the secure boot implementation that do not require a signature file, so a server check is appropriate for that case, correct?

src/utils/image-contents.ts
kb2ma: See balena-io-modules/balena-image-fs#84

kb2ma:
Updated for hopefully final changes to balena-io-modules/balena-image-fs#84. Will update once more after that PR merges to use the published package version, and also dedupe at that time.

src/commands/preload/index.ts
```diff
@@ -161,6 +162,37 @@ Can be repeated to add multiple certificates.\
...
+		const { explorePartition, BalenaPartition } = await import(
+			'../../utils/image-contents'
+		);
+		const isSecureBoot = await explorePartition<boolean>(
+			params.image,
+			BalenaPartition.ROOTA,
+			async (fs: typeof Fs): Promise<boolean> => {
+				try {
+					const files = await fs.promises.readdir('/opt');
+					return files.some((el) => el.endsWith('balenaos-img.sig'));
```
thgreasi, line 176: @kb2ma can you give an example of the file names that we try to match?
Could we just do `files.some((el) => el === 'balenaos-img.sig');` or `files.some((el) => el === '/opts/balenaos-img.sig');`?

src/commands/preload/index.ts
```diff
@@ -161,6 +162,37 @@ Can be repeated to add multiple certificates.\
...
+					return false;
+				}
+				return false;
+			},
+		);
+		// Next verify that config.json enables secureboot.
+		if (isSecureBoot) {
+			const { read } = await import('balena-config-json');
+			const config = await read(params.image, '');
+			if (config.installer?.secureboot === true) {
```
thgreasi, line 191: How about the case that `config.installer?.secureboot === true` but it's not a secure boot image?
Should we log or throw something in that case?

thgreasi (approved):
Approving to unblock you sooner than later, since all my comments can be addressed if needed at a later point, or before merging.
Kind reminder (I know you already remember) to squash your commits (it's ok to leave the `Dedupe dependencies` one as a separate w/o a change-type, to make clear that these changes were unrelated with the bump).

src/commands/preload/index.ts
kb2ma: @thgreasi Here is the content of the flash-rootA directory in the flasher image for generic-amd64 device type. There must be an `/opt` directory, and I don't think we want to read all of those other directories. As you can see the signature filename includes the device type. I think the current check is specific enough to reasonably exclude false positives.

```
$ ls -l tmp
total 92
drwxr-xr-x  2 root root         4096 Mar  9  2018 bin
drwxr-xr-x  3 root root         4096 Mar  9  2018 boot
drwxr-xr-x  2 root root         4096 Mar  9  2018 dev
drwxr-xr-x 38 root root         4096 Mar  9  2018 etc
drwxr-xr-x  8 1001 sentinelone  4096 Mar  9  2018 flash-boot
drwxr-xr-x  3 root root         4096 Mar  9  2018 home
drwxr-xr-x  9 root root         4096 Mar  9  2018 lib
lrwxrwxrwx  1 root root           19 Mar  9  2018 linuxrc -> /bin/busybox.nosuid
drwx------  2 root root        16384 Feb 12 12:23 lost+found
drwxr-xr-x  2 root root         4096 Mar  9  2018 media
drwxr-xr-x  4 root root         4096 Mar  9  2018 mnt
drwxr-xr-x  2 root root         4096 Mar  9  2018 opt
dr-xr-xr-x  2 root root         4096 Mar  9  2018 proc
drwxr-xr-x  3 root root         4096 Mar  9  2018 quirks
drwxr-xr-x  2 root root         4096 Mar  9  2018 run
drwxr-xr-x  2 root root         4096 Mar  9  2018 sbin
drwxr-xr-x  2 root root         4096 Mar  9  2018 srv
dr-xr-xr-x  2 root root         4096 Mar  9  2018 sys
drwxrwxrwt  2 root root         4096 Mar  9  2018 tmp
drwxr-xr-x 10 root root         4096 Mar  9  2018 usr
drwxr-xr-x  8 root root         4096 Mar  9  2018 var

$ ls -l tmp/opt
total 866420
-rw-r--r-- 1 root root 2378170368 Feb 28 22:03 balena-image-generic-amd64.balenaos-img
-rw-r--r-- 1 root root        512 Mar  9  2018 balena-image-generic-amd64.balenaos-img.sig
```

src/commands/preload/index.ts
kb2ma: There may be other mechanisms in the future to identify whether an image is secure boot. So I would not want to add code now to question the use of the flag in config.json is the image does not have the `.sig` file.

src/commands/preload/index.ts
thgreasi: @kb2ma I think that if you could just add a comment right above `eg: balena-image-generic-amd64.balenaos-img.sig` we remember why it's like this when anyone checks it in the future it would be great imo.

PR #2921, opened 2025-02-28
Major 21

myarmolinsky:
---

### DIFF
--- a/src/commands/api-key/generate.ts
+++ b/src/commands/api-key/generate.ts
@@ -17,7 +17,16 @@
 
 import { Args, Command } from '@oclif/core';
 import { ExpectedError } from '../../errors';
-import { getBalenaSdk, stripIndent } from '../../utils/lazy';
+import { getBalenaSdk, getCliForm, stripIndent } from '../../utils/lazy';
+import {
+	formatDuration,
+	intervalToDuration,
+	isValid,
+	parseISO,
+} from 'date-fns';
+
+// In days
+const durations = [1, 7, 30, 90];
 
 async function isLoggedInWithJwt() {
 	const balena = getBalenaSdk();
@@ -41,12 +50,20 @@ export default class GenerateCmd extends Command {
 		This key can be used to log into the CLI using 'balena login --token <key>',
 		or to authenticate requests to the API with an 'Authorization: Bearer <key>' header.
 `;
-	public static examples = ['$ balena api-key generate "Jenkins Key"'];
+	public static examples = [
+		'$ balena api-key generate "Jenkins Key"',
+		'$ balena api-key generate "Jenkins Key" 2025-10-30',
+		'$ balena api-key generate "Jenkins Key" never',
+	];
 
 	public static args = {
 		name: Args.string({
 			description: 'the API key name',
 			required: true,
+		}),
+		expiryDate: Args.string({
+			description:
+				'the expiry date of the API key as an ISO date string, or "never" for no expiry',
 		}),
 	};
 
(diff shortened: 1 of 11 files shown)


thgreasi (changes requested):
You probably need to either squash the `Drop support for OS versions <2.0.3` commit into the `Address SDK major v21 breaking changes` one, or if you want to be more explicit change it to `<=2.14.0`.

My point is, SDK V21 already dropped support for OS versions <= 2.14.0, so if here we state `<2.0.3`, it will look misleading (ie the cli somehow still manages to support OS between 2.0.3 and 2.14.0, even though the SDK doesn't)

src/commands/api-key/generate.ts
```diff
@@ -41,23 +45,69 @@ export default class GenerateCmd extends Command {
...
+			description: 'the expiry date of the API key, or "never" for no expiry',
+		}),
 	};
 
 	public static authenticated = true;
 
 	public async run() {
 		const { args: params } = await this.parse(GenerateCmd);
 
+		let expiryDateResponse: string | number = params.expiryDate ?? '';
```
thgreasi, line 69: I would prefer undefined instead of empty string.
```suggestion
		let expiryDateResponse: string | number | undefined = params.expiryDate;
```
You can then also change the `if (params.expiryDate === undefined)` to use the new `expiryDateResponse` var, and then TS should be able to understand that `expiryDateResponse` can't be undefined after the `if () {}`.

src/commands/api-key/generate.ts
```diff
@@ -41,23 +45,69 @@ export default class GenerateCmd extends Command {
...
+			let expiryDate: string | null = null;
+			if (expiryDateResponse === 'never') {
+				expiryDate = null;
+			} else if (expiryDateResponse === 'custom') {
+				expiryDate = await getCliForm().ask({
+					message: 'Please enter an expiry date for the API key',
+					type: 'input',
+				});
+			} else if (typeof expiryDateResponse === 'string') {
+				expiryDate = expiryDateResponse;
```
thgreasi, line 101: Do we want to also verify somehow the value it holds in these cases (eg require them to follow the ISO format) or are we fine passing them as-is to pine?

src/commands/preload/index.ts
```diff
@@ -295,7 +295,7 @@ Can be repeated to add multiple certificates.\
 		owns__release: {
 			$select: ['id', 'commit', 'end_timestamp', 'composition'],
 			$expand: {
-				contains__image: {
+				release_image: {
```
thgreasi, line 298: Please double check that this works, since I don't see any related code that consumes this result being updated.

src/utils/image-manager.ts
```diff
@@ -102,12 +102,7 @@ export const isImageFresh = async (deviceType: string, version: string) => {
 		return false;
 	}
 
-	const balena = getBalenaSdk();
-	const lastModifiedDate = await balena.models.os.getLastModified(
-		deviceType,
-		version,
-	);
-	return lastModifiedDate < createdDate;
+	return true;
```
thgreasi, line 105: I guess you could even delete this whole function since there is no point on fetching the image's creation date in the cache folder.

src/commands/preload/index.ts
myarmolinsky: Seems to me this is used in `balena-preload`. Opened a PR updating the SDK and addressing breaking changes there as well: https://github.com/balena-io-modules/balena-preload/pull/299

src/commands/api-key/generate.ts
myarmolinsky: I would say the SDK should be verifying the value. Currently it does not verify the value. If the value is invalid, it throws a `BalenaNotLoggedIn` error, which is also wrong.

thgreasi: Ignoring the invalid `BalenaNotLoggedIn` error for a moment, the sdk tries to support whatever pine allows, so in that context it doesn't have more checks, and that's fine since the sdk is consumed from within other software, when both iso date strings, and numeric timestamps, and Date objects are expected to work.

On the other hand though, the CLI is expecting user input from a terminal, so passing numeric hand-written timestamps there doesn't sound like a real use case. Given that, I was proposing to make the allowed input format of the expiry date to require iso date strings, since that strictness would rule out completely a number of edge cases (eg do you remember the issues that we had to go through when users could also pass numeric ids in commands and we had to parse them and try whether they were indeed ids?).

thgreasi (approved):

src/commands/api-key/generate.ts
```diff
@@ -41,23 +50,83 @@ export default class GenerateCmd extends Command {
...
+				} while (!isValid(expiryDate));
+			} else if (typeof expiryDateResponse === 'string') {
+				expiryDate = parseISO(expiryDateResponse);
+				if (!isValid(expiryDate)) {
+					throw new Error(
+						'Invalid date format, please use a valid ISO date string',
+					);
+				}
+			} else {
+				expiryDate = new Date(
```
thgreasi, line 122: NIT: I might have preferred handing the case of a numeric response higher and live the string case last, since the string case has a parseISO validation in any case.
But this might be just me being paranoid :)

PR #2928, opened 2025-04-01
Add comment with example of signature file in a flasher image

kb2ma:
Follow up to #2914. Adds a comment to illustrate presence of signature file in a flash-rootA partition. The `preload` command must determine if this file is present.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2682,9 +2682,9 @@
       }
     },
     "node_modules/@inquirer/core/node_modules/@types/node": {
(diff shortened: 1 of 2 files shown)


thgreasi:
@kb2ma How about
* removing the change-type from the `Add comment with secure boot signature file example for preload` commit
* then bumping (eg) balena-config-json in a new commit w/ a `Change-type: patch`

That's in an attempt to avoid an changelog line that doesn't have much to offer to users.

thgreasi (approved):

PR #2929, opened 2025-04-02
Update balena-config-json to rely on the balena-image-fs helpers

thgreasi:
Change-type: patch


kb2ma (changes requested):
Changes look good so far. Able to test read, write, and inject.

Asked for changes because I expect you also want to fix [this instance](https://github.com/balena-io/balena-cli/blob/42d9cbb48dbe2902d01394fc5ff75059a048ae2f/src/commands/preload/index.ts#L195) in the preload command as well.

However, unable to test reconfigure. A command like below always asks to login, even if I'm already logged in. This really is a separate issue, but can you confirm this behavior?
```
sudo balena config reconfigure --drive /dev/sdd
```

thgreasi:
It probably asks for login b/c while sudo-ing, the terminal will be checking the home directory of the sudo user. Try using v
```
sudo BALENARC_DATA_DIRECTORY=/home/full/path/to/your/.balena/file /home/full/path/to/balena-cli/bin/balena
```

JSReds (approved):

Issue #2924, opened 2025-03-20
`balena build` with podman-desktop fails to pull source images


msherman64:
I was checking if it was possible to use the balena-cli with podman-desktop, and its docker integration.

However, when running `balena build` or `balena deploy` with any combination of fleets and dockerfiles, I get an `Image manifest data unavailable` error. It seems that Podman is stricter about the query parameters it expects, and is error-ing on:
`(HTTP code 400) unexpected - failed to parse query parameter 'labels': "": unexpected end of JSON input`

Is it possible that balena-cli is sending the query key `labels` with an empty value when pulling images?

The dockerfile to reproduce is:
```
FROM ubuntu:24.04

```

And the output of build --debug is below.

```
❯ balena build  -f chi-edge-dev2 --debug
[debug] new argv=[/opt/homebrew/Cellar/node@20/20.19.0_1/bin/node,/opt/homebrew/bin/balena,build,-f,chi-edge-dev2] length=5
[debug] Deprecation check: 0.03644 days since last npm registry query for next major version release date.
[debug] Will not query the registry again until at least 7 days have passed.
[debug] Event tracking error: Timeout awaiting 'response' for 0ms
[Debug]   Parsing input...
[Debug]   Loading project...
[Debug]   Resolving project...
[Info]    No "docker-compose.yml" file found at "/Users/shermanm/Developer/scratch/chi-edge-dev2"
[Info]    Creating default composition with source: "/Users/shermanm/Developer/scratch/chi-edge-dev2"
[Debug]   Creating project...
[Build]   Building services...
[Build]   main Preparing...
[Info]    Building for aarch64/raspberrypi4-64
[Debug]   Found build tasks:
[Debug]       main: build [.]
[Debug]   Resolving services with [raspberrypi4-64|aarch64]
[Debug]   Found project types:
[Debug]       main: Standard Dockerfile
[Debug]   Prepared tasks; building...
[Debug]   main: Image manifest data unavailable for ubuntu:24.04
[Build]   Built 1 service in 0:00
[Error]   Build failed.
(HTTP code 400) unexpected - failed to parse query parameter 'labels': "": unexpected end of JSON input

Error: (HTTP code 400) unexpected - failed to parse query parameter 'labels': "": unexpected end of JSON input
    at /opt/homebrew/Cellar/balena-cli/21.1.0/libexec/lib/node_modules/balena-cli/node_modules/docker-modem/lib/modem.js:383:17
[...]

PR #2930, opened 2025-04-03
Update dockerode/docker-modem dependencies for cachefrom fix

kb2ma:
The docker-modem v5.0.6 release broke the dockerode `buildImage` operation due to handling of the `cachefrom` parameter. So, we pinned docker-modem and dockerode in #2916 to known good versions.

The dockerode v4.0.5 release fixes the problem in the right place, so we are updating those dependencies and unpinning.


kb2ma:
Moved to draft status pending triage of #2924.

kb2ma:
Based on Felipe's assessment of #2924 and my manual testing here, I'll move this out of draft again.

pipex (approved):

PR #2947, opened 2025-05-28
Update `nock` to 14.0.4 & Move DeviceAPI off `request` 

myarmolinsky:
---
Please check the CONTRIBUTING.md file for relevant information and some
guidance. Keep in mind that the CLI is a cross-platform application that runs
on Windows, macOS and Linux. Tests will be automatically run by balena CI on
all three operating systems, but this will only help if you have added test
code that exercises the modified or added feature code.

Note that each commit message (currently only the first line) will be
automatically copied to the CHANGELOG.md file, so try writing it in a way
that describes the feature or fix for CLI users.

If there isn't a linked issue or if the linked issue doesn't quite match the
PR, please add a PR description to explain its purpose or the features that it
implements. Adding PR comments to blocks of code that aren't self explanatory
usually helps with the review process.

If the PR introduces security considerations or affects the development, build
or release process, please be sure to highlight this in the PR description.

Thank you very much for your contribution!

PR #2954, opened 2025-06-02
Update nock to 15.0.0 to work with native fetch

thgreasi:
Nock 15 adds support for native fetch.
Unblocks https://github.com/balena-io/balena-cli/pull/2953

Change-type: patch
See: https://github.com/nock/nock/blob/beta/migration_guides/migrating_to_15.md
See: https://balena.fibery.io/Work/Project/Reduce-the-polyfills-that-the-SDK-uses-1568

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5799,9 +5799,9 @@
       }
     },
     "node_modules/@balena/lint/node_modules/@eslint/js": {
-      "version": "9.32.0",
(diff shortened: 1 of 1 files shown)


otaviojacobi:
@thgreasi FYI nock 14 added support to native fetch api (and we migrated here https://github.com/balena-io/balena-cli/pull/2947/files), and nock 15 seems to be just improvign the api for it

PR #2958, opened 2025-06-18
Update balena-sdk to v22

otaviojacobi:
Bump balena-sdk from 21.3.0 to 22.0.0

Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/src/commands/config/generate.ts
+++ b/src/commands/config/generate.ts
@@ -24,7 +24,14 @@ import {
 	devModeInfo,
 	secureBootInfo,
 } from '../../utils/messages';
-import type { BalenaSDK, PineDeferred } from 'balena-sdk';
+import type { Application, BalenaSDK, Pine, PineDeferred } from 'balena-sdk';
+
+const getApplicationOptions = {
+	$select: 'slug',
+	$expand: {
+		is_for__device_type: { $select: 'slug' },
+	},
+} as const;
 
 export default class ConfigGenerateCmd extends Command {
 	public static description = stripIndent`
@@ -121,14 +128,20 @@ export default class ConfigGenerateCmd extends Command {
 
 	public static authenticated = true;
 
-	public async getApplication(balena: BalenaSDK, fleet: string) {
+	public async getApplication(
+		balena: BalenaSDK,
+		fleet: string,
+	): Promise<
+		NonNullable<
+			Pine.OptionsToResponse<
+				Application['Read'],
+				typeof getApplicationOptions,
+				string
+			>
+		>
+	> {
 		const { getApplication } = await import('../../utils/sdk');
-		return await getApplication(balena, fleet, {
-			$select: 'slug',
-			$expand: {
-				is_for__device_type: { $select: 'slug' },
-			},
-		});
+		return await getApplication(balena, fleet, getApplicationOptions);
 	}
 
 	public async run() {
--- a/src/commands/deploy/index.ts
+++ b/src/commands/deploy/index.ts
@@ -24,13 +24,13 @@ import {
 	registrySecretsHelp,
 	buildArgDeprecation,
 } from '../../utils/messages';
+import type { getAppWithArch } from '../../utils/helpers';
 import * as ca from '../../utils/common-args';
 import * as compose from '../../utils/compose';
 import type {
 	BuiltImage,
 	ComposeCliFlags,
 	ComposeOpts,
-	Release as ComposeReleaseInfo,
 } from '../../utils/compose-types';
 import type { BuildOpts, DockerCliFlags } from '../../utils/docker';
 import {
@@ -41,14 +41,7 @@ import {
 	parseReleaseTagKeysAndValues,
 } from '../../utils/compose_ts';
 import { dockerCliFlags } from '../../utils/docker';
-import type { ApplicationType, DeviceType, Release } from 'balena-sdk';
-
-interface ApplicationWithArch {
-	id: number;
-	arch: string;
-	is_for__device_type: [Pick<DeviceType, 'slug'>];
-	application_type: [Pick<ApplicationType, 'slug' | 'supports_multicontainer'>];
-}
+import type { Release } from 'balena-sdk';
 
 // TODO: For this special one we can't use Interfaces.InferredFlags/InferredArgs
 // because of the 'registry-secrets' type which is defined in the actual code
@@ -229,7 +222,7 @@ ${dockerignoreHelp}
 		logger: import('../../utils/logger'),
 		composeOpts: ComposeOpts,
 		opts: {
-			app: ApplicationWithArch; // the application instance to deploy to
+			app: Awaited<ReturnType<typeof getAppWithArch>>; // the application instance to deploy to
 			appName: string;
 			image?: string;
 			dockerfilePath?: string; // alternative Dockerfile
@@ -324,7 +317,7 @@ ${dockerignoreHelp}
 					},
 			);
 
-			let release: Release | ComposeReleaseInfo['release'];
+			let release: Pick<Release['Read'], 'commit' | 'id'>;
 			if (appType.slug === 'legacy-v1' || appType.slug === 'legacy-v2') {
 				const { deployLegacy } = require('../../utils/deploy-legacy');
 
@@ -355,7 +348,7 @@ ${dockerignoreHelp}
 				);
 
 				release = await sdk.models.release.get(releaseId, {
-					$select: ['commit'],
+					$select: ['commit', 'id'],
 				});
 			} else {
 				release = await $deployProject(
--- a/src/commands/device/index.ts
+++ b/src/commands/device/index.ts
@@ -76,7 +76,6 @@ export default class DeviceCmd extends Command {
 		const { args: params, flags: options } = await this.parse(DeviceCmd);
 
 		const balena = getBalenaSdk();
-
 		let device: ExtendedDevice;
 		if (options.json) {
 			const [deviceBase, deviceComputed] = await Promise.all([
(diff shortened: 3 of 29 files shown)


src/commands/config/generate.ts
```diff
@@ -138,8 +155,7 @@ export default class ConfigGenerateCmd extends Command {
 		await this.validateOptions(options);
 
 		let resourceDeviceType: string;
-		let application: Awaited<ReturnType<typeof this.getApplication>> | null =
```
thgreasi, line 141: How come and you decided to remove the explicit type?
Or was it balena-lint that removed it?
(I do realize that as soon as the assignment happens this has the correct types, just saying)

src/commands/config/generate.ts
```diff
@@ -121,14 +121,31 @@ export default class ConfigGenerateCmd extends Command {
...
-			$expand: {
-				is_for__device_type: { $select: 'slug' },
-			},
-		});
+		return await getApplication(
+			balena,
+			fleet,
+			ConfigGenerateCmd.getApplicationOptions,
+		);
 	}
```
thgreasi, line 149: NIT: So that we can be sure that the object is only allocated when needed & GCed after that.
```suggestion
	public async getApplication(
		balena: BalenaSDK,
		fleet: string,
	) {
		const { getApplication } = await import('../../utils/sdk');
		const options = {
			$select: 'slug',
			$expand: {
				is_for__device_type: { $select: 'slug' },
			},
		}
		const result: NonNullable<
			Pine.OptionsToResponse<
				Application['Read'],
				typeof options,
				string
			>
		> = await getApplication(
			balena,
			fleet,
			options,
		);
		return result;
}
```

src/commands/config/generate.ts
otaviojacobi: This was a mistake, thanks!

src/commands/device/index.ts
```diff
@@ -77,58 +77,67 @@ export default class DeviceCmd extends Command {
...
-				}),
-				balena.models.device.get(params.uuid, {
-					$select: [
-						'overall_status',
-						'overall_progress',
-						'should_be_running__release',
-					],
-				}),
+				balena.models.device.get(params.uuid, jsonExpand),
+				balena.models.device.get(params.uuid, jsonComputedFieldsSelect),
```
thgreasi, line 129: Was there a reason to not keep the param objects inline like they were before?
It feels a bit harder to follow now that we have to scroll up to check what's fetched.

src/commands/config/generate.ts
otaviojacobi: @thgreasi This does not work as for some particular cases (such as this one) we need the response type defined explicitly

src/utils/sdk.ts
```diff
@@ -136,9 +139,9 @@ export async function getOwnOrganizations(
 						},
 					},
 				},
-				$orderby: 'name asc',
+				$orderby: { name: 'asc' },
 			},
-			options,
+			options ?? {},
```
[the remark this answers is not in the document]
otaviojacobi: This was a mistake, we don't even need the cast anymore

src/commands/config/generate.ts
```diff
@@ -121,14 +121,31 @@ export default class ConfigGenerateCmd extends Command {
 
 	public static authenticated = true;
 
-	public async getApplication(balena: BalenaSDK, fleet: string) {
+	private static readonly getApplicationOptions = {
```
thgreasi, line 124: `static` 🚓 heads up

src/commands/device/index.ts
otaviojacobi: I was mostly following the (new) pattern, but agree here and reverted

src/commands/deploy/index.ts
```diff
@@ -355,7 +347,7 @@ ${dockerignoreHelp}
 				);
 
 				release = await sdk.models.release.get(releaseId, {
-					$select: ['commit'],
+					$select: ['commit', 'id'],
```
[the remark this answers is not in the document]
otaviojacobi: yes

src/commands/config/generate.ts
otaviojacobi: @thgreasi  It can't be inlined as it is used in two places - the explicit function type return (otherwise the generated typings need a inner dep to pinejs-client-core -> abstract-sql-to-typescript to be able to be generated)

I don't mind moving it to top level, that should be fine

src/utils/promote.ts
```diff
@@ -299,29 +299,38 @@ async function getOrSelectApplication(
 		$expand: {
 			is_for__device_type: { $select: ['slug', 'is_of__cpu_architecture'] },
 		},
-	} satisfies BalenaSdk.PineOptions<BalenaSdk.Application>;
+	} satisfies BalenaSdk.Pine.ODataOptionsWithoutCount<
+		BalenaSdk.Application['Read']
+	>;
```
myarmolinsky, line 304: Why `satisifies` instead of `options:  BalenaSdk.Pine.ODataOptionsWithoutCount<BalenaSdk.Application['Read']>`?

src/utils/sdk.ts
```diff
@@ -108,19 +108,22 @@ export async function getFleetSlug(
...
 /**
  * Wraps the sdk organization.getAll method,
  * restricting to those orgs user is a member of
  */
 export async function getOwnOrganizations(
 	sdk: BalenaSDK,
-	options?: PineOptions<Organization>,
-): Promise<Organization[]> {
+	options?: Pine.ODataOptionsWithoutCount<Organization['Read']>,
+): Promise<Array<Organization['Read']>> {
```
myarmolinsky, line 126: I don't believe we need this return type

src/utils/promote.ts
myarmolinsky: > They are equivalent so I don't have any strong preference

The difference would be that we don't have to cast everywhere

src/utils/patterns.ts
```diff
@@ -210,24 +202,25 @@ const selectApplicationPineOptions = {
...
+} as const;
 
-type SelectApplicationResult = PineTypedResult<
-	Application,
-	typeof selectApplicationPineOptions
+type SelectApplicationResult = Pine.OptionsToResponse<
+	Application['Read'],
+	typeof selectApplicationPineOptions,
+	undefined
 >;
```
thgreasi, line 211: NIT: Since `selectApplication` returns 1 application, could we make this type be the object instead of an array?
```suggestion
type SelectApplicationResult = Pine.OptionsToResponse<
	Application['Read'],
	typeof selectApplicationPineOptions,
	undefined
>[number];
```

src/utils/promote.ts
otaviojacobi: in fact, the best way here is to use `as const` simply so that the strong inference can infer the follow up types :+1:

src/utils/patterns.ts
```diff
@@ -210,24 +202,25 @@ const selectApplicationPineOptions = {
...
+	filter?: T | ((app: SelectApplicationResult[number]) => boolean),
 	errorOnEmptySelection = false,
-) {
+): Promise<SelectApplicationResult[number]> {
 	const balena = getBalenaSdk();
 	let apps = (await balena.models.application.getAllDirectlyAccessible({
 		...selectApplicationPineOptions,
 		...(filter != null && typeof filter === 'object' && { $filter: filter }),
-	})) as SelectApplicationResult[];
+	})) satisfies SelectApplicationResult;
```
thgreasi, line 223: As far as I tried it locally, this isn't necessary at all any more
```suggestion
	}));
```

src/utils/promote.ts
```diff
@@ -299,29 +299,36 @@ async function getOrSelectApplication(
...
+	} as const;
 
 	// Check for a fleet slug of the form `user/fleet` and update the API query.
 	let name: string;
 	const match = appName.split('/');
 	if (match.length > 1) {
 		// These will match at most one fleet
-		(options as BalenaSdk.PineOptions<BalenaSdk.Application>).$filter = {
+		(
+			options as BalenaSdk.Pine.ODataOptionsWithoutCount<
```
thgreasi, line 310: If you are up to if, you can get rid of these casts completely if you define the opts as:
```
	const options = {
		$expand: {
			is_for__device_type: { $select: ['slug', 'is_of__cpu_architecture'] },
		},
		$filter: {}
	} satisfies BalenaSdk.Pine.ODataOptionsWithoutCount<
		BalenaSdk.Application['Read']
	>;
```

src/utils/promote.ts
myarmolinsky: I don't see the follow up types getting inferred, the call to `await sdk.models.application.getAllDirectlyAccessible(options);` below still expects `options` to only be the type from when we initialized it

otaviojacobi: Yes @myarmolinsky and for that call to have strong typed response (which we want, as this variable is used after on the filter) we need the typing to be a const readonly type (as const) rather than simply satisfying ODataOptionsWithoutCount. The `options as...` in the middle is just because we are adding $filters on the fly and that won't work with const (but that is fine as $filter does not currently affect the response type)

src/utils/patterns.ts
otaviojacobi: Yep! Thanks!

src/utils/sdk.ts
```diff
@@ -40,8 +40,8 @@ export async function getApplication<TP extends PineOptions<Application>>(
 export async function getApplication(
 	sdk: BalenaSDK,
 	nameOrSlugOrId: string | number,
-	options?: PineOptions<Application>,
-): Promise<Application> {
+	options?: Pine.ODataOptionsWithoutCount<Application['Read']>,
+): Promise<Application['Read']> {
```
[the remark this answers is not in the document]
otaviojacobi: We can remove the overload and the explicit return type, but we still need one overload in order to keep the inner options typing working (as it is has a modified $filter)

thgreasi (approved):

PR #2973, opened 2025-08-21
Add release_asset model

otaviojacobi:
Change-type: minor

Resolves: #   
Depends-on: <url>   
See: <url>   

---


src/commands/release-asset/upload.ts
```diff
@@ -0,0 +1,128 @@
...
+import { commitOrIdArg } from '../release';
+import * as fs from 'fs/promises';
+import * as path from 'path';
+
+export default class ReleaseAssetUploadCmd extends Command {
+	public static description = stripIndent`
+		Upload a release asset.
+
+		Upload a file as a release asset with the specified key. If the asset already exists,
+		use the --overwrite flag to replace it. You can customize the upload behavior with
```
myarmolinsky, line 29: nit: 
```suggestion
		you can use the --overwrite flag to replace it. You can customize the upload behavior with
```

Currently it sounds like an instruction, adding `you can` makes it sound like informing of an option

src/commands/release-asset/upload.ts
```diff
@@ -0,0 +1,128 @@
...
+
+	public static examples = [
+		'$ balena release-asset upload 1234567 ./path/to/config.json --key config.json',
+		'$ balena release-asset upload a777f7345fe3d655c1c981aa642e5555 ./app.tar.gz --key app.tar.gz --overwrite',
+		'$ balena release-asset upload 1234567 ./file.bin --key large-file.bin --chunk-size 10485760 --parallel-chunks 10',
+	];
+
+	public static args = {
+		commitOrId: commitOrIdArg({
+			description: 'the commit or ID of the release',
```
myarmolinsky, line 41: I keep forgetting, do we want to still allow IDs? If not then I assume the ones we currently allow are just remnants, and of course then we don't have to support that here cc @thgreasi

src/commands/release-asset/upload.ts
```diff
@@ -0,0 +1,128 @@
...
+		const release = await balena.models.release.get(args.commitOrId, {
+			$select: ['id'],
+		});
+
+		const absolutePath = path.resolve(args.filePath);
+
+		const visuals = getVisuals();
+		const bar = new visuals.Progress(`Uploading release asset '${flags.key}'`);
+		const spinner = new visuals.Spinner(
+			`Uploading release asset '${flags.key}' (size unknown)`,
```
myarmolinsky, line 91: Is it meant to be always `size unknown`? If so, what's the benefit of communicating this?

src/commands/release-asset/list.ts
```diff
@@ -0,0 +1,106 @@
...
+
+import { Command } from '@oclif/core';
+import { getBalenaSdk, getVisuals, stripIndent } from '../../utils/lazy';
+import { commitOrIdArg } from '../release';
+import * as cf from '../../utils/common-flags';
+import { jsonInfo } from '../../utils/messages';
+
+export default class ReleaseAssetListCmd extends Command {
+	public static description = stripIndent`
+		List all release assets.
```
myarmolinsky, line 26: nit: 
```suggestion
		List all release assets for a specific release.
```

src/commands/release-asset/list.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	public async run() {
+		const { args, flags } = await this.parse(ReleaseAssetListCmd);
+		const balena = getBalenaSdk();
+
+		const releaseAssets = await balena.models.release.asset.getAllByRelease(
+			args.commitOrId,
+			{
+				$select: ['id', 'asset_key', 'asset'],
+				$top: 20,
```
myarmolinsky, line 59: Why limit to top 20? I would expect to see all release assets

src/commands/release-asset/download.ts
```diff
@@ -0,0 +1,145 @@
...
+
+	public static flags = {
+		key: Flags.string({
+			description: 'the key of the release asset to download',
+			required: true,
+		}),
+		output: Flags.string({
+			char: 'o',
+			description: 'output path for the downloaded file',
+		}),
```
myarmolinsky, line 55: This combines changing download directory and renaming the file. Perhaps we should keep them separate? If I just wanted to change the file name, I would be a bit annoyed that I need to go get the full path I want to send the file to just to download with a new name

src/commands/release-asset/download.ts
```diff
@@ -0,0 +1,145 @@
...
+				release: release.id,
+				asset_key: flags.key,
+			},
+			{
+				$select: ['id', 'asset'],
+			},
+		);
+
+		if (releaseAsset.asset == null) {
+			throw new Error(`Release asset ${releaseAsset.id} is empty`);
```
myarmolinsky, line 79: Or it doesn't exist, right?

src/commands/release-asset/download.ts
myarmolinsky: But looking at line 89 I see it could indeed be used to only rename. Unclear to me at first, but I'm not sure if that's just me

src/commands/release-asset/download.ts
```diff
@@ -0,0 +1,145 @@
...
+
+		const dir = path.dirname(outputPath);
+		await fs.promises.mkdir(dir, { recursive: true });
+
+		const visuals = getVisuals();
+		const bar = new visuals.Progress(
+			`Downloading release asset '${flags.key}'`,
+		);
+		const spinner = new visuals.Spinner(
+			`Downloading release asset '${flags.key}' (size unknown)`,
```
myarmolinsky, line 103: Again I question the intent and usefulness of a constant `size unknown`

myarmolinsky: Especially as it appears we have the size below. But I also don't expect people even want to see the total size they are downloading, but a fraction indicating how much of the total size they have downloaded now (i.e. 250MB/3GB)

myarmolinsky: > But I also don't expect people even want to see the total size they are downloading, but a fraction indicating how much of the total size they have downloaded now (i.e. 250MB/3GB)

That is of course what the progress bar is for, so again I wonder if there's any point even showing a size here

tests/commands/release-asset/download.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+	afterEach(async () => {
+		await fs.promises.rm(tempDir, { recursive: true, force: true });
+		api.done();
+	});
+
+	itSS('should download a release asset successfully', async () => {
+		const testContent = 'test file content for download';
+
+		api.expectGetRelease();
+		api.expectGetReleaseAsset({ found: true, assetId: 123 });
```
myarmolinsky, line 49: nit since `123` is the default
```suggestion
		api.expectGetReleaseAsset({ found: true });
```

src/commands/release-asset/delete.ts
```diff
@@ -0,0 +1,77 @@
...
+		commitOrId: commitOrIdArg({
+			description: 'the commit or ID of the release',
+			required: true,
+		}),
+	};
+
+	public static flags = {
+		key: Flags.string({
+			description: 'the key of the release asset to delete',
+			required: true,
```
myarmolinsky, line 45: Same question as in a previous comment about how we determine whether a required value should be an arg or flag

src/commands/release-asset/upload.ts
```diff
@@ -0,0 +1,128 @@
...
+		filePath: Args.string({
+			description: 'path to the file to upload',
+			required: true,
+		}),
+	};
+
+	public static flags = {
+		key: Flags.string({
+			description: 'the key for the release asset',
+			required: true,
```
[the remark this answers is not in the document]
otaviojacobi: I don't think we have a fixed rule. I went with the release as an arg as it is common to all commands and the key as a flag as it is not common to all (e.g. list does not have it)  and to avoid having two args

src/commands/release-asset/upload.ts
otaviojacobi: This is basically the pattern of while the progress bar hasn't started  we should this spinner (which is what we are doing for os download) but I agree, it can be improved as for the case of release assets we know the size of the release asset before hand I think it is not necessary. I am removing the spinners all together and switching for bar only :+1:

tests/commands/release-asset/upload.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+		}
+	});
+
+	itSS('should upload a small release asset successfully', async () => {
+		api.expectGetRelease();
+		api.expectGetReleaseAsset({ found: false });
+		api.expectPostReleaseAsset({ assetKey: 'small-file.txt', assetId: 123 });
+
+		const { out, err } = await runCommand(
+			`release-asset upload 27fda508c ${smallFilePath} --key small-file.txt`,
```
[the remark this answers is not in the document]
otaviojacobi: It is the default commit used and mocked all over the CLI mocks

tests/commands/release-asset/upload.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+		);
+
+		const lines = cleanOutput(out);
+		expect(lines.join(' ')).to.contain('uploaded successfully');
+		expect(err).to.be.empty;
+	});
+
+	it.skip('should upload a large release asset using multipart upload', async () => {
```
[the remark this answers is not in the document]
otaviojacobi: Yes, they are broken and I am investigating why (and thus why the PR is still on draft but I already asked for review as I knew at least a few things I would have to change)

src/commands/release-asset/download.ts
otaviojacobi: As I mentioned before, I am removing the spinner altogether :+1:

src/commands/release-asset/delete.ts
otaviojacobi: Again I don't think the CLI has a strong guidance over it. Both (key and release) are mandatory. I think having two args is not ideal (as they are positional and can't be passed using preferred key/value). As release is mandatory in all 4 commands and key in only 3, I preferred the first as an argument (so that all commands feel similar like `balena release-asset list 123` `balena release-asset upload 123 ...`  ... and the key only on the commands that require it (e.g. list does not)

src/commands/release-asset/download.ts
```diff
@@ -0,0 +1,126 @@
...
+		);
+
+		const totalSize = releaseAsset.asset.size;
+		let downloadedBytes = 0;
+
+		const progressStream = new Transform({
+			transform(chunk, _encoding, callback) {
+				downloadedBytes += chunk.length;
+
+				if (totalSize && totalSize > 0) {
```
myarmolinsky, line 110:
```suggestion
				if (totalSize) {
```

src/commands/release-asset/download.ts
```diff
@@ -0,0 +1,126 @@
...
+			throw new Error(`Release asset ${releaseAsset.id} is empty`);
+		}
+
+		const stream = await balena.models.release.asset.download({
+			release: release.id,
+			asset_key: flags.key,
+		});
+
+		let outputPath: string;
+		if (flags.output) {
```
thgreasi, line 88: How about making the output path necessary or at least double checking first that there is no file with such a name already there?

tests/nock/balena-api-mock.ts
```diff
@@ -209,6 +209,273 @@ export class BalenaAPIMock extends NockMock {
...
+								asset: {
+									href: `https://balena-asset-downloads.s3.amazonaws.com/release-asset-${assetId}.bin`,
+									size: 1024,
+									content_type: 'application/octet-stream',
+								},
+							},
+						],
+					},
+				];
+			}
```
thgreasi, line 322: NIT:
```suggestion
			}

			return [
				200,
				{
					d: [
						{
							id: assetId,
							asset: {
								href: `https://balena-asset-downloads.s3.amazonaws.com/release-asset-${assetId}.bin`,
								size: 1024,
								content_type: 'application/octet-stream',
							},
						},
					],
				},
			];
```

src/commands/release-asset/download.ts
otaviojacobi: Double checking is a good idea, thanks!

src/commands/release-asset/list.ts
```diff
@@ -0,0 +1,105 @@
...
+					if (size < 1024) {
+						sizeStr = `${size} B`;
+					} else if (size < 1024 * 1024) {
+						sizeStr = `${(size / 1024).toFixed(2)} KiB`;
+					} else if (size < 1024 * 1024 * 1024) {
+						sizeStr = `${(size / 1024 / 1024).toFixed(2)} MiB`;
+					} else {
+						sizeStr = `${(size / 1024 / 1024 / 1024).toFixed(2)} GiB`;
+					}
+				}
```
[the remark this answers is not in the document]
thgreasi: I think that we need to re-raise that w/ them since this would be the only place that would be using these different units.
The UI, Builder & CLI all use MB all over the place, just like how google, docker and filesystems do nowadays. Even release assets in the UI atm use MB.
Since AWS is an implementation details, I don't think it justifies us using different units just for release assets.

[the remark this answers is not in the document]
otaviojacobi: The 1024 discussion was for the limits we impose (10MiB) - I can move them to MB in here (but I don't think we want to use the library)

src/commands/release-asset/download.ts
```diff
@@ -0,0 +1,126 @@
...
+				downloadedBytes += chunk.length;
+
+				if (totalSize != null && totalSize > 0) {
+					const percentage = (downloadedBytes / totalSize) * 100;
+					bar.update({ percentage, eta: null });
+				}
+
+				callback(null, chunk);
+			},
+		});
```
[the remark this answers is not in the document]
otaviojacobi: IMHO the current version reads better...

src/commands/release-asset/download.ts
otaviojacobi: I added a --overwrite flag, similar to the one on the upload with a prompt.

thgreasi (approved):

PR #2997, opened 2025-09-26
Update @balena/compose, tar-utils & tar-stream

thgreasi:
Update @balena/compose from 7.0.10 to 7.2.1
Update tar-utils from 3.0.3 to 3.1.1

Change-type: patch
See: https://balena.fibery.io/Work/Project/1926

### DIFF
--- a/src/utils/compose_ts.ts
+++ b/src/utils/compose_ts.ts
@@ -37,6 +37,7 @@ import type {
 	BuiltImage,
 	ComposeOpts,
 	ComposeProject,
+	Release,
 	TaggedImage,
 	TarDirectoryOptions,
 } from './compose-types';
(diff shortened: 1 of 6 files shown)


src/utils/compose_ts.ts
```diff
@@ -1552,9 +1552,7 @@ export function createRunLoop(tick: (...args: any[]) => void) {
 
 async function getContractContent(
 	filePath: string,
-): Promise<
-	import('@balena/compose/dist/release/models').ReleaseModel['contract']
-> {
+): Promise<import('@balena/compose').release.Request['contract']> {
```
cywang117: Just a TODO for another PR which I noted in my [PR](https://github.com/balena-io/balena-cli/pull/3041/changes#diff-f5de550fbc6503a54a23055935c717c21a95341de8bb43476a780228d9f9e5beR105) as well, but it looks like the contract typings in the SDK are incorrect (`{ [key: string]: JSONType } | JSONType[] | null`), and SDK really should be the source of typings

PR #3003, opened 2025-10-17
Fix and enable `help` tests

otaviojacobi:
Change-type: patch

Resolves: #   
Depends-on: <url>   
See: <url>   

---

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -148,7 +148,17 @@ async function oclifRun(command: string[], options: AppOptions) {
 
 	const { trackPromise } = await import('./hooks/prerun');
 
-	await Promise.all([trackPromise, deprecationPromise, runPromise]);
+	// Help command and others may not trigger the prerun hook, so we need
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -148,7 +148,17 @@ async function oclifRun(command: string[], options: AppOptions) {
 
 	const { trackPromise } = await import('./hooks/prerun');
 
-	await Promise.all([trackPromise, deprecationPromise, runPromise]);
+	// Help command and others may not trigger the prerun hook, so we need
+	// to handle the case where trackPromise never resolves. We use Promise.race
```
thgreasi: I would argue that `trackResolve` should always settle to something, and how we have it atm isn't really great. It seems that we really care about is whether we need to still wait for tracking events to be sent, and not whether an event was really sent or not w/ the current command. This way it seems that `trackResolve` should be in a `finally` block in the hooks, so that it's always called even if the currently issued command didn't call the tracking code at all.

PR #3007, opened 2025-11-11
Fix linter issues with balena lint v9

cywang117:
Changes:
- Use nullish coalescing: `@balena/lint` v9 has started to warn about use of `||` over `??`
- `getExpandedProp` in `utils/pine.ts` has the wrong TypeScript-inferred return type
- Re-enable JS linting - this was disabled with move to `@balena/lint` v9, re-enable it and fix linting errors using JSDoc annotations to add type information

Change-type: patch

### DIFF
--- a/src/commands/device/restart.ts
+++ b/src/commands/device/restart.ts
@@ -17,7 +17,7 @@
 
 import { Flags, Args, Command } from '@oclif/core';
 import { getBalenaSdk, getCliUx, stripIndent } from '../../utils/lazy';
-import type { BalenaSDK } from 'balena-sdk';
+import type { BalenaSDK, CurrentService } from 'balena-sdk';
 
 export default class DeviceRestartCmd extends Command {
 	public static description = stripIndent`
@@ -122,7 +122,7 @@ export default class DeviceRestartCmd extends Command {
 			// Each service is an array of `CurrentServiceWithCommit`
 			// because when service is updating, it will actually hold 2 services
 			// Target commit matching `device.is_running__release`
-			const serviceContainer = service.find((s) => {
+			const serviceContainer = service.find((s: CurrentService) => {
 				return s.commit === activeRelease;
 			});
 
--- a/src/commands/device/start-service.ts
+++ b/src/commands/device/start-service.ts
@@ -17,7 +17,7 @@
 
 import { Args, Command } from '@oclif/core';
 import { getBalenaSdk, getCliUx, stripIndent } from '../../utils/lazy';
-import type { BalenaSDK } from 'balena-sdk';
+import type { BalenaSDK, CurrentService } from 'balena-sdk';
 
 export default class DeviceStartServiceCmd extends Command {
 	public static description = stripIndent`
@@ -104,7 +104,7 @@ export default class DeviceStartServiceCmd extends Command {
 			// Each service is an array of `CurrentServiceWithCommit`
 			// because when service is updating, it will actually hold 2 services
 			// Target commit matching `device.is_running__release`
-			const serviceContainer = service.find((s) => {
+			const serviceContainer = service.find((s: CurrentService) => {
 				return s.commit === activeReleaseCommit;
 			});
 
--- a/src/commands/device/stop-service.ts
+++ b/src/commands/device/stop-service.ts
@@ -17,7 +17,7 @@
 
 import { Args, Command } from '@oclif/core';
 import { getBalenaSdk, getCliUx, stripIndent } from '../../utils/lazy';
-import type { BalenaSDK } from 'balena-sdk';
+import type { BalenaSDK, CurrentService } from 'balena-sdk';
 
 export default class DeviceStopServiceCmd extends Command {
 	public static description = stripIndent`
@@ -104,7 +104,7 @@ export default class DeviceStopServiceCmd extends Command {
 			// Each service is an array of `CurrentServiceWithCommit`
 			// because when service is updating, it will actually hold 2 services
 			// Target commit matching `device.is_running__release`
-			const serviceContainer = service.find((s) => {
+			const serviceContainer = service.find((s: CurrentService) => {
 				return s.commit === activeReleaseCommit;
 			});
 


src/utils/compose_ts.ts
```diff
@@ -491,6 +491,7 @@ async function qemuTransposeBuildStream({
 	task.buildStream = (await transpose.transposeTarStream(
 		task.buildStream,
 		transposeOptions,
+		// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
 		dockerfilePath || undefined,
```
joshbwlng:
```suggestion
		dockerfilePath,
```

I don't think we need a fallback here at all since `dockerfilePath` is already typed as `string | undefined`

automation/build-bin.ts
```diff
@@ -253,7 +253,7 @@ async function signWindowsInstaller() {
 			'-sha1',
 			process.env.SM_CODE_SIGNING_CERT_SHA1_HASH,
 			'-tr',
-			process.env.TIMESTAMP_SERVER || 'http://timestamp.comodoca.com',
+			process.env.TIMESTAMP_SERVER ?? 'http://timestamp.comodoca.com',
```
myarmolinsky: I think `process.env.X` can be `''` as well in which case we want to use `||` for all of them, not `??`

src/commands/device/rename.ts
```diff
@@ -49,11 +49,11 @@ export default class DeviceRenameCmd extends Command {
 		const balena = getBalenaSdk();
 
 		const newName =
-			params.newName ||
+			params.newName ??
```
myarmolinsky: As with `options.X`, the question is: can this be `''` and if it can, how do we want to treat that? In this case, I believe a device name _can_ actually be `''`. If we want to block that, we should do so at the API level as well

src/hooks/command-not-found/suggest.ts
```diff
@@ -46,14 +46,14 @@ const hook: Hook<'command-not-found'> = async function (
 	}
 
 	const suggestions: string[] = [];
-	suggestions.push(closest(commandId).replace(':', ' ') || '');
+	suggestions.push(closest(commandId).replace(':', ' ') ?? '');
```
myarmolinsky: The need for `??` and `||` conflicts with what I see in the `closest` function above 🤔

src/utils/compose_ts.ts
```diff
@@ -696,7 +697,7 @@ export async function getServiceDirsFromComposition(
 			let dir =
 				(typeof service.build === 'string'
 					? service.build
-					: service.build?.context) || '.';
+					: service.build?.context) ?? '.';
```
myarmolinsky: Please confirm whether `''` is possible and whether that's okay. Also (not part of your change), odd that `service.build` is type `never`. The typing does not indicate that it can ever be a string 🤔

src/utils/compose_ts.ts
```diff
@@ -708,7 +709,7 @@ export async function getServiceDirsFromComposition(
 			// remove './' prefix (or '.\\' on Windows)
 			dir = dir.startsWith(relPrefix) ? dir.slice(2) : dir;
 
-			serviceDirs[serviceName] = dir || '.';
+			serviceDirs[serviceName] = dir ?? '.';
```
myarmolinsky: The typing does not indicate that this can be a non-string, so I would assume `dir` can be `''` and the goal was to replace that with `'.'`

src/utils/docker.ts
```diff
@@ -215,7 +215,7 @@ export function getDefaultDockerModemOpts(
 	const originalDockerHost = process.env.DOCKER_HOST;
 	try {
 		if (opts.dockerHost) {
-			process.env.DOCKER_HOST ||= opts.dockerPort
+			process.env.DOCKER_HOST ??= opts.dockerPort
```
myarmolinsky: Not certain this can't be `''`

src/utils/promote.ts
```diff
@@ -73,7 +73,7 @@ export async function leave(
 	deviceHostnameOrIp?: string,
 ): Promise<void> {
 	logger.logDebug('Determining device...');
-	deviceHostnameOrIp = deviceHostnameOrIp || (await selectLocalDevice());
+	deviceHostnameOrIp = deviceHostnameOrIp ?? (await selectLocalDevice());
```
myarmolinsky: As above

automation/update-module.ts
```diff
@@ -93,7 +93,7 @@ async function $main() {
 		throw new Error(getUsage(upstreams, upstreamName));
 	}
 
-	const packageName = upstream.module || upstream.repo;
+	const packageName = upstream.module ?? upstream.repo;
```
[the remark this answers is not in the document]
cywang117: I guess in the CLI specifically, upstream is [loaded](https://github.com/balena-io/balena-cli/blob/3ae0d79983621a085060f44026625b6e6ed93240/automation/update-module.ts#L55) from local `repo.yml`: https://github.com/balena-io/balena-cli/blob/6e05a07a53635084cf8a51559a93a7e7e5659c2b/repo.yml#L7, wherein there are no empty string declarations for `upstream.module`. I think an empty `upstream.module` is also unlikely to pass code review, so using `??` should be safe, but I added a comment about `upstream.module` being an empty string within this commit

```
    // upstream.module could be an empty string, but requires it be defined as
	// such within repo.yml, which won't happen with proper code review.
```

src/commands/build/index.ts
```diff
@@ -164,7 +164,7 @@ ${dockerignoreHelp}
 			{
 				dockerfilePath: opts.dockerfile,
 				noParentCheck: opts['noparent-check'] || false,
-				projectPath: opts.source || '.',
+				projectPath: opts.source ?? '.',
```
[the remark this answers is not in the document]
cywang117: Copilot analysis of @oclif/core says:

Short answer

If an optional arg/flag is not provided, the parser does not set it to the empty string — it is left undefined / omitted from the returned result.
- For flags: undefined flags are omitted from the returned flags object (see tests that assert undefined flags are omitted).
- For args: optional positional args that aren’t given are effectively absent (types show them as possibly undefined).

When you will see an empty string

Only when the user explicitly provided an empty value. Examples that produce an empty string value:
  - --opt= (equals with nothing after it) — the parser will treat the value as the empty string.
  - --opt "" (shell passes an empty string token) — the positional/option value is the empty string.
  - Passing an empty quoted positional argument (program "") results in the positional arg being the empty string.
  - If a flag/arg default is explicitly set to '' (or a default function returns ''), the final value will be ''.

So the only way is if the user explicitly passes in an empty string. I reverted back to `||` for safety

src/utils/compose_ts.ts
cywang117: A user could explicitly set context to '' which then defaults to cwd in native Docker compose, so it appears an empty string is valid.

Regarding the typing being `never`, that's actually because balena-compose types BuildConfig wrongly. I'm fixing that with my next PR to integrate a newer balena-compose into CLI

src/utils/patterns.ts
```diff
@@ -356,7 +356,7 @@ export async function getOnlineTargetDeviceUuid(
 		type: 'list',
 		default: devices[0].uuid,
 		choices: devices.map((device) => ({
-			name: `${device.device_name || 'Untitled'} (${device.uuid.slice(0, 7)})`,
+			name: `${device.device_name ?? 'Untitled'} (${device.uuid.slice(0, 7)})`,
```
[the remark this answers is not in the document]
cywang117: If keep it means to revert to `||`, I've done so

bin/dev.js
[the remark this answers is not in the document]
cywang117: Done [here](https://github.com/balena-io/balena-cli/pull/3007/files#diff-e9d5048dfd542830b17402ee1476f31f41d6feccb2a7eedb592b6cc967315da2R33)

src/auth/utils.ts
```diff
@@ -50,6 +50,7 @@ export async function getDashboardLoginURL(
  * @returns whether the login was successful or not
  */
 export async function loginIfTokenValid(token?: string): Promise<boolean> {
+	// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
thgreasi: I think that we don't need this and can use `??`, since `if (!token)` right after will preserve the original behavior for any falsy value afaict.
```suggestion
```

bin/dev.js
```diff
@@ -30,7 +30,7 @@ const rootDir = path.join(__dirname, '..');
 
 // Allow balena-dev to work with oclif by temporarily
 // pointing oclif config options to src/ instead of build/
-modifyOclifPaths();
+modifyOclifPaths(false);
```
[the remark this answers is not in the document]
cywang117: Josh commented to pass in the param for code clarity, although I agree it's not needed

src/commands/api-key/generate.ts
```diff
@@ -75,26 +75,24 @@ export default class GenerateCmd extends Command {
...
-												start: 0,
-												end: duration * 24 * 60 * 60 * 1000,
-											}),
-										)
-									: 'Custom expiration',
-						value: duration,
-					})),
-				});
-			}
+			expiryDateResponse ??= await getCliForm().ask({
```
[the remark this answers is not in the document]
cywang117: Doing so brings up the interactive menu for selecting expiration date from a list, so it looks like passing in an empty string still results in the `??` getting triggered, which I didn't expect and doesn't seem correct.

I'll revert this to `||=` for safety though and we can revisit

src/commands/device/note.ts
```diff
@@ -61,7 +61,8 @@ export default class DeviceNoteCmd extends Command {
 			throw new ExpectedError('Missing note content');
 		}
 
-		options.device = options.device || options.dev;
+		// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+		options.device ||= options.dev;
```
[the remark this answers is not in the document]
cywang117: Shall I remove as a separate commit in this PR or would you prefer to do with with v23?

src/commands/device/rename.ts
```diff
@@ -49,10 +49,12 @@ export default class DeviceRenameCmd extends Command {
 		const balena = getBalenaSdk();
 
 		const newName =
+			// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
 			params.newName ||
 			(await getCliForm().ask({
 				message: 'How do you want to name this device?',
 				type: 'input',
+				// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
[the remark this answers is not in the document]
cywang117: > Not sure in which case the lefthand would end up nullish though

I think that case isn't possible, nor is any falsey output from getCliForm aside from empty string, but in the case of empty string, it doesn't matter whether the second operator is `??`/`||` as either way will end up with an empty string device name. I think we can keep `||` for both for now, and revisit later in a dedicated PR to remove support for empty strings as an edge case.

src/hooks/command-not-found/suggest.ts
cywang117: I see what you mean by your original comment now. In that case I don't think either `||` or `??` are necessary. `closest` asserts that it cannot return undefined with `!` which is accurate, as the only case in which `closest` can return undefined is if the commandIDs array is empty. This array can't be empty though, as `version` is always pushed to it. I removed the `||` after both instances of `closest` in the following commit

src/commands/support/index.ts
```diff
@@ -91,11 +91,12 @@ export default class SupportCmd extends Command {
...
 		// Calculate expiry ts
 		const durationDefault = '24h';
+		// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
 		const duration = options.duration || durationDefault;
 		const expiryTs = Date.now() + this.parseDuration(duration);
 
-		const deviceUuids = options.device?.split(',') || [];
-		const appNames = options.fleet?.split(',') || [];
+		const deviceUuids = options.device?.split(',') ?? [];
+		const appNames = options.fleet?.split(',') ?? [];
```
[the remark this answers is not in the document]
cywang117: Done [here](https://github.com/balena-io/balena-cli/pull/3007/commits/cc0853a51db2e80da5d1fa065af49207359a1646#diff-0d655fcae8d267ed7e2046da51b135e7c6e43e4f8c7b26d6d3d2c82710cf0a99L97-R101)

src/errors.ts
```diff
@@ -73,7 +73,7 @@ export function instanceOf(err: any, klass: any): boolean {
 	if (err instanceof klass) {
 		return true;
 	}
-	const name: string | undefined = err.name || err.constructor?.name;
+	const name: string | undefined = err.name ?? err.constructor?.name;
```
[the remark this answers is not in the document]
cywang117: Done [here](https://github.com/balena-io/balena-cli/pull/3007/commits/cc0853a51db2e80da5d1fa065af49207359a1646#diff-55db69e02ff5714d444d8081ec6ecac5d9833fb29fda64d1e829e5766434fdc0L76-R77)

src/hooks/command-not-found/suggest.ts
```diff
@@ -46,6 +46,7 @@ const hook: Hook<'command-not-found'> = async function (
 	}
 
 	const suggestions: string[] = [];
+
```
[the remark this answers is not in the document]
cywang117: This was, but `npm run lint` has the annoying habit of sometimes removing the eslint comment even though `removeComments: false` is configured in tsconfig.dev.json. In any case, it doesn't matter for this instance now, as I removed the `||` operator after both calls to `closest`

src/auth/utils.ts
cywang117: Reverted to `??`

PR #3008, opened 2025-11-12
Drop some usage of lodash in favor of native equivalents

myarmolinsky:
Fibery: https://balena.fibery.io/Work/Project/We-should-migrate-from-lodash-to-es-toolkit-(UI-CLI-etc.)-1895
Change-type: patch

---

### DIFF
--- a/src/commands/deploy/index.ts
+++ b/src/commands/deploy/index.ts
@@ -280,13 +280,12 @@ ${dockerignoreHelp}
 			// multibuild takes in a composition and always attempts to
 			// build or pull all services. we workaround that here by
 			// passing a modified composition.
-			const compositionToBuild = _.cloneDeep(project.composition);
-			compositionToBuild.services = _.omit(
-				compositionToBuild.services,
-				servicesToSkip,
-			);
+			const compositionToBuild = structuredClone(project.composition);
+			for (const serviceToSkip of servicesToSkip) {
+				delete compositionToBuild.services[serviceToSkip];
+			}
 			let builtImagesByService: Dictionary<BuiltImage> = {};
-			if (_.size(compositionToBuild.services) === 0) {
+			if (Object.keys(compositionToBuild.services).length === 0) {
 				logger.logInfo(
 					'Everything is up to date (use --build to force a rebuild)',
 				);
--- a/src/utils/cloud.ts
+++ b/src/utils/cloud.ts
@@ -94,11 +94,7 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
 		| [DeviceWithOptions, undefined]
 	> => {
 		const device = await sdk.models.device.get(deviceUUID, deviceOptions);
-		const apps = device.belongs_to__application;
-		if (_.isEmpty(apps) || _.isEmpty(apps[0])) {
-			return [device, undefined] as const;
-		}
-		return [device, apps[0]] as const;
+		return [device, device.belongs_to__application[0]] as const;
 	},
 	// Memoize the call based on UUID
 	(_sdk, deviceUUID) => deviceUUID,
--- a/src/utils/compose.ts
+++ b/src/utils/compose.ts
@@ -31,6 +31,7 @@ import type {
 import Logger = require('./logger');
 import type { ProgressCallback } from 'docker-progress';
 import { getCliUx } from './lazy';
+import { pick } from './helpers';
 
 export function generateOpts(options: {
 	source?: string;
@@ -130,7 +131,6 @@ export const createRelease = async function (
 	contract: import('@balena/compose/dist/release/models').ReleaseModel['contract'],
 	imgDescriptors: ImageDescriptor[],
 ): Promise<Release> {
-	const _ = require('lodash') as typeof import('lodash');
 	const crypto = require('crypto') as typeof import('crypto');
 	const releaseMod =
 		require('@balena/compose/dist/release') as typeof import('@balena/compose/dist/release');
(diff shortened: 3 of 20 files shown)


tests/helpers.ts
```diff
@@ -393,8 +392,11 @@ export function deepTemplateReplace(
...
 				return data.map((i) => deepTemplateReplace(i, templateVars));
 			}
-			return _.mapValues(data, (value) =>
-				deepTemplateReplace(value, templateVars),
+			return Object.fromEntries(
+				Object.entries(data).map(([key, value]) => [
+					key,
+					deepTemplateReplace(value, templateVars),
+				]),
 			);
```
Page-: I think this might be better as
```suggestion
			for (const key of Object.keys(data)) {
				data[key] = deepTemplateReplace(data[key], templateVars);
			}
			return data;
```
because it will avoid so many intermediate allocations, assuming we don't mind mutating the input object (I suspect we don't mind, but it is the main concern)

src/utils/logger.ts
```diff
@@ -90,7 +89,7 @@ class Logger {
 			livepush: logger.createLogStream('live'),
 		};
 
-		_.forEach(this.streams, function (stream, key) {
+		Object.entries(this.streams).forEach(([key, stream]) => {
```
Page-:
```suggestion
		for (const [key, stream] of Object.entries(this.streams)) {
```

src/utils/compose.ts
```diff
@@ -171,25 +170,32 @@ export const createRelease = async function (
...
+		serviceImages: Object.fromEntries(
+			Object.entries(serviceImages).map(([key, serviceImage]) => {
+				if ('created_at' in serviceImage) {
+					delete serviceImage.created_at;
+				}
+				if ('is_a_build_of__service' in serviceImage) {
+					delete serviceImage.is_a_build_of__service;
+				}
+				return [key, serviceImage];
+			}),
```
Page-: The new version already mutates the existing entries (which might be a problem, but assuming it's not) so you can simplify this to purely use that mutation via:
```suggestion
	Object.values(serviceImages).map((serviceImage) => {
		if ('created_at' in serviceImage) {
			delete serviceImage.created_at;
		}
		if ('is_a_build_of__service' in serviceImage) {
			delete serviceImage.is_a_build_of__service;
		}
	});

	...

	return {
		...,
		serviceImages,
```
And it might also be possible to drop those `in` checks, but I don't know

src/utils/cloud.ts
```diff
@@ -95,7 +95,7 @@ export const getDeviceAndMaybeAppFromUUID = _.memoize(
 	> => {
 		const device = await sdk.models.device.get(deviceUUID, deviceOptions);
 		const apps = device.belongs_to__application;
-		if (_.isEmpty(apps) || _.isEmpty(apps[0])) {
+		if (!apps.length || !apps[0]) {
```
Page-: Can you make these checks better than purely `!`, I also suspect you don't need both checks and they were historically here because we didn't know if `device.belongs_to__application` would be a deferred object like `{ __id: 123 }` or an expanded array like `[{ id: 123 }]`

src/utils/device/deploy.ts
```diff
@@ -611,9 +611,9 @@ export function generateTargetState(
 
 	const services: { [serviceId: string]: any } = {};
 	let idx = 1;
-	_.each(composition.services, (opts, name) => {
+	Object.entries(composition.services).forEach(([name, opts]) => {
```
Page-:
```suggestion
	for (const [name, opts] of Object.entries(composition.services)) {
```

src/utils/compose.ts
myarmolinsky: > The new version already mutates the existing entries (which might be a problem

`serviceImages` is retrieved from a Promise directly above and not used after this, so I think mutating here should be fine, but if you can think of a reason it should not be, please let me know

> And it might also be possible to drop those in checks, but I don't know

Sadly no, I initially tried without them and it errors with `Property 'X' does not exist on type 'ImageModel'.` I tried it again just to be sure and it indeed seems still necessary

src/utils/cloud.ts
myarmolinsky: Looks like we should just be able to remove this `if`, as an empty array's `apps[0]` should be `undefined` which means `return [device, apps[0]] as const;` will yield the same result as this `if`

src/commands/deploy/index.ts
```diff
@@ -279,13 +279,12 @@ ${dockerignoreHelp}
...
-				compositionToBuild.services,
-				servicesToSkip,
-			);
+			const compositionToBuild = structuredClone(project.composition);
+			for (const serviceToSkip of servicesToSkip) {
+				delete compositionToBuild.services[serviceToSkip];
+			}
 			let builtImagesByService: Dictionary<BuiltImage> = {};
-			if (_.size(compositionToBuild.services) === 0) {
+			if (!Object.keys(compositionToBuild.services).length) {
```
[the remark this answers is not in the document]
myarmolinsky: That's fair, but also `0` is the only possible falsey value in this case 😅 It's _technically_ a catch-all here but I meant it as a synonym. Accepted your suggestion in any case

tests/docker-build.ts
```diff
@@ -94,8 +93,12 @@ export async function inspectTarStream(
 		sourceTarStream.pipe(extract);
 	});
 
-	const $expected = _.mapValues(expectedFiles, (v) =>
-		_.omit(v, 'testStream', 'contents'),
+	const $expected = Object.fromEntries(
+		Object.entries(expectedFiles).map(([key, v]) => {
+			delete v.testStream;
+			delete v.contents;
```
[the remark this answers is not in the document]
myarmolinsky: It looks like this is the end of the line in all cases where it's used, so I think we're safe

src/utils/cloud.ts
Page-: Sounds good

PR #3020, opened 2025-11-19
Include sub-commands in --help

thgreasi:
Depends-on: https://github.com/balena-io/balena-cli/pull/3018
Change-type: minor
See: https://balena.fibery.io/Work/Project/2064

### DIFF
--- a/src/help.ts
+++ b/src/help.ts
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { Command, Interfaces } from '@oclif/core';
+import type { Command } from '@oclif/core';
 import { Help } from '@oclif/core';
 import * as indent from 'indent-string';
 import type { ResolvableReturnType } from 'balena-sdk/typings/utils';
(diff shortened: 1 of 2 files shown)


src/help.ts
```diff
@@ -76,9 +76,42 @@ export default class BalenaHelp extends Help {
 	public async showCommandHelp(command: Command.Loadable): Promise<void> {
 		const name = command.id;
 		const depth = name.split(':').length;
+		// We are suppressing how oclif prints sub-commands & sub-topics
+		// so that we can add a header for them right after.
```
myarmolinsky: It looks like oclif should be adding a header already: https://github.com/oclif/core/blob/ec07566d96fb96836cb1d6a0533bab0fba6febfc/src/help/index.ts#L167-L184

I know we did not see it, but we should see if we can figure out why we don't see it. And if we do need to override how the subtopics are displayed in order to attach our own header, I expect we should be able to just override `formatTopics` instead of dealing with the suppression and mirroring

myarmolinsky (approved):

PR #3041, opened 2026-01-05
Bump @balena/compose to 7.3.0

cywang117:
This bump adds support for `io.balena.requires.sw.linux` and `io.balena.requires.sw.balena-os` container contract labels.

compose v7.0.12 removed ImageModel and ReleaseModel typings in favor of getting them from balena-sdk, so this bump necessitates changing some type sourcing from compose to SDK as well.

Change-type: patch

### DIFF
--- a/src/commands/os/configure.ts
+++ b/src/commands/os/configure.ts
@@ -208,7 +208,7 @@ export default class OsConfigureCmd extends Command {
 			})) as DeviceWithDeviceType & {
 				belongs_to__application: BalenaSdk.PineDeferred;
 			};
-			deviceTypeSlug = device.is_of__device_type[0].slug;
+			deviceTypeSlug = device.is_of__device_type[0].slug!;
 		} else if (fleetSlugOrId != null) {
 			app = await getApplication(balena, fleetSlugOrId, {
 				$select: 'slug',
@@ -226,7 +226,7 @@ export default class OsConfigureCmd extends Command {
 
 		const deviceTypeManifest = await helpers.getManifest(
 			params.image,
-			deviceTypeSlug,
+			deviceTypeSlug!,
 		);
 
 		const { normalizeOsVersion } = await import('../../utils/normalization');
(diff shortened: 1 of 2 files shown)


src/utils/compose-types.d.ts
```diff
@@ -82,23 +86,32 @@ export interface ComposeProject {
...
+// TODO: The balena contract model for release.contract is `{ [key: string]: JSON } | JSON[] | null`
+// which appears to be incorrect, as a contract is represented as `{ [key: string]: any }`
+// (more specifically, `{ type: string, [key: string]: any }`) when querying a release for its contract.
+// @balena/contrato also defines a ContractObject as the type below, see:
+// https://github.com/balena-io/contrato/blob/543e891249431aa98474e17fecc66617ab9ca8f3/lib/types.ts#L15-L17
+export type Contract = {
+	type: string;
+	[key: string]: any;
+};
+
```
cywang117: `{ [key: string]: JSON } | JSON[]` has never been the valid type for a contract from what I could find. Even the release contract field generated from `balena.yml` is the type as specified above. For example, tested on a device:
```
contract: {
  description: "Test app for Intel NUC",
  name: "NUCTest",
  type: "sw.application",
  version: "1.0.0",
  data: {...}
}
```

Where is the `{ [key: string]: JSON } | JSON[]` type being derived from?

src/utils/compose.ts
```diff
@@ -165,16 +166,18 @@ export const createRelease = async function (
...
 	for (const serviceImage of Object.values(serviceImages)) {
 		if ('created_at' in serviceImage) {
-			delete serviceImage.created_at;
+			// created_at is not an optional field in the SDK, but we need to delete it
+			serviceImage.created_at = undefined as any;
 		}
 		if ('is_a_build_of__service' in serviceImage) {
-			delete serviceImage.is_a_build_of__service;
+			// is_a_build_of__service is not an optional field in the SDK, but we need to delete it
+			serviceImage.is_a_build_of__service = undefined as any;
```
cywang117: As mentioned above, these two fields are omitted in the ServiceImage type, unsure of reason but assuming that they're API-computed fields. However, this type error is due to balena-compose's releaseMod.create function returning the wrong type, so we should fix it there, then consider removing these deletes.

PR #3082, opened 2026-03-25
Update @balena/compose to v8

cywang117:
This update switches to using [@balena/compose-parser](https://github.com/balena-io-modules/balena-compose-parser) directly for compose file parsing. See [here](https://github.com/balena-io-modules/balena-compose-parser/pull/1) for compose features that are now supported with the switch to compose-parser.

Some compose features enabled by compose-parser are not implemented with this bump, as they are separate features:

* [Multiple compose file merging](https://docs.docker.com/reference/compose-file/merge/). Pass the files as an array to `parse`, with later files overriding earlier files.
* Build-time variable interpolation from shell env
    * **TODO compose-parser**: Variables from .env file in the same dir as the compose file aren't being picked up

## Release Notes
This release adds CLI support for a number of compose fields in the latest docker compose spec. The fields are listed below, however note that balena Supervisor support needs to be added for the fields to be fully supported by balena. These fields will be rejected by the Supervisor via the container contract mechanism until Supervisor support is implemented. See https://github.com/balena-io-modules/balena-compose-parser/pull/19 where the compose container contract is injected if new fields are detected.

### Dev overlay file merging
While in local mode, all top-level fields in `docker-compose.dev.yml` are merged into the main compose file. Before, only top-level `service` fields were merged. A bare Dockerfile with no docker-compose.yml will no longer apply docker-compose.dev.yml as an overlay, so make sure to create a docker-compose file if you wish for dev overlays to be applied.

#### Compose features:
* [Fragments](https://docs.docker.com/reference/compose-file/fragments/), [extensions](https://docs.docker.com/reference/compose-file/extension/) (pre-v8 balena-compose already supported this, but using a custom implementation. balena-compose v8 moves to the official implementation used by docker compose (compose-go)
* [`include`](https://docs.docker.com/reference/compose-file/include/) directive

#### `services.${serviceName}`:

NOTE: "logged / not logged" refers to Supervisor logging `Ignoring unsupported or unknown compose fields`

| field | SV behavior |
| --- | --- |
| annotations | ignored, logged |
| attach | ignored, logged |
| cgroup | ignored, logged |
| cpu_rt_runtime | ignored, logged |
| cpu_rt_period | ignored, logged |
| cpus | ignored, not logged |
| device_cgroup_rules | ignored, logged |
| env_file | N/A - compose-go folds this into `environment` during parsing |
| extends | N/A - compose-go folds this into the composition during parsing |
| healthcheck.start_interval | ignored, not logged |
| healthcheck.start_period | supported |
| ipc | **conditional error loop** if `ipc: service:${serviceName}`, supports `ipc: shareable` |
| label_file | N/A - compose-go folds this into `labels` during parsing |
| network_mode=service:${serviceName} | supported (this was supported before but compose-go adds a depends_on dependency which wasn't present before) |
| networks.aliases | supported |
| networks.ipv4_address | supported |
| networks.ipv6_address | ignored, not logged |
| networks.link_local_ips | **error loop** |
| networks.mac_address | ignored; Supervisor configures it in the wrong place, `Config.MacAddress`, instead of `NetworkSettings.MacAddress` |
| networks.driver_opts | ignored, not logged |
| networks.gw_priority | ignored, not logged |
| networks.priority | ignored, not logged |
| pid=service:${serviceName} | ignored, not logged |
| pids_limit | **conditional error loop** if negative `pids_limit` |
| post_start | ignored, logged |
| pre_stop | ignored, logged |
| read_only | supported |
| uts | ignored, logged |
| volumes_from | ignored, logged |

#### Reject unsupported fields with clear messages:

* Top-level `secrets` & `configs`
* Service fields defined in [SERVICE_CONFIG_DENY_LIST](https://github.com/balena-io-modules/balena-compose-parser/blob/e5a71ebb8c7a7b70ad62d73501456e8f03c74e36/lib/compose.ts#L198)
* Service.build fields defined in [BUILD_CONFIG_DENY_LIST](https://github.com/balena-io-modules/balena-compose-parser/blob/e5a71ebb8c7a7b70ad62d73501456e8f03c74e36/lib/compose.ts#L445)
* `container:${containerId}` for service.network_mode
* `container:${containerId}` for service.pid
* `container:${containerId}` for service.volumes_from
* All service.security_opt as unsupported except `no-new-privileges`
* Network subfields `attachable`, `external`, `name`, `driver` that's not `bridge` or `default`, `enable_ipv4`, `enable_ipv6`, `ipam.config.aux_addresses`
* Volume subfields `external`, `name`, `driver` that's not `local` or `default`
* Long syntax volumes of type `bind`, `npipe`, `cluster`, `image`, unless specified bind mount is in [allowedBindMounts](https://github.com/balena-io-modules/balena-compose-parser/blob/e5a71ebb8c7a7b70ad62d73501456e8f03c74e36/lib/compose.ts#L231) (corresponding to balena's supported [labels](https://docs.balena.io/reference/supervisor/docker-compose/#labels))
* Labels using balena namespace `io.balena.private`

#### Ignore fields that are required in balena-compose:

* Top-level `version` is now optional and won't throw an error if not present and equal to one of
  `2.0`, `2.1`, `2.2`, `2.3`, or `2.4`

#### Normalize fields parsed by compose-go to be more compatible with balena or in line with current balena-compose behavior:

* Remove null entrypoint, which in Docker means that the default entrypoint from the image is used,
  but in balena, overrides any ENTRYPOINT directive in the Dockerfile.
* Convert long syntax ports to short syntax, as all port definitions are converted to long syntax by compose-go,
  but aren't supported by the Supervisor. For similar reasons, also convert long syntax depends_on, devices, and volumes.
* Compose converts relative context paths to absolute paths, convert back to relative paths to be compatible with our build system.
* For any allowed bind mount specified in service.volumes, remove them from volumes and add to feature labels.
* Add image as build tag if image & build both present. However, uncertain if this behavior is necessary.
* Move `service.volumes` of type `tmpfs` to `service.tmpfs`
* Reject service.volume where:
        * `source` or `target` are not defined
        * volume is not defined in top-level volumes
        * it specifies volume [options](https://docs.docker.com/reference/compose-file/services/#long-syntax-6) (long syntax only)
* Reject service.tmpfs if it defines tmpfs [options](https://docs.docker.com/reference/compose-file/services/#long-syntax-6) (long syntax only)
* Allow short syntax tmpfs with options (not supported in Supervisor however, this is in line with previous balena-compose behavior, however is likely an erroneous allowance)
* Reject `build.network` that's not defined at top-level

#### Log warnings:

* service.expose is informational only, warn & remove it from composition
* Warn if oom_score_adj is set to a value under OOM_SCORE_ADJ_WARN_THRESHOLD (-900)
* `com.docker.network.bridge.name` network.driver_opt which may interfere with current Supervisor firewall implementation, which relies on default naming scheme

Change-type: major
Depends-on: https://github.com/balena-io-modules/balena-compose/pull/97
See: https://balena.fibery.io/Work/Project/Integrate-compose-go-using-Go-binary-into-balena-compose,-CLI,-and-builder-1746

### DIFF
--- a/automation/build-bin.ts
+++ b/automation/build-bin.ts
@@ -201,6 +201,22 @@ export async function signFilesForNotarization() {
 		'Developer ID Application: Balena Ltd (66H43P8FRG)',
 		'node_modules/macmount/bin/macmount',
 	]);
+	console.log('running command:', 'codesign', [
+		'-d',
+		'-f',
+		'--options=runtime',
+		'-s',
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/@balena/compose-parser/bin/balena-compose-parser',
+	]);
+	await whichSpawn('codesign', [
+		'-d',
+		'-f',
+		'--options=runtime',
+		'-s',
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/@balena/compose-parser/bin/balena-compose-parser',
+	]);
 }
 
 export async function runOclifCommand(cmd: string, opts: string[]) {
--- a/src/commands/deploy/index.ts
+++ b/src/commands/deploy/index.ts
@@ -16,7 +16,7 @@
  */
 
 import { Args, Flags, Command } from '@oclif/core';
-import type { ImageDescriptor } from '@balena/compose/dist/parse';
+import type { ImageDescriptor } from '@balena/compose-parser';
 import { ExpectedError } from '../../errors';
 import { getBalenaSdk, getCliUx, stripIndent } from '../../utils/lazy';
 import {
--- a/src/commands/deploy/index.ts
+++ b/src/commands/deploy/index.ts
@@ -263,7 +263,9 @@ ${dockerignoreHelp}
 					}
 					try {
 						await docker
-							.getImage((isBuildConfig(d.image) ? d.image.tag : d.image) ?? '')
+							.getImage(
+								(isBuildConfig(d.image) ? d.image.tags?.[0] : d.image) ?? '',
+							)
 							.inspect();
 
 						return d.serviceName;
@@ -309,7 +311,7 @@ ${dockerignoreHelp}
 				(d) =>
 					builtImagesByService[d.serviceName] ?? {
 						serviceName: d.serviceName,
-						name: (isBuildConfig(d.image) ? d.image.tag : d.image) ?? '',
+						name: (isBuildConfig(d.image) ? d.image.tags?.[0] : d.image) ?? '',
 						logs: 'Build skipped; image for service already exists.',
 						props: {},
 					},
--- a/src/utils/compose.ts
+++ b/src/utils/compose.ts
@@ -19,7 +19,7 @@ import type { Renderer } from './compose_ts';
 import type * as SDK from 'balena-sdk';
 import type * as Dockerode from 'dockerode';
 import * as path from 'path';
-import type { Composition, ImageDescriptor } from '@balena/compose/dist/parse';
+import type { Composition, ImageDescriptor } from '@balena/compose-parser';
 import type { RetryParametersObj } from 'pinejs-client-core';
 import type {
 	BuiltImage,
@@ -128,7 +128,7 @@ export const createRelease = async function (
 	composition: Composition,
 	draft: boolean,
 	semver: string | undefined,
-	contract: import('@balena/compose/dist/release/models').ReleaseModel['contract'],
+	contract: Dictionary<any> | null,
 	imgDescriptors: ImageDescriptor[],
 ): Promise<Release> {
 	const crypto = require('crypto') as typeof import('crypto');
@@ -165,16 +165,17 @@ export const createRelease = async function (
 		commit: crypto.pseudoRandomBytes(16).toString('hex').toLowerCase(),
 		semver,
 		is_final: !draft,
-		contract,
+		contract: contract ?? undefined,
 		imgDescriptors,
 	});
 
 	for (const serviceImage of Object.values(serviceImages)) {
 		if ('created_at' in serviceImage) {
-			delete serviceImage.created_at;
+			delete (serviceImage as Partial<typeof serviceImage>).created_at;
 		}
 		if ('is_a_build_of__service' in serviceImage) {
-			delete serviceImage.is_a_build_of__service;
+			delete (serviceImage as Partial<typeof serviceImage>)
+				.is_a_build_of__service;
 		}
 	}
 
--- a/src/utils/compose.ts
+++ b/src/utils/compose.ts
@@ -54,37 +54,34 @@ export function generateOpts(options: {
 	}));
 }
 
-/** Parse the given composition and return a structure with info. Input is:
+/** Create a ComposeProject from a Composition object.
  * - composePath: the *absolute* path to the directory containing the compose file
- *  - composeStr: the contents of the compose file, as a string
+ * - composition: the parsed Composition object
  */
 export function createProject(
 	composePath: string,
-	composeStr: string,
+	composition: Composition,
 	projectName = '',
 	imageTag = '',
 ): ComposeProject {
-	const yml = require('js-yaml') as typeof import('js-yaml');
-	const compose =
-		require('@balena/compose/dist/parse') as typeof import('@balena/compose/dist/parse');
-
-	// both methods below may throw.
-	const rawComposition = yml.load(composeStr);
-	const composition = compose.normalize(rawComposition);
+	const { toImageDescriptors } =
+		require('@balena/compose-parser') as typeof import('@balena/compose-parser');
 
 	projectName ||= path.basename(composePath);
 
-	const descriptors = compose.parse(composition).map(function (descr) {
+	const descriptors = toImageDescriptors(composition).map(function (descr) {
 		// generate an image name based on the project and service names
 		// if one is not given and the service requires a build
 		if (
 			typeof descr.image !== 'string' &&
 			descr.image.context != null &&
-			descr.image.tag == null
+			!descr.image.tags?.length
 		) {
 			const { makeImageName } =
 				require('./compose_ts') as typeof import('./compose_ts');
-			descr.image.tag = makeImageName(projectName, descr.serviceName, imageTag);
+			descr.image.tags = [
+				makeImageName(projectName, descr.serviceName, imageTag),
+			];
 		}
 		return descr;
 	});
--- a/src/utils/compose_ts.ts
+++ b/src/utils/compose_ts.ts
@@ -106,6 +106,19 @@ export async function applyReleaseTagKeysAndValues(
 const LOG_LENGTH_MAX = 512 * 1024; // 512KB
 const compositionFileNames = ['docker-compose.yml', 'docker-compose.yaml'];
 
+export async function parseComposePaths(
+	paths: string | string[],
+): Promise<Composition> {
+	const composeParser = await import('@balena/compose-parser');
+	const pathStr = Array.isArray(paths) ? paths.join(', ') : paths;
+	try {
+		return await composeParser.parse(paths);
+	} catch (err: any) {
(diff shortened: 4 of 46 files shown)


src/utils/compose.ts
```diff
@@ -128,7 +125,7 @@ export const createRelease = async function (
 	composition: Composition,
 	draft: boolean,
 	semver: string | undefined,
-	contract: import('@balena/compose/dist/release/models').ReleaseModel['contract'],
+	contract: Dictionary<any> | null,
```
myarmolinsky, line 128: Is this intentionally left as `Dictionary<any>`?

src/utils/compose_ts.ts
```diff
@@ -1556,15 +1528,13 @@ export function createRunLoop(tick: (...args: any[]) => void) {
 
 async function getContractContent(
 	filePath: string,
-): Promise<
-	import('@balena/compose/dist/release/models').ReleaseModel['contract']
-> {
+): Promise<Dictionary<any> | null> {
```
myarmolinsky, line 1539: Is this intentionally `Dictionary<any>`?

src/utils/compose.ts
cywang117: Yeah, the `contract` type from the SDK (something like `import('balena-sdk').BalenaModel['release']['Read']['contract']` or similar, can't recall exact type path) is too permissive, as it's `JSON | JSON[] | null`. A contract is never `JSON[]`. Contracts have already been verified as `JSON` by [`isContract`](https://github.com/balena-io/balena-cli/blob/6018bc072515156663b533c8695971b7d2f52203/src/utils/compose_ts.ts#L1561) which narrows the type to `Dictionary<any>` before it gets here, so this type here is fine IMO, but I'm not entirely comfortable with it either. Maybe a separate change to more accurately reflect the contract typing?

automation/build-bin.ts
```diff
@@ -201,6 +201,22 @@ export async function signFilesForNotarization() {
...
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/@balena/compose-parser/bin/balena-compose-parser',
+	]);
+	await whichSpawn('codesign', [
+		'-d',
+		'-f',
+		'--options=runtime',
+		'-s',
+		'Developer ID Application: Balena Ltd (66H43P8FRG)',
+		'node_modules/@balena/compose-parser/bin/balena-compose-parser',
```
klutchell: The existing `macmount` signing above uses `-d` too so this is consistent, ignore my earlier thought on that. But worth checking: is `-d` (display signature info) intentional in the signing step? It doesn't affect signing, just adds stdout noise.

src/utils/compose_ts.ts
```diff
@@ -118,118 +118,93 @@ export async function loadProject(
...
+					);
+					composition = await composeParser.parse([
+						composePath,
+						devOverlayPath,
+					]);
+				} else {
+					composition = await composeParser.parse(composePath);
+				}
+			} else {
+				composition = await composeParser.parse(composePath);
```
klutchell: The old `resolveProject` wrapped `fs.readFile` in a try/catch that added the compose file path to the error. Now if the parser binary can't read the file (permissions, encoding), users get a raw error from the Go binary with no CLI-side context.
Worth wrapping these `parse()` calls so the error includes the file path, similar to what was removed?

src/utils/compose_ts.ts
```diff
@@ -118,118 +118,93 @@ export async function loadProject(
...
+					opts.projectPath,
+					'docker-compose.dev.yml',
+				);
+				if (await exists(devOverlayPath)) {
+					logger.logInfo(
+						'Docker compose dev overlay detected (docker-compose.dev.yml) - merging.',
+					);
+					composition = await composeParser.parse([
+						composePath,
+						devOverlayPath,
```
klutchell: The dev overlay behavior changed here: old code merged only `services` from `docker-compose.dev.yml`, new code passes both files to `composeParser.parse()` which does a full top-level merge (networks, volumes, etc.). The PR description calls this out as intentional, but there's no test coverage for it. No fixture with a `docker-compose.dev.yml`, no `isLocal: true` test path.

Also, if `composeName` is falsy (default composition from a bare Dockerfile), the dev overlay is now skipped entirely. The old code applied it regardless. Not sure if that's intentional but either way it should have a test.

src/commands/deploy/index.ts
```diff
@@ -263,7 +263,9 @@ ${dockerignoreHelp}
 					}
 					try {
 						await docker
-							.getImage((isBuildConfig(d.image) ? d.image.tag : d.image) ?? '')
+							.getImage(
+								(isBuildConfig(d.image) ? d.image.tags?.[0] : d.image) ?? '',
```
klutchell, line 267: The `tag` to `tags?.[0]` migration here and at line 314 has no behavioural test assertions in `deploy.spec.ts` (only type annotation changes).

If `tags` is `[]` instead of `undefined`, `?.[0]` returns `undefined` and the fallback produces an empty image name, which will cause a confusing Docker error downstream instead of a clear message.

cywang117: `tags` in practice should never be `[]`, as it gets populated in 2 places:
- [src/utils/compose.ts](https://github.com/balena-io/balena-cli/blob/balena-compose-v8/src/utils/compose.ts#L83): in createProject, when toImageDescriptors produces a descriptor with a BuildConfig that has no tags, it sets `tags = [makeImageName(projectName, serviceName, imageTag)]`. This happens early, when the project is first created from the composition.
- [src/utils/compose_ts.ts](https://github.com/balena-io/balena-cli/blob/balena-compose-v8/src/utils/compose_ts.ts#L421): in setTaskAttributes, after multibuild resolves the build task, `d.image.tags = [task.tag]` is set. This reassigns the tag to match whatever the build task resolved to.

That being said, there is no test for this case, so I added one that asserts that the tag array is populated in [3adbf19](https://github.com/balena-io/balena-cli/pull/3082/commits/3adbf194cf59c42215cb978f6c70c6da019d841c)

src/utils/compose_ts.ts
cywang117: Re: dev overlay skipped for bare Dockerfile: this is intentional. The new parse() API works with file paths, and the default composition with just bare Dockerfile doesn't have a file on disk to pass to the parser. The old behavior of applying a dev overlay to a default composition looked to be incidental. It can be argued that docker-compose.dev.yml without a docker-compose.yml is not a realistic use case. Thoughts?

Otherwise, added a test case for the dev overlay merging in [b2cbd97](https://github.com/balena-io/balena-cli/pull/3082/commits/b2cbd979f5c7a8b362673f1f6001179b15548f59)

src/utils/compose_ts.ts
cywang117: Good catch, added a parseComposePaths() wrapper that annotates errors with the file path, used in both loadProject and getServiceDirsFromComposition. Compose-parser already wraps Go binary errors as a human-readable ComposeError with some path info, but OS-level errors (permissions, missing binary) would still propagate without context. See fffedbec.

src/commands/deploy/index.ts
klutchell: SGTM

src/utils/compose_ts.ts
klutchell: That's a fair assessment, but it is a change in behaviour. Worth noting that it's intentional I guess as part of this PR.

src/utils/compose_ts.ts
cywang117: See https://github.com/balena-io/balena-cli/pull/3082#discussion_r3059792516

src/utils/compose_ts.ts
cywang117: I've noted this in the release notes section of the PR!

> A bare Dockerfile with no docker-compose.yml will no longer apply docker-compose.dev.yml as an overlay, so make sure to create a docker-compose file if you wish for dev overlays to be applied.

klutchell (approved):

PR #3085, opened 2026-04-07
Update TypeScript to 6.0.2

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Project/2389

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,6 @@
 		"module": "commonjs",
 		"target": "es2018",
 		"outDir": "build",
-		"strict": true,
 		"strictPropertyInitialization": false,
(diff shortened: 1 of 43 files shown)


tsconfig.json
```diff
@@ -13,6 +12,7 @@
 		"removeComments": true,
 		"sourceMap": true,
 		"skipLibCheck": true,
+		"types": ["node"],
```
Page-, line 15: This probably isn't necessary?

Page- (commented):
I think a lot of these changes could be done as a separate PR(s) to strip this one down, since the changes are only because ts6 now forces them but equally they are valid and even the expected way to do it in ts5 (eg all the `require` related changes)

PR #3094, opened 2026-05-01
Fix handleError exit code fallback

cywang117:
Use `|| 1` instead of `?? 1` to fall back to 1 even if process.exitCode
was already 0, as handleError implies failure.

Introduced in CLI [v23.0.1](https://github.com/balena-io/balena-cli/pull/3007) with https://github.com/balena-io/balena-cli/pull/3007

See: https://balena.fibery.io/Inputs/Pattern/CLI-exits-with-success-(0-exit-code)-even-though-remote-build-faills-5312

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -275,7 +275,8 @@ export async function handleError(error: Error | string) {
 			? 0
 			: Number.isFinite(truncatedCode)
 				? truncatedCode
-				: (process.exitCode ?? 1);
+				: // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+					process.exitCode || 1;
 
 	// Prepare message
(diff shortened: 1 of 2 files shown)


src/errors.ts
```diff
@@ -275,7 +275,8 @@ export async function handleError(error: Error | string) {
 			? 0
 			: Number.isFinite(truncatedCode)
 				? truncatedCode
-				: (process.exitCode ?? 1);
+				: // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+					process.exitCode || 1;
```
klutchell, line 279: We should note the WHY in the comments and the commit body (e.g. because handleError implies we should ignore a current exitCode of 0)

klutchell (approved):

PR #3110, opened 2026-06-17
Fix deploy progress feedback

shaunmulligan:
While working on the Nvidia hostapp-extension PoC, I noticed there are a few issues with the UX of `balena deploy` that are very ugly when pushing large AI images. The most notable is the bug for when the images are pushed to the registry. The progress bar there jumps immediately to 100% because the logic is backwards and then the user is left staring at 100% for 30+ minutes (I often thought the process had failed/hung)

I also added a elapsed time counter to the progress bar to improve the UX, this way the user knows that the process is still ticking along and hasn't failed.

Finally, this PR also changes the log after Pushing images to registry... which says `Saving image details for registry2.balena-cloud.com/v2/c089c421fb2336d0475166fbf3d0f9fa'` from debug to info, because it is useful to know that things are being finalized.

This is what the new UX looks like
Screenshot 2026-06-18 at 11 55 45

### DIFF
--- a/src/utils/compose.ts
+++ b/src/utils/compose.ts
@@ -315,27 +315,50 @@ const formatDuration = (seconds: number): string => {
 };
 
 const renderProgressBar = function (percentage: number, stepCount: number) {
-	percentage = Math.max(Math.min(percentage, 0), 100);
+	percentage = Math.min(Math.max(percentage, 0), 100);
 	const barCount = Math.floor((stepCount * percentage) / 100);
 	const spaceCount = stepCount - barCount;
 	const bar = `[${'='.repeat(barCount)}>${' '.repeat(spaceCount)}]`;
 	return `${bar} ${`${percentage}`.padStart(3)}%`;
 };
 
+/**
+ * Creates a progress callback for image push operations.
+ * Supports both TTY (in-place line replacement) and non-TTY (throttled newline output).
+ */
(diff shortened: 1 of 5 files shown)


src/utils/compose.ts
```diff
@@ -315,27 +315,59 @@ const formatDuration = (seconds: number): string => {
...
 };
 
+const formatElapsedForProgress = (elapsedSeconds: number): string => {
+	if (elapsedSeconds < 60) {
+		return ` (${elapsedSeconds}s)`;
+	}
+	const minutes = Math.floor(elapsedSeconds / 60);
+	const remainingSeconds = elapsedSeconds % 60;
+	return ` (${minutes}:${remainingSeconds.toString().padStart(2, '0')})`;
+};
```
otaviojacobi: @shaunmulligan can you confirm with claude if there's nothing else in the codebase that already does such things for us? I vaguely recall have similar things around, we just might need to find where (but I am not very confident)

shaunmulligan: Oh yeah, thats true, I see there is actually `formatDuration` that already does basically the same thing. Let me switch to re-using that 👍

shaunmulligan: used the existing `formatDuration` and amended that commit, much cleaner now.

otaviojacobi (approved):

